PR #169, opened 2019-11-15
Feat/ktabs

travispamaral:


docs/components/tabs.md
```diff
@@ -1,15 +1,16 @@
...
-<KTabs :tabs="tabs">
-  <template v-slot:tab1>
-    <p>Tab 1 content</p>
-  </template>
-  <template v-slot:tab2>
-    <p>Tab 2 content</p>
-  </template>
-</KTabs>
+**KTabs** - A mindblowing tabs component
+<ClientOnly>
```
darrenjennings: SSR tabs could be useful in the future if we ever use in another site that is vuepress or nuxt powered. This brings up an API question, should the tabs component be able to have a tab selected based on a prop or data attribute in the `tabs` list? Then you could "remember" the tab from something other than comparing the tab list with the `window.location.hash`. This would mean that you'd need to not reference window in the `data`, but check for SSR like we do with kicons.

travispamaral: I like the idea of being able to say which is open using a prop and theres a part of me that doesn't want the tabs to know about window at all. But at the same time keeping the window hash lookup in the tabs will mean they just work if there is a hash.

darrenjennings: Yes I agree, what I mean is that it uses both. So if it's server rendered, then it can select a tab before window is available, and then on mount, if it sees a window hash, then it can update there.

travispamaral: I removed the hash logic all together and added a `default-tab` prop. This way you can handle setting/using hash _outside_ of the component if thats the feature you want.

docs/components/tabs.md
```diff
@@ -0,0 +1,292 @@
...
+    </template>
+    <template v-slot:tab2>
+      <p>Tab 2 content</p>
+    </template>
+  </KTabs>
+</ClientOnly>
+
+```vue
+<KTabs
+  default-tab-hash="#tab2"
```
darrenjennings: I had a prop called "initialValue" with vue-autosuggest, only to later realize that users preferred `v-model`. I think we should do the same here. It means it would be reactive as well so you could `v-model="$route.hash"` and wouldn't have to worry about the change event.

darrenjennings: some prior art:
https://bootstrap-vue.js.org/docs/components/tabs
https://quasar.dev/vue-components/tabs

travispamaral: updated to use v-model dawg

docs/components/tabs.md
```diff
@@ -0,0 +1,278 @@
...
+        { hash: '#movies', title: 'Movies' },
+        { hash: '#books', title: 'Books' },
+      ]
+    }
+  }
+}
+</script>
+```
+
+### default-tab
```
darrenjennings: this is no longer a prop, it's all v-model now

docs/components/tabs.md
```diff
@@ -0,0 +1,278 @@
...
+</KTabs>
+```
+
+## Slots
+In order to actually see your tabbed content you must slot it using the tab hash property without the hash mark.
+
+<ClientOnly>
+  <KTabs :tabs="slottedTabs">
+    <template v-slot:pictures>
+      <p>Wow look Pictures!</p>
```
darrenjennings: looks like this example isn't working

docs/components/tabs.md
```diff
@@ -0,0 +1,278 @@
...
+        { hash: '#books', title: 'Books' },
+      ]
+    }
+  }
+}
+</script>
+```
+
+## Usage
+### Window Hash
```
darrenjennings: the example doesn't use window.location.hash anymore so might call this "Vue Router" and you could mention `window.location.hash` is also available. I don't think this is working btw. when I click a tab, the hash changes, but not the highlighted tab.

travispamaral: updated to be more generic. Also seems to be working for me
image

darrenjennings: for the hash demo to reproduce the issue: click "Books", refresh the page, then click "Pictures"

darrenjennings:
Looks like the tabs are inaccessible, we need to add aria attributes and allow users to be able to key through them: 

Reference docs:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role
https://www.w3.org/TR/wai-aria/#tablist

travispamaral:
> Looks like the tabs are inaccessible, we need to add aria attributes and allow users to be able to key through them:
> 
> Reference docs:
> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role
> https://www.w3.org/TR/wai-aria/#tablist

Attempted to format the MDN example into what has been built in this PR but it was unsuccessful. Instead I added aria labels, a tabindex and a handler to select the focused tab

PR #172, opened 2019-11-18
feat(styles): adds missing utilities & add important declarations 

travispamaral:
### Summary
Adds missing flex, display, width utilities. Add `!important` to classes to ensure they take effect regardless of where the stylesheet is imported.

### DIFF
--- a/docs/style-guide/utilities.md
+++ b/docs/style-guide/utilities.md
@@ -1,3 +1,7 @@
+---
+pageClass: utilities
+sidebarDepth: 1
+---
 
 # Utilities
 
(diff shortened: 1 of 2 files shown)


travispamaral:
> It would be nice if we structured our stylesheets better so that important isnt needed but if this is the only way to do it without getting too complicated then lgtm

Im not following, the real only way to structure would to build everything out into their own files but because we have so little amount of styles probably not worth it. If you only want forms you can import the forms scss file, etc

tzou-kong:
> > It would be nice if we structured our stylesheets better so that important isnt needed but if this is the only way to do it without getting too complicated then lgtm
> 
> Im not following, the real only way to structure would to build everything out into their own files but because we have so little amount of styles probably not worth it. If you only want forms you can import the forms scss file, etc

well think about why important is needed. If its needed its because we needed to override some existing styles that were already on the element right? Well why would we need to do that? Ideally if we structured our styles appropriately we wouldn't need to. It's often because we have styles that conflict with other then one needs to be overriden. Why are conflicts happening? Well you are already refactoring the third party libs like bootstrap so that will get rid of alot of potential conficts. But we have other conflicts internally then that means we aren't cascading or structuring our css efficiently. but maybe I'm old school I've always been taught that `!important` should never be necessary in code if styles were structured efficiently.

travispamaral:
Ah your worried about the important. The reason is because we can not guarantee where a user of the stylesheet will import or what other styles they will have applied in their project. By adding !important it ensures that our utility classes are applied as one might expect. This is the same approach bootstrap takes for utilities. Right now if you look at the bootstrap branch you’ll see utility classes being overridden because of something set in the app. We always want utility classes to take effect regardless. If you don’t want the utility don’t use it.

tzou-kong:
> Ah your worried about the important. The reason is because we can not guarantee where a user of the stylesheet will import or what other styles they will have applied in their project. By adding !important it ensures that our utility classes are applied as one might expect. This is the same approach bootstrap takes for utilities. Right now if you look at the bootstrap branch you’ll see utility classes being overridden because of something set in the app. We always want utility classes to take effect regardless. If you don’t want the utility don’t use it.

Right, forgot for a moment that this was for kongponents not our codebase, my bad. Yeah looks good then

PR #176, opened 2019-12-13
style(KTable) Remove button padding

travispamaral:
### Summary
It was realized that when using a `KButton` in `KTable` the padding of the button dramatically increased the table row height. This is undesired so this fix will remove any padding. If a design/user calls for it they can override in their project

#### Before
image

### After
image

#### Changes made:
* Remove button top and bottom padding

### PR Checklist

### DIFF
--- a/packages/KTable/KTable.vue
+++ b/packages/KTable/KTable.vue
@@ -110,8 +110,8 @@ table.k-table {
       }
       button,
       .k-button {
-        padding-top: 0;
-        padding-bottom: 0;
+        --KButtonPaddingY: 0;
+        --KButtonPaddingX: 0;
       }
     }
   }
--- a/packages/KTable/KTable.vue
+++ b/packages/KTable/KTable.vue
@@ -108,6 +108,11 @@ table.k-table {
           text-decoration: underline;
         }
       }
+      button,
+      .k-button {
+        padding-top: 0;
+        padding-bottom: 0;
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/KTable/KTable.vue
```diff
@@ -108,6 +108,11 @@ table.k-table {
           text-decoration: underline;
         }
       }
+      button,
+      .k-button {
+        padding-top: 0;
+        padding-bottom: 0;
```
darrenjennings: re: our discussion offline, we should hook into a css var for this one. e.g. a new KButton var: `--KButtonPaddingY`.

travispamaral: Added to button and replaced here. Works like a charm 🎉

packages/KTable/KTable.vue
```diff
@@ -108,6 +108,11 @@ table.k-table {
           text-decoration: underline;
         }
       }
+      button,
+      .k-button {
```
darrenjennings: we should be able to just declare these vars on the table itself, then you don't have to worry about k-button class ever changing

travispamaral: I did do that at first but seemed cleaner to apply to the the actual button. Then when inspecting the code you see the override where its taking place. IDC though I can move if you think that makes more sense

darrenjennings: we can leave the specificity for now sure thing. we've changed k-button class once before, but doubt we'd do it again.

PR #177, opened 2019-12-17
fix(KTable) Swap vars for negative margin

travispamaral:
Swap out css var overrides for negative margin so buttons look normal but row is short

### DIFF
--- a/packages/KTable/KTable.vue
+++ b/packages/KTable/KTable.vue
@@ -110,7 +110,8 @@ table.k-table {
       }
       button,
       .k-button {
-        margin: calc(-1 * var(--KButtonPaddingY, var(--spacing-xs))) 0;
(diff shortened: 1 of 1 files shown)


packages/KTable/KTable.vue
```diff
@@ -110,8 +110,7 @@ table.k-table {
       }
       button,
       .k-button {
-        --KButtonPaddingY: 0;
-        --KButtonPaddingX: 0;
+        margin: calc(-1 * var(--KButtonPaddingY, var(--spacing-xs))) 0;
```
darrenjennings: not sure we want 0 margin on left and right

travispamaral: yes you're right oversight. Just pushed to only do top and bottom

PR #181, opened 2020-01-14
fix(*) enable postcss on UMD the build process - INTF-1522

jeffyongtaotang:
### Summary
The `Postcss` was disable on the UMD build process, so the css var fallback were not injected.

##### before
image

##### after
image

#### Changes made:

* extracts `postcss-custom-propreties` plugin config
* injects `posscss` plugin config into the build process

### PR Checklist


jeffyongtaotang:
image

we can’t use the `importForm` because of the triple css-var in KButton.
the `importForm` means postcss loads css var from  packages/styles/styles.css, once postcss got the css-var like --spacing-xs , then it’s will only inject the 1st fallback, but not the 2nd fallback(the hardcoded color)

PR #184, opened 2020-01-16
style(kslideout) adds hardcoded color fallback

jeffyongtaotang:
### Summary
We need to add a hardcoded color fallback to a css-var, then the UMD module can be use in IE11.

image

#### Changes made:

* adds color fallback

### PR Checklist

### DIFF
--- a/packages/KSlideout/KSlideout.vue
+++ b/packages/KSlideout/KSlideout.vue
@@ -87,7 +87,7 @@ export default {
     right: 0;
     bottom: 0;
     left: 0;
-    background: var(--tblack-45, color(tback-45));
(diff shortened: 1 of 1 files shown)


packages/KSlideout/KSlideout.vue
```diff
@@ -87,7 +87,7 @@ export default {
     right: 0;
     bottom: 0;
     left: 0;
-    background: var(--tblack-45);
+    background: var(--tblack-45, color(tback-45));
```
xitongzou: tblack, not tback

jeffyongtaotang: Epic typo >_<

In fact, `color()` is not working, let go back to `hardcoded` style.

PR #188, opened 2020-01-24
feat(kcard): add permanent shadow property

darrenjennings:
### Summary
adds permanent shadow property to card even when not hovering.

### PR Checklist

### DIFF
--- a/packages/KCard/KCard.vue
+++ b/packages/KCard/KCard.vue
@@ -1,6 +1,6 @@
 <template>
   <div
-    :class="[borderVariant, {'hover': hasHover }]"
+    :class="[borderVariant, {'hover': hasHover, 'kcard-shadow': hasShadow }]"
     class="kong-card">
     <div
       v-if="title || $scopedSlots.title || $scopedSlots.actions || $slots.title"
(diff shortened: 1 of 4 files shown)


packages/KCard/KCard.vue
```diff
@@ -66,7 +71,6 @@ export default {
 
 .kong-card {
   padding: var(--KCardPaddingY, 1rem) var(--KCardPaddingX, 1rem);
-  margin-bottom: var(--KCardMarginY, 1rem) var(--KCardMarginX, 1rem);
```
jeffyongtaotang: image


@xitongzou fixed `margin-bottom` for above, but it seems should be `margin-bottom: var(--KCardMarginY, 1rem) ;`
(https://github.com/Kong/kongponents/pull/185/files#diff-870dbd37faad1d06526ecdc0d78ca840R69)

darrenjennings: I removed this because cards shouldn't worry about margin. If a user wants to add margin then they can do it like:

`<KCard class="mb-2">`

This helps with composition.

jeffyongtaotang: ok, I will create a PR to KM, after this got merge.

darrenjennings: yea we'll need to review some things such as immunity alert cards as well

PR #197, opened 2020-03-11
feat(ktable): add sorting for ktable - fti-1291

xitongzou:
### Summary
added logic to handled sorting of different columns

#### Changes made:
*added two props, sortState and sortKey which allows the parent to pass in the data needed for the KTable to sort the appropriate column. Clicking the appropriate column will emit and event which is handled by the parent. The sorting logic is left up to the parent component.
*added exported defaultSorter function which can be optionally implemented via a helper in the parent component

### PR Checklist


packages/KTable/KTable.vue
```diff
@@ -93,6 +117,19 @@ table.k-table {
       text-align: left;
       font-size: var(--KTableHeaderSize, var(--type-sm, type(sm)));
       font-weight: 500;
+      cursor: pointer;
```
darrenjennings: this is a little confusing for when a column cannot be sorted, so we probably shouldn't put a cursor for all tables and all header cells. Not sure how it should work right now as click the headers in the vuepress docs don't do anything

xitongzou: yeah not sure what is a good example for the vue docs since the sort logic is left up to the parent implementation for it to actually do anything

darrenjennings: we should add an example in the docs of a parent component using ktable sort

packages/KTable/KTable.vue
```diff
@@ -80,6 +97,11 @@ table.k-table {
 .k-table {
   width: 100%;
   max-width: 100%;
+
+  &.sortable th {
```
darrenjennings: looks like the class is on the table, so every single column has a pointer cursor. We should define a prop "sortable" on `options.headers` objects that tells ktable if a given column should be clickable (i.e. cursor: pointer) and if it should emit the sort event.

packages/KTable/KTable.vue
```diff
@@ -65,6 +68,20 @@ export default {
     isSmall: {
       type: Boolean,
       default: false
+    },
+    /**
+     * the sort state for the table.
+     */
+    sortState: {
```
darrenjennings: would prefer this to be `sortOrder` so it's a little more obvious.

packages/KTable/KTable.vue
```diff
@@ -65,6 +68,20 @@ export default {
...
+      type: String,
+      default: 'ascending'
```
darrenjennings: add validator here:

```js
validator: function (value) {
  return ['ascending', 'descending'].indexOf(value) > -1
}
```

docs/components/table.md
```diff
@@ -85,6 +85,80 @@ Lessen the table cell padding
...
+        this.tableOptions.data.sort((a, b) => {
+          if (a[key] < b[key]) return -1
+          if (a[key] > b[key]) return 1
+          return 0
+        })
+      } else {
+        this.tableOptions.data.reverse()
+      }
+      this.sortState = this.sortState === 'ascending' ? 'descending' : 'ascending'
+    }
```
darrenjennings: we should add a `sort` function to the `KTable` kongponent itself and then a user can just use that. It can be an exported function at the top of the KTable `<script>` tag similar to how we export appearances in `KButton`. Right now, none of our application specific logic uses anything special so a default function returning different comparators based on the data type even if it's just client side would be nice.

packages/KTable/KTable.vue
xitongzou: I'm confused. a table only has `sortable` if a sortKey is passed in. Yes every column is clickable if that happens but that's how it should be right? is there a situation where we would want only some columns to be sortable but others not? I thought it was either everything is sortable or everything isn't sortable.

docs/components/table.md
```diff
@@ -85,6 +85,77 @@ Lessen the table cell padding
...
+    return {
+      sortOrder: 'ascending',
+      sortKey: 'name',
+      sortField: sortField,
+      tableOptions: {
+        headers: [
+          { label: 'Name', key: 'name', sortable: true },
+          { label: 'ID', key: 'id', sortable: true },
+          { label: 'Enabled', key: 'enabled', sortable: true },
+          { key: 'actions', hideLabel: true }
```
darrenjennings: unused headers `actions` in this example

docs/components/table.md
```diff
@@ -85,6 +85,77 @@ Lessen the table cell padding
...
+<template>
+<KTable
+    :options="tableOptions"
+    :sort-order="sortOrder"
+    :sort-key="sortKey"
+    @sort-field="sortFieldHelper"
+    />
+</template>
+<script>
+import { sortField } from '../../packages/KTable/KTable'
```
darrenjennings: for the example, we should use `'@kongponents/KTable'` to be closer to what a user of kongponents will use. Also, calling the helper "sortField" is a little confusing because the event name is also sortField. How about "defaultSorter"? Can we somehow reduce the boilerplate of a defaultSorter usage case? e.g.
```vue
<KTable
    :options="tableOptions"
    :sort-order="sortOrder"
    :sort-key="sortKey"
    @sort-field="key => defaultSorter(key, sortKey, sortOrder)"
    />
```

And then have defaultSorter actually mutate the value of sortOrder/sortKey directly without needing to do it in a `method`. How do you feel about renaming "sort-field" to just be `@sort`?

docs/components/table.md
```diff
@@ -85,6 +85,77 @@ Lessen the table cell padding
...
+            enabled: 'true'
+          },
+          {
+            name: 'Website Desktop',
+            id: '328027447731198',
+            enabled: 'false'
+          },
+          {
+            name: 'Android App',
+            id: '405383051040955',
```
darrenjennings: Let's make an example have a number id so we can see that the default sort works with strings and numbers.

docs/components/table.md
```diff
@@ -85,6 +85,77 @@ Lessen the table cell padding
...
+            enabled: 'true'
```
darrenjennings: what if this is a true boolean, does the default sort work?

xitongzou: how should booleans be sorted anyways?

xitongzou: right now its sorted as a string e.g alphabetically

docs/components/table.md
xitongzou: yeah i tried this method out already it seems that mutating the value in the method doesn't really work and the changes do not actually persist, thats why I had to make the helper method to set the values onto `this`

docs/components/table.md
```diff
@@ -230,6 +300,51 @@ An Example of changing the hover background might look like.
...
 ```
 
+<script>
+import { defaultSorter } from '../../packages/KTable/KTable'
+export default {
+  data() {
+    return {
+      sortOrder: 'ascending',
+      sortKey: 'name',
+      defaultSorter: defaultSorter,
```
darrenjennings: this is a reference to a function so it works for `data`, but would prefer it to be in methods e.g.
```js
methods: {
    sortFieldHelper (key) {...},
    defaultSorter
  },

```

packages/KTable/KTable.vue
```diff
@@ -39,6 +42,38 @@
 </template>
 
 <script>
+export const defaultSorter = (key, sortKey, sortOrder, items) => {
```
darrenjennings: could use jsdoc here. I'm a little confused on difference between key and sortKey. is sortKey basically "the previous key used to sort"? I would imagine that sortKey is what you would be sorting on but afaict `key` is actually what is being used to sort.

packages/KTable/__snapshots__/KTable.spec.js.snap
```diff
@@ -4,16 +4,16 @@ exports[`KTable has hover class when passed 1`] = `
 <table class="k-table has-hover">
   <thead>
     <tr>
-      <th>
+      <th class="sortable ">
         Name
       </th>
-      <th>
+      <th class=" ">
```
darrenjennings: looks like we're injecting a random space here in the class. we should trim the string or change `"sortable"` to `" sortable"` (prefixed with a space) in the :class definition.

packages/KTable/KTable.spec.js
```diff
@@ -68,4 +68,49 @@ describe('KTable', () => {
...
+
+    expect(items[0].username).toEqual('bobby')
+    expect(sortKey1).toEqual('username')
+    expect(sortOrder1).toEqual('ascending')
+
+    let {sortKey: sortKey2, sortOrder: sortOrder2} = defaultSorter('username', 'username', sortOrder1, items)
+
+    expect(items[0].username).toEqual('zach')
+    expect(sortKey2).toEqual('username')
+    expect(sortOrder2).toEqual('descending')
```
darrenjennings: can we add a `defaultSorter(): sorts the items by string, number` test

docs/components/table.md
```diff
@@ -85,6 +85,76 @@ Lessen the table cell padding
   isSmall />
 ```
 
+### Sorting
+There are two props used to make the table sortable; ```sortOrder```, which is either 'ascending' or 'descending' and ```sortKey```, which tells the table which column is currently being sorted. If a sortKey exists, then clicking the
+table header will emit an Event called ```sort``` which must be handled by the parent component to implement the
+actual sorting logic. A basic implementation of ```sort``` called ```defaultSorter``` is included in KTable and can be imported separately and used with a helper function in the parent.
```
travispamaral: this is confusing to me just coming in to this. `Sort` is the handler? and `sortKey` is the key in which the handler sorts the list by? I feel like `sort` and `defaultSorter` should be renamed to be more clear. The word _sort_ feels like a boolean to me

darrenjennings: maybe `@sorter`? then `defaultSorter` would be more obvious what it's used for.

darrenjennings: actually, @sorter wouldn't make much sense, because it's an action. `@sort` is read as "on sort". If it was a prop then it would make more sense to interpret as a boolean

travispamaral: Im not necessarily saying its wrong, Im just saying looking at this for the first time its confusing to me to interpret what Im supposed to do/what things mean

xitongzou: `sort` is the name of the event action unless you have a better name for it

packages/KTable/KTable.vue
xitongzou: yeah, i think i will rename it `sortKey` and `previousKey` it makes more sense

xitongzou: actually i know its confusing, but the param in the `defaultSorter` was renamed to `previousKey` - there's another `sortKey` thats passed in as a prop and that one is more in line with what the name says - its the current or default sort key so i kept that one the same

docs/components/table.md
travispamaral: Bootstrap uses [sort-compare](https://bootstrap-vue.js.org/docs/components/table/#custom-sort-compare-routine), Im not sure of a better name I just feel sort isn't clear. Again I understand it _now_ but at first read it was not clear. Especially when the function is named "sorter". Maybe `sort-function`, `sort-handler`, `sort-comparer`

xitongzou: `sort-compare` is the function / prop name in that example, and `sort` in KTable is the event name. it makes sense to me at least for the action to be `sort` and the function to act on it be the `sorter`

PR #206, opened 2020-04-09
fix(ktable): beef up defaultSorter to handle null, undefined and array values

xitongzou:
### Summary
Added additional functionality to KTable's defaultSorter to handle null, undefined and array values

#### Changes made:
*Add type checking for ktable to properly sort null, defined and array values
*added associated unit tests
*Updated docs to have array example

### PR Checklist

### DIFF
--- a/packages/KTable/KTable.vue
+++ b/packages/KTable/KTable.vue
@@ -53,26 +53,35 @@
  */
 export const defaultSorter = (key, previousKey, sortOrder, items) => {
   let comparator = null
-  const type = typeof items[0][key]
+  const numberComparator = (a, b) => a && b && a - b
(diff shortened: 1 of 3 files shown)


packages/KTable/KTable.vue
```diff
@@ -53,19 +53,27 @@
...
       comparator = (a, b) => {
-        if (a[key] < b[key]) return -1
-        if (a[key] > b[key]) return 1
+        const transformer = (val) => {
+          if (val === undefined || val === null) {
+            return ''
+          }
+
+          if (Array.isArray(val)) {
+            return String(val[0])
```
darrenjennings, line 69: the first item in the array could also be of any type e.g. undefined, null, number, string.

PR #207, opened 2020-04-20
feat(kskeleton): add full screen loader

darrenjennings:
### Summary
Add walking gruce 🦍full screen loader to KSkeleton.

### PR Checklist


docs/components/skeleton.md
```diff
@@ -132,3 +132,98 @@ This loading state is used for a spinner, which can be used for a wide variety o
   <KSkeleton type="spinner" />
 </template>
 ```
+
+## Full Screen Loading State
+The full screen loading state is used to display a full screen loader typically 
+during initial render of an app to avoid any FOUC while the app tries to figure
```
xitongzou, line 138: you might want to say `FOUC (Flash of Unstyled Content` here just to save potential googling by people

packages/KSkeleton/FullScreenKongSkeleton.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <div
+    class="fullscreen-loading-container"
+    data-testid="full-screen-loader">
```
xitongzou, line 4: i wonder if this needs to be in here. this is Kongponents and `data-testid` is for Kong specific testing right?

darrenjennings: data-testid is actually just a testing strategy so that implementation (normal dom attributes) can change, but you have selectors explicitly for testing. Eventually we'd be able to use sugar methods provided by the testing library: https://github.com/vuejs/vue-test-utils/issues/1500

xitongzou: ok, this is the first time we've used it in Kongponents so wanted to check

PR #213, opened 2020-05-15
feat(skeleton): add box primitive

darrenjennings:
### Summary
Introduce a new skeleton component `<KSkeletonBox />` which allows users to compose skeletons to form their own placeholder content. This was born from needing to create a custom placeholder that looks more like the content it is place-holding. For example, Kong Manager there are multiple cards with different layouts, so instead of making the cards look generic everywhere, allow them to be composable and specialized to the content. e.g. plugin select cards, workspace dashboard cards. Check new docs in this PR for more information on usage.

#### Changes made:
* export kskeleton as a plugin, auto-importing components
* use KSkeletonBox in all the skeleton variants
* slot content in skeleton card
* allow KSkeletonBox to be used as a primitive outside of KSkeleton

### PR Checklist

### DIFF
--- a/docs/.vuepress/enhanceApp.js
+++ b/docs/.vuepress/enhanceApp.js
@@ -26,6 +26,7 @@ import KToggle from '../../packages/KToggle/KToggle.js'
 import Komponent from '../../packages/Komponent/Komponent.js'
 import KSlideout from '../../packages/KSlideout/KSlideout.vue'
 import KSkeleton from '../../packages/KSkeleton/KSkeleton.vue'
+import KSkeletonBox from '../../packages/KSkeleton/KSkeletonBox.vue'
 
 export default ({
   Vue,
@@ -56,6 +57,7 @@ export default ({
   Vue.component('KToggle', KToggle)
   Vue.component('KSlideout', KSlideout)
   Vue.component('KSkeleton', KSkeleton)
+  Vue.component('KSkeletonBox', KSkeletonBox)
 
   Vue.prototype.$icons = Object.keys(icons)
 }
--- a/docs/components/skeleton.md
+++ b/docs/components/skeleton.md
@@ -6,7 +6,9 @@
 ### type
 - `type`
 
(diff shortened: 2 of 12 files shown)


docs/components/skeleton.md
```diff
@@ -165,12 +167,137 @@ to perform any expensive querying on first load.
...
+
+KSkeleton package uses a component to render the placeholder content 
+`<KSkeletonBox>`. It can be used as a component primitive to create your own
+custom placeholder components.
+
+<KSkeletonBox />
+<KSkeletonBox width="2" height="2"/>
+<KSkeletonBox width="5" height="2"/>
+<KSkeletonBox width="50" height="1"/>
+<KSkeletonBox width="100" height="2"/>
```
xitongzou, line 180: you should probably mention here the accept values for width and height since u have a validator there for certain values

travispamaral:
Can you take a look at the overflow of cards while you're here? If you shrink to real small it breaks out of the card
image

docs/.vuepress/enhanceApp.js
```diff
@@ -26,6 +26,7 @@ import KToggle from '../../packages/KToggle/KToggle.js'
 import Komponent from '../../packages/Komponent/Komponent.js'
 import KSlideout from '../../packages/KSlideout/KSlideout.vue'
 import KSkeleton from '../../packages/KSkeleton/KSkeleton.vue'
+import KSkeletonBox from '../../packages/KSkeleton/KSkeletonBox.vue'
```
travispamaral, line 29: I’m wondering if this should just be a helper class instead. I wonder if adding another component just to add classes is worth the trouble? Like really we’re just adding the ability to add whatever width and height div inside the card right?

travispamaral: Though at the end of the day I guess it doesn’t matter. We’re just exposing the underlying div. 🤔

darrenjennings: the idea is that we can describe our own DSL outside the context of css and it will allow users to easily create placeholder skeletons themselves without having to think too much about how it's implemented under the hood.

darrenjennings:
> Can you take a look at the overflow of cards

should be responsive now

PR #220, opened 2020-05-26
fix(kicon): remove gear kicon's opacity setting

xitongzou:
### Summary
the hard-coded fill-opacity on the gear kicon is preventing us from overriding it in other applications

#### Changes made:
*removed hard coded fill-opacity

### PR Checklist


darrenjennings:
seems we have inconsistency between kicons on fill-opacity and opacity. Any reason to not just remove fill-opacity from all kicons and have opacity be 0.5 across the board?

travispamaral:
I’m thinking all of the icons need some work. We should update to Make then so the same size, same color, same fill, etc. what are your thoughts on waiting to make this change and doing it in an overhaul?

xitongzou:
we're just going to use the same icons for now until we get better direction from UX

PR #222, opened 2020-06-12
feat(KInlineEdit): Adds KInlineEdit component

travispamaral:
### Summary
Adds inline edit. Currently only supports text input editing per feature request. More input types to come.

Screen Recording 2020-06-12 at 4 30 20 PM

### PR Checklist

### DIFF
--- a/docs/components/inline-edit.md
+++ b/docs/components/inline-edit.md
@@ -39,15 +39,13 @@ export default {
 </script>
 ```
 
-<Komponent :data="{ inlineText: 'Cool Text', emitedVal: '' }" v-slot="{ data }">
 <KCard>
   <template slot="body">
-    <div class="mb-4">Emit Value: {{ data.emitedVal }}</div>
+    <div class="mb-4">Emit Value: {{ emittedVal }}</div>
     <label class="k-input-label">Click to edit</label>
-    <KInlineEdit @changed="val => data.emitedVal = val"><p class="mt-0 mb-0">{{ data.emitedVal || data.inlineText }}</p></KInlineEdit>
+    <KInlineEdit @changed="handleEmitChange"><p class="mt-0 mb-0">{{ emittedDefault }}</p></KInlineEdit>
   </template>
 </KCard>
-</Komponent>
 
 ## Slots
 - `default` - Content to be edited
@@ -69,4 +67,33 @@ An HTML element must be passed in the slot. An error will be thrown if not passe
 :::
 
 ## Theming
-:lipstick: To theme, reference [KInput](/components/input.html#theming).
+:lipstick: To theme, reference [KInput](/components/input.html#theming). The input takes up 100% of its parent container. To change, add a class or width styling to the wrapping component.
+
+<Komponent :data="{ inlineText: 'Im 50%!' }" v-slot="{ data }">
+  <KInlineEdit class="w-50" @changed="newVal => data.inlineText = newVal"><h3>{{ data.inlineText }}</h3></KInlineEdit>
+</Komponent>
+
+```vue
+<KInlineEdit
+  class="w-50"
+  @changed="newVal => text = newVal">
+  <h3>{{ text }}</h3>
+</KInlineEdit>
+```
+
+<script>
+export default {
+  data() {
+    return {
+      emittedVal: '',
+      emittedDefault: 'Cool Text'
+    }
+  },
+  methods: {
+    handleEmitChange(val) {
+      this.emittedVal = val
+      this.emittedDefault = val
+    }
+  }
+}
+</script>
--- a/docs/components/inline-edit.md
+++ b/docs/components/inline-edit.md
@@ -17,35 +17,39 @@
 ```
 
 ## Props
-- `@changed` - Emitted blurred away from the editing input
+- `@changed` - Emitted blurred away from the editing input or when enter is pressed.
 
-When clicking out of the input `@changed` will fire and emit the value. Can be used to reset the outside data
+When clicking out of the input `@changed` will fire and emit the value. Can be used to reset the outside data.
 
-```vue
-<template>
-  <label>Click this text</label>
-  <EditableInput @changed="newValue => editableText = newValue ">
-    <h4>{{ editableText }}</h4>
-  </EditableInput>
-</template>
-<script>
-export default {
-  data() {
-    return {
-      editableText: 'Cool Text'
-    }
-  }
-}
-</script>
-```
+:::tip
+While the component itself does not protect against returning empty an empty value, it is advised to do a check at the implementation layer to ensure empty & validation are accounted for.
+:::
 
 <KCard>
   <template slot="body">
-    <div class="mb-4">Emit Value: {{ emittedVal }}</div>
-    <label class="k-input-label">Click to edit</label>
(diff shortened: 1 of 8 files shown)


packages/KInlineEdit/KInlineEdit.vue
```diff
@@ -0,0 +1,131 @@
...
+        }
+      }
+    }
+  },
+
+  mounted () {
+    if (this.$slots.default[0].tag === undefined) {
+      console.warn(`KInlineEdit expects slotted HTML tag.
+
+Example usage:
```
xitongzou, line 56: this formatting looks off?

travispamaral: The formatting has to be like this to look correct in the browser console. This is the same as KToggle https://github.com/Kong/kongponents/blob/master/packages/KToggle/KToggle.js#L34

packages/KInlineEdit/KInlineEdit.vue
```diff
@@ -0,0 +1,131 @@
...
+  <div
+    id="editable-wrapper"
+    class="k-inline-edit">
+    <input
+      v-if="isEditing"
+      ref="input"
+      :style="styles"
+      v-model="inputText"
+      class="k-input"
+      @blur="handleSave"
```
darrenjennings, line 11: When I remove the text in the vuepress example, it emits an empty value '', but the kinput is reset to what it was before.
Jun-15-2020 15-11-21

docs/components/inline-edit.md
```diff
@@ -0,0 +1,72 @@
+# Inline Edit
+
+**KInlineEdit** - A wrapper which adds inline edit capability. Currently only supports single text input.
+
+<Komponent :data="{ inlineText: 'Click to edit me' }" v-slot="{ data }">
+  <KInlineEdit @changed="newVal => data.inlineText = newVal"><h3>{{ data.inlineText }}</h3></KInlineEdit>
```
darrenjennings, line 6: does the component work with full-width? We will probably need to have it expand beyond the text to be full-width of the container e.g. when you hover over a far right portion of the inline-edit container:
image

packages/KInlineEdit/KInlineEdit.vue
```diff
@@ -0,0 +1,131 @@
...
+  name: 'KInlineEdit',
+  data () {
+    return {
+      isEditing: false,
+      inputText: '',
+      styles: {}
+    }
+  },
+  computed: {
+    listeners () {
```
darrenjennings, line 42: this isn't being used?

packages/KInlineEdit/KInlineEdit.vue
travispamaral: hmm good find, thanks

docs/components/inline-edit.md
travispamaral: it doesn't currently but it should by default actually. Then you can control the overall width at the top level. Will update

packages/KInlineEdit/KInlineEdit.vue
travispamaral: fixed this, wrong logic in the example

docs/components/inline-edit.md
travispamaral: Changed this to be 100% by default and added an example to make smaller

packages/KInlineEdit/KInlineEdit.vue
travispamaral: nope, old code, not needed with the `@blur` in the template

packages/KInlineEdit/KInlineEdit.vue
```diff
@@ -0,0 +1,122 @@
...
+      overflow: hidden;
+      transition: background-color 200ms ease;
+    }
+    &:hover > * {
+      background-color: var(--grey-92);
+    }
+  }
+  .k-input {
+    display: inline-flex;
+    width: 100%;
```
darrenjennings, line 114: this makes the k-input 100% once you click it, but the hoverable-area is still limited to just the target element width itself.

travispamaral: ah I thought you meant just the input. I'll move up to the wrapper to so it inherits from the overall parent.

travispamaral: made everything fullwidth

packages/KInlineEdit/KInlineEdit.vue
```diff
@@ -0,0 +1,124 @@
+<template>
+  <div
+    id="editable-wrapper"
+    class="k-inline-edit">
+    <input
```
darrenjennings, line 5: How does a user declare that an input is required? For example, you can't have an empty service package name, so are we going to be able to handle that?

travispamaral: right now you can't so open to suggestions. Could do something like autosuggest with `inputProps` and bind the object. Thoughts?

travispamaral: another option could be if you pass a required true prop we could disallow empty inputs. And if you remove all text and click away it negates your change and doesn't make any updates.

darrenjennings: @travispamaral is there even a use case for removing a value entirely? I think for the first iteration we can assume it's required and reset the value to the original, but still emit an event of empty text so that users (aka "us") could alert in some way that it's required e.g. toaster/kalert/red border or something. We can address what that looks like in implementation pr. Interestingly, if "empty" == "reset to original", then we have a new feature of reseting the input.

travispamaral: IDK if sending empty makes sense. We don't "reset" or even manage the text internally. The text is always passed up and then reassigned via the element in the slot. To me I feel the implementation should handle the rules. 

```js
// handler in implementing component
handleEmitChange(val) {
  if (val.length === 0) {
    $toaster.notify({message: "Can't be empty", type: "danger" })
  } else {
    this.text = val
  }
}

Comment (travispamaral) [reply packages/KInlineEdit/KInlineEdit.vue:5]: Another option is we could add a required function and if met emits/changes input to red. Then in the parent you can do whatever you want if the conditions are met.

File: docs/components/inline-edit.md [hunk 7]
```diff
@@ -0,0 +1,99 @@
...
+```
+
+## Props
+- `@changed` - Emitted blurred away from the editing input
+
+When clicking out of the input `@changed` will fire and emit the value. Can be used to reset the outside data
+
+```vue
+<template>
+  <label>Click this text</label>
```
darrenjennings, line 26: example not in sync with actual code below

packages/KInlineEdit/KInlineEdit.vue
darrenjennings: yea after playing with it a little bit I see what you mean, that we should make the example check the truthiness of the updated value. Here's an example:

```vue
<Komponent :data="{ inlineText: 'Click to edit me' }" v-slot="{ data }">
  <div>
    Updated: {{ data.inlineText }}
    <KInlineEdit @changed="newVal => { if(newVal.length) { data.inlineText = newVal } }">
      <h3>{{ data.inlineText }}</h3>
    </KInlineEdit>
  </div>
</Komponent>
```

PR #225, opened 2020-06-23
feat(*) Update Colors & Component Styles

travispamaral:
### Summary
Design has updated the colors across all components and this PR covers both updates to color references as well as general docs clean up and theming updates.

I also went ahead and added a custom prismJS theme to match the new styles. Every code example across all of Kong is light so I made a light theme using only colors from the new palette.

image

#### Changes made:
* New Color Palette
* New Vuepress theme
* KButton gets new colors across the board and focus states
* Update docs
* KModal header border removed
* KPop title border removed
* Vuepress and Vuepress live updated

### PR Checklist
- [x]  **Cleanliness:** does not have formatting issues, unused code (e.g., console.logs), or leftover comments

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -1,7 +1,6 @@
 const path = require('path')
 
 module.exports = {
-  port: 8082,
   title: 'Kongponents',
   description: 'Kong UI Components & Style Guide',
   themeConfig: {
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -1,6 +1,7 @@
 const path = require('path')
 
 module.exports = {
+  port: 8082,
   title: 'Kongponents',
   description: 'Kong UI Components & Style Guide',
   themeConfig: {
--- a/docs/.vuepress/styles/prism-theme.min.css
+++ b/docs/.vuepress/styles/prism-theme.min.css
@@ -48,14 +48,14 @@ pre[class*="language-"] {
   padding: 1em;
   margin: .5em 0;
   overflow: auto;
-  background: var(--grey-98);
+  background: var(--grey-100);
 }
 pre[class*="language-"] code,
 :not(pre) > code[class*="language-"] {
   padding: .1em .3em;
   border-radius: .3em;
   color: var(--blue-600) !important;
-  background: var(--grey-92);
+  background: var(--grey-200);
 }
 /*********************************************************
 * Tokens
--- a/docs/.vuepress/styles/prism-theme.min.css
+++ b/docs/.vuepress/styles/prism-theme.min.css
@@ -1,14 +1,25 @@
-/* Night Owl + Custom */
+/* Generated with http://k88hudson.github.io/syntax-highlighting-theme-generator/www */
+/* http://k88hudson.github.io/react-markdocs */
+/**
+ * @author k88hudson
+ *
+ * Based on prism.js default theme for JavaScript, CSS and HTML
+ * Based on dabblet (http://dabblet.com)
+ * @author Lea Verou
+ */
+/*********************************************************
+* General
+*/
 
-code[class*='language-'],
-pre[class*='language-'] {
-  color: #d6deeb;
-  font-family: Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace;
+div[class*="language-"],
+pre[class*="language-"],
+code[class*="language-"] {
+  color: var(--black-70);
+  font-family: 'Roboto Mono', Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace;
   text-align: left;
   white-space: pre;
   word-spacing: normal;
   word-break: normal;
-  word-wrap: normal;
   line-height: 1.5;
   -moz-tab-size: 4;
   -o-tab-size: 4;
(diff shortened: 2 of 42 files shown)


darrenjennings:
The light... THE LIGHT

travispamaral:
Tests are failing which is interesting because it seems they fail on the dependency of KButton in KMultiselect 🤔
https://github.com/Kong/kongponents/pull/225/checks?check_run_id=801263225#step:5:613

darrenjennings:
> The build is failing which is interesting because it seems they fail on the dependency of KButton in KMultiselect 🤔
https://github.com/Kong/kongponents/pull/225/checks?check_run_id=801263225#step:5:613

This is because KMultiselect is still using an old version of KButton, but it is using the latest styles colors, so KMultiselect pulls kbutton@0.1.16 down from npm to bundle, and then can't find the color `blue-base` anywhere. Bumping KMultiselect's kbutton fixes it.

travispamaral:
Yeah realized it probably needed bumping last night but didn’t have a chance to update. I’ll push an update shortly 👨🏻‍💻

travispamaral:
aren't we still going to have a problem though even after bumping the KButton version? The color changes are done all in one go so regardless of what version of KButton it currently pulls none will be referencing these colors

docs/.vuepress/config.js
```diff
@@ -1,6 +1,7 @@
 const path = require('path')
 
 module.exports = {
+  port: 8082,
```
darrenjennings, line 4: Shouldn't need to adjust this?

docs/.vuepress/styles/prism-theme.min.css
```diff
@@ -18,108 +29,116 @@ pre[class*='language-'] {
...
+  font-weight: bold;
 }
 .token.italic {
   font-style: italic;
 }
+.token.entity {
+  cursor: help;
+}
+/*********************************************************
+* Line highlighting
```
darrenjennings, line 123: Line highlighting is fugly
```
```bash{1}
```
image

docs/.vuepress/config.js
travispamaral: I did it because I find myself needing to run one of our other applications at the same time after I've already started the docs. So then I have to stop and restart because the other apps expect to be on `:8080`. I was thinking of instead using a 3000 something port which we never use. I think there is no harm in defining a port to run on

docs/.vuepress/styles/prism-theme.min.css
travispamaral: oh very weird, I actually themed that!
image

travispamaral: fixed numbers and highlighting
image

docs/.vuepress/config.js
darrenjennings: vuepress automatically detects when a port is in use. For example, I run apps on :8080 and then vuepress will start iterating from 8080 until it finds a port using `portfinder` package https://github.com/vuejs/vuepress/blob/7e9d0c1f42196dee52daabfaf796e0cdaa7bd6eb/packages/%40vuepress/core/lib/node/dev/index.js#L290

travispamaral: I understand but when you run docs on 8080 and then start manager to link a component or something manager doesn't run because it expects 8080. So if docs NEVER run on 8080 we never have to worry about it

packages/KButton/KButton.vue
```diff
@@ -155,72 +162,97 @@ export default {
 
   /* Apperance Variations */
   &.secondary {
-    border-color: var(--KButtonSecondaryBorder, color(grey-88));
-    background-color: var(--KButtonSecondaryBase, var(--secondary, color(grey-98)));
+    border-color: var(--KButtonSecondaryBorder, var(--black-10, color(black-1)));
+    background-color: var(--KButtonSecondaryBase, var(--white, color(white)));
```
darrenjennings, line 166: curious about styling here... Secondary buttons no longer look like buttons, and links now have background colors?
image

travispamaral: are you applying a state (active, hover, etc)? I can't seem to replicate. I just pushed a typo fix for the secondary although it should have been still working
image

steph-kalacis: Yeah good catch – this isn't what we discussed (and I'm sure you know that @travispamaral)

For reference: https://share.goabstract.com/a2d058fe-b21c-4961-95c8-919442e0d9bb?mode=design&sha=8fc822e6ef087e75c1edb228f937419c78640290

docs/.vuepress/config.js
darrenjennings: Yea if it helps your workflow, it's not hurting anyone else!

docs/style-guide/colors.md
```diff
@@ -37,13 +37,14 @@ export default {
...
     this.$page.colors = {
-      Blues: colors.filter(i => i.includes('blue')),
-      Reds: colors.filter(i => i.includes('red')),
-      Greens: colors.filter(i => i.includes('green')),
-      Yellows: colors.filter(i => i.includes('yellow')),
-      Blacks: colors.filter(i => i.includes('black')),
-      Greys: colors.filter(i => i.includes('grey')),
-      Whites: colors.filter(i => i.includes('white'))
+      Blue: colors.filter(i => i.includes('blue')),
+      Steal: colors.filter(i => i.includes('steal')),
```
darrenjennings, line 41: blue, steal... 😂

docs/style-guide/colors.md
```diff
@@ -37,13 +37,14 @@ export default {
...
 
     this.$page.colors = {
-      Blues: colors.filter(i => i.includes('blue')),
-      Reds: colors.filter(i => i.includes('red')),
-      Greens: colors.filter(i => i.includes('green')),
-      Yellows: colors.filter(i => i.includes('yellow')),
-      Blacks: colors.filter(i => i.includes('black')),
-      Greys: colors.filter(i => i.includes('grey')),
-      Whites: colors.filter(i => i.includes('white'))
+      Blue: colors.filter(i => i.includes('blue')),
```
darrenjennings, line 40: what is the rhyme reason to the numbers? I would expect them all to be 100-700 but black and gray seem to diverge from this pattern so assuming that's because they reflect the opacity values, but should we just make them all 100-700 to make it easier to remember? Also, should we be using transparency instead of hex for black and gray? Seems like it will be hard to use these values for backgrounds:

image

travispamaral: there is no particular reason, they are numbered lowest to highest of the needed color. Because there has never been a need for other yellows or greens there is less of them. Black and grey were left alone. Grey's could be made into a scale and could have been from the beginning I suppose but we're also not touched during this. Blacks I believe should remain the same because they are the only color changed by opacity.

travispamaral: Initially the idea was only change new colors but after discussing with design further we’ve decided to rename the greys to follow the other colors. Blacks will remain titled with their opacity.

packages/KSegmentedControl/KSegmentedControl.vue
```diff
@@ -74,6 +74,9 @@ export default {
   margin-left: -1px;
   flex: 1;
 }
+.segmented-control .k-button:focus {
+  z-index: 1;
```
darrenjennings, line 78: thank you!

PR #229, opened 2020-06-29
feat(KIcon) Adds icons for konnect

travispamaral:
Adds new and replaces some existing icons for the Konnect application


darrenjennings:
Why are the icons off-centered by default? Can they fill their svg containers? I know we revisit this often, just worried every user will have to do the viewBox/size thing every time they use one of these.
image

travispamaral:
yes we revisit this everytime lol. The default viewBox is 0 0 24 24. Currently the KIcon viewBox logic looks for a prop or will fall back to the svg. This poses a problem because we have a default value for the prop https://github.com/Kong/kongponents/blob/b92e2b71e5ecfbafe1d618b14abbec3f1bed7807/packages/KIcon/KIcon.vue#L59.

All of the icons are a total mess and need to be cleaned up like we did colors. I think thats out of scope at the moment

travispamaral:
Ideally all icons the same size and have the same viewBox. Then you can override if needed but shouldn't have to in a lot of cases.

travispamaral:
while we're at it we need to add support for elements that aren't paths. Like `<circle />`

PR #231, opened 2020-06-30
Feat/intf 2482 add table hover

xitongzou:
### Summary
Add ability to add custom row bindings via a passed in function

#### Changes made:
*Added rowAttrs, a custom function that adds row bindings onto KTable
*Added hasSideBorder, a prop that adds a default side border onto KTable

### PR Checklist

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -95,6 +95,111 @@ Adds `cursor: pointer` and `user-select: none` styling.
   isClickable />
 ```
 
+### Custom Row Attributes
+
+There is a prop called `hasSideBorder` that when passed in, enables a default side border on each row.
+This prop is on by default.
+These colors can be overridden and customized by using theme variables.
+
+Example with `hasSideBorder` set to false:
+
+<template>
+  <KTable
+    :options="tableOptions"
+    :hasSideBorder="false"
+    />
+</template>
+
+```vue
+<template>
+  <KTable
+    :options="tableOptions"
+    :hasSideBorder="false"
+    />
+</template>
+```
+
+Add custom properties to individual rows. The row object is passed as a param.
+
+- `rowAttrs` - is a Function that returns an object comprising the attributes.
+
+Example below:
+
+<template>
+  <KTable
+    :options="tableOptionsRowAttrs"
+    :rowAttrs="rowAttrsFn"
+    />
+</template>
+
+```vue
+<template>
+  <KTable
+    :options="tableOptions"
+    :rowAttrs="rowAttrsFn"
+    />
+</template>
+<script>
+import { defaultSorter } from '@kongponents/KTable'
+export default {
+  data() {
+    return {
+      row: null,
+      eventType: '',
+      tableOptions: {
+        headers: [
+          { label: 'Type', key: 'type' },
+          { label: 'Value', key: 'value' },
+          { label: 'Enabled', key: 'enabled'}
+        ],
+        data: [
+          {
+            type: 'desktop',
+            value: 'Windows 10',
+            enabled: 'true'
+          },
+          {
+            type: 'phone',
+            value: 'LineageOS',
+            enabled: 'false'
+          },
+          {
+            type: 'tablet',
+            value: 'ipadOS',
+            enabled: 'true'
+          }
+        ]
+      }
+    }
+  },
+  methods: {
+    rowAttrsFn (rowItem) {
+      return {
+        class: {
+          'enabled': rowItem.enabled === 'true',
+          'disabled': rowItem.enabled === 'false'
+        },
+        'data-testid': 'row-item'
+      }
+    }
+  }
+}
+</script>
+<style>
+.k-table {
+  tr.enabled {
+    --KTableHover: var(--green-200, #ccffe1);
+    --KTableBorder: var(--green-400, #19a654);
+  }
+
+  tr.disabled {
+    --KTableHover: var(--yellow-100, #fff9e6);
+    --KTableBorder: var(--yellow-200, #ffdc73);
+  }
+}
+</style>
+```
+
 ## Events
 Bind any DOM [events](https://developer.mozilla.org/en-US/docs/Web/Events) to
 various parts of the table.
@@ -262,111 +367,6 @@ export default {
   }
 }
 </script>
-```
-
-## Custom Row Attributes
-
-There is a prop called `hasSideBorder` that when passed in, enables a default side border on each row, using the color `steal-300` by default.
-This prop is on by default.
-These colors can be overridden by themes or by using custom row attributes.
-Example with `hasSideBorder` set to false:
-
-<template>
-  <KTable
-    :options="tableOptions"
-    :hasSideBorder="false"
-    />
-</template>
-
-```vue
-<template>
-  <KTable
-    :options="tableOptions"
-    :hasSideBorder="false"
-    />
-</template>
-```
-
-We can also add custom row attributes via a prop called `rowAttrs` - this prop is a Function that returns an object comprising the conditional custom attributes
-that you want the rows to have.
-In the following example, the rows have a custom green color if enabled is true, and yellow if enabled is false.
-We can also add a custom data-testid onto each row using these attributes (you can inspect it to see there is a custom data-testid attribute with `rowItem` on each row).
-
-<template>
-  <KTable
-    :options="tableOptionsRowAttrs"
-    hasSideBorder
-    :rowAttrs="rowAttrsFn"
-    />
-</template>
-
-```vue
-<template>
-  <KTable
-    :options="tableOptions"
-    :rowAttrs="rowAttrsFn"
-    hasSideBorder
-    />
-</template>
-<script>
-import { defaultSorter } from '@kongponents/KTable'
-export default {
-  data() {
-    return {
-      row: null,
-      eventType: '',
-      tableOptions: {
-        headers: [
-          { label: 'Type', key: 'type' },
-          { label: 'Value', key: 'value' },
-          { label: 'Enabled', key: 'enabled'}
-        ],
-        data: [
-          {
-            type: 'desktop',
-            value: 'Windows 10',
-            enabled: 'true'
-          },
-          {
-            type: 'phone',
-            value: 'LineageOS',
-            enabled: 'false'
-          },
-          {
-            type: 'tablet',
-            value: 'ipadOS',
-            enabled: 'true'
-          }
-        ]
-      }
-    }
-  },
-  methods: {
-    rowAttrsFn (rowItem) {
-      return {
-        class: {
-          'enabled': rowItem.enabled === 'true',
-          'disabled': rowItem.enabled === 'false'
-        },
-        'data-testid': 'row-item'
-      }
-    }
-  }
-}
-</script>
-<style>
-.k-table {
-  tr.enabled {
-    --KTableHover: var(--green-200, #ccffe1);
-    --KTableBorder: var(--green-400, #19a654);
-  }
-
-  tr.disabled {
-    --KTableHover: var(--yellow-100, #fff9e6);
-    --KTableBorder: var(--yellow-200, #ffdc73);
-  }
-}
-</style>
 ```
 
 ## Slots
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -115,8 +115,10 @@ The below example demonstrates the disabled state:
 </template>
 ```
 
-## rowAttrs
+## Custom Row Attributes
 Add custom properties to individual rows. The row object is passed as a param.
+
+## rowAttrs - Function that returns an object comprising the attributes.
 
 Example below:
 
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -21,7 +21,7 @@ tableOptions:
       id: 405383051040955
       enabled: true
 ---
-# Table
+# Table 
 Pass an object of headers & data to build a slot-able table.
 
 <KTable :options="$frontmatter.tableOptions" />
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -110,7 +110,20 @@ various parts of the table.
 
 - `@cell:<event>` - returns the `Event`, the cell value, and the type: `cell`
 
-<KTable :options="$frontmatter.tableOptions" @cell:mouseout="actionRow" @cell:mousedown="actionRow" />
+<template>
+  <div>
+    <div v-if="eventType">
+      {{eventType}} on: {{row}}
+    </div>
+    <div v-else>Waiting</div>
+    <KTable
+      :options="tableOptions"
+      is-clickable
+      @row:click="actionRow"
+      @cell:mouseover="actionRow"
+    />
+  </div>
+</template>
 
 #### Example
 
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -118,7 +118,8 @@ The below example demonstrates the disabled state:
 ## Custom Row Attributes
 Add custom properties to individual rows. The row object is passed as a param.
 
-## rowAttrs - Function that returns an object comprising the attributes.
+## rowAttrs
+Function that returns an object comprising the attributes.
 
 Example below:
 
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -66,13 +66,13 @@ export default {
 ```
 ## Props
 ### Hover
-Highlight the table row on hover
+Highlight the table row on hover. By default this is set to true. In the example we can set it to false as well.
 
-<KTable :options="$frontmatter.tableOptions" hasHover />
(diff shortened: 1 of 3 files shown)


packages/KTable/KTable.vue
```diff
@@ -164,6 +171,13 @@ export default {
     sortKey: {
       type: String,
       default: ''
+    },
+    /**
+     * A function that conditionally specifies the row classes on the rows
+     */
+    rowClasses: {
```
darrenjennings, line 178: I think it would make our lives easier if we v-bind on `tr` and call this `rowAttributes`. For example, if a user wants to add anything e.g. `data-testid` to the rows then we wouldn't need to add a new property.

xitongzou:
@darrenjennings @travispamaral run the docs to see the example. There's a lot of leeway on the `hasSideborder` prop as well, I gave it a default color and hover and made it a pre-req to use the `rowAttrs` but this can all be changed around or renamed to something better

packages/KTable/KTable.vue
```diff
@@ -184,6 +216,19 @@ export default {
 
 table.k-table {
   border-collapse: collapse;
+
+  &.side-border {
+    border-collapse: separate;
+    border-spacing: 0 5px;
+
+    tr:hover {
```
travispamaral, line 224: Idk that this should be a different color than the default hover value. Let’s double check with design but my thought is just use the predefined hover color and we can change it if it’s different per app

packages/KTable/KTable.vue
```diff
@@ -148,6 +149,12 @@ export default {
       type: Boolean,
       default: false
     },
+    /**
+     */
+    isClickable: {
```
travispamaral, line 154: I know I added this but if we’re adding prop descriptions can you add one here? “Adds hover and non selectable styling “ etc

packages/KTable/KTable.vue
```diff
@@ -164,6 +171,20 @@ export default {
...
+     * A function that conditionally specifies row attributes on each row
+     */
+    rowAttrs: {
+      type: Function,
+      default: () => ({})
+    },
+    /**
+     * A prop that enables a side border with a themable color to it.
+     */
+    hasSideBorder: {
```
travispamaral, line 185: I think @darrenjennings was thinking this should be true by default. I agree if all tables have it

packages/KTable/KTable.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <table
-    :class="{'has-hover': hasHover, 'is-small': isSmall}"
+    :class="{'has-hover': hasHover, 'is-small': isSmall, 'is-clickable': isClickable, 'side-border': hasSideBorder}"
```
travispamaral, line 3: I’m wondering if this should move to a computed haha it’s getting big!

docs/components/table.md
```diff
@@ -257,6 +255,125 @@ export default {
...
+          'disabled': rowItem.enabled === 'false'
+        },
+        'data-testid': 'row-item'
+      }
+    }
+  }
+}
+</script>
+<style>
+table.k-table {
```
travispamaral, line 351: This is confusing. If you can theme by css var why demo all this css overrides?

xitongzou: the theming is only overridable for the default border color and hover color but in this case we want to have multiple colors for border and hover so I'm confused how would we use multiple colors here with theming?

travispamaral: You're adding a specific class so instead of redefining every css trait, just set what the var is for each class.

```scss
.k-table
  &.enabled {
    --KTableBorder: var(--green-400);
    --KTableHover: var(--green-100);
  }
    &.disabled {
    --KTableBorder: var(--yellow-400);
    --KTableHover: var(--yellow-100);
  }
}
```

docs/components/table.md
```diff
@@ -139,37 +149,25 @@ export default {
   }
 }
 </script>
-<style>
-.clickable-rows tr {
-  cursor: pointer;
-}
-</style>
 ```
```
travispamaral, line 152: can you add vue to this code block?
image

packages/KTable/KTable.vue
xitongzou: the predefined hover color doesn't exist in KTable natively though

docs/components/table.md
```diff
@@ -66,7 +66,7 @@ export default {
...
 ## Props
 ### Hover
-highlight the table row on hover
+Highlight the table row on hover
```
travispamaral, line 69: now that highlight is true by default, lets change this example and doc to explain its always on and you can set it to false `hasHover="false"`

docs/components/table.md
```diff
@@ -115,14 +125,30 @@ various parts of the table.
...
-      class="clickable-rows"
       :options="tableOptions"
+      is-clickable
       @row:click="actionRow" 
       @cell:mouseover="actionRow"
     />
   </div>
 </template>
-
+<template>
```
travispamaral, line 135: this example is no longer showing

docs/components/table.md
```diff
@@ -257,6 +254,113 @@ export default {
 </script>
 ```
 
+### Custom Row Attributes
```
travispamaral, line 257: I think this should be moved out of the "Events" section and be its own before "slots". Also lets not make the border required to pass attrs. I would move `hasSideBorder` up above next to `clickable`,`small`, and `hover` and just say "by default each row has a left border. You can disable by passing false."

Then explain `rowAttrs` is a fn that allows you to bind any attribute to the table row. It passes the row details up so you can conditionally apply attributes. Then show the class example. 
"Here we demonstrate theming the `tableHover` variable based on row data., etc"

docs/components/table.md
```diff
@@ -257,6 +254,113 @@ export default {
...
+        headers: [
+          { label: 'Type', key: 'type' },
+          { label: 'Value', key: 'value' }
+        ],
+        data: [
+          {
+            name: 'Basic Auth',
+            id: '517526354743085',
+            enabled: 'true',
+            themeColors: []
```
travispamaral, line 318: lets remove this key, to me its confusing when in an example about theming and its not used to theme

docs/components/table.md
```diff
@@ -257,6 +254,113 @@ export default {
...
+          'disabled': rowItem.enabled === 'false'
+        },
+        'data-testid': 'row-item'
+      }
+    }
+  }
+}
+</script>
+<style>
+table.k-table.side-border {
```
travispamaral, line 350: here lets just style `.k-table` and not `side-border`

docs/components/table.md
```diff
@@ -468,4 +580,16 @@ export default {
   .table-wrapper {
   --KTableHover: lavender;
   }
+
+  table.k-table.side-border {
```
travispamaral, line 584: same as above. `.k-table tr.enabled/.disabled`

packages/KTable/KTable.vue
```diff
@@ -184,6 +217,19 @@ export default {
 
 table.k-table {
   border-collapse: collapse;
+
+  &.side-border {
```
travispamaral, line 221: I think its weird we have these two .k-table definitions. Lets combine everything into one nested css obj. This is left over formatting from before scss I think.

packages/KTable/KTable.vue
```diff
@@ -184,6 +217,19 @@ export default {
...
+    border-collapse: separate;
+    border-spacing: 0 5px;
```
travispamaral, line 223: this should be 4px.
image

packages/KTable/KTable.vue
```diff
@@ -184,6 +217,19 @@ export default {
...
+
+    tr:hover {
+      background-color: var(--KTableHover, var(--steal-100, color(steal-100)));
+    }
+
+    td:first-child {
+      border-left: 3px solid var(--KTableBorder, var(--steal-200, color(steal-200)));
```
travispamaral, line 230: in stead of adding to the first td, lets instead set border: none/0 to the whole TR and then use a box shadow here as its a little more flexible and also won't eat into the content. It should also be 2px & steal-300 per the design. Lets also remove this hover declaration and use steal-100 for the `.has-hover` class

```css
tr {
  border: none;
  box-shadow: -2px 0 0 var(--KTableBorder, var(--steal-300, color(steal-300)));
}
```

docs/components/table.md
xitongzou: question here: `border-spacing: 0 4px` and `border-collapse: separate` should we make this the default css then?

xitongzou: not having any side borders would then turn the border-collapse and border-spacing back to the way it was

packages/KTable/KTable.vue
xitongzou: re: my comment above for the border-spacing I think it was originally just to add that. it seems we always want `border-collapse: separate` now though, but that will actually remove the borders from each row, which will also affect KM as well and everything that uses `KTable` if we are fine with that

packages/KTable/KTable.vue
xitongzou: I defer to the CSS master here

packages/KTable/KTable.vue
travispamaral: The default should be collapse. And when `side-border` is applied its `separate`. so no need to remove the border as I mentioned in another comment, that will be done automatically. For KM we will add `hasBorder="false"` by default so they retain the borders

xitongzou: yeah but for `rowAttrs` we also need the border to be separate too so should we just repeat some styles there? the reason I had `sideBorder` on to require rowAttrs was this reason

travispamaral: Why do you need it separate for rowAttrs? technically rowAttrs has nothing to do with the border. It’s just a helper function to style based on row data.

xitongzou: ok, but in the example we would want to have different classes on each row and each border to give it a different color. So in that case, we have to make the border collapse separate. You're right that `rowAttrs` could be anything, not just related to borders. But in the case we do use it for borders which is the example case, I think we have to add this in. So I guess I'll just repeat it just for the example then.

travispamaral: I think it’s we’re just misunderstanding each other. You had said in order for `rowAttrs` to be applied you needed `sideBorder` and my point was they aren’t related. Just in the KHCP use case we’re using one to style the other. Happy to discuss over a call 🤓

docs/components/table.md
```diff
@@ -257,6 +264,111 @@ export default {
 </script>
 ```
 
+## Custom Row Attributes
+
+There is a prop called `hasSideBorder` that when passed in, enables a default side border on each row, using the color `steal-300` by default.
```
travispamaral, line 269: Please move this to the top props section https://github.com/Kong/kongponents/pull/231/files#r449151214

docs/components/table.md
```diff
@@ -257,6 +264,111 @@ export default {
...
+```vue
+<template>
+  <KTable
+    :options="tableOptions"
+    :hasSideBorder="false"
+    />
+</template>
+```
+
+We can also add custom row attributes via a prop called `rowAttrs` - this prop is a Function that returns an object comprising the conditional custom attributes
```
travispamaral, line 290: Lets move this description up (below "Custom Row Attrs") and update the description to not be a continuation.

```suggestion
Add custom properties to individual rows. The row object is passed as a param.

- `rowAttrs` - returns an object comprising the attributes
```

docs/components/table.md
```diff
@@ -257,6 +264,111 @@ export default {
...
+    :rowAttrs="rowAttrsFn"
+    />
+</template>
+
+```vue
+<template>
+  <KTable
+    :options="tableOptions"
+    :rowAttrs="rowAttrsFn"
+    hasSideBorder
```
travispamaral, line 308:
```suggestion

```

docs/components/table.md
travispamaral: I also don't think you need to call out the default color. Just say its customizable with a theme var

packages/KTable/KTable.vue
```diff
@@ -173,22 +195,39 @@ export default {
...
   max-width: 100%;
+  border-collapse: collapse;
+
+  &.side-border {
+    border-collapse: separate;
+    border-spacing: 0 4px;
+
+    tbody tr {
+      border: none;
+      box-shadow: -2px 0 0 var(--KTableBorder, var(--steal-300, color(steal-300)));
```
travispamaral, line 228: Actually looks like the default changes on hover. So maybe if `hasHover` then default should be `steal-200` and `steal-300` on hover
https://share.goabstract.com/edbb70e6-5126-4195-b523-80ab300e1459?mode=build&sha=5447ee919f352c8b9e0029181760b55ac98b844e

packages/KTable/KTable.vue
```diff
@@ -173,22 +195,43 @@ export default {
...
-table.k-table {
-  border-collapse: collapse;
-}
-
 .k-table {
   width: 100%;
   max-width: 100%;
+  border-collapse: collapse;
+
+  &.side-border.has-hover {
```
travispamaral, line 222: This should remain the way it was but. IF it has hover then add the hover styles (steal-300)

docs/components/table.md
travispamaral: Still not seeing this in the docs. Looks like its fixed on the other PR https://github.com/Kong/kongponents/pull/237/files#diff-4975fee17748787a06e99b3259d78e9fR311

docs/components/table.md
```diff
@@ -85,6 +85,121 @@ Lessen the table cell padding
...
+- `rowAttrs` - is a Function that returns an object comprising the attributes.
+
+Example below:
+
+<template>
+  <KTable
+    :options="tableOptionsRowAttrs"
+    :rowAttrs="rowAttrsFn"
+    />
+</template>
```
travispamaral, line 133:
```suggestion
## hasSidebar
Adds left border to each table row. By default set to true, the below example demos disabling. 

 ```vue
 <KTable
    :options="tableOptions"
    :has-sidebar="false" />
 ``` 

 <KTable
    :options="tableOptions"
    :has-sidebar="false" />

### Custom Row Attributes
Allows the ability to pass additional attributes to each table row.

- `rowAttrs` - is a Function that returns an object comprising the attributes.<template>

File: docs/components/table.md [hunk 23]
```diff
@@ -85,6 +85,121 @@ Lessen the table cell padding
...
+  }
+
+  tr.disabled {
+    --KTableHover: var(--yellow-100, #fff9e6);
+    --KTableBorder: var(--yellow-200, #ffdc73);
+  }
+}
+</style>
+```
+
```
travispamaral, line 202: Re-add example below

```suggestion
<KTable
    :options="tableOptionsRowAttrs"
    :rowAttrs="rowAttrsFn"
    />
</template>
```

docs/components/table.md
```diff
@@ -85,24 +85,169 @@ Lessen the table cell padding
...
+```vue
+<template>
+  <KTable
+    :options="tableOptions"
+    :hasSideBorder="false"
+    />
+</template>
+```
+
```
travispamaral, line 119:
```suggestion
### hasSidebar
Adds left border to each table row. By default set to true. The colors can be overridden by themes.
The below example demonstrates the disabled state:

<template>
  <KTable
    :options="tableOptions"
    :hasSideBorder="false"
    />
</template>

```vue
<template>
  <KTable
    :options="tableOptions"
    :hasSideBorder="false"
    />
</template>
```

## Custom Row Attributes
```

File: docs/components/table.md [hunk 25]
```diff
@@ -85,24 +85,169 @@ Lessen the table cell padding
...
 ## Events
 Bind any DOM [events](https://developer.mozilla.org/en-US/docs/Web/Events) to
 various parts of the table.
 
 ### Rows
 - `@row:<event>` - returns the `Event`, the row item, and the type: `row`
 
+<template>
```
travispamaral, line 207: This isn't needed. Can you revert to match what currently exists for the events section [here](https://kongponents.netlify.app/components/table.html#events). 
```
## Events
### rows
### cells
`example`
```

docs/components/table.md
travispamaral: can you copy this formatting to https://github.com/Kong/kongponents/pull/237

docs/components/table.md
```diff
@@ -85,6 +85,116 @@ Lessen the table cell padding
...
+  <KTable
+    :options="tableOptions"
+    :hasSideBorder="false"
+    />
+</template>
+```
+
+## rowAttrs
+Add custom properties to individual rows. The row object is passed as a param.
```
travispamaral, line 119:
```suggestion
## Custom Row Attributes
Add custom properties to individual rows. The row object is passed as a param.

## rowAttrs - Function that returns an object comprising the attributes.
```

PR #232, opened 2020-06-30
style(KButton) Update btn-link and anchor to blue-500

travispamaral:


darrenjennings:
Looks like ktable link styling also needs updating https://github.com/Kong/kongponents/blob/master/packages/KTable/KTable.vue#L245

PR #237, opened 2020-07-06
feat(ktable): add clickable rows and links - khcp-106

xitongzou:
### Summary
Added clickable rows and links to KTable.

#### Changes made:
*Made rows clickable to navigate to another page
*Made inline links in rows
*Added metakey support for clicking

### PR Checklist


xitongzou:
https://github.com/Kong/kongponents/pull/231 needs to be merged first

packages/KTable/KTable.vue
```diff
@@ -35,7 +36,15 @@
               :name="value.key"
               :row="row"
               :rowKey="rowIndex"
-              :rowValue="row[value.key]">{{ row[value.key] }}</slot>
+              :rowValue="row[value.key]">
+              <a
```
travispamaral, line 40: I think that you either need to wrap the entire row, which if so need to make sure you can still click buttons in the row without navigating. If that doesn't work I think we decided to leave as is and not use a native anchor tag

packages/KTable/KTable.vue
```diff
@@ -35,7 +36,15 @@
...
+                v-if="row[value.key] && row[value.key].href"
+                :href="row[value.key].href">{{ row[value.key].label }}</a>
+              <span v-else-if="row[value.key] && row[value.key].submenu">&#8230;</span>
```
travispamaral, line 43: what is a sub menu? Why would we use this?

docs/components/table.md
```diff
@@ -439,6 +742,155 @@ export default {
             themeColors: ['green', 'yellow']
           }
         ]
+      },
```
travispamaral, line 745: We need to show a demo of clickable links inside a row thats clickable (If I have an external link inside a clickable row it shouldn't open the row too). 

We also need an example of a button with an action (kpop, alert window, etc) to demo you can click a button without opening the row

packages/KTable/KTable.vue
xitongzou: why do we need to wrap the entire row? the row click is handled by `handleRowClick` which is implementation dependent but that way we can click the entire row as well as click any individual links inside of that row if we use this anchor tag here. Also wrapping the whole row with an `a` tag doesn't work well.

packages/KTable/KTable.vue
xitongzou: Leftover code, will remove

docs/components/table.md
xitongzou: We have a demo here.

image

I'll add an example with button too

packages/KTable/KTable.vue
travispamaral: let's chat with @darrenjennings. My understanding when we chatted was we were going to try and wrap the row if the prop was present _instead_ of the click handler to get native functionality. We could always just slot the anchor, theres no need to v-if one in the template. The thought was if a native link around the row doesn't work then we won't implement that. And we would document the user can add meta key support in their implementation

xitongzou: whats the benefit of having the native click? this implementation works and fulfills all the requirements doesn't it? do we need to make it more complex? yeah in the example the meta key support is all handled in the implementation but that is going by the row click handler

travispamaral: The idea of the anchor tag was to to get native right click/meta click support, and things like link url preview, etc. The discussion was _try_ to wrap in a link. If it doesn't work don't do it and just use the @row/@column handler

xitongzou: yeah, I think the main problem with wrapping it is it completely messes up the table styling so I went with the row handler

travispamaral: yeah I mentioned that might be a problem with wrapping the entire row. I wasn't sure what would happen but it might not be good. If its completely jacked with wrapping I'd say remove and just leave the updated click handler. Just add documentation to show how a user could hijack meta click

docs/components/table.md
xitongzou: Example with buttons added. Note that everything uses the same row click handler; anything more fine tuned can also be added in the row data's metadata (like if the button goes to a different href) but its all implementation dependent. 

image

travispamaral: This doesn't show that clicking the button doesn't fire the row click event. The button should instead fire an alert, console.log, Kpop, etc. Something other than a row click.

Overall I think these docs are getting more complicated then needed. Lets pair tomorrow to make sure we're all on the same page

xitongzou: How many examples do these docs need to show? I mean we can show a link, a button etc doing something entirely different too, but like I said this is all implementation specific. For this example, it fires the row click. If we want the button or link to do something else, we can just add an `@click` handler to them and then in the row event handler, add something like `if (e.target.tagName !== 'BUTTON')` to prevent the row click for buttons etc but again this is all up to how complicated and how many examples we need in the doc

travispamaral: Only one example is needed. The idea is to show that  while a row has a `row:click` event, having another action (like an actions menu) does not fire the row event. An example of this would be having the row click event open a new window and the button in a cell fires an alert, *or* KPop showing the new window does not open (row event doesn't fire). We do not want to add an  `e.target !==` in implementation. It is currently developed so that is not needed.

xitongzou: `It is currently developed so that is not needed.` I don't understand what is meant by this. Currently the way it is, having any link or button inside a row with a row handler will also trigger the row handler, so we need to handle that case to prevent the row handler from firing the row event. That's the whole point of the example right? But there's nothing in KTable that would inherently prevent this, so we need to deal with this in the implementation.

travispamaral: this code [here](https://github.com/Kong/kongponents/pull/237/files#diff-158ff5639cfcb0ec94ce4226ff60c18dR204) is supposed to not fire the TD click. So a link or a button _inside_ of a row should not fire the event. @darrenjennings helped write this code before this ticket was started

xitongzou: I don't think that should be inside `KTable` it should be implementation specific because we don't know what tags or listeners the end user will put in. `TD` and `mouseup` are assumptions based on something that should be implementation specific. In the example we would have to add an additional check for `A` or `BUTTON` since `TD` is not the target element in those cases, and an additional `click` listener to pluck. We can discuss with @darrenjennings but my feeling is once it gets this complex and specific then it should be left to the implementation level.

travispamaral: The click target is always TD unless it’s a link or a button. So that’s why it’s implemented in the table. If you click a link or a button it should not trigger the row event. The only implementation addition should be meta key if the user wants. This already was discussed and implemented

xitongzou: I see, I didn't have the entire background on the implementation of `trlisteners` but it seems like it does work the way you described once I changed `mouseup` to `click` because in the examples we are using click listeners, not mouseup. I'll revise the examples.

docs/components/table.md
```diff
@@ -85,25 +85,247 @@ Lessen the table cell padding
...
+Adds `cursor: pointer` and `user-select: none` styling. 
+
+<KTable :options="$frontmatter.tableOptions" isClickable />
+```vue
+<KTable
+  :options="tableOptions"
+  isClickable />
+```
+
+### Custom Row Attributes
```
darrenjennings, line 98:
```suggestion
```

docs/components/table.md
```diff
@@ -209,6 +209,82 @@ various parts of the table.
 <KTable @row:click="rowHandler" @row:dblclick="rowHandler">
 ```
 
+- The rows can be clicked and metakeys are also supported in the row click handler. Hyperlinks and Buttons can also be clicked separately, as in the below example:
```
travispamaral, line 212:
```suggestion
On top of the individual row being clickable, we support being able to click elements inside a row, like buttons or hyperlinks without triggering the row event. You can also add logic to check for metakey to support cmd/ctrl clicking a row.
```

docs/components/table.md
```diff
@@ -209,6 +209,82 @@ various parts of the table.
...
+export default {
+  data() {
+    return {
+      row: null,
+      eventType: '',
+      tableOptions: {
+        headers: [
+          { label: 'Company', key: 'company' }
+        ],
+        data: [
```
travispamaral, line 247: this data doesn't look like anything we'd pass in, can you update with a more common set of items?

docs/components/table.md
```diff
@@ -371,6 +447,111 @@ export default {
...
+  }
+
+  tr.disabled {
+    --KTableHover: var(--yellow-100, #fff9e6);
+    --KTableBorder: var(--yellow-200, #ffdc73);
+  }
+}
+</style>
+```
+
```
travispamaral, line 554:
```suggestion
```

packages/KTable/KTable.vue
```diff
@@ -200,9 +200,9 @@ export default {
 
         return {
           ...pluckedListeners,
-          mouseup (e) {
-            if (e.target.tagName === 'TD' && pluckedListeners['mouseup']) {
-              pluckedListeners['mouseup'](e, entity, type)
+          click (e) {
```
travispamaral, line 203: double checked this works on mobile 👍

docs/components/table.md
xitongzou: I thought this was just meant to be an example. What is a more 'common' set of items? we just need an example with a set of data with clickable links right?

PR #240, opened 2020-07-13
Style(KButton) Adds isOpen prop for dropdown caret 

travispamaral:
### Summary
More and more designs are including a button with a dropdown caret similar. Instead of adding one off instances I added a utility class to KButton. I also updated KMultiselect to use the caret 💄

#### Changes made:
* Add KButton classes
* Use class in KMultiselect
* Add example of class to button docs

### PR Checklist

### DIFF
--- a/docs/components/button.md
+++ b/docs/components/button.md
@@ -121,6 +121,24 @@ KButton supports using an icon either before the text or without text.
 ```
 
 ## Theming
+KButton includes a utility class `.has-caret` which displays a dropdown caret to the right hand side. this is useful for buttons that control dropdowns and popovers. You can rotate the caret (active state) by applying `.is-active` as an additional class.
+
+<Komponent :data="{ isActive: false}" v-slot="{ data }">
+  <KButton appearance="primary" class="has-caret" :class="{'is-active': data.isActive}" @click="data.isActive = !data.isActive">I'm a button</KButton> 
+</Komponent>
+
+> The `Komponent` component is used in this example to create state.
+```vue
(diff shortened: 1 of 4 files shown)


docs/components/button.md
```diff
@@ -121,6 +121,24 @@ KButton supports using an icon either before the text or without text.
 ```
 
 ## Theming
+KButton includes a utility class `.has-caret` which displays a dropdown caret to the right hand side. this is useful for buttons that control dropdowns and popovers. You can rotate the caret (active state) by applying `.is-active` as an additional class.
```
darrenjennings, line 124: Needing to know about a class is an implementation detail that we should obscure from users as you can't use the class outside k-button. I am thinking it would be easier to use a single prop `:isOpen` which would then toggle the `.has-caret` and `.is-active` classes. The absence of the prop would be "no caret".

travispamaral: how about `hasCaret` prop? How would `is-open`/`is-active` be set then? Now you need two props instead of two classes. I thought about going the prop route but you still need a way to toggle the state

darrenjennings: I was thinking something like (pseudo code):
```js
let classes = ""
if (isOpen !== undefined) {
  classes = isOpen ? ".has-caret .is-active" : ".has-caret"
}
```

travispamaral: Eh this seems like added complexity for a caret. I'd rather add the utility class to the general css (truncate, cursor, etc) instead

PR #250, opened 2020-09-01
feat(KInlineedit) Adds props for overriding styles & ignore value

travispamaral:
### Summary
Blocks https://konghq.atlassian.net/browse/KHCP-331

- Adds prop for ignoring initial value when clicking to edit
- Adds prop for adding custom styles to the input which will override those inherited from the element when clicking to edit.

In Konnect we implemented this component to display slightly differently when there was no value. This became a problem when you would click to edit for the first time, the styling from the initial element would transfer in. We also want to display text when empty but NOT pass that text in on click (similar to input placeholder).   

#### Changes made:
* Adds props
* Adds tests
* Updates snapshots
* Adds doc

### PR Checklist

### DIFF
--- a/docs/components/inline-edit.md
+++ b/docs/components/inline-edit.md
@@ -17,7 +17,49 @@
 ```
 
 ## Props
-- `@changed` - Emitted blurred away from the editing input or when enter is pressed.
+### ignoreValue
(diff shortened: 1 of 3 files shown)


travispamaral:
haha yeah I thought about `ignoreInitialValue`? But its not necessarily initial only. It really should just be a flag to say when to ignore so I left

PR #253, opened 2020-09-02
feat(KIcon) add fileEmpty icon and default fill to file icons

nateslo:
### Summary
Add default fill to file svgs so they can be overwritten with `color` prop. Also adds missing newline at the end of the svgs

#### Changes made:
* add default fill to file icons
* add missing newline to end of file svgs


travispamaral (approved):
:shipit:

PR #258, opened 2020-10-06
feat(kpop): kpop adds an optional callback function when popover is already opened

xitongzou:
### Summary
Adding a callback function prop that triggers when the popover has already been opened

#### Changes made:
*Adding a callback function prop that triggers when the popover has already been opened
*Rename isShow to isOpen

### PR Checklist

### DIFF
--- a/docs/components/popover.md
+++ b/docs/components/popover.md
@@ -264,12 +264,35 @@ You can pass in an optional flag to not show the caret on the edge of the popove
 
 ### onPopoverClick
 You can pass in an optional callback function to trigger when the popup is already open and the trigger method is click.
+The callback function can optionally return a boolean, which will show or hide the popup depending on the value of the boolean.
+
+<KPop title="Cool header" :on-popover-click="toggle">
+  <KButton>button</KButton>
+  <div slot="content">
+  The first time you click the button, I will close when you click here once.
+  The second time you click the button, I will close when you click here twice.</div>
+</KPop>
 
 ```vue
-<KPop title="Cool header" on-already-open="onPopoverClick()">
+<KPop title="Cool header" :on-popover-click="toggle">
   <KButton>button</KButton>
   <div slot="content">Click me and you will get an alert if I am already opened!</div>
 </KPop>
+<script>
+  export default {
+    data () {
+      return {
+        isToggled: true
+      }
+    },
+    methods: {
+      toggle () {
+        this.isToggled = !this.isToggled
+        return this.isToggled
+      }
+    }
+  }
+</script>
 ```
 
 ### isSVG Flag
@@ -377,6 +400,7 @@ This is the slot that takes in the content of the popover.
           'pending': 'idle'
         },
         count: 0,
+        isToggled: true,
         timeout: null
       }
     },
@@ -387,7 +411,6 @@ This is the slot that takes in the content of the popover.
           'idle': 'Load something'
         }[this.currentState]
       },
-      
       message () {
         return {
           'pending': `Loading ${this.count}...`,
@@ -403,14 +426,16 @@ This is the slot that takes in the content of the popover.
           this.transition()
         }, 2000)
       },
-      
+      toggle () {
+        this.isToggled = !this.isToggled
+        return this.isToggled
+      },
       onClose () {
         clearTimeout(this.timeout)
         if (this.currentState == 'pending') {
           this.transition()
         }
       },
-      
       transition() {
         this.currentState = this.states[this.currentState]
       }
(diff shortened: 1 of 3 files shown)


docs/components/popover.md
```diff
@@ -330,11 +330,10 @@ This is the slot that takes in the content of the popover.
...
 ### Events / Loading Content
 
-- `opened` - emitted once the popover has been opened
-- `closed` - emitted when the popover has been triggered closed (emits on all
-triggers)
+- `opened` - emitted once the popover has been opened, also passes in whether or not the popover is currently being displayed
+- `closed` - emitted when the popover has been triggered closed (emits on all triggers)
 
-<KPop @opened="loadSomething" @closed="onClose">
+<KPop @opened="(isShow) => loadSomething(isShow)" @closed="onClose">
```
dimitropoulos, line 336: is `isShow` a common Vue name?  I've seen `isOpen`/`open` and `isVisible`/`visible` used a lot more.  that's not even a nit (we can call it whatever you think is best), but rather just a curious question.

xitongzou: right, maybe I should use `isOpen`

dimitropoulos: I'm completely cool with whatever, was just curious :smile:

xitongzou:
@dimitropoulos decided to revise it so that we can pass in a custom callback prop which is much more versatile for our needs

docs/components/popover.md
```diff
@@ -262,6 +262,17 @@ You can pass in an optional flag to not show the caret on the edge of the popove
 </KPop>
 ```
 
+### onAlreadyOpen
+You can pass in an optional callback function to trigger when the popup is already open (usually when the trigger method is by clicking).
+This can be useful in situations where for example, you need to close the popup.
+
+```vue
+<KPop title="Cool header" on-already-open="closePopup()">
```
darrenjennings, line 270: what is `closePopup`? It seems like we are making the user think too much about implementation. Why not have a prop called `is-open` that is reactive and then the user can decide when they want the popover opened or not? This is one of the problems I've had with vue-autosuggest, so we implemented `should-render-suggestions` which allows users to hook into the state of the autosuggest and close it based on whatever criteria they select, but they can also see how `vue-autosuggest` decides to close it. https://github.com/darrenjennings/vue-autosuggest#shouldrendersuggestions 
I think users generally prefer a reactive `is-open` prop because it's straightforward.

xitongzou: I don't have the actual implementation for `closePopup` the same reason why I don't have it for `zoom` in the docs is because its left up to the user, an example component might be too complex for the docs.

The reason why it's a callback and not a boolean is explained here: https://github.com/Kong/khcp/pull/561#issuecomment-704757189

darrenjennings: yea but as the docs currently stand, it's not apparent how a user can control the open/close state of kpop. i.e. closePopup should not be a user implemented function.

xitongzou: what do you recommend then? `closePopup should not be a user implemented function.`  I think you are misunderstanding the use of this prop. Because this prop isn't meant to be a close popup function all the time, it's a custom callback function. Using `hidePopper` is just one way the user can do it, but they can choose to do something else when the popup is already opened, like toggling something. I can put in a simple example.

darrenjennings: the el.hidePopper is the part that is what I want to avoid. `kpopRef.hidePopper` is an escape hatch that is subject to change based on the implementation of kpop.

xitongzou: yeah, but this prop doesn't have to be used for closing the popup. That's just how we choose to use it in the Dropdown. Since Dropdown has no visibility into the click handler implementation of Kpop, I dont see a better way of doing it than using a callback right now. I'm on all ears for a better solution but I spent 3 hours up to midnight last night racking my brain on how to do this and this is was the best solution I could find that was as general of a fix as possible for our dropdowns.

docs/components/popover.md
```diff
@@ -262,6 +262,39 @@ You can pass in an optional flag to not show the caret on the edge of the popove
...
+  <KButton>button</KButton>
+  <div slot="content">
+  The first time you click the button, I will close when you click here once.
+  The second time you click the button, I will close when you click here twice.</div>
+</KPop>
+
+```vue
+<KPop title="Cool header" :on-popover-click="toggle">
+  <KButton>button</KButton>
+  <div slot="content">Click me and you will get an alert if I am already opened!</div>
```
darrenjennings, line 279: this is out of sync now

darrenjennings (approved):
lgtm! (besides the small docs example being out of sync)

PR #259, opened 2020-10-08
refactor(kpop) on-popover-click return value

darrenjennings:
@xitongzou check this out and lemme know what you think. Adding as part of work for dropdown

### DIFF
--- a/packages/KPop/KPop.vue
+++ b/packages/KPop/KPop.vue
@@ -281,7 +281,10 @@ export default {
           this.createInstance()
         }
       } else if (this.$refs.popper && this.$refs.popper.contains(e.target) && this.onPopoverClick) {
-        this.onPopoverClick()
+        const shouldShow = !this.onPopoverClick()
+        if (shouldShow !== undefined) {
+          shouldShow ? this.showPopper() : this.hidePopper()
+        }
       } else if (this.$refs.popper && this.$refs.popper.contains(e.target)) {
         this.showPopper()
       } else if (this.isOpen) { this.hidePopper() }
--- a/packages/KPop/KPop.vue
+++ b/packages/KPop/KPop.vue
@@ -281,9 +281,9 @@ export default {
           this.createInstance()
         }
       } else if (this.$refs.popper && this.$refs.popper.contains(e.target) && this.onPopoverClick) {
-        const shouldShow = !this.onPopoverClick()
-        if (shouldShow !== undefined) {
-          shouldShow ? this.showPopper() : this.hidePopper()
+        const isOpen = this.onPopoverClick()
+        if (isOpen !== undefined) {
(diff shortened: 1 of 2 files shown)


packages/KPop/KPop.vue
```diff
@@ -281,7 +281,10 @@ export default {
           this.createInstance()
         }
       } else if (this.$refs.popper && this.$refs.popper.contains(e.target) && this.onPopoverClick) {
-        this.onPopoverClick()
+        const shouldShow = !this.onPopoverClick()
```
xitongzou, line 284: I dont understand this line. So is `onPopoverClick()` assumed to return a boolean? cause we always know it's defined at this section of the code.

packages/KPop/KPop.vue
```diff
@@ -281,7 +281,10 @@ export default {
...
+        if (shouldShow !== undefined) {
+          shouldShow ? this.showPopper() : this.hidePopper()
```
xitongzou, line 286: when would we want to keep it shown? also I guess we are turning this prop into one that will always control hiding and showing, which is a little less versatile I think, like there are cases where you might want to pass a callback function on click but not want to do anything with showing or hiding

xitongzou: Like its useful for a dropdown implementation but we are assuming the user will always have a dropdown implementation for using this.. what if they just wanted to execute a custom callback whenever the popover is already opened? And their implementation isn't a dropdown? maybe its a checkbox or multi select or something.

darrenjennings: > we are turning this prop into one that will always control hiding and showing

Updated it slightly didn't realize I was accidentally checking the falsity of the callback's return. It will not always control hiding/showing. It works the exact same as your current PR except that if the callback provided returns a boolean then it will decide to hide or show. If the callback returns `void`, then it will not do anything. It allows the implementation to stay minimal without needing to know about popper or refs. I can help update some documentation if you agree this is preferable.

>  what if they just wanted to execute a custom callback whenever the popover is already opened? And their implementation isn't a dropdown? maybe its a checkbox or multi select or something.

The callback still always fires, just its return type is now meaningful.

packages/KPop/KPop.vue
darrenjennings: Here's how the typescript prop would look
```ts
onPopoverClick: {
  type: Function as PropType<() => boolean|undefined>,
  required: false,
  default: () => undefined
}
```

packages/KPop/KPop.vue
xitongzou: yeah i think having a few examples in the docs will help, in this case there's two return types that do different things. returning a boolean triggers a show or hide and returning a void just calls the callback

xitongzou (approved):

PR #261, opened 2020-10-09
feat(kicon): force re-publish

darrenjennings:
forces re-publish of kicon. I will deprecate the version prior to this once this build successfully. I will also bump major version.

### DIFF
--- a/packages/KIcon/package.json
+++ b/packages/KIcon/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@kongponents/kicon",
   "version": "0.5.4",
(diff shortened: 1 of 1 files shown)


packages/KIcon/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@kongponents/kicon",
   "version": "0.5.4",
-  "description": "Icon component",
+  "description": "Icon component.",
```
xitongzou, line 4: the dreaded DOT

xitongzou (approved):

PR #271, opened 2020-12-01
style(kbadge) auto width by default, add vars, truncate

darrenjennings:
### Summary

Add width "auto" + sane min/max widths by default. Add css vars support for overriding, and truncate with ellipsis.

### PR Checklist

Demo

image

### DIFF
--- a/packages/KBadge/KBadge.vue
+++ b/packages/KBadge/KBadge.vue
@@ -2,7 +2,7 @@
   <div
     :class="[appearance !== 'custom' && `kbadge-${appearance}`]"
     :style="color && backgroundColor && {backgroundColor, color}"
-    class="k-badge"
+    class="k-badge truncate"
   >
     <slot />
   </div>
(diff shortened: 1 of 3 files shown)


packages/KBadge/KBadge.vue
```diff
@@ -49,14 +49,17 @@ export default {
 @import '~@kongponents/styles/_variables.scss';
 
 .k-badge {
-  display: inline-flex;
+  display: inline-block;
```
travispamaral, line 52: You can remove justification as it won't work in block

packages/KBadge/KBadge.vue
```diff
@@ -49,14 +49,17 @@ export default {
...
   font-weight: 500;
+  vertical-align: top;
   font-size: var(--KBadgeFontSize, var(--type-sm, type(sm)));
-  height: 27px;
-  align-items: center;
-  width: var(--KBadgeWidth, 71px);
-  padding: var(--KBadgePaddingY, 2px) var(--KBadgePaddingX, 4px);
+  height: auto;
+  text-align: center;
+  min-width: var(--KBadgeMinWidth, 71px);
```
travispamaral, line 59: So we even need a min width/height? I feel it should be set automatically by padding. And then max width to keep it from growing huge

packages/KBadge/KBadge.vue
darrenjennings: oh thought I removed that thanks!

packages/KBadge/KBadge.vue
darrenjennings: there is no min height (heigh auto could possibly be removed?). I added a min-width so that the immunity alert badge styles wouldn't regress.

travispamaral (approved):
🚢

PR #272, opened 2020-12-01
chore(style/checkbox) force publish

darrenjennings:
NPM token was invalidated so need to push a trivial change to force publish styles up to the registry.


xitongzou:
Did this bump the npm package properly? https://www.npmjs.com/package/@kongponents/styles I still see `2.0.0` here

darrenjennings:
@xitongzou no we need to make a change to npm packages to accept automation tokens

PR #298, opened 2021-04-20
fix(KCheckbox): wrap label slot in span

dustinryerson:
### Summary
Wrap the whole KCheckbox slot in a span tag, rather than just the `label` prop

#### Changes made:
* Wraps the whole KCheckbox slot in a span tag, rather than just the label prop. https://github.com/Kong/kongponents/pull/296 works when the `label` prop is used, but if the default slot is used instead, the span is overridden. Wrapping the slot will ensure it's not overridden, and will still work if the `label` prop is used.

### PR Checklist


kaiarrowood (changes requested):
I was looking at this the first time and went in and had wondered about that. I think we should update the docs to have a visible example similar to how we do it for the KCard component. Let's put some actual custom HTML in the slot for an example.

image

dustinryerson:
@kaiarrowood I've added the requested example to the docs:

Screen Shot 2021-04-28 at 10 30 34 AM

kaiarrowood:
nonblocking: While you're in this component, can we update the TOC to put this in the right place alphabetically? This has been driving me nuts for a long time.

image

Code lives in the `.vupress` config file.

image

Otherwise this looks awesome!

dustinryerson:
> nonblocking: While you're in this component, can we update the TOC to put this in the right place alphabetically? This has been driving me nuts for a long time.

Definitely, fixed the ordering of `Icon` too.

Screenshots:

**Before**
Screen Shot 2021-05-04 at 3 35 32 PM

**After**
Screen Shot 2021-05-04 at 3 33 26 PM

kaiarrowood (approved):
This is good to go! Approved!

PR #301, opened 2021-04-29
feat(KTable): add cellAttrs prop for customization

dustinryerson:
### Summary
Add `cellAttrs` prop for greater table customization

This allows customization at a much more granular level. Cells, columns and rows can be customized via this new prop.

Here's an example I've added to the docs. This table has been customized in the following ways:
1. Two columns of cells,`Name` and `Description`, have been given different `max-width` values.
2. Two columns of cells, `Name` and `Description`, have been given the `truncate` class.
3. The cells in the last row have been given different background colors based on data in the row (in this case, based on Google being the value in the `Company` column)

Screen Shot 2021-04-29 at 12 56 28 PM

#### Changes made:
* Add `cellAttrs` prop to KTable component
* Update docs to include details about new `cellAttrs` prop

### PR Checklist

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -282,19 +282,6 @@ export default {
       }
 
       /**
-       * Sets cell max-width based on the column headerKey
-       */
-      const colWidth = () => {
-        if (headerKey === 'description') {
-          return '50ch'
-        } else if (headerKey === 'name') {
-          return '22ch'
-        }
-
-        return '25ch'
-      }
-
-      /**
        * Returns an object of attributes to be applied to cells
        */
       return {
@@ -303,7 +290,7 @@ export default {
         },
         'datatest-id': `row-${rowIndex + 1}-col-${headerKey}`,
         style: {
-          'maxWidth': colWidth(),
+          'maxWidth': headerKey==='description' ? '50ch' : headerKey === 'name' ? '22ch' : '25ch',
           'backgroundColor': backgroundColor(),
         },
       }
@@ -879,19 +866,6 @@ export default {
       }
 
       /**
-       * Sets cell max-width based on the column headerKey
-       */
-      const colWidth = () => {
-        if (headerKey === 'description') {
-          return '50ch'
-        } else if (headerKey === 'name') {
-          return '22ch'
-        }
-
-        return '25ch'
-      }
-
-      /**
        * Returns an object of attributes to be applied to cells
        */
       return {
@@ -900,7 +874,7 @@ export default {
         },
         'datatest-id': `row-${rowIndex + 1}-col-${headerKey}`,
         style: {
-          'maxWidth': colWidth(),
+          'maxWidth': headerKey==='description' ? '50ch' : headerKey === 'name' ? '22ch' : '25ch',
           'backgroundColor': backgroundColor(),
         },
       }
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -199,11 +199,9 @@ export default {
 
 Add custom properties to individual table cells or groups of cells. The cell attributes object is passed as a param.
 
-### `cellAttrs` Description
+`cellAttrs` - A function that returns an object comprising the attributes.
 
-A function that returns an object comprising the attributes.
-
-### `cellAttrs` Function Parameters
+**Parameters**
 
 | Parameter | Description
 |:-------- |:-------
@@ -304,7 +302,7 @@ export default {
           'truncate': headerKey === 'description' || headerKey === 'name',
         },
         'datatest-id': `row-${rowIndex + 1}-col-${headerKey}`,
-        style : {
+        style: {
           'maxWidth': colWidth(),
           'backgroundColor': backgroundColor(),
         },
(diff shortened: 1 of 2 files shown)


docs/components/table.md
```diff
@@ -195,6 +195,125 @@ export default {
...
+
+      /**
+       * Returns an object of attributes to be applied to cells
+       */
+      return {
+        class: {
+          'truncate': headerKey === 'description' || headerKey === 'name',
+        },
+        'datatest-id': `row-${rowIndex + 1}-col-${headerKey}`,
+        style : {
```
kaiarrowood, line 307:
```suggestion
        style: {
```

Super tiny nit: remove unnecessary space.

docs/components/table.md
```diff
@@ -712,7 +860,53 @@ export default {
...
```
kaiarrowood, line 904:
```suggestion
        style: {
```
nit: space

docs/components/table.md
```diff
@@ -195,6 +195,125 @@ export default {
 </style>
 ```
 
+## Cell Attributes
+
+Add custom properties to individual table cells or groups of cells. The cell attributes object is passed as a param.
+
+### `cellAttrs` Description
```
kaiarrowood, line 202: suggestion: The heading style looks nice but the TOC entries don't, do we want to keep it like this?

image

I'm thinking follow this style for the Function Description section, we don't really need a TOC entry for it.

image

For the parameters, maybe we just make the heading `Parameters`? Or at the lease `cellAttrs Parameters` and remove `Function` so it's a bit less wordy.

What do you think?

dustinryerson: Totally agree, I've made these changes in the latest commit.

docs/components/table.md
dustinryerson: Fixed in latest commit along with other requested changes, thanks for catching it!

docs/components/table.md
dustinryerson: Fixed in latest commit along with other requested changes, thanks for catching it!

docs/components/table.md
dustinryerson: **New README appearance**
Screen Shot 2021-05-04 at 2 37 49 PM

**TOC**
Screen Shot 2021-05-04 at 2 38 07 PM

docs/components/table.md
```diff
@@ -712,7 +858,53 @@ export default {
...
+       * Returns an object of attributes to be applied to cells
+       */
+      return {
+        class: {
+          'truncate': headerKey === 'description' || headerKey === 'name',
+        },
+        'datatest-id': `row-${rowIndex + 1}-col-${headerKey}`,
+        style: {
+          'maxWidth': colWidth(),
+          'backgroundColor': backgroundColor(),
```
jackkav, line 904: I think this could be difficult to read as an example as it creates unnecessary intermediary functions, and else condition with early return.
```
style:{
  'maxWidth': headerKey==='description' ? '50ch' : headerKey === 'name' ? '22ch' : '25ch'
}
```
This approach has less potential for side effect from a change in conditions or use of variables out of scope.

Aside from that, KTable traditionally used slots and classes, which I feel discourages css in js. I feel like this might not be the most idiomatic vue but not a blocking concern. Just my two cents.

jackkav (commented):
Would like to see some prior art from another component library. In the past segmentio's evergreen has been an inspiration for component API decisions. https://evergreen.segment.com/components/table
In this case I didn't find anything, but I'm curious if this change is based on similar inspiration?

docs/components/table.md
kaiarrowood: Looking good!

docs/components/table.md
dustinryerson: > Would like to see some prior art from another component library. In the past segmentio's evergreen has been an inspiration for component API decisions. https://evergreen.segment.com/components/table
> In this case I didn't find anything, but I'm curious if this change is based on similar inspiration?

@jackkav thanks for your review and feedback. I've removed the unnecessary intermediary function from `maxWidth` in the example. I made this change based on the `rowAttributes` KTable prop. In our use case, we believe adding the ability to customize the `max-width` value for columns will allow us to utilize more of the currently empty space in our data tables. I felt that passing information about the row and column would enable developers to customize with more specificity. As an optional prop, it's there for developers to leverage, but can simply be ignored if they don't need it.

As for examples from another component library, these are the closest examples I could find from Segment:
1. https://github.com/segmentio/evergreen/blob/master/src/table/src/TextTableCell.js#L29
2. https://github.com/segmentio/evergreen/blob/master/src/typography/src/Text.js#L32-L36
3. https://github.com/segmentio/evergreen/blob/master/src/table/src/TableCell.js#L106

Others (these are just similar examples, not identical use cases):
1. Microsoft Fluent UI: https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist/customitemcolumns.
2. Elastic EUI: https://elastic.github.io/eui/#/tabular-content/data-grid-styling-and-control

kaiarrowood (approved):
This looks good to me. Verified behaving as expected in docs. 🐉 🎇

PR #306, opened 2021-05-25
feat(klabel): add help property with icon + tooltip

mheap:
### Summary
Some inputs require additional context to explain what the user should provide. This feature uses `KIcon` to add a help icon and `KoolTip` to show a tooltip if the `help` property is provided.

`KoolTip` uses the prop `label` to provide the text. I have chosen `help` in this component so that there is no confusion between the `KLabel` slot and the `KoolTip` prop.

[Demo](https://deploy-preview-306--kongponents.netlify.app/components/label.html#with-help)

#### Changes made:
* Add `help` property to `KLabel`
* Add tests for `KLabel`

### PR Checklist

### DIFF
--- a/docs/components/label.md
+++ b/docs/components/label.md
@@ -1,0 +1,57 @@
+# Label
+
+**KLabel** provides a wrapper around general `label` tags.
+
+## Standard Input
+
+<br />
+
+<KLabel>Input Title</KLabel>
+
+```vue
+<KLabel>Input Title</KLabel>
+```
+
+## With Help
+
+<br />
+
+<KLabel help="This is an example">Input Title</KLabel>
+
+```vue
+<KLabel help="This is an example">Input Title</KLabel>
+```
+
+<br />
+
+<KLabel help="This is a really long tooltip. Hopefully we won't have anything this long but we might. I wonder how it handles long inputs">Long Input Title</KLabel>
+
+```vue
+<KLabel
+  help="This is a really long tooltip. Hopefully we won't have anything this long but we might. I wonder how it handles long inputs"
+>
+  Long Input Title
+</KLabel>
+```
+
+## With Info
+
+<br />
+
+<KLabel info="This is an example">Input Title</KLabel>
+
+```vue
+<KLabel info="This is an example">Input Title</KLabel>
+```
+
+## Sample input with a tooltip
+
+<br />
+
+<KLabel help="A service is an API that you want to offer">Service Name</KLabel>
+<KInput />
+
+```vue
+<KLabel help="A service is an API that you want to offer">Service Name</KLabel>
+<KInput />
+```
--- a/packages/KLabel/KLabel.vue
+++ b/packages/KLabel/KLabel.vue
@@ -23,7 +23,6 @@
         icon="info"
         height="16"
         width="16"
-        color="black"
         hide-title />
     </Kooltip>
     <slot v-else />
--- a/packages/KLabel/KLabel.vue
+++ b/packages/KLabel/KLabel.vue
@@ -1,12 +1,48 @@
 <template functional>
   <label class="k-input-label">
-    <slot />
+    <KoolTip
+      v-if="props.help"
+      :label="props.help"
+      class="label-tooltip"
+    >
+      <slot />
+      <KIcon
+        icon="help"
+        height="16"
+        width="16"
+        view-box="0 0 16 16"
+      />
+    </Kooltip>
+    <KoolTip
+      v-else-if="props.info"
+      :label="props.info"
+      class="label-tooltip"
+    >
+      <slot />
+      <KIcon
+        icon="info"
+        height="16"
+        width="16"
+        color="black"
+      />
+    </Kooltip>
+    <slot v-else />
   </label>
 </template>
 
 <script>
 export default {
-  name: 'KLabel'
+  name: 'KLabel',
+  props: {
+    help: {
+      type: String,
+      default: undefined
+    },
+    info: {
+      type: String,
+      default: undefined
+    }
+  }
 }
 </script>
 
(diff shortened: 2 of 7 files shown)


packages/KLabel/__snapshots__/KLabel.spec.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`KLabel renders a tooltip when \`help\` is provided 1`] = `<label class="k-input-label" help="This is a tooltip">Full Name</label>`;
```
mheap, line 3: This snapshot doesn't provide as much confidence as I'd like, but the current version of `@vue/test-utils` does not support stubs. Upgrading to `v1.2.0` allows for a much more complete snapshot with `stubs` like so:

```
   <label class="k-input-label">
      <kooltip-stub label="This is a tooltip">Full Name <kicon-stub icon="help" height="16" width="16"></kicon-stub>
      </kooltip-stub>
    </label>
```

However, upgrading to 1.2.0 broke a lot of existing tests that I do not have the context to fix at this time

mheap:
@dustinryerson @kaiarrowood Anything I can do to keep this moving?

packages/KLabel/KLabel.vue
```diff
@@ -1,12 +1,27 @@
...
+    <slot v-else />
   </label>
 </template>
 
 <script>
 export default {
-  name: 'KLabel'
+  name: 'KLabel',
+  props: {
+    help: {
```
kaiarrowood, line 20: Can we also add the ability to specify an `info` prop that uses the info icon?

docs/components/label.md
```diff
@@ -0,0 +1,42 @@
...
+>
+  Long Input Title
+</KLabel>
+```
+
+## Sample with an input + tooltip
+
+<br />
+
+<KLabel tooltip="A service is an API that you want to offer">Service Name</KLabel>
```
kaiarrowood, line 41: Not seeing any handling in `KLabel` for a `tooltip` prop. Please also include a `vue` code section so the user can see the code.

image

docs/components/label.md
```diff
@@ -0,0 +1,42 @@
...
+
+```vue
+<KLabel
+  help="This is a really long tooltip. Hopefully we won't have anything this long but we might. I wonder how it handles long inputs"
+>
+  Long Input Title
+</KLabel>
+```
+
+## Sample with an input + tooltip
```
kaiarrowood, line 37:
```suggestion
## Sample input with a tooltip
```

docs/components/label.md
```diff
@@ -0,0 +1,42 @@
+# Label
+
+**KLabel** provides a wrapper around general `label` tags
```
kaiarrowood, line 3:
```suggestion
**KLabel** provides a wrapper around general `label` tags.
```

kaiarrowood (changes requested):

docs/components/label.md
dustinryerson, line 41:
```suggestion
<KLabel help="A service is an API that you want to offer">Service Name</KLabel>
```

dustinryerson, line 41: When I used `help` locally and hovered over the icon for +2 seconds, I saw 2 tooltips - one being the intended tooltip and the other "help" coming from somewhere else. I think we need to find and suppress the unintended tooltip:

Screen Shot 2021-06-02 at 8 03 51 PM

dustinryerson (changes requested):
@mheap I like this idea! Thanks for taking the time to work on it. Apologies for the delayed review.

I left a suggested change and a comment - please let me know if I can clarify further or if I can help with anything else. I think if all looks good once the comments/requested changes are addressed we should be good to go 👍 

Thanks!

docs/components/label.md
mheap: Good catch, the prop was renamed to `help`

docs/components/label.md
kaiarrowood: Please use correct punctuation in the documentation.

kaiarrowood (changes requested):

packages/KLabel/KLabel.vue
mheap: Great idea. I've added this

docs/components/label.md
mheap: Good catch! This was related to the `<title>` property in `KIcon`. I've added an attribute to disable it (default to `false` to keep backwards compatibility)

packages/KLabel/KLabel.vue
```diff
@@ -1,12 +1,48 @@
...
+      v-else-if="props.info"
+      :label="props.info"
+      class="label-tooltip"
+    >
+      <slot />
+      <KIcon
+        icon="info"
+        height="16"
+        width="16"
+        color="black"
```
kaiarrowood, line 26: This is looking great @mheap , is there a reason why we are making this one explicitly black?

kaiarrowood:
Looks good @mheap , let mr know about the `black` coloring and then we can go ahead and get this merged in. 👍🏻

packages/KLabel/KLabel.vue
mheap: @kaiarrowood Based on the intended context (KHCP) black is all that's needed. I'm happy to make this a prop with a default if you think it would be useful

mheap:
Just added a missing test for the `info` prop

packages/KLabel/KLabel.vue
kaiarrowood, line 26:
```suggestion
```

packages/KLabel/KLabel.vue
```diff
@@ -1,12 +1,47 @@
...
+    <KoolTip
+      v-else-if="props.info"
+      :label="props.info"
+      class="label-tooltip"
+    >
+      <slot />
+      <KIcon
+        icon="info"
+        height="16"
+        width="16"
```
kaiarrowood, line 25:
```suggestion
        width="16"
        color="#A3BBCC"
```

PR #307, opened 2021-06-04
feat(KSwitch) add optional tooltip when disabled

kaiarrowood:
### Summary
Add the ability to display a tooltip on hover if `KSwitch` is `disabled`.

#### Changes made:
* Add new property `disabledTooltipText`
* Added separate section in docs for `disabled` property

image

### PR Checklist

### DIFF
--- a/docs/components/switch.md
+++ b/docs/components/switch.md
@@ -41,7 +41,7 @@ state of toggle switch. You can read more about passing values via `v-model`
 
 ### label
 
-Will place label text to the right of the switch. Can also be [slotted](#slots). You can also add `disabled` to the input to disallow interactivity. 
+Will place label text to the right of the switch. Can also be [slotted](#slots).  
 
 - `label`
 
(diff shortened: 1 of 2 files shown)


docs/components/switch.md
```diff
@@ -51,12 +51,30 @@ Will place label text to the right of the switch. Can also be [slotted](#slots).
...
 <KInputSwitch v-model="labelPropChecked" label="disabled" disabled />
 
+### disabledTooltipText
+
+You can specify tooltip text to be displayed when the switch is disabled.
+
+- `disabledTooltipText`
+
+```vue
+<KInputSwitch v-model="checked" label="disabled" disabled disabledTooltipText="I'm disabled!" />
```
dustinryerson, line 73: When I viewed this code example on the page, it gets cutoff - could we format it so that it's on a few lines like this:

```
<KInputSwitch
  v-model="checked"
  label="disabled"
  disabled
  disabledTooltipText="I'm disabled!"
/>
```

Screen Shot 2021-06-04 at 4 14 51 PM

dustinryerson (approved):
Left a non-blocking comment about code example formatting.

One question I had is should we also provide the ability to change the positioning of the tooltip? That could provide more flexibility since we don't know where the switch will be positioned, or what will be present around it.

docs/components/switch.md
kaiarrowood: Good catch! I didn't even notice.

Fixed.

image

PR #310, opened 2021-06-25
feat(KModal): add support for help text

kaiarrowood:
### Summary
Allow users to specify help text to display under the modal title.

#### Changes made:
* Add new slot `help` and property `helpText`

image

### PR Checklist

### DIFF
--- a/packages/KModal/KModal.vue
+++ b/packages/KModal/KModal.vue
@@ -12,7 +12,10 @@
         class="modal-dialog"
         @click.stop>
         <div class="modal-content">
-          <div class="modal-header">
+          <div
+            :class="helpText ? 'mb-3' : 'mb-4'"
+            class="modal-header"
+          >
             <slot name="header-content">{{ title }}</slot>
           </div>
           <div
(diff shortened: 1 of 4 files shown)


packages/KModal/KModal.vue
```diff
@@ -15,6 +15,11 @@
           <div class="modal-header">
```
dustinryerson, line 15:
```suggestion
          <div class="modal-header" :class="helpText ? 'mb-3' : 'mb-4'">
```

Our use case has a `12px` bottom margin here, so I'd suggest adding this conditional bottom margin and removing the styling in the style tags:
```
margin-bottom: var(--spacing-md, spacing(md));
```

dustinryerson (approved):
Looks good! Added one suggestion (will reapprove if you take it).

packages/KModal/KModal.vue
kaiarrowood: @dustinryerson Having trouble visualizing what you mean here. May be better to chat on zoom.

kaiarrowood: Fixed

dustinryerson:
@kaiarrowood looks good! 🎉  Also looks like I don't need to reapprove?

kaiarrowood:
> @kaiarrowood looks good! 🎉 Also looks like I don't need to reapprove?

Yeah, I was confused by that too. I think the repo settings need to be adjusted a bit. I don't like that we can merge without anyone approving.

PR #311, opened 2021-06-28
feat(styles/utilities): CSS utility class updates

kaiarrowood:
### Summary
Add new utility classes.

#### Changes made:
* Fix `py-0` - previously had same definition as `px-0`
* Margin/padding attributes `all` support: adds `.ma-#` and `.pa-#` 
* Text transformations: `.capitalize`, `.lowercase`, `.uppercase`
* Font-Weight: `.bold-500`, `.bold-600`, `.bold-700`
* Truncation: `.truncate-multi` and example config
* Height: `.h-100`, `.h-auto`, `.h-screen`
* Float: `.float-right`, `.float-left`, `.float-center`
* Overflow: `.overflow-auto`

### PR Checklist

### DIFF
--- a/packages/styles/_utils.scss
+++ b/packages/styles/_utils.scss
@@ -12,11 +12,24 @@ $sides: (top, right, bottom, left);
   margin-top: auto !important;
   margin-bottom: auto !important;
 }
+.ma-auto {
+  margin-left: auto !important;
+  margin-right: auto !important;
+  margin-top: auto !important;
+  margin-bottom: auto !important;
+}
+
 .mx-0 {
   margin-left: 0px !important;
   margin-right: 0px !important;
 }
 .my-0 {
+  margin-top: 0px !important;
+  margin-bottom: 0px !important;
+}
+.ma-0 {
+  margin-left: 0px !important;
+  margin-right: 0px !important;
   margin-top: 0px !important;
   margin-bottom: 0px !important;
 }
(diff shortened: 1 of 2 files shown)


packages/styles/_utils.scss
```diff
@@ -110,8 +149,39 @@ $sides: (top, right, bottom, left);
...
+  text-transform: lowercase !important;
+}
+.uppercase {
+  text-transform: uppercase !important;
+}
+.bold-500 {
+  font-weight: 500 !important;
+}
+.bold-600 {
+  font-weight: 600 !important;
```
chrisvdp, line 180: **thought:** ‏The Roboto that we use doesn't have a 600 weight. I'm not sure which weight would get rendered in the case that this was used. Possibly confusing. https://fonts.google.com/specimen/Roboto

kaiarrowood: I don't think it matters? It's like someone trying to use `font-weight` `700` when only up to `600` is supported. It ends up rendering the highest it supports?

`font-weight` `500` exists

image


`font-weight` `600` exists

image


`font-weight` `700` not supported and looks the same as `600`

image

kaiarrowood:
Reminder looking for a review @chrisvdp (when you get back from vaca) and @dustinryerson

chrisvdp (approved):

PR #313, opened 2021-06-29
fix(KCard): add bottom margin to help text element

dustinryerson:
### Summary
This adds `16px` bottom margin to the KCard help text element. This way, users of the prop won't always have to add top margin to KCard body content.

#### Changes made:
* Adds bottom margin to KCard help text element

(no screenshot because it looks the same)

### PR Checklist

### DIFF
--- a/docs/components/card.md
+++ b/docs/components/card.md
@@ -95,7 +95,7 @@ String positioned closely under the title to serve as help text
   help-text="A confirmation email will be sent to the specified email address"
 >
   <template slot="body">
-    <div class="mt-5">
+    <div>
       <KLabel>Email Address</KLabel>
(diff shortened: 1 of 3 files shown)


kaiarrowood:
Hey @dustinryerson, not sure what this is doing. Can you add a before/after pic?

dustinryerson:
> Hey @dustinryerson, not sure what this is doing. Can you add a before/after pic?

@kaiarrowood this change is just adding a default bottom margin to the helptext element and removing the top-margin from the card bodies in the examples (which don't need top margin now that the helptext has a bottom margin). There's no visual change as the margin is the same, just in a different place.

kaiarrowood (approved):
LGTM

PR #314, opened 2021-07-07
fix(KTable) fix side border display

kaiarrowood:
### Summary
Fix CSS color variable name so that side border will display on rows.

#### Changes made:
*

### PR Checklist


dustinryerson (approved):
LGTM 👍 

Non-blocking, but while we're at it, should we update https://github.com/Kong/kongponents/blob/master/docs/.vuepress/styles/prism-theme.min.css#L136 and https://github.com/Kong/kongponents/blob/master/docs/.vuepress/styles/prism-theme.min.css#L142 too? Looks like those values were manually entered here: https://github.com/Kong/kongponents/pull/257/files

kaiarrowood:
Good catch @dustinryerson . Fixed.

PR #315, opened 2021-07-07
fix(KCard): adjust helpText margins

dustinryerson:
### Summary
Adjusts KCard helpText margins. Fixes an issue where unwanted margin was added when helpText was used via component slot rather than a prop. Adds `24px` bottom margin to the helpText element.

#### Changes made:
* Fixes an issue where unwanted margin was added when helpText was used via component slot rather than a prop
* Adds `24px` bottom margin to the helpText element
* Adds an example of using helpText via slot
* Updates test snapshot

#### Screenshots:
**Before:**
Screen Shot 2021-07-07 at 3 08 55 PM

**After:**
Screen Shot 2021-07-07 at 3 07 54 PM

### PR Checklist

### DIFF
--- a/docs/components/card.md
+++ b/docs/components/card.md
@@ -118,6 +118,44 @@ String positioned closely under the title to serve as help text
 </KCard>
 ```
 
+Example of a KCard with helpText slot
+
+<KCard title="Invite a New User">
+  <template slot="helpText">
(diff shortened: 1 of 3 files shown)


kaiarrowood (approved):
LGTM!

PR #318, opened 2021-07-16
feat(kinput): fun with labels

kaiarrowood:
#### Changes made:
* Parameterize `help` text and add a built in `label` prop
* Maintain support for `.help` utility class and combining with `KLabel`
* Maintain attribute binding
* Update snapshots

Help section:

image

Labeling options:

image

image

### PR Checklist

### DIFF
--- a/docs/components/input.md
+++ b/docs/components/input.md
@@ -36,10 +36,10 @@ String to be used as the input label. Make sure that if you are using the built 
 
 - `label`
 
-<KInput label="Name" />
+<KInput label="Name" placeholder="I'm labelled!" />
 
 ```vue
-<KInput label="Name" />
+<KInput label="Name" placeholder="I'm labelled!" />
 ```
 
 If the label is omitted it can be handled with another component, like **KLabel**. This is meant to be used before **KInput** and will be styled appropriately. 
@@ -54,7 +54,7 @@ If the label is omitted it can be handled with another component, like **KLabel*
 </template>
 ```
 
-### Bound attributes
+### Attribute Binding
 You can pass any input attribute and it will get properly bound to the element.
 
 <KInput class="mb-2" placeholder="placeholder" />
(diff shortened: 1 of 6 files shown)


dustinryerson:
@kaiarrowood should we add some tests for these changes as well?

dustinryerson (commented):
@kaiarrowood should we add some tests for these changes as well?

kaiarrowood:
> @kaiarrowood should we add some tests for these changes as well?

@dustinryerson Definitely, added a couple tests, not sure if the coverage doing it this was is good enough?

docs/components/input.md
```diff
@@ -8,6 +8,53 @@
...
+<KInput help="I can help with that" placeholder="Need help?" />
+
+```
+<KInput help="I can help with that" placeholder="Need help?" />
+```
+
+You also have the option of using the `.help` utility class. This is meant to be used after **KInput** and will be styled appropriately. 
+
+<KInput type="text" placeholder="Need help?" />
+<p class="help">I can help with that</p>
```
dustinryerson, line 25: Non-blocking question: is this supposed to be styled like the input help above?

Screen Shot 2021-07-26 at 2 08 50 PM

dustinryerson (approved):
Left a non-blocking question, but this looks good, great work!

docs/components/input.md
kaiarrowood: That's really peculiar I'm not seeing this in the netlify build. Are you seeing this locally? If it's local, you have to run the CSS generator command `yarn build:postcss` to pick up the newly generated styles.

image

dustinryerson: Ah thanks! I also had to run `yarn build:styles`, but that seems to have fixed it. Thanks for the tip!

kaiarrowood: Ah, sorry! Yeah that's the one to run, `yarn build:styles` calls the postcss script I mentioned above too!

PR #323, opened 2021-08-24
style(kcheckbox,kradio): fix label space

aanchalm01:
### Summary

Radio and Checkbox button has different spacing between button and label depending on how the label is specified. When the label is passed through the prop, the spacing is smaller:  
```
<KRadio label="Option 1" />
<KCheckbox :label="checkbox1 ? 'on' : 'off'" />
```

When the label is passed through a slot, the spacing is bigger:
```
<KRadio>Option 1</KRadio>
<KCheckbox v-model="checkbox1">Label goes here.</KCheckbox>
```

#### Changes made:
* Added `d-inline-block` class for label
* Added `float-left` class for input and span
* Added `label-text` class for checkbox span

### PR Checklist

### DIFF
--- a/docs/components/input-checkbox.md
+++ b/docs/components/input-checkbox.md
@@ -82,11 +82,11 @@ Any valid attribute will be added to the input. You can read more about `$attrs`
 
 ```vue
 <KCheckbox v-model="checkbox1">
-  Label goes here. The checkbox is {{ checkbox1 ? 'checked' : 'not checked' }}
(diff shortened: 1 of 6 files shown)


chrisvdp:
I think the ideal fix here is to not have a space. If a space is needed it can be added, but if not, it would require the addition of CSS to add a negative margin.

chrisvdp (approved):

PR #334, opened 2021-09-02
Update deprecated slots to new v-slot syntax

Pan-Bubr:
### Summary
    
> In 2.6.0, we introduced a new unified syntax (the v-slot directive) for named and scoped slots. It replaces the slot and slot-scope attributes, which are now deprecated, but have not been removed and are still documented here. The rationale for introducing the new syntax is described in this RFC.

    https://vuejs.org/v2/guide/components-slots.html

Kongponents is using the old scope syntax.

This will make Kongponents compatible with Vue 2.6+ and Vue 3 for slots.

#### Changes made:
* `slot="name"` syntax were switched to `<template v-slot:name>`
* structural and non-functional `div`s and `spans` has been simplified

### PR Checklist

### DIFF
--- a/docs/components/alert.md
+++ b/docs/components/alert.md
@@ -171,20 +171,20 @@ Fixes KAlert to the top of the container.
 - `alertMessage` - Default message slot
 
 <KAlert appearance="info">
-  <template slot="alertIcon">
+  <template v-slot:alertIcon>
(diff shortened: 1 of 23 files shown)


kaiarrowood:
Just force-pushed to pick up the update pushing all versions to 4.0.

PR #340, opened 2021-09-09
style(kcheckbox): fix label alignment

aanchalm01:
### Summary
Konnect Registration page has Checkbox input/span alignment issue.
#### Changes made:
*

### PR Checklist

### DIFF
--- a/docs/components/input-checkbox.md
+++ b/docs/components/input-checkbox.md
@@ -118,15 +118,6 @@ like:
 .KCheckbox-wrapper {
   --KCheckboxPrimary: blueviolet;
 }
-
-.k-checkbox {
-  .k-input + span {
-    &.float-left {
-      position: relative;
-      padding: 3px;
-    }
-  }
-}
 </style>
 
 <script>
--- a/docs/components/input-checkbox.md
+++ b/docs/components/input-checkbox.md
@@ -40,33 +40,6 @@ state of the input. You can read more about passing values via `v-model`
 ```vue
 <KCheckbox v-model="checked" />
 ```
-
-### label
-
-Will place label text to the right of the input. Can also be [slotted](#slots).
-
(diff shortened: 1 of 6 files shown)


chrisvdp:
@aanchalm01 looking at the Percy results, the labels are misaligned in Radio docs now. https://deploy-preview-340--kongponents.netlify.app/components/input-radio.html

aanchalm01:
@chrisvdp does the checkbox look okay to you now ? I can do the scoped style to make sure that radio style is not updated.

chrisvdp:
My issue with overly specific CSS rules is that it tends to create a CSS selector arms race. The styles are very hard to override if you need to. This isn't a great approach in websites and it's especially bad approach in style guides as these are consumed by webapps and they are very real use cases where they need to be overridden.

It seems that a [render function](https://vuejs.org/v2/guide/render-function.html) may be needed here to avoid this. A render function should give us access to the raw string so we can trim white space before it makes it to the DOM.

docs/components/input-checkbox.md
```diff
@@ -145,6 +118,15 @@ like:
 .KCheckbox-wrapper {
   --KCheckboxPrimary: blueviolet;
 }
+
+.k-checkbox {
```
chrisvdp, line 122: **question:** ‏do we still need this CSS?

chrisvdp:
@aanchalm01 this should be marked as a breaking change

https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with--to-draw-attention-to-breaking-change

docs/components/input-checkbox.md
aanchalm01: No! Removed it. Thanks for catching this.

chrisvdp (approved):
LGTM

PR #351, opened 2021-09-20
feat(kinput): redesign kinput

kaiarrowood:
### Summary
Update `KInput` to new metalab designs.

#### Changes made:
* Add `size` property that can take values `small`, `medium` (default), `large`
* Add border updates on `hover` and `focus`
* Add `--KInputHover` theming var for hover border / label color
* Update snaps / tests

Figma: https://www.figma.com/file/sfSZhMfpzCzddEhd8tdk9s/Kong_Design-library?node-id=209%3A1429

image

### PR Checklist


packages/KInput/KInput.vue
```diff
@@ -2,24 +2,35 @@
...
+      }"
       v-on="listeners">
 
     <div
       v-else
+      :class="`k-input-label-${size}`"
       class="col-md-4 mt-5">
       <div class="text-on-input">
-        <label :class="{ focused: isFocused }">
+        <label :class="{ focused: isFocused, hovered: isHovered }">
```
chrisvdp, line 20: The label here is missing the `for` attr. I was playing with a library called `vue-unique-id` to handle cases like this. The other option, that I think also helps simplify the hover and focus styling is wrapping the label around the input. In this case, it doesn't require an id/for to associate the label and input.

packages/KInput/KInput.vue
```diff
@@ -2,24 +2,35 @@
   <div class="k-input-wrapper">
     <input
       v-if="!label"
```
chrisvdp, line 4: every input should have a label. It can be visually hidden for style purposes

packages/KInput/KInput.vue
kaiarrowood: Updated it so if there is a label passed in we are now using `for` attribute.

packages/KInput/KInput.vue
kaiarrowood: Will not address as we doc correct usage of `KLabel` with `KInput`.

image

packages/KInput/KInput.vue
```diff
@@ -49,11 +65,25 @@ export default {
...
       default: ''
+    },
+    size: {
+      type: String,
+      default: 'medium'
+    },
+    /**
+     * Debug mode - for testing only, strips out generated ids
+     */
+    debugMode: {
```
chrisvdp, line 76: **nitpick:** ‏is there a better name than debug mode? Seems like something that should get turned off, maybe `testingMode`? or `staticIds`?

kaiarrowood: Changed to `testMode`.

dustinryerson (dismissed):
Overall this is looking good! I have a few comments:
1. It looks like the hover and focus color changes are being applied via box-shadow rather than the border color. I think we should update it to modify the border color and remove the box-shadow. Below are some screenshots for reference:

**Kongponents**
Hover:
Screen Shot 2021-09-28 at 11 16 00 AM
hover
Focus:
Screen Shot 2021-09-28 at 11 15 33 AM
focus
Focus (error):
Screen Shot 2021-09-28 at 11 21 17 AM

**Figma Design**
Screen Shot 2021-09-28 at 11 18 08 AM

2. This is a smaller nitpick - I think the input label should have a little more padding on the x-axis.
**Kongponent**
Screen Shot 2021-09-28 at 11 56 48 AM
**Figma**
Screen Shot 2021-09-28 at 11 56 25 AM

3. I think it might improve the UX to add a slight transition when changing the color on hover and focus, that way the color change isn't so immediate (which I think some users may experience as a little jarring)

4. I may have just missed it, but I couldn't find any design mockup for the input field error state border color and label styling. If it doesn't exist, it would be helpful to get it added to Figma so that we can reference it during the review.

kaiarrowood:
Hey @dustinryerson , thanks for the review. I have addressed the label horizontal padding and fixed the error state display. I am using `box-shadow` instead of `border` so that I can use the `inset` property. `border` does not differentiate between if it is an inner border or an outer border - in the new designs the focus "border" needs a width of `2px`, so to avoid screen shift on focus/hover I am using an `inset` `box-shadow`.

I could add a transition, but I'm not familiar with that CSS prop; do you have a suggestion on what you think may be a good one?

Lastly, no, there isn't any Figma related to error state at this point.

dustinryerson (approved):

PR #354, opened 2021-09-22
feat(kcard): kcard redesign [khcp-1895]

kaiarrowood:
### Summary
Update `KCard` to be able to handle new slots for the redesign.

#### Changes made:
* Add new package `vue-uuid` for generating ids for aria properties
* Added `status` prop and `statusHat` slot - Small text directly above the title
* Added `notifications` slot - Right side of the body section. Used for badges, icons, etc
* Remove `helpText` prop and slot
* Add `debugMode` prop for testing with aria prop ids
* Update docs
* Update tests/snap

Figma: https://www.figma.com/file/sfSZhMfpzCzddEhd8tdk9s/Kong_Design-library?node-id=1668%3A11038

image

### PR Checklist


packages/KCard/KCard.vue
```diff
@@ -3,14 +3,22 @@
     :class="[borderVariant, {'hover': hasHover, 'kcard-shadow': hasShadow }]"
```
chrisvdp, line 3: I think the wrapper element of the card is a good use case for the `section` element https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

packages/KCard/KCard.vue
```diff
@@ -1,35 +1,42 @@
 <template>
-  <div
+  <section
     :class="[borderVariant, {'hover': hasHover, 'kcard-shadow': hasShadow }]"
```
chrisvdp, line 3: **thought:** ‏I think we can add [`aria-labeledby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) here and have it point to the title attribute. This might require the unique id generator though.

Also, [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute) and point it to the body.

chrisvdp (approved):

packages/KCard/KCard.vue
kaiarrowood: The logic now uses `aria-label` if the `title` prop is specified. If not, it will generate a unique id for the title `div` and use that as the `aria-labelledby`. Generate a unique id for content and use as `aria-describedby`.

aanchalm01 (commented):
LGTM! Great work. :star2:

dustinryerson (approved):
This looks great, nice work Kai!

Small, non-blocking nit: the icon and text for the statusHat slot don't seem to be horizontally aligned:

Before:
Screen Shot 2021-09-27 at 10 21 39 AM

After:
Screen Shot 2021-09-27 at 10 22 13 AM

I'm calling this a non-blocking nit because I'm not sure how easy it would be to fix this since the slot seems to be meant to be a flexible and have the ability to contain a variety of content.

PR #356, opened 2021-09-24
feat(kmodal): style redesign [khcp-1897]

kaiarrowood:
### Summary
Update KModal with new design.
For khcp-1897.

#### Changes made:
* Default `KButton` `appearance` property to `outline`
* Remove default values for modal `title` and `content`
* Make `title` required for accessibility
* Add `hideTitle` prop if they don't want the title displayed
* Remove `helpText` prop and `help` slot
* Swap order of action and cancel buttons (cancel comes first)
* Add `action-buttons` slot
* Update style classes to prefix with `k-` (class without `k-` left in for backwards compatibility)
* Update snaps

Figma: https://www.figma.com/file/sfSZhMfpzCzddEhd8tdk9s/Kong_Design-library?node-id=209%3A1086

image

### PR Checklist

### DIFF
--- a/docs/components/modal.md
+++ b/docs/components/modal.md
@@ -43,8 +43,8 @@ export default {
 - `@canceled` - Emitted when cancel/close button is clicked
 
 **Content**
-- `title` - Text to display in header if not using slot
-- `helpText` - Help text to display below header if not using slot
+- `title` - Text displayed in header if not using slot (**Note**: this field is still required for accessibility reasons even if using the slot)
+- `hideTitle` - If not using the `header-content` slot, tells the component whether or not to display the title
 - `content` - Text to display in content if not using slot
 
 **Buttons & Appearance**
@@ -54,10 +54,9 @@ export default {
 - `cancelButtonAppearance` - Change the [appearance](/components/button.html#props) of the cancel
 
 ## Slots
-There are 5 designated slots you can use to display content in the modal.
+There are 4 designated slots you can use to display content in the modal.
 
 - `header-content`
-- `help` - Light text between title and content.
 - `body-content`  
 - `footer-content` - Contains cancel & action buttons by default.
 - `action-buttons` - Contains action buttons which are right-aligned. This slot will not exist if using `footer-content` slot.
@@ -65,7 +64,8 @@ There are 5 designated slots you can use to display content in the modal.
 ---
 ### Usage
 
-Using both the provided props and slot options we will now demonstrate how to customize the modal for a delete confirmation.
+Using both the provided props and slot options we will now demonstrate how to customize the modal for a delete confirmation. 
+Notice that even though we are using the `header-content` slot we still specify the `title` attribute for accessibility.
 
 <KButton @click="slottedIsOpen = true">Open Delete Modal</KButton>
 <template>
@@ -73,9 +73,12 @@ Using both the provided props and slot options we will now demonstrate how to cu
     :isVisible="slottedIsOpen"
     actionButtonText="Delete"
     actionButtonAppearance="danger"
-    @canceled="slottedIsOpen = false">
-    <template v-slot:header-content>Delete Item</template>
-    <template v-slot:help>Take this action to delete</template>
+    @canceled="slottedIsOpen = false"
+    title="Delete Item">
+    <template v-slot:header-content>
+      <KIcon icon="dangerCircle" color="red" class="mr-2" />
+      Delete Item
+    </template>
     <template v-slot:body-content>Are you sure you want to delete this item? This action can not be undone.</template>
     <template v-slot:action-buttons>
       <KButton appearance="outline" class="mr-2">Back</KButton>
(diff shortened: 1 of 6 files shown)


packages/KModal/KModal.vue
```diff
@@ -5,39 +5,43 @@
...
+      class="k-modal-backdrop modal-backdrop"
       @click="close">
 
       <div
-        class="modal-dialog"
+        class="k-modal-dialog modal-dialog"
         @click.stop>
-        <div class="modal-content">
+        <div class="k-modal-content modal-content">
           <div
```
chrisvdp, line 15: This should be a heading tag.

packages/KModal/KModal.vue
```diff
@@ -5,39 +5,43 @@
     class="k-modal"
```
chrisvdp, line 5: We should add an `aria-labelledby` and point it to the title

packages/KModal/KModal.vue
```diff
@@ -5,39 +5,43 @@
...
           </div>
-          <div class="modal-footer">
+          <div class="k-modal-footer modal-footer d-flex">
             <slot name="footer-content">
-              <KButton
-                :appearance="actionButtonAppearance"
-                @click="proceed">
-                {{ actionButtonText }}
-              </KButton>
               <KButton
```
chrisvdp, line 31: this should have a keyboard event, like onEnter

packages/KModal/KModal.vue
```diff
@@ -5,39 +5,43 @@
...
                 :appearance="cancelButtonAppearance"
                 @click="close">
                 {{ cancelButtonText }}
               </KButton>
+              <div class="k-modal-action-buttons">
+                <slot name="action-buttons">
+                  <KButton
```
chrisvdp, line 38: this should have a keyboard event, like onEnter

chrisvdp (commented):
When the modal is set to `isVisible true` we should also set the focus to the first focusable element. In this case that's the close button.

packages/KModal/KModal.vue
kaiarrowood: I don't really want to mess with styles associated with headings. Is there a `role` I can use?

packages/KModal/KModal.vue
kaiarrowood: Using `@keyup.enter`, love the idea thanks!

packages/KModal/KModal.vue
kaiarrowood: Using `@keyup.esc`, thanks for the idea!!

packages/KModal/KModal.vue
kaiarrowood: Now using `aria-modal`, but `aria-labelledby` requires an `id`? Instead using `aria-label` with the `title` prop.... if they instead use the `header-content` slot with some sort of complex HTML, this won't work though, and wouldn't work for an `aria-labelledby` would it? Thinking of making `title` possibly a required prop?

packages/KModal/KModal.vue
kaiarrowood: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role will work?

packages/KModal/KModal.vue
kaiarrowood: I have made the `title` prop required. It is always used for the `aria-label`. Added a new prop `hideTitle` if they don't want to display it for some reason.

docs/components/modal.md
```diff
@@ -39,41 +43,48 @@ export default {
...
-- `helpText` - Help text to display below header if not using slot
+- `title` - Text displayed in header if not using slot (**Note**: this field is still required for accessibility reasons even if using the slot)
+- `hideTitle` - If not using the `header-content` slot, tells the component whether or not to display the title
 - `content` - Text to display in content if not using slot
 
 **Buttons & Appearance**
 - `actionButtonText` - Change the text content of the submit/proceed button
 - `actionButtonAppearance` - Change the [appearance](/components/button.html#props) of the submit/proceed
 - `cancelButtonText` - Change the text content of the close/cancel button
+- `cancelButtonAppearance` - Change the [appearance](/components/button.html#props) of the cancel
```
dustinryerson, line 54:
```suggestion
- `cancelButtonAppearance` - Change the [appearance](/components/button.html#props) of the cancel button
```

docs/components/modal.md
```diff
@@ -39,41 +43,48 @@ export default {
...
-- `cancelButtonAppearance` - Change the [appearance](/components/button.html#props) of the close/cancel
-- `title` - Text to display in header if not using slot
-- `helpText` - Help text to display below header if not using slot
+- `title` - Text displayed in header if not using slot (**Note**: this field is still required for accessibility reasons even if using the slot)
+- `hideTitle` - If not using the `header-content` slot, tells the component whether or not to display the title
 - `content` - Text to display in content if not using slot
 
 **Buttons & Appearance**
 - `actionButtonText` - Change the text content of the submit/proceed button
 - `actionButtonAppearance` - Change the [appearance](/components/button.html#props) of the submit/proceed
```
dustinryerson, line 52:
```suggestion
- `actionButtonAppearance` - Change the [appearance](/components/button.html#props) of the submit/proceed button
```

dustinryerson (approved):
LGTM, great work! 🎉

PR #358, opened 2021-09-27
feat(kemptystate): redesign [khcp-1990]

kaiarrowood:
### Summary
Update `KEmptyState` design.

#### Changes made:
* Add `icon` prop for customizing the displayed icon
* Add theming support for `--KEmptyTitleColor`, `--KEmptyContentColor`, `--KEmptyBackground`
* Update tests

Figma: https://www.figma.com/file/sfSZhMfpzCzddEhd8tdk9s/Kong_Design-library?node-id=257%3A1640

### PR Checklist

### DIFF
--- a/packages/KEmptyState/KEmptyState.vue
+++ b/packages/KEmptyState/KEmptyState.vue
@@ -72,7 +72,7 @@ export default {
   padding: 42px 0;
   text-align: center;
   border-radius: 3px;
-  background-color: var(--KEmptyBackground);
+  background-color: var(--KEmptyBackground, var(--white));
 }
 .empty-state-wrapper h5 {
   color: var(--KEmptyTitleColor, var(--black-500));
--- a/packages/KEmptyState/KEmptyState.vue
+++ b/packages/KEmptyState/KEmptyState.vue
@@ -3,11 +3,11 @@
     <div class="empty-state-title">
       <div
         v-if="isError"
-        class="warning-icon card-icon mb-4"
+        class="warning-icon empty-state-icon card-icon mb-4"
       >
         <KIcon
           :size="iconSize"
-          icon="warning"
+          :icon="icon"
           view-box="0 0 50 50" />
       </div>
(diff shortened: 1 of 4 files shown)


packages/KEmptyState/KEmptyState.vue
```diff
@@ -2,13 +2,13 @@
   <div class="empty-state-wrapper">
```
chrisvdp, line 2: Probably a good usage of `<section>`

chrisvdp (changes requested):

packages/KEmptyState/KEmptyState.vue
```diff
@@ -65,22 +68,22 @@ export default {
...
-  color: rgba(0,0,0,.70);
-  border: 1px solid rgba(151,151,151,.1);
   border-radius: 3px;
-  background-color: rgba(0,0,0,.03);
+  background-color: var(--KEmptyBackground, var(--white));
 }
 .empty-state-wrapper h5 {
-  margin: 0 0 .75rem;
+  color: var(--KEmptyTitleColor, var(--black-500));
+  margin: 0 0 14px;
```
aanchalm01, line 79: Any specific reason for why we are switching from rem to px?

packages/KEmptyState/KEmptyState.vue
kaiarrowood: Fixed

packages/KEmptyState/KEmptyState.vue
kaiarrowood: This is a good question. For awhile `rem` was necessary due to `px` not scaling well, but now most browsers use zoom to increase scale so rem is actually no longer necessary. The first answer here has a pretty good explanation: https://stackoverflow.com/questions/11799236/should-i-use-px-or-rem-value-units-in-my-css

chrisvdp (approved):

PR #363, opened 2021-09-29
Redesign KIcon

Pan-Bubr:
### Summary
The old implementation of KIcon only allowed for using `<path>` object in SVG files. Now it is expanded to allow for `<circle>` and `<rect>` tags.

The change also unifies the primary color of the stock icons along with adding the new ones that will be needed for the Kongponent redesign.

#### Changes made:
* Implement support for Circle and Rect svg objects in KIcon svg scrapper
* Unify the collor pallette of icons
* Add new icons to the list
* Update test snapshots

### PR Checklist


kaiarrowood (approved):
Awesome work Adrian!! 🐉 🔥

PR #365, opened 2021-10-01
fix(*): misc issue fixes

kaiarrowood:
### Summary
Miscellaneous kongponent redesign fixes

#### Changes made:
* Update `KBadge` styles: `font-weight`, `font-size`, `background-color` of `default` badge
* `Krumbs` has a dependency on `KIcon`
* Fix layout in `KInput` docs
* Add left margin to `KCard` `footer-action` content
* Update Typography page to display new style classes
* Update `KToggle` doc `KModal` implementation

image

### PR Checklist


packages/KCard/KCard.vue
```diff
@@ -39,7 +39,7 @@
       </div>
       <div
         v-if="$scopedSlots.notifications"
-        class="k-card-notifications">
+        class="k-card-notifications pl-3">
```
dustinryerson, line 42:
```suggestion
        class="k-card-notifications ml-3">
```

dustinryerson: I think margin would be better here instead of padding.

dustinryerson (changes requested):
Noticed a few things:
1. The SVGs on the Breadcrumbs page are broken:

**Original**
Screen Shot 2021-10-04 at 1 34 46 PM

**PR**
Screen Shot 2021-10-04 at 1 56 17 PM

2. For the [content styles](http://localhost:8080/style-guide/type.html#content-styles) section, I'm not seeing a difference in style between the `Heading` and `Body`. Also not seeing any content for either section:
Screen Shot 2021-10-04 at 1 41 45 PM

3. Can we use a different background color for the [shape](http://localhost:8080/components/badge.html#shape) section? IMHO the yellow background makes it hard to see the different styles:
Screen Shot 2021-10-04 at 1 26 44 PM

dustinryerson (approved):
After rebuilding the styles, everything looks great. Apologies for the confusion. Nice work! 🚀

PR #367, opened 2021-10-07
feat(kselect): add new kongponent

kaiarrowood:
### Summary
Add new KSelect/KDropdown component

#### Changes made:
* Has two styles `select` and `dropdown`
* Supports filtering
* Support keyboard nav

Dropdown:

image

Select:

image

Figma: https://www.figma.com/file/sfSZhMfpzCzddEhd8tdk9s/Kong_Design-library?node-id=3326%3A14211

### PR Checklist

### DIFF
--- a/packages/KSelect/KSelectItem.vue
+++ b/packages/KSelect/KSelectItem.vue
@@ -66,6 +66,7 @@ export default {
     background-color: var(--white);
     border-radius: 4px;
     text-align: left;
+    font-weight: 200;
 
     .k-select-item-label {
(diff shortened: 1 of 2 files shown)


chrisvdp:
When I search, if there are no results then the dropdown appears empty. Should we have a blank state?

image

aanchalm01:
@kaiarrowood: For the `select` style, the check icon is always showing up with first item value, is that intentional ?

Screen Shot 2021-10-11 at 2 29 04 PM

PR #370, opened 2021-10-14
feat(kmenu): add new kongponent

aanchalm01:
### Summary
Create new `KMenu` kongponent.
#### Changes made:
Add new component `KMenu`, and child component `KMenuItem`
Supports use of `items` prop or `body` slot which can instantiate `KMenuItem`
Add docs
Add tests

### PR Checklist


packages/KMenu/KMenuItem.vue
```diff
@@ -0,0 +1,156 @@
...
+    transform: translateX(-57%);
+    bottom: 0;
+    width: 88%;
+    height: 1px;
+    background-color: var(--grey-300);
+  }
+}
+
+.k-menu ul:not(:last-child):after {
+    content: " ";
```
chrisvdp, line 118: **nitpick:** ‏indentation

packages/KMenu/KMenuItem.vue
```diff
@@ -0,0 +1,156 @@
+<template>
+  <div
+    :data-testid="item ? `${item.title.replace(' ', '-')}-menu-item` : 'menu-item'"
+    :test-mode="testMode"
+    :class="[isOpen && expandable ? 'title-dark' : '', {'expand-item' : expandable}]"
+    class="k-menu-item">
+    <div
+      class="menu-title"
+      @click="isOpen = !isOpen">
```
chrisvdp, line 9: **issue:** ‏we should not have click events on non clickable elements. This should be an <a> or <button>

chrisvdp: **issue:** ‏any click event should also have a keyboard equivalent

chrisvdp: can we add aria attributes for this

aria-controls
aria-expanded

packages/KMenu/KMenuItem.vue
```diff
@@ -0,0 +1,156 @@
...
+        class="span-icon-container">
+        <KIcon
+          v-if="expandable"
+          :icon="isOpen ? 'chevronUp' : 'chevronDown'"
+          color="var(--grey-400)"
+          size="16"
+        />
+      </span>
+    </div>
+    <div
```
chrisvdp, line 27: can we add aria attributes for this

aria-labelledby

chrisvdp (changes requested):
This seems like an accordion to me. Is there a reason we went with the name menu?

packages/KMenu/KMenuItem.vue
aanchalm01: Fixed!

packages/KMenu/KMenuItem.vue
aanchalm01: So this click event needs to be removed ?

chrisvdp: we can have the click event, but the element should be changed to a button

aanchalm01:
@chrisvdp: The name Menu is selected because it's going to be used from inside a table for filtering/sort purposes. Created it like an accordion because that was the requirement.

packages/KMenu/KMenuItem.vue
chrisvdp: for the keyboard event see https://github.com/maranran/eslint-plugin-vue-a11y/blob/master/docs/rules/click-events-have-key-events.md

aanchalm01: Okay! I see. I will add button and keyboard equivalent and aria attributes.

packages/KMenu/KMenuItem.vue
```diff
@@ -0,0 +1,174 @@
+<template>
+  <div
+    :data-testid="item ? `${item.title.replace(' ', '-')}-menu-item` : 'menu-item'"
+    :test-mode="testMode"
+    :class="[isOpen && expandable ? 'title-dark' : '', {'expand-item' : expandable}]"
+    aria-labelledby="menu-item"
+    class="k-menu-item">
+    <button
```
chrisvdp, line 8: is there a reason not to us KButton here?

packages/KMenu/KMenuItem.vue
```diff
@@ -0,0 +1,174 @@
+<template>
+  <div
+    :data-testid="item ? `${item.title.replace(' ', '-')}-menu-item` : 'menu-item'"
+    :test-mode="testMode"
+    :class="[isOpen && expandable ? 'title-dark' : '', {'expand-item' : expandable}]"
+    aria-labelledby="menu-item"
```
chrisvdp, line 6: I can't tell what this is associated with. It should be the id of an element.

chrisvdp: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

chrisvdp (dismissed):

packages/KMenu/KMenuItem.vue
aanchalm01: The button here does not have any styles associated with what we have in KButton, therefore used general button element. Does that make sense ?

kaiarrowood: Would `non-visual-button` class help here?

image

kaiarrowood:
> @chrisvdp: The name Menu is selected because it's going to be used from inside a table for filtering/sort purposes. Created it like an accordion because that was the requirement.

It's also not an accordion because we allow multiple entries to be expanded simultaneously.

packages/KMenu/KMenuItem.vue
aanchalm01: I tried using `.non-visual-button` but it has `cursor: pointer !important;` and if the menu item is not expandable I need the cursor not to be a pointer. I will see if I can change the specificity and make it non-cursor but also: is it okay to have !important for a specific styling in KMenuItem.vue file ?

chrisvdp:
> > @chrisvdp: The name Menu is selected because it's going to be used from inside a table for filtering/sort purposes. Created it like an accordion because that was the requirement.
> 
> It's also not an accordion because we allow multiple entries to be expanded simultaneously.

That's not always a constraint of an accordion, it's just an implementation detail. Bootstrap for example has an option to allow multiples to be open. https://getbootstrap.com/docs/5.0/components/accordion/#always-open

I think we should avoid thinking about a single use case only in a design system and think about potential reuse. We can use this component as a menu, but that doesn't mean it's the only use case.

packages/KMenu/KMenuItem.vue
kaiarrowood: I think in this scenario using `!important` will be ok provided it is on a custom class we don't see outside of KMenu (or inside a scoped SCSS block).

chrisvdp: I don't think you need to render a button in the case where it is not expandable.

aanchalm01: There is a named slot inside KButton that I am using to display the item title, if I use v-if for the button, I won't be able to display the title using same named slot.

packages/KMenu/KMenuItem.vue
kaiarrowood: Done

packages/KMenu/KMenuItem.vue
kaiarrowood: All menu items are considered clickable. Fire 'clicked' event if non-expando, fire 'toggled' event if expando.

packages/KMenu/KMenuItem.vue
kaiarrowood: Fixed.

chrisvdp (approved):

PR #378, opened 2021-10-20
chore(kicon): add title attribute

kaiarrowood:
#### Changes made:
* Update existing KIcon's with `title` attribute and use better logic for determining title if not provided
* Remove duplicate cog icon
* Update docs

image

image

### PR Checklist


chrisvdp:
We should also update https://github.com/Kong/kongponents/blob/master/docs/contributing/adding-icons-to-kicon.md to add a blurb about including `<title>` in new icons

PR #397, opened 2021-11-11
fix(kinput): apply attributes to input [KHCP-2108]

adamdehaven:
### Summary

Apply attributes to the child input element rather than the wrapper to avoid duplicate elements with the same id.

#### Changes made:
* Add `inheritAttrs` property to component definition to disable attribute inheritance on the wrapper element.
* Alter `v-bind` to use `$attrs` instead of `attrs` (to spec).

### PR Checklist

### DIFF
--- a/packages/KInput/KInput.vue
+++ b/packages/KInput/KInput.vue
@@ -3,7 +3,7 @@
     <input
       v-if="!label"
       :value="value"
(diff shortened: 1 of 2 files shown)


kaiarrowood (approved):
Nice work Adam! LGTM!

PR #405, opened 2021-11-16
fix(kicon): wrap kicon with tag [KHCP-2186]

adamdehaven:
The `KIcon` component utilizes the `DOMParser` Web API to parse through svg documents; however, this API is not available in Vue until the `mounted()` hook is called. 

When combined with a `Kooltip` component (more specifically, the underlying `KPop`), the `KPop` component had nothing to bind to during render as the DOM was still loading, meaning the svg was not yet in the DOM because the `KIcon` component element could not yet be appended to, causing the following error:

```js
DOMException: Failed to execute 'appendChild' on 'Node': This node type does not support this method.
```

If a user landed on the docs site on a page that did not include a `KIcon` with `Kooltip` (or a `KLabel` with `help/info` attribute, etc.) the site would continue to work properly even after visiting other routes.

If a user landed directly on a page with the combination of components as described above, or if the user refreshed on one of these pages, navigation and the Vue instance became unusable, and the user was unable to navigate away from the given route or interact with any components on the page.

#### Changes made:

1. Wrap the `KIcon` in a wrapper `<span>` that always renders and does not wait for the mounted event, and trigger the svg itself to load on the `$nextTick` after the `mounted()` hook fires. This gives `KPop` a node to bind to regardless of the state of the DOM.
2. Update styles that were directly applied to svg elements and rewrite them to the `.kong-icon` class (which is now applied to the wrapper `<span>` instead of the svg itself).

The solution resolves the issue both for local development with `yarn docs:dev` as well as for the production build.

### PR Checklist


adamdehaven:
Abandoning this in favor of a new PR since the source branch was already rebased.

PR #406, opened 2021-11-16
feat(pagination): support in table and catalog

kaiarrowood:
### Summary
Update KPagination to allow only prev/next page nav support.

#### Changes made:
* Add new prop `disablePageJump`

### PR Checklist

### DIFF
--- a/docs/components/card-catalog.md
+++ b/docs/components/card-catalog.md
@@ -155,7 +155,15 @@ Set this to `true` to limit pagination navigation to `previous` / `next` page on
 
 <KCardCatalog
   :fetcher="fetcher" 
-  :disablePaginationPageJump="true" />
+  :disablePaginationPageJump="true"
+  :paginationPageSizes="[4, 5, 6]"
+  :initial-fetcher-params="{
+    pageSize: 4,
+    page: 1,
+    query: '',
+    sortColumnKey: '',
+    sortColumnOrder: ''
+  }" />
 
 ```vue
 <template>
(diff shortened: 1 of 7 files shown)


dustinryerson (approved):
Nice work!

A couple non-blocking questions:
1. I noticed when clicking the next and previous buttons (although I think this happens in most of the examples) that the buttons and text on the left shift slightly when clicking through. Is that something we should fix here or on a separate PR?
2. What do you think about adding text to the arrow buttons when using the `disablePaginationPageJump` prop? Similar to the screenshots below:
Screen Shot 2021-11-16 at 9 43 01 AM
Screen Shot 2021-11-16 at 9 44 46 AM

kaiarrowood:
> Nice work!
> 
> A couple non-blocking questions:
> 
> 1. I noticed when clicking the next and previous buttons (although I think this happens in most of the examples) that the buttons and text on the left shift slightly when clicking through. Is that something we should fix here or on a separate PR?
> 2. What do you think about adding text to the arrow buttons when using the `disablePaginationPageJump` prop? Similar to the screenshots below:
>    Screen Shot 2021-11-16 at 9 43 01 AM
>    Screen Shot 2021-11-16 at 9 44 46 AM

Thanks @dustinryerson I fixed the pagination text shifting the buttons on page change. Design has already approved this approach so will leave out the "prev"/"next" text for now.

dustinryerson:
> Thanks @dustinryerson I fixed the pagination text shifting the buttons on page change. Design has already approved this approach so will leave out the "prev"/"next" text for now.

Awesome! 🚀

PR #415, opened 2021-11-17
fix(*): general fixes

kaiarrowood:
### Summary
General fixes from the redesign

#### Changes made:
* Update `KTable` docs to remove sortable when not sorting
* `KTable` - fix client sort and add experimental `enableClientSort` prop
* `KtextArea`
  * Add prop `disableCharLimit`
  * Fix alignment of char limit text
  * Emit an event if the character limit is exceeded
  * Update docs / tests

### PR Checklist

### DIFF
--- a/docs/components/textarea.md
+++ b/docs/components/textarea.md
@@ -31,22 +31,38 @@ If the label is omitted it can be handled with another component, like **KLabel*
 
 ## Props
 ### Size
+
 You can specify `rows`, `cols` for the textarea size.
 
-<KTextArea :rows=2 :cols=15 placeholder="rows:2, cols:15" />
-<br>
 <KTextArea label="Size" :rows=3 :cols=20 placeholder="I'm labelled and customized!" />
 <br>
 <KTextArea :rows=8 :cols=25 placeholder="rows:8, cols:25"  />
 
 ```vue
-<KTextArea :rows=2 :cols=15 placeholder="rows:2, cols:15" />
-<br>
 <KTextArea label="Size" :rows=3 :cols=20 placeholder="I'm labelled and customized!" />
 <br>
 <KTextArea :rows=8 :cols=25 placeholder="rows:8, cols:25"  />
 ```
 
+### characterLimit
+
+Use this prop to specify a character limit for the textarea, defaults to `2048`.
+
+<KTextArea :characterLimit="500" />
+
+```vue
+<KTextArea :characterLimit="500" />
+```
+
+### disableCharacterLimit
+
+Use this prop to remove the character limit on the textarea. Defaults to `false`.
+
+<KTextArea :disableCharacterLimit="true" />
+
+```vue
+<KTextArea :disableCharacterLimit="true" />
+```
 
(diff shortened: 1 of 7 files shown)


docs/components/textarea.md
```diff
@@ -31,22 +31,38 @@ If the label is omitted it can be handled with another component, like **KLabel*
...
+```
+
+### disableCharacterLimit
+
+Use this prop to remove the character limit on the textarea. Defaults to `false`.
+
+<KTextArea :disableCharacterLimit="true" />
+
+```vue
+<KTextArea :disableCharacterLimit="true" />
```
adamdehaven, line 64: Unless the standard is to explicitly show props with a `true` value, this could just be written as

``` html
<KTextArea disableCharacterLimit />

<!-- or -->

<KTextArea disable-character-limit />
```

adamdehaven (commented):
If we're providing a `characterLimit` prop, should we not be enforcing that character limit by limiting the text the user can enter into the field, or at least providing a way to hook into whether it's valid? 

It looks like this is more just for user feedback, but if I was using this component I would possibly assume that the character limit functionality would also trim the value of the field (e.g. anything over the `characterLimit` is trimmed from the end of the value).

At the least, I would think the component would provide a way to hook into whether or not the value is valid (i.e. if the value is within the `characterLimit` parameters), at least with a computed property that indicates whether the character limit has been violated, or some sort of `$emit`.

packages/KTextArea/KTextArea.vue
```diff
@@ -118,28 +118,37 @@ export default {
...
-  resize: none;
+.k-input-wrapper {
+  width: fit-content;
+  display: grid;
 
-  &::placeholder {
-    color: var(--grey-500);
-  }
+  textarea.form-control {
+    font-family: 'Maison Neue';
```
aanchalm01, line 126:
```suggestion
    font-family: var(--font-family-sans);
```

kaiarrowood: Got it

docs/components/textarea.md
kaiarrowood: We don't really have a standard, and do a mix throughout the docs, but I went ahead and changed this instance.

adamdehaven (approved):

kaiarrowood:
> If we're providing a `characterLimit` prop, should we not be enforcing that character limit by limiting the text the user can enter into the field, or at least providing a way to hook into whether it's valid?
> 
> It looks like this is more just for user feedback, but if I was using this component I would possibly assume that the character limit functionality would also trim the value of the field (e.g. anything over the `characterLimit` is trimmed from the end of the value).
> 
> At the least, I would think the component would provide a way to hook into whether or not the value is valid (i.e. if the value is within the `characterLimit` parameters), at least with a computed property that indicates whether the character limit has been violated, or some sort of `$emit`.

Good idea @adamdehaven I've added a `char-limit-exceeded` event that is emitted when the character limit is exceeded and updated the docs.

PR #418, opened 2021-11-17
fix(ktextarea): limit exceeded event

kaiarrowood:
#### Changes made:
* Update `char-limit-exceeded` event to only fire if switching between exceeding and not exceeding
* Return more data on the event

### PR Checklist

### DIFF
--- a/packages/KTextArea/KTextArea.vue
+++ b/packages/KTextArea/KTextArea.vue
@@ -117,7 +117,7 @@ export default {
   watch: {
     charLimitExceeded (newval, oldval) {
       if (newval !== oldval) {
-        this.$emit('char-limit-exceed', {
+        this.$emit('char-limit-exceeded', {
           value: this.currValue,
           length: this.currValue.length,
           characterLimit: this.characterLimit,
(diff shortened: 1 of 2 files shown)


packages/KTextArea/KTextArea.vue
```diff
@@ -117,6 +109,21 @@ export default {
...
       return listeners
+    },
+    charLimitExceeded () {
+      return !this.disableCharacterLimit && this.currValue.length > this.characterLimit
+    }
+  },
+  watch: {
+    charLimitExceeded (newval, oldval) {
+      if (newval !== oldval) {
+        this.$emit('char-limit-exceed', {
```
adamdehaven, line 120: Incorrect event name, should match the docs: `char-limit-exceeded`

adamdehaven (changes requested):

packages/KTextArea/KTextArea.vue
kaiarrowood: Thank you! Fixed.

adamdehaven (approved):
LGTM

PR #434, opened 2021-11-22
fix(*): fix deps and build process

adamdehaven:
#### Vue Composition API

When utilizing the `@vue/composition-api` inside of a Kongponent package, the `@vue/composition-api` needs to be added as a `peerDependency` to the component package rather than a `dependency`.

> Currently, installing with yarn with the peer flag `yarn add {package-name} --peer` [is broken](https://github.com/yarnpkg/yarn/issues/5287)

You will need to manually add the `peerDependencies` entry inside the individual component's `package.json`. For example, inside `./packages/KTable/package.json` this is what was added

``` json
{
  "peerDependencies": {
    "@vue/composition-api": "^1.2.4"
  }
}
```

The **_parent application_** must then manually add `@vue/composition-api` to it's `package.json` as a dependency if utilizing any of the components depending on the package (this is now outlined in the docs for each corresponding component, currently `KTable` and `KCardCatalog`).

This is required as the component _must_ use the same instance of the plugin as the parent application, meaning that the plugin has to be added to the parent application's dependencies. [This is further outlined in an issue on the composition api repo](https://github.com/vuejs/composition-api/issues/792).

> Once Vue 2.7 is released, `@vue/composition-api` will be directly shipped with it, so this extra step will no longer be needed.

The quote above, taken from the issue, means that we should plan to upgrade Konponents (and any parent Kong-owned applications, such as `khcp`, `kong-admin`, etc.) to `2.7` when released, which will make utilizing the library easier on the user.

#### Build Process

Update the build process to copy necessary config files in a separate step, leaving them in place until `yarn build:cli` finishes, then remove from the individual `/package/*` directories, as other steps in the build process will fail if `postcss.config.js` files are not left in place (e.g. in `KIcon` package) until the completion of all package builds.

#### Docs

Update docs for `KTable` and `KCardCatalog` for usage with Composition API.

### PR Checklist

### DIFF
--- a/docs/components/card-catalog.md
+++ b/docs/components/card-catalog.md
@@ -1,5 +1,14 @@
 # Card Catalog
 
+:::tip Note
+The `KCardCatalog` component requires the [`@vue/composition-api`](https://github.com/vuejs/composition-api) package as a `peerDependency`. You must **manually** add the package to your host project by running the following
+
+``` shell
+yarn add @vue/composition-api
+```
+
+:::
+
 **KCardCatalog** - A grid view of KCards
 
 <KCardCatalog :items="getItems(5)" />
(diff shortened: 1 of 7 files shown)


kaiarrowood (approved):
This is great!! Thanks Adam!!

PR #436, opened 2021-11-23
Fix reloading pagination component in KTable

Pan-Bubr:
### Summary
Changing page size triggers a reload in KTable which set KPagination back to initial values. Now it doesn't happen.

#### Changes made:
* added initial page size setting for KPagination
* fixed setting a correct initial page size in KTable after KPagination rerenders during loading

### PR Checklist


kaiarrowood (approved):
This works great! Thanks for the super fast turnaround. 🐉  🔥 🔥

PR #439, opened 2021-11-23
refactor(kicon): remove domparser api from kicon

adamdehaven:
### Summary

Refactor `KIcon.vue` to remove the `DOMParser` Web API while keeping all existing functionality ( 🟢  &nbsp; backwards-compatible). 

> **Consuming applications may need to alter the `viewbox` prop on any `KIcon` elements after upgrading to ensure proper display. In addition, check your consuming project's styles for any CSS rules that target `svg` elements**

#### Changes made:
* Refactored the `<slot/>` functionality -- it's a little more DOM manipulation; however, all elements are valid HTML (including the temporary `svg > svg` structure.
* Updated tests to reflect new `KIcon` structure.
* Swapped the `color`/`secondaryColor` svg properties in `icn-warning.svg` to produce the expected result; also updated `KEmptyState.vue` when in error state.
* Moved a single SCSS rule outside of the `scoped` block so that it can properly target `<g>` elements that are contained within the component-external SVG.
* Removes the requirement of adding svg rules specifically for Kongponent icons in a consuming project's `vue.config.js` or webpack config.

### PR Checklist


kaiarrowood (approved):
Looks great, thanks Adam!!

PR #445, opened 2021-11-29
chore(kmodal): remove aria-hidden as it isnt necessary (TDX-1659)

davidma415:
### Summary
The KModal has an aria-hidden property, but the component is not rendered when isVisible is false
which means we shouldn't need this property.
#### Changes made:
* Remove `aria-hidden` attribute in `KModal`

### PR Checklist

### DIFF
--- a/packages/KModal/KModal.vue
+++ b/packages/KModal/KModal.vue
@@ -1,7 +1,6 @@
 <template>
   <div
     v-if="isVisible"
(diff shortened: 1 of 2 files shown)


Mierenga (approved):
LGTM

adamdehaven (approved):
LGTM

For context: `aria-hidden` is not needed in this case since the modal is removed from the DOM when `isVisible` is set to false.

PR #448, opened 2021-11-29
chore(kmodal): add aria-modal and aria-level to kmodal

davidma415:
### Summary
Screen Shot 2021-11-29 at 5 13 32 PM
Screen Shot 2021-11-29 at 5 13 37 PM

Addresses a few a11y issues in the `KModal`

#### Changes made:
* fixed blank `aria-modal`
* added `aria-level` to `role="heading"`

### PR Checklist

### DIFF
--- a/packages/KModal/KModal.vue
+++ b/packages/KModal/KModal.vue
@@ -16,8 +16,7 @@
           <div
             v-if="$scopedSlots.title || !hideTitle"
             class="k-modal-header modal-header mb-5"
-            role="heading"
-            aria-level="1">
+            role="heading">
             <slot name="header-content">{{ title }}</slot>
           </div>
           <div class="k-modal-body modal-body">
--- a/packages/KModal/KModal.vue
+++ b/packages/KModal/KModal.vue
@@ -4,7 +4,7 @@
     :aria-label="title"
     class="k-modal"
     role="dialog"
-    aria-modal>
+    aria-modal="true">
     <div
       class="k-modal-backdrop modal-backdrop"
       @click="close">
@@ -16,7 +16,8 @@
           <div
             v-if="$scopedSlots.title || !hideTitle"
             class="k-modal-header modal-header mb-5"
-            role="heading">
+            role="heading"
+            aria-level="1">
             <slot name="header-content">{{ title }}</slot>
           </div>
           <div class="k-modal-body modal-body">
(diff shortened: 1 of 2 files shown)


packages/KModal/KModal.vue
```diff
@@ -16,7 +16,8 @@
           <div
             v-if="$scopedSlots.title || !hideTitle"
             class="k-modal-header modal-header mb-5"
-            role="heading">
+            role="heading"
+            aria-level="1">
```
adamdehaven, line 20: I'm not sure this is necessary or that it will even be picked up since there is an `aria-label` on the parent `.k-modal` element. (See the warning [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role#accessibility_concerns)).

Either way, the value should be lower than 1 since it isn't the page title.
> A level 1 heading, indicated with `aria-level="1"`, usually indicates the main heading of a page

davidma415: @adamdehaven that's a good point, I didn't notice the `aria-label` with the same value. I think it should be fine if we just remove the `role` and `aria-level` attribute?

adamdehaven: You can leave the `role` as that will still call out the text is a heading to screen readers; however, you can remove the `aria-level` attribute.

> If there is more than one heading on the page and the heading hierarchy is defined through the visual presentation, the aria-level attribute should be used to indicate the hierarchical level of the heading.

Since this is the case, the heading is the only one inside the modal (and it matches the label) meaning it wouldn't need a level. 

Really, we should be using `aria-labeledby` as [shown here](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html), but we have an open ticket to improve accessibility that will cover a lot of this.

davidma415: There are some asks for a clean aXe report in the portal, which would likely require having the `aria-level` attribute present. I think for the time being before the change to use `aria-labeledby`, we can set this `aria-level` to 2?

adamdehaven: `aria-level` defaults to `2` if not provided

davidma415: The aXe tool will throw the warning if it's not explicitly provided - would it be okay to include it?

adamdehaven: Sure -- I'm thinking maybe with a value of `3` since it's a modal, but if you want to stick with `2` that's fine as well. Don't forget the test snapshots will need to be updated as well.

Mierenga (approved):
LGTM

PR #452, opened 2021-12-10
fix(kalert): fix invalid html structure

adamdehaven:
## Summary

Fix invalid HTML structure within `KAlert`. Slot elements should not be wrapped with `<span>` tags as the content for a `KAlert` slot is typically HTML, meaning a `<div>` or `<p>` tag inside the slot renders invalid HTML, which is also adversely impacting styles.

#### Changes made:
* Change `slot` wrapping elements to `<div>`
* Add CSS rule to remove margin of last `<p>` tag inside `.k-alert-msg`
* Updated test snapshots to match tag structure.

### PR Checklist

### DIFF
--- a/packages/KAlert/KAlert.vue
+++ b/packages/KAlert/KAlert.vue
@@ -60,20 +60,20 @@
         class="k-alert-icon" />
     </span>
     <div class="k-alert-msg-text">
-      <span
+      <div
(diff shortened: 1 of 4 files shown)


kaiarrowood (approved):
LGTM!

PR #457, opened 2021-12-17
feat(*): fix dependencies and styles

kaiarrowood:
### Summary
Fix dependencies of all kongponents.

#### Changes made:
* Fix docs / pagination example in `KCardCatalog`
* Update dependencies for all kongponents
* Emit `input` event in `KSelect` for `v-bind`
* Adjust imports of scss partials to use partials syntax
* Add `index.js` for `KMenu` to allow using subcomponents
* Make sure to include import of `_variables.scss` if it is used in the style blocks

### PR Checklist

### DIFF
--- a/packages/KSelect/KSelect.vue
+++ b/packages/KSelect/KSelect.vue
@@ -295,6 +295,7 @@ export default {
       })
       this.filterStr = this.appearance === 'dropdown' ? '' : item.label
       this.$emit('selected', item)
+      this.$emit('input', item)
     },
     clearSelection () {
       this.items.forEach(anItem => {
@@ -315,6 +316,8 @@ export default {
 </script>
 
 <style lang="scss">
+@import '~@kongponents/styles/variables';
+
 .k-select {
   width: fit-content; // necessary for correct placement of popup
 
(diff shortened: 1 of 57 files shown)


packages/KSelect/KSelect.vue
```diff
@@ -295,6 +295,7 @@ export default {
       })
       this.filterStr = this.appearance === 'dropdown' ? '' : item.label
       this.$emit('selected', item)
+      this.$emit('input', item)
```
adamdehaven, line 298: Since you're adding an emit for the `input` event, should you add an emit on `change` as well since `change` and `input` are both events fired by `<select>` elements?

kaiarrowood: Good idea, I'll add it!

adamdehaven: It looks like you also need to update the Events section on `select.md` for all new events

adamdehaven (approved):
LGTM, with select event changes (emit and document the events)

packages/KSelect/KSelect.vue
kaiarrowood: Fixed!

PR #458, opened 2021-12-17
chore(node): use node 14

adamdehaven:
### Summary

Use node `14.x` instead of `12.x`

### PR Checklist


chrisvdp:
Looks like you also need to update the `.nvmrc` file. I wonder if we should switch to an approach like https://github.com/actions/setup-node/issues/32#issuecomment-525791142

zekth:
> Looks like you also need to update the `.nvmrc` file. I wonder if we should switch to an approach like [actions/setup-node#32 (comment)](https://github.com/actions/setup-node/issues/32#issuecomment-525791142)

That's the way to do it. +1 on this

chrisvdp (approved):
nice work on the gh action

adamdehaven:
Could have done it without you @chrisvdp; just wouldn't have been as exciting 😂

Issue #466, opened 2021-12-21
[KSkeleton] Problem with styles for card type in the KSkeleton component


LukeSwierlik:
The problem is when a user uses a **KSkeleton** with **card** type and counts for **>3**.

Link for the komponent: 
https://kongponents.netlify.app/components/skeleton.html#cardcount

In the attachment there is screen:
![Screen Shot 2021-12-21 at 11 29 24 AM](https://user-images.githubusercontent.com/40358447/146915050-3eb75b94-c969-4bd6-a411-6c5f4f612b68.png)

PR #467, opened 2021-12-21
style(kskeleton): repair styles for card type

LukeSwierlik:
Issue: #466

#### Changes made:
*

### PR Checklist


adamdehaven:
@LukeSwierlik - thanks for the contribution! I rolled your changes into #468 but needed to revert the class refactor. I also had to remove the `justify-content: space-between` rule as it produces an undesired result if `count` is set to an uneven value, e.g. `5`

5 skeleton cards

PR #470, opened 2022-01-07
fix(kinput): allow numbers to be passed in as value

dustinryerson:
### Summary
Sets the KInput value prop type to `[String, Number]` to prevent warnings in the browser console when using inputs with `type="number"`.

#### Changes made:
* Small addition to KInput value prop type

### PR Checklist

### DIFF
--- a/packages/KInput/KInput.vue
+++ b/packages/KInput/KInput.vue
@@ -63,7 +63,7 @@ export default {
   inheritAttrs: false,
   props: {
     value: {
-      type: String,
+      type: [String, Number],
       default: ''
     },
     label: {


adamdehaven (approved):
Should be used in conjunction with `v-model.number` on the frontend if `Number` is the intended data type.

dustinryerson:
Percy is failing because the default branch was set to `master` in the Percy config. I'm not sure how to bump the Percy check to get it to compare to `main` and not `master` without closing and reopening the PR. @chrisvdp any suggestions?

chrisvdp:
> Percy is failing because the default branch was set to `master` in the Percy config. I'm not sure how to bump the Percy check to get it to compare to `main` and not `master` without closing and reopening the PR. @chrisvdp any suggestions?

You can change it within Percy and then re-run the test suite.

dustinryerson:
> You can change it within Percy and then re-run the test suite.

It's not obvious to me how to do this, I looked both on GitHub and on the Percy dashboard. Do you happen to know how to do it?

PR #493, opened 2022-02-15
feat(ktable): add empty-state-action-button-icon

adamdehaven:
### Summary

Adds a new `emptyStateActionButtonIcon` prop to `KTable` which accepts a `string` name of an icon to display next to the action button. (optional).
  
image

### PR Checklist

### DIFF
--- a/packages/KTable/KTable.vue
+++ b/packages/KTable/KTable.vue
@@ -52,6 +52,15 @@
               appearance="primary"
               @click="$emit('ktable-empty-state-cta-clicked')"
             >
+              <template
+                v-slot:icon
+                v-if="emptyStateActionButtonIcon">
+                <KIcon
+                  :icon="emptyStateActionButtonIcon"
+                  color="white"
+                  view-box="0 0 20 20"
+                  size="16" />
+              </template>
               {{ emptyStateActionMessage }}
             </KButton>
           </template>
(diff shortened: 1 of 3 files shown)


packages/KTable/KTable.vue
```diff
@@ -292,6 +301,13 @@ export default defineComponent({
       type: String,
       default: ''
     },
+    /**
+     * A prop to pass in a custom empty state action message
```
dustinryerson, line 305:
```suggestion
     * A prop to pass in a custom empty state action button icon
```

kaiarrowood (approved):
LGTM!

packages/KTable/KTable.vue
```diff
@@ -52,6 +52,15 @@
               appearance="primary"
               @click="$emit('ktable-empty-state-cta-clicked')"
             >
+              <template
+                v-slot:icon
+                v-if="emptyStateActionButtonIcon">
+                <KIcon
+                  :icon="emptyStateActionButtonIcon"
+                  color="white"
```
dustinryerson, line 60: Non-blocking question: will we always want `white` for the icon color? Hypothetically if for some reason we wanted an icon in a button with the outline style, I don't think the icon would be visible. Possibly an unreasonable edge-case but just wanted to surface it.

dustinryerson (changes requested):
Looking good! Left a non-blocking question and change suggestion.

dustinryerson (approved):
LGTM 🚀

PR #504, opened 2022-02-22
feat(kprompt): add actionPending prop

kaiarrowood:
### Summary
Add `actionPending` prop.

image

#### Changes made:
* Add tests
* Update snaps

### Vue 3 Upgrade

We are currently in the process of upgrading Kongponents to Vue 3. If changes are made to a component or doc file on the `main` branch, a corresponding PR needs to be made into the `next` branch that includes:

- The component feature/fix, updated for Vue 3 and the Composition API.
- Documentation updates for the component changes, as well as updating examples and usage to Vue 3 and the Composition API.
- Updates to the corresponding `.spec.ts` test file(s) to utilize [Cypress Component Testing](https://docs.cypress.io/guides/component-testing/introduction).

**Does your PR modify a component [that already exists on the `next` branch](https://github.com/Kong/kongponents/tree/next/src/components)?**

  - [ ] **Yes**, and there is a corresponding PR to update the component on the `next` branch
    - `LINK_TO_PR_ON_NEXT_BRANCH` (**required**)
  - [x] **No**, the component does not yet exist on `next` branch.

### PR Checklist

### DIFF
--- a/packages/KPrompt/KPrompt.vue
+++ b/packages/KPrompt/KPrompt.vue
@@ -72,6 +72,7 @@
               slot="icon"
               icon="spinner"
               size="16"
+              color="var(--white)"
             />
             {{ actionButtonText }}
           </KButton>
--- a/packages/KPrompt/KPrompt.vue
+++ b/packages/KPrompt/KPrompt.vue
@@ -46,6 +46,7 @@
 
             <KInput
               v-model="confirmationInput"
+              data-testid="confirmation-input"
               class="pt-2" />
           </div>
         </div>
@@ -66,6 +67,12 @@
             :disabled="disableProceedButton"
             class="k-prompt-proceed"
             @click="proceed">
+            <KIcon
+              v-if="actionPending"
+              slot="icon"
+              icon="spinner"
+              size="16"
+            />
             {{ actionButtonText }}
           </KButton>
         </slot>
@@ -111,6 +118,14 @@ export default {
       type: String,
       default: 'Cancel'
     },
+    /**
+     * Boolean to disable action buttons while a submission is occurring. Display
+     * spinner on action button.
+     */
+    actionPending: {
+      type: Boolean,
+      default: false
+    },
     isVisible: {
       type: Boolean,
       default: false
@@ -144,6 +159,10 @@ export default {
       return this.capitalize(this.type)
     },
     disableProceedButton () {
+      if (this.actionPending) {
+        return true
+      }
+
       if (!this.confirmationText.length) {
         return false
       }
--- a/docs/components/prompt.md
+++ b/docs/components/prompt.md
@@ -100,6 +100,31 @@ Change the text content of the close/cancel button.
 />
 ```
 
+### actionPending
+
+This boolean indicates if an action is being taken on the dialog and we should disable the action button
+to prevent spam clicking.
+
+<KButton appearance="primary" @click="pendingIsOpen = true">Prompt</KButton>
+
+<KPrompt
+  :is-visible="pendingIsOpen"
+  message="Click Cancel to close me"
+  :action-pending="true"
+  @canceled="pendingIsOpen = false"
+  @proceed="pendingIsOpen = false"
+/>
+
+```vue
+<KPrompt
+  :is-visible="pendingIsOpen"
+  message="Click Cancel to close me"
+  :action-pending="true"
+  @canceled="pendingIsOpen = false"
+  @proceed="pendingIsOpen = false"
+/>
+```
+
 ### type
 
 This prompt determines the look and feel of the dialog. Can be `danger`, `warning`, or `info`. Defaults to `info`.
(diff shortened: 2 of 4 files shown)


packages/KPrompt/KPrompt.vue
```diff
@@ -66,6 +67,12 @@
             :disabled="disableProceedButton"
             class="k-prompt-proceed"
             @click="proceed">
+            <KIcon
+              v-if="actionPending"
+              slot="icon"
+              icon="spinner"
+              size="16"
+            />
```
adamdehaven, line 75: nit: The icon looks a little small in the button and the gray color is a little hard to see when the button is disabled

image

packages/KPrompt/KPrompt.vue
```diff
@@ -144,6 +159,10 @@ export default {
       return this.capitalize(this.type)
     },
     disableProceedButton () {
+      if (this.actionPending) {
+        return true
+      }
```
adamdehaven, line 164: When the Proceed button is disabled, is your thought that they clicked the button to proceed and it's finishing a process before closing the prompt?

If yes, I believe you should disable the cancel button, `esc` press, and close button click to prevent them from closing the prompt before the process finishes as well.

adamdehaven (changes requested):

packages/KPrompt/KPrompt.vue
kaiarrowood: I had thought about this but we don't currently do this anywhere else in the UI. I think we are only trying to prevent spam clicking, not closing the prompt.

packages/KPrompt/KPrompt.vue
kaiarrowood: The font size is 16, I can't make the spinner any bigger without getting weird positioning of things. I DID change the color to white to improve visibality.

image

packages/KPrompt/KPrompt.vue
kaiarrowood: The action completes in the background. We don't want to interrupt or cancel the call at that stage.

adamdehaven: That's why I'm thinking we would disable everything. If the user closes the prompt before it completes, we will then be emitting 2 events back to the caller, `canceled` and `proceed` (assuming the `proceed` will still be called after the action completes). 

If the consuming app has logic to respond to both events, we could potentially be giving them a "false cancel" even though the user actually did choose to proceed. 

Actually, now I'm curious -- how can the consuming UI ever set `actionPending`? The user clicks the `proceed` button, and immediately the modal is closed unless the `actionPending` prop is already somehow set to true?

adamdehaven:
I'm also unsure if the default `enter` keypress should be hooked up to the `proceed` action since that would automatically dismiss a danger/warning modal. Maybe that should only be the case if the `type` === `info`?

packages/KPrompt/KPrompt.vue
adamdehaven: ^ more on this (commenting here since you're in meetings all day)

> This boolean indicates if an action is being taken on the dialog and we should disable the action button to prevent spam clicking.

Since the consuming app is only notified when they click the button (not before) how would this prop get switched to true to show the pending state on click?

kaiarrowood: This is interesting, so the typical implementation I am expecting is there will be state transitions happening in the submit button's action handler like:

image

`actionPending` is determined by this state:

image

Also, typically it is ok if both `proceed` and `cancel` events are triggered because in general the only thing happening in a cancel is 
- the dialog is closed
- form values are cleared
- we may send a state change event (which would be handled correctly because it's a no-op if sending `CANCEL` event when in `pending` state, we only take action `RESOLVE` or `REJECT` events from that state)

State machine:

image

kaiarrowood:
> I'm also unsure if the default `enter` keypress should be hooked up to the `proceed` action since that would automatically dismiss a danger/warning modal. Maybe that should only be the case if the `type` === `info`?

I am against this change because it isn't accessible. Clicking `OK` is equivalent to hitting `Enter` for a keyboard only user.

adamdehaven:
> I am against this change because it isn't accessible. Clicking OK is equivalent to hitting Enter for a keyboard only user.

So the common pattern I've seen is that instead of binding the click event to the modal itself when it has focus, instead you would bind the default focus to the cancel button. If the user pressed enter, it would trigger cancel by default instead of proceed. If they want to proceed, they would tab to that button (so still accessible).

Regardless, I'm probably overthinking it for now.

packages/KPrompt/KPrompt.vue
adamdehaven: Thanks for the context. 

So based on your example, the `@proceed` emitted event doesn't actually close the modal -- the user has to handle that themselves, right? I was missing that piece; I was thinking clicking the proceed button also closed the prompt when I read through this at the end of the day.

adamdehaven (approved):
LGTM -- nice work! 🚀

PR #525, opened 2022-03-18
fix(kselect): set up v-model

kaiarrowood:
### Summary
Fix `v-model` to work correctly on `KSelect`, should hook up to just the item's value you not the entire object.

#### Changes made:
* Fix v-model
* Update docs/tests

### Vue 3 Upgrade

### PR Checklist


adamdehaven (changes requested):
If I type in the `v-model` test field on the docs site, it updates the value to whatever I type, even if I don't select an item from the list. This allows me to set the `v-model` to any value I'd like. I think the logic needs to be updated to only update the model value when an item from the list is selected or else the user can set the value to anything just by typing in the input.

image

adamdehaven (approved):
LGTM

PR #563, opened 2022-04-08
fix(ktable): add sortKey option to headers

dustinryerson:
### Summary
Adds a `sortKey` option to the table headers object, which is useful when you want to sort the column based on a different key in the data.

#### Changes made:
* Adds a `sortKey` option to the table headers object
* Adds a usage example to the table headers documentation

### Vue 3 Upgrade

### PR Checklist

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -292,6 +292,7 @@ Pass in an array of header objects for the table.
 | `label` | string | The label displayed on the table for the column
 | `sortable` | boolean | Enables or disables server-side sorting for the column (`false` by default)
 | `hideLabel`| boolean | Hides or displays the column label (useful for actions columns)
+| `sortKey` | string | A key returned in fetcher response data used to sort column
 
 :::tip Note
 `sortable` columns emit a `sort` event when clicked, returns:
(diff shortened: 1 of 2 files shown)


dustinryerson:
@adamdehaven @kaiarrowood I didn't add a test for this because the client-side sorting doesn't seem to be tested in `KTable.spec.js`. Let me know if you'd like to see a test and if so, the best place to put it. Thanks!

dustinryerson:
Also, I'll open a PR against the `next` branch once this merges, just in case any changes are requested for this PR.

docs/components/table.md
```diff
@@ -326,6 +327,86 @@ Example headers array:
 </script>
 ```
 
+#### Note about the `sortKey` option
+
+There may be times when data returned in a fetcher is not human-readable, but still needs to be displayed and  sorted in a table in a human-readble format. We cannot depend on the human-readable format sorting the column correctly - in these cases it is useful to specify which key in the data should be used to sort the column. An example of this would be when a date/time is returned in milliseconds.
```
kaiarrowood, line 332:
```suggestion
There may be times when data returned in a fetcher is not human-readable, but still needs to be displayed and sorted in a table in a human-readble format. We cannot depend on the human-readable format sorting the column correctly - in these cases it is useful to specify which key in the data should be used to sort the column. An example of this would be when a date/time is returned in milliseconds.
```

kaiarrowood (commented):
This is an interesting concept, but I wonder if we could make this a little bit more flexible if we instead offered a `sortTransformer` prop that could take a function. The function takes the row of data as a parameter and returns the value that should be sorted by.

I am currently implementing my own sort function to handle scenarios where I have to transform the data. In some scenarios I may want to transform the field data into something that can be sorted, or even enable sorting on a column that doesn't actually have an equivalent prop in the row data at all.

image

In the example above, I have added an `applied_to` column in the plugin `headers` entry, but the value displayed for that column is actually determined by three different properties on the row. If I could pass a transformer function I would be able to handle this scenario. What do you think?

https://github.com/Kong/khcp-ui/blob/main/client/src/pages/Configuration/Plugins/List.vue#L325

dustinryerson:
@kaiarrowood good call, closing this PR in favor of https://github.com/Kong/kongponents/pull/566

PR #580, opened 2022-04-15
feat(ktable): add sortHandlerFn support

kaiarrowood:
### Summary
Adds a `sortHandlerFn` prop allowing custom sorting on table columns.

#### Changes made:
* Adds `sortHandlerFn` prop
* Adds `useSortHandlerFn` property to `headers` object
* Updates docs to detail the new prop and provide a usage example

Ports changes from https://github.com/Kong/kongponents/pull/566

### Vue 3 Upgrade

### PR Checklist


docs/components/table.md
```diff
@@ -66,9 +66,130 @@ Set this to `true` to disable ablity to sort.
...
+```vue
+<template>
+  <KTable
+    :fetcher="fetcher"
+    :headers="headers"
+    :sortHandlerFn="sortHandlerFn"
+    enable-client-sort
+  />
+</template>
+<script>
```
adamdehaven, line 108: nit: can you add an empty line before the script tag?

Also, we should show `lang="ts"` for all the examples with script tags.

docs/components/table.md
```diff
@@ -66,9 +66,130 @@ Set this to `true` to disable ablity to sort.
...
+export default {
```
adamdehaven, line 109: question: should this example use `defineComponent` and the `setup()` instead?

src/components/KTable/KTable.vue
```diff
@@ -207,6 +207,8 @@ export interface TableHeader {
   label: string
   sortable: boolean
   hideLabel: boolean
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  useSortHandlerFn: Function
```
adamdehaven, line 211: TBH this isn't a valid type 

Could do something like

`useSortHandlerFn: (payload: Record<string, any>) => void`

If you wanted to get real fancy, you could add a type interface for payload instead of using this default object.

src/components/KTable/KTable.vue
```diff
@@ -812,44 +818,37 @@ export default defineComponent({
...
     }
+  }
 
-    &.has-hover {
-      tbody tr:hover td:first-child {
-        border-left: 3px solid var(--KTableBorder, var(--steel-300, color(steel-300)));
+  tbody {
+    td {
+      button,
+      .k-button {
```
adamdehaven, line 849: You may need to wrap this is a deep selector

```suggestion
      :deep(.k-button) {
```

It's typically needed when you're targeting a class in a child component in Vue 3. 

I didn't check if it's needed for the caret icons above.

adamdehaven (changes requested):

src/components/KTable/KTable.vue
kaiarrowood: I'm confused by this. I thought the way to access child components is to use an `unscoped` SCSS block. Do you need `:deep` INSTEAD of `unscoping` or AS WELL AS?

docs/components/table.md
kaiarrowood: Seems like unnecessary overcomplication in the docs

src/components/KTable/KTable.vue
adamdehaven: TL;DR - if it's not `scoped` then I think it's fine.

I think we want to continue to use scoped where we can instead of unscoped (since this is a component library).

If the block isn't scoped (I didn't check when I wrote this) then it's probably fine, but just verify the rule is actually being applied?

Vue 3 handles scss within child components a little differently if `scoped`: https://vuejs.org/api/sfc-css-features.html#deep-selectors

kaiarrowood: A lot of styles were lost when you originally ported of `KTable` due to the scoping. One of the things that I did in this PR was to split the SCSS into two blocks, one scoped, one unscoped. Only the necessary styles were moved into the unscoped section 👍🏻

src/components/KTable/KTable.vue
kaiarrowood: Fixed

docs/components/table.md
kaiarrowood: Fixed

adamdehaven (approved):

PR #603, opened 2022-04-20
fix(ktable): add data-testid to empty/error state buttons

dustinryerson:
### Summary
Adds a `data-testid` attribute to empty and error state action buttons using the values passed into the `emptyStateActionMessage` and `errorStateActionMessage` props. The data-testid will reflect the text of the button: if an action button reads "Create New Service", the attribute will be `data-testid="create-new-service"`. This allows the data-testids to be created on the action buttons without needing to manually build out the empty/error states via slots and specify the data-testids by hand.

#### Changes made:
- Add `data-testid` to error state and empty state action buttons.

### Vue 3 Upgrade

### PR Checklist

### DIFF
--- a/packages/KTable/KTable.vue
+++ b/packages/KTable/KTable.vue
@@ -665,7 +665,7 @@ export default defineComponent({
     }
 
     const getTestIdString = (message) => {
-      return message.toLowerCase().replaceAll(' ', '-')
+      return message.toLowerCase().replace(/[^[a-z0-9]/gi, '-')
     }
 
     watch(() => props.searchInput, (newValue) => {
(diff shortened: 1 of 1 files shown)


packages/KTable/KTable.vue
```diff
@@ -662,6 +664,10 @@ export default defineComponent({
       }
     }
 
+    const getTestIdString = (message) => {
+      return message.toLowerCase().replaceAll(' ', '-')
```
adamdehaven, line 668: suggestion: Just in case someone passes in something crazy. This will replace any non-alphanumeric character.
```suggestion
      return message.toLowerCase().replace(/[^[a-z0-9]/gi, '-')
```

adamdehaven (dismissed):
Looks good; left 1 non-blocking suggestion

packages/KTable/KTable.vue
dustinryerson: @adamdehaven good suggestion, I committed it via GitHub, so let's see if things are happy now

adamdehaven (approved):
LGTM - build should pass this time

PR #608, opened 2022-04-21
fix(ktable,kcatalog): fix issues

kaiarrowood:
### Summary
Fix miscellaneous issues with `KTable` and `KCatalog`.

#### Changes made:
- remove code causing duplicate fetch on initial load
- fix row/cell click handling for Vue3
- trigger row clicks on the `td` instead of the `tr`

### Vue 3 Upgrade

### PR Checklist

### DIFF
--- a/packages/KTable/KTable.vue
+++ b/packages/KTable/KTable.vue
@@ -131,7 +131,6 @@
             :tabindex="isClickable ? 0 : null"
             :role="isClickable ? 'link' : null"
             v-bind="rowAttrs(row)"
-            v-on="hasSideBorder ? tdlisteners(row, row) : null"
           >
             <template>
               <td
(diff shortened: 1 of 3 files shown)


packages/KTable/KTable.vue
```diff
@@ -131,14 +131,14 @@
             :tabindex="isClickable ? 0 : null"
             :role="isClickable ? 'link' : null"
             v-bind="rowAttrs(row)"
-            v-on="trlisteners(row, 'row')"
+            v-on="hasSideBorder ? tdlisteners(row, row) : null"
```
adamdehaven, line 134: question: why would the click events only be bound if hasSideBorder is true? 

I’m probably missing something, and I know you’re going to review Dustin’s changes, but I want to understand this lol

adamdehaven (approved):

packages/KTable/KTable.vue
kaiarrowood: You had mentioned that if we didn't have `border-collapse` (which `hasSideBorder` doesn't) might cause space between cells. Seems like a non-issue though, so backed out the code.

PR #609, opened 2022-04-21
fix(ktable): create listeners even when td is not clicked target

dustinryerson:
### Summary
This resolves the table row click issue we've seen on a number of our tables.

#### Changes made:
- Removed a condition that limited `@row:click` event listening to just `<td></td>` elements, and not any child elements that may exist within `<td></td>` elements.

### Vue 3 Upgrade

### PR Checklist

### DIFF
--- a/packages/KTable/KTable.vue
+++ b/packages/KTable/KTable.vue
@@ -535,7 +535,7 @@ export default defineComponent({
         return {
           ...pluckedListeners,
           click (e) {
-            if (e.target.tagName === 'TD' && pluckedListeners['click']) {
+            if (pluckedListeners['click']) {
               pluckedListeners['click'](e, entity, type)
             }
           }


xitongzou:
https://github.com/Kong/kongponents/pull/239

You may want to do some more testing on this, I think there was a reason why I limited this to TD https://github.com/Kong/kongponents/pull/244 
https://konghq.atlassian.net/browse/INTF-2543 was the original bug which caused me to add back in this condition
This may not be a problem anymore but just double check all the individual elements inside a ktable row are still firing their events as expected

kaiarrowood:
> #239
> 
> You may want to do some more testing on this, I think there was a reason why I limited this to TD #244 https://konghq.atlassian.net/browse/INTF-2543 was the original bug which caused me to add back in this condition This may not be a problem anymore but just double check all the individual elements inside a ktable row are still firing their events as expected

Thanks @xitongzou I am going to try the fix in the `beta` branch and use it in the kongponents sandbox repo to verify this fix won't retrigger the issue you linked.

kaiarrowood (approved):
Verified this fixes the issue in kongponents-sandbox.

Nice find @dustinryerson !!

PR #613, opened 2022-05-03
fix(ktable): row click handling

kaiarrowood:
### Summary
Fix handling of row click events.

#### Changes made:
- Ignore clicks from `a`, `button`, and `input`
- Ignore clicks from popover content
- Update docs

### Vue 3 Upgrade

### PR Checklist

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -607,7 +607,7 @@ Bind any DOM [events](https://developer.mozilla.org/en-US/docs/Web/Events) to va
 
 `@row:<event>` - returns the `Event`, the row item, and the type: `row`
 
-To avoid firing row clicks by accident, the row click handler ignores events coming from `a`, `button`, `input`, and `select` elements (unless they are `disabled`). As such click handlers attached to these element types do not require stopping propagation via `@click.stop`.
+To avoid firing row clicks by accident, the row click handler ignores events coming from `a`, `button`, `input`, and `select` elements (unless they have the `disabled` attribute). As such click handlers attached to these element types do not require stopping propagation via `@click.stop`.
 
 <KTable
   :headers="tableOptionsLinkHeaders"
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -601,24 +601,24 @@ export default {
 
 ## Events
 
-Bind any DOM [events](https://developer.mozilla.org/en-US/docs/Web/Events) to
-various parts of the table. We support events on both table rows and cells in addition to click elements inside a row (ie. buttons or hyperlinks) without triggering the a row or cell event. You can also add logic to check for `metakey` to support cmd/ctrl when clicking. Examples highlighted below.
+Bind any DOM [events](https://developer.mozilla.org/en-US/docs/Web/Events) to various parts of the table. We support events on both table rows and cells, but must be careful with clickable content in rows when row click is enabled. You can also add logic to check for `metakey` to support cmd/ctrl when clicking. Examples highlighted below.
 
 ### Rows
 
 `@row:<event>` - returns the `Event`, the row item, and the type: `row`
 
 ```vue
-<KTable @row:click="rowHandler" @row:dblclick="rowHandler">
+<KTable @row:click="rowHandler">
 ```
+
+To avoid firing row clicks by accident, the row click handler ignores events coming from `button`, `input`, and `a` tags. As such click handlers attached to these element types do not require stopping propagation via `@click.stop`.
 
 <KTable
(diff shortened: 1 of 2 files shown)


packages/KTable/KTable.vue
```diff
@@ -523,6 +523,7 @@ export default defineComponent({
       return (entity, rowData) => {
         const rowListeners = pluckListeners('row:', ctx.listeners)(rowData, 'row')
         const cellListeners = pluckListeners('cell:', ctx.listeners)(entity, 'cell')
+        const ignoredElements = ['button', 'input', 'a']
```
xitongzou, line 526: is there any other element that could possible exist inside a row that could fire events besides these? just making sure we have everything covered

kaiarrowood: Also added `select`, I can't think of any others and it's suprisingly hard to google for.

adamdehaven: Not to be the party-pooper, but what if the user wants a click event on a `div` ? or an `icon` (whatever element that would be)

kaiarrowood: Not sure I follow @adamdehaven , these are the elements that will **NOT** trigger row click event.

kaiarrowood: As shown in the docs, just add `@click.stop`.

`<template v-slot:wrapped>
    <div>Row click event? <div @click.stop="buttonHandler"><KBadge appearance="success">Button click</KBadge></div></div>
  </template>`

adamdehaven: Ohhhh so the list `button, a, input` is a list of "we already know we should ignore these things? 

If yes, that makes more sense 👍🏼

docs/components/table.md
```diff
@@ -601,24 +601,24 @@ export default {
...
-various parts of the table. We support events on both table rows and cells in addition to click elements inside a row (ie. buttons or hyperlinks) without triggering the a row or cell event. You can also add logic to check for `metakey` to support cmd/ctrl when clicking. Examples highlighted below.
+Bind any DOM [events](https://developer.mozilla.org/en-US/docs/Web/Events) to various parts of the table. We support events on both table rows and cells, but must be careful with clickable content in rows when row click is enabled. You can also add logic to check for `metakey` to support cmd/ctrl when clicking. Examples highlighted below.
 
 ### Rows
 
 `@row:<event>` - returns the `Event`, the row item, and the type: `row`
 
 ```vue
-<KTable @row:click="rowHandler" @row:dblclick="rowHandler">
+<KTable @row:click="rowHandler">
```
adamdehaven, line 611: You need to put a "fake" fetcher here since we made it a required prop

docs/components/table.md
```diff
@@ -601,24 +601,24 @@ export default {
...
 ```
 
+To avoid firing row clicks by accident, the row click handler ignores events coming from `a`, `button`, `input`, and `select` elements. As such click handlers attached to these element types do not require stopping propagation via `@click.stop`.
+
```
adamdehaven, line 615: What if the user still wants the row click to happen even if there is one of these elements that is clicked?

Maybe we change the logic to "if one of those elements AND it does not have a `disabled` attribute"?

packages/KTable/KTable.vue
```diff
@@ -532,7 +533,8 @@ export default defineComponent({
           ...rowListeners,
           ...cellListeners,
           click (e) {
-            if (e.target.tagName === 'TD' && (rowListeners.click || cellListeners.click)) {
+            const isPopoverContent = e.target.className.includes('k-popover')
+            if (!ignoredElements.includes(e.target.tagName.toLowerCase()) && !isPopoverContent && (rowListeners.click || cellListeners.click)) {
```
adamdehaven, line 537: Again, I think we should only ignore if the `ignoredElements` do not have a `disabled` attribute (this allows the user a way to bypass if desired)

adamdehaven: Something like `&& !e.target.hasAttribute('disabled')`

(Untested code... just an idea)

docs/components/table.md
kaiarrowood: Good call, will add.

packages/KTable/KTable.vue
kaiarrowood: Added

docs/components/table.md
kaiarrowood: Added

docs/components/table.md
kaiarrowood: Removed this example since there is a fleshed out example beneath

PR #623, opened 2022-05-10
fix(kicon): watch icon color to apply updates

dustinryerson:
### Summary
This is part two of a two part fix that resolves an issue where the icon color within a KButton was not being updated when the disabled attribute on the KButton was toggled.

When the value of the `color` prop was updated, it was not being applied to the icon.

Part one of this fix is here: https://github.com/Kong/kongponents/pull/622

### Vue 3 Upgrade

### PR Checklist

### DIFF
--- a/packages/KIcon/KIcon.vue
+++ b/packages/KIcon/KIcon.vue
@@ -137,7 +137,6 @@ export default {
     // Ensure icon color is updated when color prop value changes
     color: {
       handler () {
-        this.svg = this.$refs.svgWrapper ? this.$refs.svgWrapper.querySelector('svg:not(.slot-content)') : null
         this.recursivelyCustomizeIconColors(this.svg)
       }
     }
--- a/packages/KIcon/KIcon.vue
+++ b/packages/KIcon/KIcon.vue
@@ -133,6 +133,15 @@ export default {
       return this.svg ? (this.viewBox || (this.svg && this.svg.getAttribute('viewBox')) || DEFAULTS.viewBox) : DEFAULTS.viewBox
     }
   },
+  watch: {
+    // Ensure icon color is updated when color prop value changes
+    color: {
+      handler () {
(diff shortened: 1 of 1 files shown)


packages/KIcon/KIcon.vue
```diff
@@ -133,6 +133,15 @@ export default {
       return this.svg ? (this.viewBox || (this.svg && this.svg.getAttribute('viewBox')) || DEFAULTS.viewBox) : DEFAULTS.viewBox
     }
   },
+  watch: {
+    // Ensure icon color is updated when color prop value changes
+    color: {
+      handler () {
+        this.svg = this.$refs.svgWrapper ? this.$refs.svgWrapper.querySelector('svg:not(.slot-content)') : null
```
adamdehaven, line 140: I’m unsure about setting `this.svg` to `null` here. Unsure what else this could cause

dustinryerson: I pulled that from https://github.com/Kong/kongponents/blob/main/packages/KIcon/KIcon.vue#L140, should we be doing this differently in both places, or just on this update?

dustinryerson: @adamdehaven bumping my question above 👆

adamdehaven: Oops, I missed this.

So we setup `this.svg` to the element or`null` in the `mounted()` function.. why does it need to be set in the watcher as well? Can't the function you're calling access `this.svg` already if it is set?

adamdehaven: @dustinryerson ^

dustinryerson: @adamdehaven nope you are totally right, I don't need that line. Thanks for catching that 👍

adamdehaven (approved):
LGTM

PR #628, opened 2022-05-13
fix: add data-testids to action buttons

dustinryerson:
### Summary
Adds data-testids to action buttons.

### Vue 3 Upgrade

### PR Checklist

### DIFF
--- a/packages/KAlert/KAlert.vue
+++ b/packages/KAlert/KAlert.vue
@@ -21,6 +21,7 @@
       v-if="dismissType === 'icon'"
       type="button"
       aria-label="Close"
+      data-testid="k-alert-close-button"
       class="close"
       @click="dismissAlert">
       <KIcon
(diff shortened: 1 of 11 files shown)


adamdehaven (commented):
The changes look good; my only concern is that we do not typically add `data-testid` attributes to Kongponents. 

Typically the consuming application would add test attributes on the component when it is being used (which are applied to the top-most element, unless `inheritAttrs = false`) and then Cypress or the test library would traverse down the DOM from that element to grab what is needed (e.g. an action button). 

I'm not sure this implementation hurts anything specifically for Konnect; however, if another consumer of the component library utilizes a different testing pattern, these attributes could possibly conflict (rare).

dustinryerson:
> The changes look good; my only concern is that we do not typically add `data-testid` attributes to Kongponents.
> 
> Typically the consuming application would add test attributes on the component when it is being used (which are applied to the top-most element, unless `inheritAttrs = false`) and then Cypress or the test library would traverse down the DOM from that element to grab what is needed (e.g. an action button).
> 
> I'm not sure this implementation hurts anything specifically for Konnect; however, if another consumer of the component library utilizes a different testing pattern, these attributes could possibly conflict (rare).

Good point, closing PR.

PR #631, opened 2022-05-18
fix(kbutton) disable pointer-events on caret when button is disabled

Mierenga:
### Vue 3 Upgrade

### PR Checklist


Mierenga:
@adamdehaven your suggestion to make it more general for any icon sounds good. The issue I'm facing with this is that the tests I added aren't working. I was expecting that clicking the caret when a button is not disabled would toggle `isOpen`, but it doesn't. I guess the issue is that `isOpen` is a prop, so it is actually controlled by the parent element and not toggled inside KButton. So I'm not sure how I can write a test case for this (if we need one).

adamdehaven:
@Mierenga I'm pairing with @davidma415 now -- the issue with the test is that you were looking at the prop to change; however, `isOpen` does not change since props are not reactive within the component.

We're looking at the `is-active` class as a diff means of testing... still 👀

davidma415:
So I did debugging with this, I now have the test case where clicking it will apply `is-active` - but I think the ignore case isn't working because I don't think vue-test-utils applies styles to the component during tests. I believe it will only apply inline styles if you pass them in.

Correct me if I'm wrong - but I logged the style property `.kong-icon` and there was nothing applied, even though there are styles. So I'm not sure how we can go around this / test this.

adamdehaven:
> I don't think vue-test-utils applies styles to the component during tests. I believe it will only apply inline styles if you pass them in.

This is correct. 

Given that we’re just disabling pointer-events on the icon, I’m ok if there isn’t test coverage here for this use-case. 

When I port the change to the `beta` branch, it utilizes Cypress component test runner and we can then write a test that can check for styles with `invoke`

PR #638, opened 2022-05-20
feat(ktable): support for conditional events

kaiarrowood:
# Summary

Backport https://github.com/Kong/kongponents/pull/637

Edit: we may end up not backporting this after all

## Vue 3
Better handling for condition events

---

## PR Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,6 +22,7 @@ module.exports = {
     'object-property-newline': 'error',
     'import/no-named-default': 0,
     'object-curly-spacing': 0,
+    'no-extra-boolean-cast': 0,
     'lines-between-class-members': ['error', 'always'],
     'padding-line-between-statements': [
       "error",
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -599,20 +599,35 @@ export default {
 </script>
 ```
 
-## Events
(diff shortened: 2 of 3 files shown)


.eslintrc.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
     'object-property-newline': 'error',
     'import/no-named-default': 0,
     'object-curly-spacing': 0,
+    'no-extra-boolean-cast': 0,
```
adamdehaven, line 25: question: Was this intentional?

kaiarrowood: Yes, it was mad about the `!!` in the conditional

docs/components/table.md
```diff
@@ -599,20 +599,35 @@ export default {
 </script>
 ```
 
-## Events
+Bind DOM [events](https://developer.mozilla.org/en-US/docs/Web/Events) to various parts of the table. We support events on both table rows and cells, but must be careful with clickable content in rows when row click is enabled. You can also add logic to check for `metakey` to support cmd/ctrl when clicking. Examples highlighted below.
```
adamdehaven, line 602: It looks like you accidentally deleted the `## Events` header

docs/components/table.md
```diff
@@ -632,10 +647,12 @@ To avoid firing row clicks by accident, the row click handler ignores events com
 </KTable>
 
 ```html
+<KInputSwitch v-model="enableRowClick" :label="enableRowClick ? 'Row clicks enabled' : 'Row clicks disabled'" />
```
adamdehaven, line 650: These switches aren't working -- the row clicks are not firing

adamdehaven (changes requested):

PR #665, opened 2022-06-23
feat(kpagination/ktable): support offset-based pagination

dustinryerson:
# Summary
Adds support for offset-based pagination.

In practice, the following would need to be done in order to enable offset-based pagination:
1. Use the new `paginationType` prop for KTable: `<KTable :pagination-type="offset" />`
2. Use the new `offset` param in the fetcher function passed to KTable: `fetcher ({ offset }) {}`

**Assumptions:**
This implementation assumes that the API endpoint used in the fetcher function returns the next offset value in the response. Here is an example of the assumed format:
```json
{
  "data": [{}],
  "pagination": {
    "offset": "cGFnZS1udW1iZXI6Mg"
  },
}
```

This implementation also assumes that the API endpoint used in the fetcher function properly handles offset usage. Here's an example:
`GET /api/runtime_groups/6e30594e-9a90-4e0f-94f8-90b923696c83/services?size=100&offset=cGFnZS1udW1iZXI6Mg`

## Screenshot:
Screen Shot 2022-06-22 at 9 24 15 AM

## Vue 3

---

## PR Checklist

### DIFF
--- a/docs/components/pagination.md
+++ b/docs/components/pagination.md
@@ -188,6 +188,18 @@ export default defineComponent({
 </script>
 ```
 
+### paginationType
+
+Pass in the type of pagination to be used. Options are `default` (page/pageSize) or `offset` (offset/pageSize)
+
+### offsetPrevButtonDisabled
+
+Pass in a boolean value for whether or not the offset-based Previous button should be disabled
+
+### offsetNextButtonDisabled
+
+Pass in a boolean value for whether or not the offset-based Next button should be disabled
+
 ## Usage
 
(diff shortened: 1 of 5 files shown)


kaiarrowood (changes requested):
Seeing the same style issue as was on the `main` branch


image
**

kaiarrowood:
Need to port over the disabled state changes

kaiarrowood (dismissed):
Need to port https://github.com/Kong/kongponents/pull/671

PR #671, opened 2022-06-30
fix(kpagination): disable buttons when necessary

dustinryerson:
# Summary

This fix adds more cases for disabled state handling for the prev/next page buttons.

It's possible that when the data returned is either less than the page size or the last page of data, no pagination data will be returned in the response (i.e. a `next` URL or an `offset` value). This is our current use case, and this fix should now handle these cases.

Cases tested:
1. When the data loaded on initial page load is less than the page size
3. When the data loaded after clicking the "next" page button is less than the page size (i.e. the last page of results)
4. When the current page is not the first page, data is refetched (i.e. a search)

Note about cbdaf702a97eaaa1d6adcfe29c4aaace08833f83:
I included this change because I found the following bug:
1. `enable-client-sort` is disabled on a KTable instance
2. `sort: true` is set for one or more of the headers for that table
3. Clicking a sortable header would trigger the `revalidate()` call. When used in combination with `pagination-type="offset"`, clicking a sortable header would trigger the `revalidate()` call, fetching the next page of results

Are there cases where we would want to revalidate after clicking a sortable header?

## Vue 3

---

## PR Checklist

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -1344,7 +1344,7 @@ export default {
       row: null,
       eventType: '',
       offsetPaginationPageSize: 15,
-      offsetPaginationData: [],
+      offsetPaginationData: {},
       headers: [
         { label: 'Title', key: 'title', sortable: true },
         { label: 'Description', key: 'description', sortable: true },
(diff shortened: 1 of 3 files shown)


adamdehaven:
> Are there cases where we would want to revalidate after clicking a sortable header?

I think you would want to revalidate on sorting in case there are new/removed records after the user took an action (i.e. sorting). Are there reasons NOT to do this?

dustinryerson:
@adamdehaven I updated the PR with https://github.com/Kong/kongponents/pull/671/commits/5250a19966d646996c945c497ed1a55bd6cfe3b5 and added the revalidate back in.

kaiarrowood (approved):
LGTM! 🚀

PR #691, opened 2022-07-14
feat(kselect): misc fixes [beta]

kaiarrowood:
# Summary

- Fix `width` prop handling to ensure selected item, input, and dropdown are the same size
- Add support for `%` for `width` props
- Add `enableFilter` prop to control filter support on `KSelect` input
- Break out unscoped styles because `:deep()` wasn't correctly applying all styles

## Vue 3

---

## PR Checklist

### DIFF
--- a/src/components/KPop/KPop.vue
+++ b/src/components/KPop/KPop.vue
@@ -28,7 +28,7 @@
           :id="popoverId"
           ref="popper"
           :style="popoverStyle"
-          :class="[popoverClasses, {'hide-caret': hideCaret }, { 'pb-0': $slots.actions }]"
+          :class="popoverClassObj"
           role="region"
           class="k-popover"
         >
@@ -72,7 +72,7 @@
         :id="popoverId"
         ref="popper"
         :style="popoverStyle"
-        :class="[popoverClasses, {'hide-caret': hideCaret }, { 'pb-0': $slots.actions }]"
+        :class="popoverClassObj"
         role="region"
         class="k-popover"
       >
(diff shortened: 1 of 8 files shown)


src/components/KPop/KPop.vue
```diff
@@ -288,10 +288,13 @@ export default defineComponent({
   computed: {
     popoverStyle: function() {
       return {
-        width: this.width === 'auto' ? this.width : this.width + 'px',
-        'max-width': this.maxWidth === 'auto' ? this.maxWidth : this.maxWidth + 'px',
+        width: this.width === 'auto' || this.width.endsWith('%') || this.width.endsWith('px') ? this.width : this.width + 'px',
+        'max-width': this.maxWidth === 'auto' || this.maxWidth.endsWith('%') || this.maxWidth.endsWith('px') ? this.maxWidth : this.maxWidth + 'px',
       }
     },
+    popoverClassObj: function() {
```
kaiarrowood, line 295: Moved into a computed because having it in the template was breaking the syntax highlighting for the rest of the file.

src/components/KSelect/KSelect.vue
```diff
@@ -255,6 +257,13 @@ export default defineComponent({
       type: Function,
       default: (params: SelectFilterFnParams) => params.items.filter((item: SelectItem) => item.label.toLowerCase().includes(params.query.toLowerCase())),
     },
+    /**
+     * Control whether the input for `select` and `dropdown` appearances supports filtering.
+     */
+    enableFiltering: {
+      type: Boolean,
+      default: null,
```
adamdehaven, line 265: issue: the `default` would need to be `true` or `false` since `null` is not a valid Boolean value.

src/components/KSelect/KSelect.vue
```diff
@@ -271,18 +280,35 @@ export default defineComponent({
     const selectInputId = computed((): string => props.testMode ? 'test-select-input-id-1234' : uuidv1())
     const selectTextId = computed((): string => props.testMode ? 'test-select-text-id-1234' : uuidv1())
     const selectItems: Ref<SelectItem[]> = ref([])
+    const filterIsEnabled = computed((): boolean => {
+      if (props.enableFiltering !== null) {
```
adamdehaven, line 284: this would need to be updated once the default prop value is changed to a Boolean

src/components/KSelect/KSelect.vue
kaiarrowood: TS actually accepts this value and I need it to be `null` because I need to know if they explicitly set it to either `true` or `false` as these should override the default values (which differ based on the `appearance`).

image

https://fettblog.eu/boolean-in-javascript-and-typescript/

src/components/KSelect/KSelect.vue
kaiarrowood: See above

PR #707, opened 2022-07-19
fix(kselect): disabled and readonly states [beta]

adamdehaven:
# Summary

- Fixes `.k-input` `disabled` and `readonly` styles so that they are supported both on the component and if a user only utilizes the CSS classes
- Updates `KSelect` `disabled` and `readonly` logic to both be styled correctly given these attributes, but also to respond to or ignore the filter and text input accordingly.
    - Adds a `readonly` KSelect to the `overviewLabel` docs just to display a readonly example

image

image

---

## PR Checklist


kaiarrowood (approved):
LGTM! ✨

PR #717, opened 2022-07-25
feat(kselect): add autosuggest support [KHCP-4227]

Leopoldthecoder:
# Summary

Add autosuggest support for KSelect. @adamdehaven 

## Vue 3

- [x] **No**, because this PR should be approved before porting to `beta`

---

## PR Checklist

### DIFF
--- a/docs/components/select.md
+++ b/docs/components/select.md
@@ -396,6 +396,7 @@ function getItems(count) {
 
 const allItems = new Array(10).fill().map((_, i) => ({
   label: `Item ${i}`,
+  description: `This is the description for item ${i}.`,
   value: `autosuggest-item-${i}`
 }));
 
--- a/docs/components/select.md
+++ b/docs/components/select.md
@@ -298,23 +298,37 @@ const allItems = new Array(10).fill().map((_, i) => ({
 export default {
   data() {
     return {
+      defaultItems: [],
       items: [],
-      query: '',
       loading: false,
     }
   },
   methods: {
     onQueryChange (val) {
+      if (val === '' && !this.defaultItems.length) {
+        this.loading = true;
+        // If query is empty and default items are not fetched, fetch them
+        setTimeout(() => {
+          this.defaultItems = allItems;
+          this.items = this.defaultItems.map(item => Object.assign({}, item));
+          this.loading = false;
+        }, 200);
+        return;
+      }
+      if (val === '') {
+        // If query is empty and default items are fetched, use the default items
+        this.items = this.defaultItems.map(item => Object.assign({}, item));
+        return;
+      }
       this.loading = true;
-      // mock API call for items that contain the keyword
+      // Otherwise fetch items that contain the keyword
       setTimeout(() => {
-        this.query = val;
         this.items =
           allItems
-            .filter(item => item.label.toLowerCase().includes(this.query.toLowerCase()) || item.description.toLowerCase().includes(this.query.toLowerCase()))
+            .filter(item => item.label.toLowerCase().includes(val.toLowerCase()) || item.description.toLowerCase().includes(val.toLowerCase()))
             .map(item => Object.assign({}, item));
         this.loading = false;
-      }, 400);
+      }, 200);
     }
   }
 }
@@ -333,7 +347,7 @@ The following is an example:
   :loading="loadingForDebounced"
   width="300px"
   appearance="select"
-  @query-change="onQueryChangeDebounced"
+  @query-change="onQueryChangeForDebounced"
 >
   <template v-slot:item-template="{ item }">
     <div class="select-item-label">{{ item.label }}</div>
@@ -374,20 +388,37 @@ const allItems = new Array(10).fill().map((_, i) => ({
 export default {
   data() {
     return {
+      defaultItems: [],
       items: [],
-      query: '',
       loading: true,
     }
   },
   methods: {
-    onQueryChange: debounce(function (val) {
+    onQueryChange (val) {
+      if (val === '' && !this.defaultItems.length) {
+        // If query is empty and default items are not fetched, fetch them
+        this.loading = true;
+        setTimeout(() => {
+          this.defaultItems = allItems;
+          this.items = this.defaultItems.map(item => Object.assign({}, item));
+          this.loading = false;
+        }, 200);
+        return;
+      }
+      if (val === '') {
+        // If query is empty and default items are fetched, use the default items
+        this.items = this.defaultItems.map(item => Object.assign({}, item));
+        return;
+      }
+      this.debouncedHandler(val);
+    },
+    debouncedHandler: debounce(function (val) {
       this.loading = true;
-      // mock API call for items that contain the keyword
+      // Fetch items that contain the keyword
       setTimeout(() => {
-        this.query = val;
         this.items =
           allItems
-            .filter(item => item.label.toLowerCase().includes(this.query.toLowerCase()) || item.description.toLowerCase().includes(this.query.toLowerCase()))
+            .filter(item => item.label.toLowerCase().includes(val.toLowerCase()) || item.description.toLowerCase().includes(val.toLowerCase()))
             .map(item => Object.assign({}, item));
         this.loading = false;
       }, 200);
@@ -542,11 +573,11 @@ export default {
         label: '50',
         value: '50'
       }],
+      defaultItemsForAutosuggest: [],
       itemsForAutosuggest: [],
-      query: '',
       loading: false,
+      defaultItemsForDebouncedAutosuggest: [],
       itemsForDebouncedAutosuggest: [],
-      queryForDebounced: '',
       loadingForDebounced: true,
     }
   },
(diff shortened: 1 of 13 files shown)


docs/components/select.md
```diff
@@ -367,6 +450,21 @@ export default {
...
+        this.itemsForAutosuggest = allItems.filter(item => item.label.includes(this.keyword)).map(item => Object.assign({}, item));
+        this.loading = false;
+      }, 400);
+    },
+    onFocus () {
+      this.loading = true;
+      setTimeout(() => {
+        this.itemsForAutosuggest = allItems.map(item => Object.assign({}, item));
+        this.loading = false;
+      }, 200);
```
adamdehaven, line 467: question: I'm not sure I understand the purpose of this `onFocus` handler. Can you explain?

packages/KSelect/KSelect.vue
```diff
@@ -93,7 +93,7 @@
             <KInput
               :id="selectTextId"
               v-bind="$attrs"
-              v-model="filterStr"
+              :value="filterStr"
```
adamdehaven, line 96: question: why are you removing `v-model` here?

packages/KSelect/KSelect.vue
```diff
@@ -254,6 +273,20 @@ export default {
...
+    autosuggest: {
+      type: Boolean,
+      default: false
+    },
+    /**
+     * Loading state in autosuggest
+     */
+    loading: {
+      type: Boolean,
+      default: false
```
adamdehaven, line 289: Again, we'll need to document this new prop in `select.md`

Also, does this only get utilized when the `autosuggest` prop is set to `true`? If yes, that needs to be included in the documentation.

packages/KSelect/KSelect.vue
```diff
@@ -347,21 +385,30 @@ export default {
...
-        if (this.appearance === 'select') {
-          this.filterStr = this.selectedItem.label
+  watch: {
+    items: {
+      handler () {
+        // items need keys
+        this.selectItems = this.items
+
+        for (let i = 0; i < this.selectItems.length; i++) {
+          this.selectItems[i].key = `${this.selectItems[i].label.replace(' ', '-')}-${i}`
```
adamdehaven, line 395: issue: There's a chance other invalid characters could be passed, so let's use a regular expression
```suggestion
          this.selectItems[i].key = `${this.selectItems[i].label.replace(' ', '-').replace(/[^a-z0-9-_]/gi, '')}-${i}`
```

packages/KSelect/KSelect.vue
```diff
@@ -412,6 +459,10 @@ export default {
 
         return false
       }
+    },
+    onKeywordChange (val) {
+      this.filterStr = val
+      this.$emit('keyword-change', val)
```
adamdehaven, line 465: issue: we need to add a `debounce` function here. We have the `useDebounce` util function inside `/packages/utils` that you may be able to use; however, if this requires importing the `@vue/composition-api` we would rather just write a custom function.

I think setting the `debounce` value to something like `400` is reasonable; however, we could also add this as a prop.

Also, if we rename the Event emit to `query-change` then we should update this function name to `onQueryChange`

docs/components/select.md
```diff
@@ -231,6 +231,80 @@ KSelect works as regular inputs do using v-model for data binding:
...
+  @keyword-change="onKeywordChange"
+  @focus="onFocus"
+>
+  <template v-slot:loading>
+    <div>Loading...</div>
+  </template>
+  <template v-slot:empty>
+    <div>No results found</div>
+  </template>
+</KSelect>
```
adamdehaven, line 253: Most of the time in Konnect, we'll be using the slotted syntax with the `autosuggest` prop. Can we update this example to utilize the slotted syntax?

You will also need to add a new `item.description` to the data object. I added suggestions in those places.

One other thing, you could update your filter function to check the `label` and the `description` for autosuggest, if you would like.

```suggestion
<KSelect
  autosuggest
  :items="itemsForAutosuggest"
  :loading="loading"
  appearance="select"
  @keyword-change="onKeywordChange"
  @focus="onFocus"
>
  <template v-slot:item-template="{ item }">
    <div class="select-item-label">{{ item.label }}</div>
    <div class="select-item-desc">{{ item.description }}</div>
  </template>
  <template v-slot:loading>
    <div>Loading...</div>
  </template>
  <template v-slot:empty>
    <div>No results found</div>
  </template>
</KSelect>
```

docs/components/select.md
```diff
@@ -231,6 +231,80 @@ KSelect works as regular inputs do using v-model for data binding:
...
+  autosuggest
+  :items="items"
+  :loading="loading"
+  appearance="select"
+  @keyword-change="onKeywordChange"
+  @focus="onFocus"
+> 
+  <template v-slot:loading>
+    <div>Loading...</div>
+  </template>
```
adamdehaven, line 266: Let's update the example to use the slotted syntax as well
```suggestion
  <template v-slot:item-template="{ item }">
    <div class="select-item-label">{{ item.label }}</div>
    <div class="select-item-desc">{{ item.label }}</div>
  </template>
  <template v-slot:loading>
    <div>Loading...</div>
  </template>
```

docs/components/select.md
```diff
@@ -231,6 +231,80 @@ KSelect works as regular inputs do using v-model for data binding:
...
+  <template v-slot:empty>
+    <div>No results found</div>
+  </template>
+</KSelect>
+
+<script>
+const allItems = new Array(10).fill().map((_, i) => ({
+  label: `Item ${i}`,
```
adamdehaven, line 274: Let's add a `description` property to support the slotted syntax
```suggestion
  label: `Item ${i}`,
  description: `This is the description for item ${i}.`,
```

docs/components/select.md
```diff
@@ -319,6 +394,11 @@ function getItems(count) {
   return myItems
 }
 
+const allItems = new Array(10).fill().map((_, i) => ({
+  label: `Item ${i}`,
```
adamdehaven, line 398: Let's add the `description` property for the slotted syntax.
```suggestion
  label: `Item ${i}`,
  description: `This is the description for item ${i}.`,
```

adamdehaven (changes requested):

packages/KSelect/KSelect.vue
Leopoldthecoder: Yes, `loading` is used when `autosuggest` is enabled. I'm adding it in the docs.

packages/KSelect/KSelect.vue
Leopoldthecoder: We need to emit `query-change` event in this `KInput`'s input event handler, so here I'm breaking `v-model` down to `value` prop and `input` event. In the `input` handler, `filterStr` is also updated, so it should have the same effect as `v-model="filterStr"`

docs/components/select.md
```diff
@@ -303,6 +377,7 @@ export default {
 | `selected` | `selectedItem` Object |
 | `input` | `selectedItem` Object or null |
 | `change` | `selectedItem` Object or null |
+| `keyword-change` | `keyword` String |
```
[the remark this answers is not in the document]
Leopoldthecoder: Sure, updated.

docs/components/select.md
Leopoldthecoder: OK, now the filter function checks both `label` and `description`.

docs/components/select.md
Leopoldthecoder: I think there might be cases where autosuggest should show options before any query is typed. For example, when creating a new SNI in KM, the autosuggest shows all existing certificates upon focus, before the end user types any keyword/query:
image

This is what the `onFocus` handler does.

packages/KSelect/KSelect.vue
Leopoldthecoder: Yes, in most cases the actual data fetching function should be debounced. But is it more suitable for autosuggest's consumer to take care of this?

For autosuggest itself, I think it's best for the `query-change` event to fire immediately after the query changes, because in some scenarios the consumer doesn't need to debounce their handlers.

For example, in both KM and Konnect we have autosuggests that do not send API requests when query changes. Instead they just filter the suggestions locally with the latest query (the suggestions are already fetched when focusing the autosuggest, like I mentioned in the previous comment). Deboucing the handler is not necessary, since no requests are sent.

The `vue-autosuggest` we currently use does not seem to have built-in debouce either, and I see Konnect handling this using `composables.useDebounce`. To me this looks like a common practice: the child component emits `change` events with every change, and the consumer component takes care of debouncing its handler.

docs/components/select.md
adamdehaven: Instead of the user having to manually add the `@focus` event handler, is there a way we could include this functionality by default in the `autosuggest` enabled `KSelect` component ?

packages/KSelect/KSelect.vue
adamdehaven: > But is it more suitable for autosuggest's consumer to take care of this?
> 
> For autosuggest itself, I think it's best for the `query-change` event to fire immediately after the query changes, because in some scenarios the consumer doesn't need to debounce their handlers.

Ok thanks for providing your thoughts. 

What if we just add a note in the docs page that tells the user something like, "You may want to implement a debounce to the `@query-change` event handler" and provide a simple example?

```
::: tip Note
Note text goes here... 
:::

Comment (Leopoldthecoder) [reply packages/KSelect/KSelect.vue]: Sure, I'll update the docs.

File: docs/components/select.md (hunk 1 above)
Comment (Leopoldthecoder) [reply docs/components/select.md:489]: OK, let me think about it.

File: packages/KSelect/KSelect.vue (hunk 5 above)
Comment (Leopoldthecoder) [reply packages/KSelect/KSelect.vue]: Updated in https://github.com/Kong/kongponents/pull/717/commits/651d924800c87a571f99dfe4caeaf4830a07a3b2

File: docs/components/select.md (hunk 1 above)
Comment (Leopoldthecoder) [reply docs/components/select.md:576]: I thought about this for a while but couldn't seem to find a better solution🤣.

We could have another prop, maybe something like `defaultSuggestions`, for users to pass in default options. But we already have `items`, and users can initialize their default options (maybe in the `created` hook) and pass them to `items`. This will have the same effect as adding a dedicated `defaultSuggestions` prop.

By the way, the example initializes default options in the `@focus` event handler instead of the `created` hook because I think this will eliminate unnecessary API requests. The end user may not click on the `KSelect` at all, so sending API requests in the `created` hook is a waste of resource. This also applies to if we have `defaultSuggestions`.

We could also turn `defaultSuggestions` into a method `getDefaultSuggestions`, and call this method inside `KSelect` itself when the `KInput` is focused. But this still requires users to pass in a method, which is not quite different from adding a `@focus` handler.

So do you have any better ideas?

Comment (adamdehaven) [reply docs/components/select.md:592]: What about this: instead of attaching the `@focus` event to the KSelect, what if inside `KSelect.vue` `mounted` we emit the `query-change` event with an empty string as the value, which will trigger an initial load of the items?

```ts
mounted () {
  const inputElem = document.getElementById(this.selectInputId)
  if (inputElem) {
    this.inputWidth = inputElem.offsetWidth
    this.$emit('query-change', '')
  }
},
```

If I remove the `@focus` handler on the `KSelect` in the docs in the `### autosuggest` section, this seems to work. What do you think?

Leopoldthecoder: Right, this could work. The downside of triggering `query-change` when `mounted` is that it still sends potentially unnecessary API request, if the end user does not interact with `KSelect` at all. That being said, I can implement this if it's acceptable.

adamdehaven: As an alternative solution, you could maybe do something like this:

1. Within `KSelect.vue` create a new `initialFocusTriggered: false` data variable
2. Detect the first focus of the field within the component; and on first focus, if `initialFocusTriggered` still equals false, trigger the event `emit` (which will load the items) and then set `initialFocusTriggered = true` to prevent it from firing again

Didn't test this, but I think it's possible

docs/components/select.md
```diff
@@ -231,6 +231,172 @@ KSelect works as regular inputs do using v-model for data binding:
...
+  :items="itemsForDebouncedAutosuggest"
+  :loading="loadingForDebounced"
+  appearance="select"
+  @query-change="onQueryChangeDebounced"
+>
+  <template v-slot:item-template="{ item }">
+    <div class="select-item-label">{{ item.label }}</div>
+    <div class="select-item-desc">{{ item.label }}</div>
+  </template>
+</KSelect>
```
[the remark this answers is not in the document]
Leopoldthecoder: OK, will do.

docs/components/select.md
Leopoldthecoder: But the event emitted with the first focus is also potentially unnecessary. If multiple `KSelect`s with autosuggest are present on a page, multiple API requests are sent once the page is loaded.

Leopoldthecoder: Great! I'll implement this solution.

Leopoldthecoder: Updated.

docs/components/select.md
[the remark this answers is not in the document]
Leopoldthecoder: This is fixed.

docs/components/select.md
Leopoldthecoder: I also updated the two examples for `autosuggest` so they look more like real life usages.

packages/KSelect/KSelect.spec.js
```diff
@@ -210,4 +216,35 @@ describe('KSelect', () => {
 
     expect(wrapper.html()).toMatchSnapshot()
   })
+
+  it('works in autosuggest mode', async () => {
```
adamdehaven, line 220: Can you write one additional test to make sure the filtering with `autosuggest` enabled still works?

Like, it emits the event (maybe) but at the least filters the list properly?

packages/KSelect/KSelect.vue
```diff
@@ -322,7 +365,8 @@ export default {
       return false
     },
     filteredItems: function () {
-      return this.filterFunc({ items: this.selectItems, query: this.filterStr })
+      // For autosuggest, items don't need to be filtered internally
```
adamdehaven, line 368: So the `filterFunc` prop does not work with `autosuggest` enabled, correct?

If this is the case, we need to note this under the documentation for `filterFunc` in `select.md`

Leopoldthecoder: Right. I added a note.

kaiarrowood (approved):
This looks really goo. Thanks @Leopoldthecoder ! 🚀

adamdehaven (approved):
LGTM - @Leopoldthecoder you're good to port to the `beta` branch.
Please include all changes from this PR

PR #724, opened 2022-08-01
feat(kdropdownmenu): new component [khcp-690]

kaiarrowood:
# Summary

- Add new `KDropdownMenu` kongponent - ported from `khcp-ui`.
- Bump `vue-router` to match version used in `khcp-ui`
- Add it as a `peerDependency` to `KButton`

For [KHCP-690](https://konghq.atlassian.net/browse/KHCP-690).

image

## Vue 3

---

## PR Checklist

### DIFF
--- a/docs/components/dropdown-menu.md
+++ b/docs/components/dropdown-menu.md
@@ -353,7 +353,7 @@ There are 3 primary item types:
 export default {
   data () {
     return {
-      selectedLabel: 'Selected an item',
+      selectedLabel: 'Select an item',
       selectedItem: '',
       menuItems: [{
         label: 'US (United States)',
--- a/docs/components/dropdown-menu.md
+++ b/docs/components/dropdown-menu.md
@@ -1,62 +1,8 @@
 # DropdownMenu
 
-**KDropdownMenu** - Dropdown Menu component
+**KDropdownMenu** is a button (or any slotted content) that is clicked to trigger a popover containing menu items beneath it.
 
-<div>
-  <KDropdownMenu :kpop-attributes="{ placement: 'bottomStart', width: '200' }">
-    <template #default="{ isOpen }">
-      <KButton
-        :is-open="isOpen"
-        appearance="creation"
-        icon="gearFilled"
-      />
-    </template>
-    <template #items>
-      <KDropdownItem
-        :item="{ label: 'I am a link', to: { path: '/' } }"
-        disabled
-      />
-      <KDropdownItem>
-        Generic item
-      </KDropdownItem>
-      <KDropdownItem
-        has-divider
-        class="danger"
-        @click="clickHandler"
-      >
-        Danger button
-      </KDropdownItem>
-    </template>
-  </KDropdownMenu>
-</div>
-
-```html
-<KDropdownMenu :kpop-attributes="{ placement: 'bottomStart', width: '200' }">
-  <template #default="{ isOpen }">
-    <KButton
-      :is-open="isOpen"
-      appearance="creation"
-      icon="gearFilled"
-    />
-  </template>
-  <template #items>
-    <KDropdownItem
-      :item="{ label: 'I am a link', to: { path: '/' } }"
-      disabled
-    />
-    <KDropdownItem>
-      Generic item
-    </KDropdownItem>
-    <KDropdownItem
-      has-divider
-      class="danger"
-      @click="clickHandler"
-    >
-      Danger button
-    </KDropdownItem>
-  </template>
-</KDropdownMenu>
-```
+HOW DOES THIS DIFFER FROM KSELECT
 
 <div>
   <KDropdownMenu
@@ -157,25 +103,148 @@
 
 ## Props
 
-### Prop1
+### label
 
-Description of prop1
+The label for the menu.
 
-Actual component using prop1
-<KDropdownMenu />
+### items
+
+An array of items containing a `label` and `to` will render a menu of router-links.
+
+<KDropdownMenu label="Documentation" :items="deepClone(defaultItemsUnselected)" />
 
 ```html
-<KDropdownMenu prop1="variation1" />
+<KDropdownMenu
+  label="Documentation"
+  :items="[
+    { label: 'Home', to: { path: '/' } },
+    { label: 'Button docs', to: { path: '/components/button.html' } },
+    { label: 'My docs', to: { path: '/components/dropdown-menu.html' } }
+  ]"
+/>
 ```
 
 ## Slots
 
-- `default` - default slot description
-- `slot1` - slot1 description
+There are 2 supported slots:
+
+- `default` - The trigger element for opening/closing the menu. Returns the `isOpen` state.
+- `items` - For an example of using the items slot see the [`KDropdownItem`](#KDropdownItem) section.
+
+<div>
+  <KDropdownMenu :items="deepClone(defaultItemsUnselected)">
+    <template #default="{ isOpen }">
+        <KButton
+          :is-open="isOpen"
+          appearance="creation"
+        >
+          Menu
+        </KButton>
+      </template>
+  </KDropdownMenu>
+</div>
 
 ```html
-<KDropdownMenu>
-  here is some slot content
+<KDropdownMenu :items="items">
+  <template #default="{ isOpen }">
+      <KButton
+        :is-open="isOpen"
+        appearance="creation"
+      >
+        Menu
+      </KButton>
+    </template>
+</KDropdownMenu>
+```
+
+## KDropdownItem
+
+**KDropdownMenu** generates a **KDropdownItem** for each item in the `items` property. At the most basic level, **KDropdownItem** is a wrapper around each item to display it correctly inside `KDropdownMenu`. You can use the `items` slot of the `KDropdownMenu` to manually create your own menu items.
+
+### Properties
+
+- `item` - the properties the router-link is built from, it expects a `label` and a `to`.
+- `disabled` - a boolean (default to `false`), whether or not to disable the item.
+- `selected` - a boolean (default to `false`), whether or not the item is selected.
+- `hasDivider` - a boolean (default to `false`), whether or not the item should have a divider bar displayed above it
+
+```html
+<KDropdownItem :item="{ label: 'Leave the page', to: { path: '/' } }" />
+```
+
+There are 3 primary item types:
+
+- `router-link`
+  - the generic type generated using the `items` prop on `KDropdownMenu`
+  - the generic type generated using the `item` prop on `KDropdownItem`
+- `button` - this item is generated if a handler is specified for the `@click` event on a `KDropdownItem`
+- `custom` - no special handling, you completely control the content
+
+<div>
+  <KDropdownMenu label="Variety">
+    <template #items>
+      <KDropdownItem :item="youAreHere" />
+      <KDropdownItem @click="clickHandler">
+        A button
+      </KDropdownItem>
+      <KDropdownItem
+        disabled
+        @click="clickHandler"
+      >
+        Disabled button
+      </KDropdownItem>
+      <KDropdownItem
+        has-divider
+        class="danger"
+      >
+        <a
+          href="http://www.google.com"
+          rel="noopener"
+          target="_blank"
+        >
+          External link
+          <KIcon
+            icon="externalLink"
+            size="12"
+            class="ml-2"
+          />
+        </a>
+      </KDropdownItem>
+    </template>
+  </KDropdownMenu>
+</div>
+
+```html
+<KDropdownMenu label="Variety">
+  <template #items>
+    <KDropdownItem :item="{ label: 'You are here', to: { path: '/components/dropdown-menu.html' } }" />
+    <KDropdownItem @click="clickHandler">
+      A button
+    </KDropdownItem>
+    <KDropdownItem
+      disabled
+      @click="clickHandler"
+    >
+      Disabled button
+    </KDropdownItem>
+    <KDropdownItem
+      has-divider
+      class="danger"
+    >
+      <a
+        href="http://www.google.com"
+        rel="noopener"
+        target="_blank"
+      >
+        External link
+        <KIcon
+          icon="externalLink"
+          size="12"
+          class="ml-2"
+        />
+      </a>
+    </KDropdownItem>
+  </template>
 </KDropdownMenu>
 ```
 
@@ -214,7 +283,13 @@ like:
 export default {
   data () {
     return {
-      defaultIsOpen: false
+      defaultIsOpen: false,
+      defaultItemsUnselected: [
+        { label: 'Home', to: { path: '/' } },
+        { label: 'Button docs', to: { path: '/components/button.html' } },
+        { label: 'My docs', to: { path: '/components/dropdown-menu.html' } }
+      ],
+      youAreHere: { label: 'You are here', to: { path: '/components/dropdown-menu.html' } }
     }
   },
   methods: {
--- a/docs/components/dropdown-menu.md
+++ b/docs/components/dropdown-menu.md
@@ -253,7 +253,7 @@ There are 2 supported slots:
 ### Properties
 
 - `item` - the properties the router-link is built from, it expects a `label` and a `to`.
-- `disabled` - a boolean (default to `false`), whether or not to disable the item.
+- `disabled` - a boolean (defaults to `false`), whether or not to disable the item.
 - `selected` - a boolean (default to `false`), whether or not the item is selected.
 - `hasDivider` - a boolean (default to `false`), whether or not the item should have a divider bar displayed above it
 - `isDangerous` - a boolean (default to `false`), whether or not to apply danger styles (text color is red)
--- a/docs/components/dropdown-menu.md
+++ b/docs/components/dropdown-menu.md
@@ -31,7 +31,7 @@ Use this prop to specify the display style for the dropdown menu. Can be either 
 The `menu` style is the standard you have seen in the example above. Uses a standard `primary` `KButton` with hover state over items and no notion of "selection".
 
 The `selectionMenu` style is good for a clearer representation of the currently selected menu item. `selected` state is handled automatically when clicking a `KDropdownItem` if used in conjunction with the `items` prop.
-If using the `items` slot, `KDropdownItem` children should use the `selectionMenuChild` prop to fire a `changed` event when clicked. You will need to manually control selectedness of each item using the `selected` prop.
+If using the `items` slot, `KDropdownItem` children should use the `selectionMenuChild` prop to fire a `changed` event when clicked. You will need to manually control the reactive state of the selected item.
 
 <div>
   <KDropdownMenu
--- a/docs/components/dropdown-menu.md
+++ b/docs/components/dropdown-menu.md
@@ -1,0 +1,135 @@
+# DropdownMenu
+
+**KDropdownMenu** - Dropdown Menu component
+
+<div>
+  <KDropdownMenu :kpop-attributes="{ placement: 'bottomStart', width: '200' }">
+    <template #default="{ isOpen }">
+      <KButton
+        :is-open="isOpen"
+        appearance="creation"
+        icon="gearFilled"
+      />
+    </template>
+    <template #items>
+      <KDropdownItem
+        :item="{ label: 'I am a link', to: { path: '/' } }"
+        disabled
+      />
+      <KDropdownItem>
+        Generic item
+      </KDropdownItem>
+      <KDropdownItem
+        has-divider
+        class="danger"
+        @click="clickHandler"
+      >
+        Danger button
+      </KDropdownItem>
+    </template>
+  </KDropdownMenu>
+</div>
+
+```html
+<KDropdownMenu :kpop-attributes="{ placement: 'bottomEnd', width: '200' }">
+  <template #default="{ isOpen }">
+    <KButton
+      :is-open="isOpen"
+      appearance="creation"
+      icon="gearFilled"
+    />
+  </template>
+  <template #items>
+    <KDropdownItem
+      :item="{ label: 'I am a link', to: '/' }"
+      disabled
+    />
+    <KDropdownItem>
+      Generic item
+    </KDropdownItem>
+    <KDropdownItem
+      has-divider
+      class="danger"
+      @click="clickHandler"
+    >
+      Danger button
+    </KDropdownItem>
+  </template>
+</KDropdownMenu>
+```
+
+## Props
+
+### Prop1
+
+Description of prop1
+
+Actual component using prop1
+<KDropdownMenu />
+
+```html
+<KDropdownMenu prop1="variation1" />
+```
+
+## Slots
+
+- `default` - default slot description
+- `slot1` - slot1 description
+
+```html
+<KDropdownMenu>
+  here is some slot content
+</KDropdownMenu>
+```
+
+## Theming
+
+| Variable | Purpose
+|:-------- |:-------
+| `--KDropdownMenuBorderColor`| KDropdownMenu border color
+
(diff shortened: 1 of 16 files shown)


packages/KDropdownMenu/KDropdownItem.vue
```diff
@@ -0,0 +1,165 @@
...
+  >
+    <router-link
+      v-if="type === 'link' && to"
+      :data-testid="label"
+      :class="{ 'disabled': disabled, 'has-divider': hasDivider }"
+      :to="!disabled ? to : $router.currentRoute.path"
+      class="k-dropdown-item-trigger"
+    >
+      <slot>{{ label }}</slot>
+    </router-link>
```
adamdehaven, line 19: Is it possible to refactor so that links are implemented via slot content instead of utilizing `vue-router` inside Kongponents? I this would be a better long-term solution and would break the peer dependency requirement (unsure if we still use it in other places, but removing would be ideal).

packages/KDropdownMenu/KDropdownItem.vue
```diff
@@ -0,0 +1,165 @@
...
+
+export default {
+  name: 'KDropdownItem',
+  components: { KButton },
+  props: {
+    item: {
+      type: Object,
+      default: null,
+      // Items must have a label
+      validator: (item) => item.hasOwnProperty('label')
```
adamdehaven, line 49:
```suggestion
      validator: (item) => Object.hasOwn(item, 'label')
```

packages/KDropdownMenu/KDropdownItem.vue
```diff
@@ -0,0 +1,165 @@
...
+    },
+    /**
+     * Use this prop to add a divider above the item
+     */
+    hasDivider: {
+      type: Boolean,
+      default: false
+    },
+    isDangerous: {
```
adamdehaven, line 58: suggestion: what about `destructive` instead? This falls in line with "destructive" actions, and by removing the "is" prefix it's more similar to the other props. 
```suggestion
    destructive: {
```

packages/KDropdownMenu/KDropdownItem.vue
```diff
@@ -0,0 +1,165 @@
...
+  data () {
+    return {
+      classList: 'btn-link k-button non-visual-button'
+    }
+  },
+  computed: {
+    listeners () {
+      return {
+        ...this.$listeners
+      }
```
adamdehaven, line 84: question: if we're not manipulating the $listeners, do we need to map them to a variable instead of just using `this.$listeners`?

packages/KDropdownMenu/KDropdownItem.vue
```diff
@@ -0,0 +1,165 @@
...
+      default: false
+    },
+    selected: {
+      type: Boolean,
+      default: false
+    },
+    selectionMenuChild: {
+      type: Boolean,
+      default: false
+    }
```
adamdehaven, line 73: question: I'm not sure I understand when this prob would be used or set to true/false. Can you document it?

packages/KDropdownMenu/KDropdownItem.vue
```diff
@@ -0,0 +1,165 @@
...
+  },
+  data () {
+    return {
+      classList: 'btn-link k-button non-visual-button'
```
adamdehaven, line 77: Since these classes are only used on a single element in the template, can we just add the classes on the item and remove this variable?

packages/KDropdownMenu/KDropdownMenu.vue
```diff
@@ -0,0 +1,239 @@
...
+          <ul class="k-dropdown-list dropdown-list">
+            <slot
+              :items="items"
+              :is-open="isToggled"
+              name="items"
+            >
+              <KDropdownItem
+                v-for="(item, idx) in items"
+                v-bind="item"
+                :key="`${item.label.replace(' ', '-')}-${idx}`"
```
adamdehaven, line 61: As long as the label is a string, you shouldn't need to strip characters. If you are going to strip them for the key, I would strip other "typical" invalid characters like quotes, etc. that normally wouldn't be allowed in an HTML attribute.

packages/KDropdownMenu/KDropdownMenu.vue
```diff
@@ -0,0 +1,239 @@
...
+  },
+  props: {
+    appearance: {
+      type: String,
+      default: 'menu',
+      validator: (value) => {
+        return [
+          'menu',
+          'selectionMenu'
+        ].indexOf(value) !== -1
```
adamdehaven, line 105: Just to make it like our typical usage 
```suggestion
        ].includes(value)
```

packages/KDropdownMenu/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@kongponents/kdropdownmenu",
+  "version": "0.1.0",
```
adamdehaven, line 3: Shouldn't this version match?

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+
+## Props
+
+### label
+
+The label for the menu.
+
+### items
+
+An array of items containing a `label` and other optional properties will render a menu of [`KDropdownItems`](#KDropdownItem) .
```
adamdehaven, line 15:
```suggestion
An array of item objects containing a `label` property and other optional properties which will render a menu of [`KDropdownItems`](#KDropdownItem) .
```

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
+# DropdownMenu
```
adamdehaven, line 1: Page title fix
```suggestion
# Dropdown Menu
```

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+    { label: 'Button docs', to: { path: '/components/button.html' } },
+    { label: 'My docs', to: { path: '/components/dropdown-menu.html' } }
+  ]"
+/>
+```
+
+### appearance
+
+Use this prop to specify the display style for the dropdown menu. Can be either `menu` (default) or `selectionMenu`.
+The `menu` style is the standard you have seen in the example above.
```
adamdehaven, line 31: Can you mention here that it renders a KButton style, etc?

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+
+The `selectionMenu` style is good for a clearer representation of what is selected. `selected` state is handled automatically when
+clicking a `KDropdownItem` if used in conjunction with the `items` prop. You will need to manually control selectedness if you are using the `items` slot.
```
adamdehaven, line 34: "selectedness" can we reword this? to something like this? (Unsure if this reads correctly 😬)
```suggestion
clicking a `KDropdownItem` if used in conjunction with the `items` prop. 

If you instead utilize the `items` slot, you will need to manually control the reactive state of the selected item.
```

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+Text to display on hover if dropdown is disabled.
+
+<div>
+  <KDropdownMenu
+    label="Documentation"
+    :disabled="true"
+    disabled-tooltip="You can't click me"
+    :items="deepClone(defaultItemsUnselected)"
+  />
+</div>
```
adamdehaven, line 205: issue: I can hover/click on the dropdown, the tooltip shows. If I then click on the page, the dropdown itself shows.

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+  disabled-tooltip="You can't click me"
+  :items="deepClone(defaultItemsUnselected)"
+/>
+```
+
+## Slots
+
+There are 2 supported slots:
+
+- `default` - The trigger element for opening/closing the menu. Returns the `isOpen` state.
```
adamdehaven, line 220: I'm not sure what this means? I think you're referring to the scoped slot data? 

> Returns the `isOpen` state

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+  label="Documentation"
+  :items="items"
+  show-caret
+/>
+```
+
+### kpopAttributes
+
+Use the `kpopAttributes` prop to configure the **KPop** [props](/components/popover.html) dropdown.
+
```
adamdehaven, line 123: I'm not sure the card example is ideal here since the Docs user isn't familiar with the "three dot" action menu UI and may not understand this example without clicking around or some other instruction?

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+## Slots
+
+There are 2 supported slots:
+
+- `default` - The trigger element for opening/closing the menu. Returns the `isOpen` state.
+- `items` - For an example of using the items slot see the [`KDropdownItem`](#KDropdownItem) section.
+
+<div>
+  <KDropdownMenu :items="deepClone(defaultItemsUnselected)">
```
adamdehaven, line 224: Similar to the changes we made on some other components, can we update the examples to not link away from the current page (eg linking to the button docs)? Maybe just show a tooltip or something instead?

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+      >
+        Menu
+      </KButton>
+    </template>
+</KDropdownMenu>
+```
+
+## KDropdownItem
+
+**KDropdownMenu** generates a **KDropdownItem** for each item in the `items` property. At the most basic level, **KDropdownItem** is a wrapper around each item to display it correctly inside `KDropdownMenu`. You can use the `items` slot of the `KDropdownMenu` to manually create your own menu items.
```
adamdehaven, line 251: Just a little rewording 
```suggestion
**KDropdownMenu** generates a **KDropdownItem** for each object in the `items` prop array. At the most basic level, **KDropdownItem** is a wrapper around each item to display it correctly inside `KDropdownMenu`. You can use the `items` slot of the `KDropdownMenu` to manually create your own menu items.
```

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+
+### Properties
+
+- `item` - the properties the router-link is built from, it expects a `label` and a `to`.
```
adamdehaven, line 255: See my comment about router-links -- ideally I think we should require using the slot for router-links so that we don't have a Vue router peer dependency

packages/KDropdownMenu/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "author": "Kong Inc",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/Kong/kongponents/issues"
+  },
+  "homepage": "https://github.com/Kong/kongponents#readme",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
```
adamdehaven, line 26: If we leave the Vue-router links, I believe we'd need to add `vue-router` as a peer dependency (not ideal)

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+- `disabled` - a boolean (default to `false`), whether or not to disable the item.
```
adamdehaven, line 256:
```suggestion
- `disabled` - a boolean (defaults to `false`), whether or not to disable the item.
```

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+- `selected` - a boolean (default to `false`), whether or not the item is selected.
```
adamdehaven, line 257:
```suggestion
- `selected` - a boolean (defaults to `false`), whether or not the item is selected.
```

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+- `hasDivider` - a boolean (default to `false`), whether or not the item should have a divider bar displayed above it
```
adamdehaven, line 258:
```suggestion
- `hasDivider` - a boolean (defaults to `false`), whether or not the item should have a divider bar displayed above it
```

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+- `isDangerous` - a boolean (default to `false`), whether or not to apply danger styles (text color is red)
```
adamdehaven, line 259:
```suggestion
- `isDangerous` - a boolean (defaults to `false`), whether or not to apply danger styles (text color is red)
```

Also see comment about this prop name

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+- `selectionMenuChild` - a boolean (default to `false`), whether the parent is a `selectionMenu` or not (used for events)
```
adamdehaven, line 260:
```suggestion
- `selectionMenuChild` - a boolean (defaults to `false`), whether the parent is a `selectionMenu` or not (used for events)
```

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+
+```html
+<KDropdownItem :item="{ label: 'Leave the page', to: { path: '/' } }" />
+```
+
+There are 3 primary item types:
+
+- `router-link`
```
adamdehaven, line 268:
```suggestion
- `link`
```

In the actual `KDropdownItem` component the type is `link`, not `router-link`

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+  - the generic type generated using the `items` prop on `KDropdownMenu`
+  - the generic type generated using the `item` prop on `KDropdownItem`
```
adamdehaven, line 270: I think one of these lines was  duplicated on accident and is not needed

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+    </KDropdownItem>
+  </template>
+</KDropdownMenu>
+```
+
+### Events
+
+| Event     | Description             |
+| :-------- | :------------------ |
+| `clicked` | Fires when a menu item is clicked |
```
adamdehaven, line 348: Event name
```suggestion
| `click` | Fires when a menu item is clicked |
```

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,394 @@
...
+| `changed` | Fires when items with `selectionMenuChild` prop are clicked; returns `selectedItem` Object or null |
+
```
adamdehaven, line 350: We also need to document the `toggleDropdown` event

adamdehaven (commented):
Initial pass. I know there's a lot here; I'm happy to meet up and go through them but wanted to go ahead and submit so that you're not waiting on me 🙂

packages/KDropdownMenu/KDropdownMenu.vue
```diff
@@ -0,0 +1,239 @@
+<template>
+  <div
+    :class="{'selection-dropdown-menu': appearance === 'selectionMenu'}"
+    class="k-dropdown k-dropdown-menu"
+  >
+    <KToggle v-slot="{toggle, isToggled}">
+      <KPop
```
adamdehaven, line 7: I think we should pass a `target` here so that for the dropdown, the popover is bound to the element instead of the body. Maybe `.k-dropdown-menu`

packages/KDropdownMenu/KDropdownItem.vue
kaiarrowood: I kind of like offering the linking and both `KButton` and `KBreadcrumb` use `router-link` - removing `vue-router` basically makes our breadcrumb component useless.

packages/KDropdownMenu/KDropdownItem.vue
kaiarrowood: I think I was copy/pasting. I'll switch over to `this.$listeners`

packages/KDropdownMenu/KDropdownItem.vue
kaiarrowood: It's documented:

image

image

I will also add a note about it in the `apearance: 'selectionMenu'` section.

packages/KDropdownMenu/KDropdownItem.vue
kaiarrowood: Yes, I will fix

packages/KDropdownMenu/KDropdownMenu.vue
kaiarrowood: If spaces won't break it I'll just remove the `replace()`

packages/KDropdownMenu/package.json
kaiarrowood: Yep, I will fix

docs/components/dropdown-menu.md
kaiarrowood: Read closer you grasshopper. Two different props from two different components.

packages/KDropdownMenu/KDropdownItem.vue
kaiarrowood: Actually realized I need this and am now manipulating it

packages/KDropdownMenu/KDropdownMenu.vue
kaiarrowood: Adding this causes the page to shift upward every time you click to open the menu. I remember seeing this on `KSelect` but can't remember how we fixed it. Need to investigate further

docs/components/dropdown-menu.md
kaiarrowood: Fixed

packages/KDropdownMenu/KDropdownItem.vue
kaiarrowood: Fixed

packages/KDropdownMenu/KDropdownMenu.vue
kaiarrowood: Done

packages/KDropdownMenu/package.json
kaiarrowood: Done

docs/components/dropdown-menu.md
kaiarrowood: Fixed

docs/components/dropdown-menu.md
kaiarrowood: Done

docs/components/dropdown-menu.md
kaiarrowood: Updated the text: "Slot provides `isOpen` - whether the menu is open or not."

packages/KDropdownMenu/KDropdownItem.vue
kaiarrowood: Reworked this a little bit. I'm not sure whether this is actually useful or not anymore. You can track which element is selected via the standard click events and the prop isn't being used for anything else. I included it originally because `Dropdown` had it I think, but it is really only useful to emit `changed` at the `KDropdownMenu` level rather than at the `KDropdownItem` level but I can't do that because they may be completely slotting the `items` content so there isn't a element that I can guarantee will be there for me to attach the click listener to I don't think.

Can you think of a better way to do this?

docs/components/dropdown-menu.md
kaiarrowood: Fixed (I think yours reads better than mine, I like to make up words 😓)

packages/KDropdownMenu/KDropdownItem.vue
adamdehaven: Yeah we already have the peer dependency, so may as well

packages/KDropdownMenu/package.json
adamdehaven: Add a peer dependecy on `vue-router@^3.5.4` and update the other peer dependencies (`KRumbs/package.json` and the root `package.json`)

docs/components/dropdown-menu.md
adamdehaven: The width of the element itself is too large and allows you to click (even not on the button) 
image

adamdehaven (commented):
A couple other issues:

- The page jumps when opening a menu
- Link styles within the dropdown are wrong; possibly bleeding over from the Docs site (only on this page though)
- The dropdown target "moves" in Safari (e.g. within KTable when we use the 3 dot icons)

packages/KDropdownMenu/package.json
kaiarrowood: Done!

packages/KDropdownMenu/KDropdownMenu.vue
adamdehaven: Don’t forget to back this out

packages/KDropdownMenu/KDropdownItem.vue
```diff
@@ -0,0 +1,165 @@
...
+    to () {
+      return (this.item && this.item.to) || undefined
+    }
+  },
+  methods: {
+    handleClick () {
+      this.$emit('click', this.item)
+
+      if (this.selectionMenuChild) {
+        this.$emit('changed', this.item)
```
adamdehaven: Should be `change`

packages/KDropdownMenu/KDropdownItem.vue
kaiarrowood: Ok, refactored again. Now we are using a `handleSelection` method from the `items` slot + `@change` event. This approach seems a lot cleaner. What do you think?

image

docs/components/dropdown-menu.md
kaiarrowood: Fixed

docs/components/dropdown-menu.md
kaiarrowood: Fixed - added instruction

image

docs/components/dropdown-menu.md
kaiarrowood: So, in order to have actionable items using the `items` prop, they have to be links. I compromised by only linking to anchor tags within the current page.

packages/KDropdownMenu/KDropdownMenu.vue
kaiarrowood: Backed out

packages/KDropdownMenu/KDropdownItem.vue
kaiarrowood: Fixed

adamdehaven (commented):
On the `selectionMenu` appearance, the button focus state has an outline it shouldn't have
image

A `<a>` tag inside `KDropdownItem` has `text-decoration: underline` on hover but shouldn't (see the external link hover)
image

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,403 @@
...
+<KDropdownMenu
+  label="Documentation"
+  :items="items"
+  show-caret
+/>
+```
+
+### width
+
+The width of the dropdown body. Currently we support numbers (will be converted to `px`), `auto`, and percentages for width.
```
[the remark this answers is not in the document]
kaiarrowood: Fixed

docs/components/dropdown-menu.md
```diff
@@ -0,0 +1,403 @@
...
+    { label: 'Slots', to: { path: '/components/dropdown-menu.html#slots' } },
+    { label: 'Top', to: { path: '/components/dropdown-menu.html' } }
+  ]"
+/>
+```
+
+### appearance
+
+Use this prop to specify the display style for the dropdown menu. Can be either `menu` (default) or `selectionMenu`.
+The `menu` style is the standard you have seen in the example above. Uses a standard `primary` `KButton` with hover state over items and no notion of "selection".
```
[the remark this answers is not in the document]
kaiarrowood: Done

adamdehaven (approved):
LGTM

PR #727, opened 2022-08-02
fix(kselect): remove is-open prop on KInput

Leopoldthecoder:
# Summary

`KInput` does not have `is-open` prop, so `KSelect` should not pass it down. Otherwise the rendered input would be:

image

@adamdehaven 

## PR Checklist


kaiarrowood (approved):
LGTM!

PR #735, opened 2022-08-04
fix(kselect): update popover width when it opens

Leopoldthecoder:
# Summary

The popover's width is only set in the `onMounted` hook. This works fine except:
1. `<KSelect>`'s visibility is controlled by `v-show`: https://codesandbox.io/s/crazy-moon-vokjk5
2. `<KSelect>` has a relative `width` and user resizes the viewport: https://codesandbox.io/s/happy-hawking-ngq8td

The perfect solution would be to use `MutationObserver`, but I think updating popover's width every time it opens can solve most of the problem (except when user resizes the viewport while the popover is open, which IMO is acceptable).

Let me know what you think. @adamdehaven @kaiarrowood 

## PR Checklist

### DIFF
--- a/src/components/KSelect/KSelect.vue
+++ b/src/components/KSelect/KSelect.vue
@@ -49,6 +49,7 @@
           @opened="() => {
             filterStr = ''
             toggle()
+            onPopoverOpen()
           }"
           @closed="() => {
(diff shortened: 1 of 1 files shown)


adamdehaven (approved):
lgtm

PR #768, opened 2022-08-22
MA-848: new date and time picker component

mihai-peteu:
https://konghq.atlassian.net/browse/MA-848

+ future ticket for adding component specs, further single date/time object support, and general CSS cleanup:
https://konghq.atlassian.net/browse/MA-896

**NOTE**
This will be closed in favor of https://github.com/Kong/kongponents/pull/785 once remaining comments are migrated.

# Summary

Figma Mockup: https://www.figma.com/file/OMIETLMAU9mOXlpKUoKCKt/Analytics?node-id=1587%3A11287

**Still need to:**
- Determine what time boundaries we allow in the Calendar, based on a user’s Tier. Will pass these “limits” in as props from the parent page/component

## Vue 3

- [x] **Yes**, here is a link to the PR: [#769](https://github.com/Kong/kongponents/pull/769)

---

## PR Checklist


mihai-peteu:
@adamdehaven somehow the build still failed, even w/ the "wait for DOM load" trick 😭

adamdehaven:
@mihai-peteu [I fixed the build](https://github.com/Kong/kongponents/pull/768/commits/50e12801cd797771f141377e58610d7b0ff7154d); however, you do still have to keep the `hasMounted` wrapper around the entire markdown file (which is fine, really)

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,287 @@
...
+    section: 'Previous',
+    timeframeText: 'Month',
+    timeframeLength: 60 * 60 * 24 * 90
+  },
+  {
+    section: 'Previous',
+    timeframeText: 'Quarter',
+    timeframeLength: 60 * 60 * 24 * 90
+  }
+]
```
filipgutica, line 142: Take what's defined in Timeframes.ts in khcp-ui and use that here, for the relative timeframes. 

Probably minus the `allowedTiers` part since that is `khcp-ui` specific

We'll have to do a mapping on the khcp-ui side for that

```
import { getUnixTime } from 'date-fns'
import {
  CHART_DATA_REQUEST_TRIM
} from './constants'

interface TimeframeOptions {
  key: string,
  timeframeText: string;
  timeframeLength: () => number;
  granularity: number;
  allowedTiers: Array<string>
}

export enum TimeframeKeys {
  FIFTEEN_MIN = '15m',
  ONE_HOUR = '1h',
  THREE_HOUR = '3h',
  SIX_HOUR = '6h',
  TWELVE_HOUR = '12h',
  ONE_DAY = '24h',
  SEVEN_DAY = '7d',
  THIRTY_DAY = '30d',
  NINTY_DAY = '90d',
  CURRENT_WEEK = 'current_week',
  CURRENT_MONTH = 'current_month',
  CURRENT_QUARTER = 'current_quarter',
  PREVIOUS_WEEK = 'previous_week',
  PREVIOUS_MONTH = 'previous_month',
  PREVIOUS_QUARTER = 'previous_quarter'
}

export class Timeframe {
  opts: TimeframeOptions

  constructor (opts: TimeframeOptions) {
    this.opts = opts
  }

  end (): number {
    return getUnixTime(Date.now()) - CHART_DATA_REQUEST_TRIM
  }

  start (): number {
    return this.end() - this.opts.timeframeLength()
  }

  deltaStart () {
    return this.end() - (2 * this.opts.timeframeLength())
  }
}

class PreviousWeek extends Timeframe {
  end (): number {
    const lastMonday = new Date()

    lastMonday.setDate(lastMonday.getDate() - (lastMonday.getDay() + 13) % 14)
    lastMonday.setHours(23, 59, 59)

    return getUnixTime(lastMonday.getTime())
  }
}

class PreviousMonth extends Timeframe {
  end (): number {
    const lastMonth = new Date()

    lastMonth.setMonth(lastMonth.getMonth() - 1)
    const lastDayOfMonth = new Date(lastMonth.getFullYear(), lastMonth.getMonth() + 1, 0)

    lastDayOfMonth.setHours(23, 59, 59)

    return getUnixTime(lastDayOfMonth.getTime())
  }

  start (): number {
    const lastMonth = new Date()

    lastMonth.setMonth(lastMonth.getMonth() - 1)
    lastMonth.setDate(1)
    lastMonth.setHours(0, 0, 0, 0)

    return getUnixTime(lastMonth.getTime())
  }
}

export const TimePeriods = new Map<string, Timeframe>([
  [
    TimeframeKeys.FIFTEEN_MIN,
    new Timeframe({
      key: TimeframeKeys.FIFTEEN_MIN,
      timeframeText: 'Last 15 minutes',
      timeframeLength: () => 60 * 15,
      granularity: 10, // every 10 seconds
      allowedTiers: ['free', 'trial', 'plus', 'enterprise']
    })
  ],
  [
    TimeframeKeys.ONE_HOUR,
    new Timeframe({
      key: TimeframeKeys.ONE_HOUR,
      timeframeText: 'Last hour',
      timeframeLength: () => 60 * 60 * 1,
      granularity: 60 * 1, // every 1 min
      allowedTiers: ['free', 'trial', 'plus', 'enterprise']
    })
  ],
  [
    TimeframeKeys.THREE_HOUR,
    new Timeframe({
      key: TimeframeKeys.THREE_HOUR,
      timeframeText: 'Last 3 hours',
      timeframeLength: () => 60 * 60 * 3,
      granularity: 60 * 1, // every 1 min
      allowedTiers: ['free', 'trial', 'plus', 'enterprise']
    })
  ],
  [
    TimeframeKeys.SIX_HOUR,
    new Timeframe({
      key: TimeframeKeys.SIX_HOUR,
      timeframeText: 'Last 6 hours',
      timeframeLength: () => 60 * 60 * 6,
      granularity: 60 * 10, // every 10 min
      allowedTiers: ['free', 'trial', 'plus', 'enterprise']
    })
  ],
  [
    TimeframeKeys.TWELVE_HOUR,
    new Timeframe({
      key: TimeframeKeys.TWELVE_HOUR,
      timeframeText: 'Last 12 hours',
      timeframeLength: () => 60 * 60 * 12,
      granularity: 60 * 10, // every 10 min
      allowedTiers: ['free', 'trial', 'plus', 'enterprise']
    })
  ],
  [
    TimeframeKeys.ONE_DAY,
    new Timeframe({
      key: TimeframeKeys.ONE_DAY,
      timeframeText: 'Last 24 hours',
      timeframeLength: () => 60 * 60 * 24,
      granularity: 60 * 10, // every 10 mins
      allowedTiers: ['free', 'trial', 'plus', 'enterprise']
    })
  ],
  [
    TimeframeKeys.SEVEN_DAY,
    new Timeframe({
      key: TimeframeKeys.SEVEN_DAY,
      timeframeText: 'Last 7 days',
      timeframeLength: () => 60 * 60 * 24 * 7,
      granularity: 60 * 60, // every 1 hour
      allowedTiers: ['trial', 'plus', 'enterprise']
    })
  ],
  [
    TimeframeKeys.THIRTY_DAY,
    new Timeframe({
      key: TimeframeKeys.THIRTY_DAY,
      timeframeText: 'Last 30 days',
      timeframeLength: () => 60 * 60 * 24 * 30,
      granularity: 60 * 60, // every 1 hour
      allowedTiers: ['trial', 'plus', 'enterprise']
    })
  ],
  [
    TimeframeKeys.CURRENT_WEEK,
    new Timeframe({
      key: TimeframeKeys.CURRENT_WEEK,
      timeframeText: 'This week',
      timeframeLength: () => {
        // Monday -> now
        const prevMonday = new Date()

        prevMonday.setDate(prevMonday.getDate() - (prevMonday.getDay() + 6) % 7)
        prevMonday.setHours(0, 0, 0, 0)

        return (new Date().getTime() - prevMonday.getTime()) / 1000
      },
      granularity: 60 * 60, // every 1 hour
      allowedTiers: ['trial', 'plus', 'enterprise']
    })
  ],
  [
    TimeframeKeys.CURRENT_MONTH,
    new Timeframe({
      key: TimeframeKeys.CURRENT_MONTH,
      timeframeText: 'This month',
      timeframeLength: () => {
        // First of the month -> now
        const firstOfTheMonth = new Date()

        firstOfTheMonth.setDate(1)
        firstOfTheMonth.setHours(0, 0, 0)

        return (new Date().getTime() - firstOfTheMonth.getTime()) / 1000
      },
      granularity: 60 * 60, // every 1 hour
      allowedTiers: ['trial', 'plus', 'enterprise']
    })
  ],
  [
    TimeframeKeys.PREVIOUS_WEEK,
    new PreviousWeek({
      key: TimeframeKeys.PREVIOUS_WEEK,
      timeframeText: 'Previous week',
      timeframeLength: () => 60 * 60 * 24 * 7,
      granularity: 60 * 60, // every 1 hour,
      allowedTiers: ['trial', 'plus', 'enterprise']
    })
  ],
  [
    TimeframeKeys.PREVIOUS_MONTH,
    new PreviousMonth({
      key: TimeframeKeys.PREVIOUS_MONTH,
      timeframeText: 'Previous month',
      timeframeLength: () => 60 * 60 * 24 * 30,
      granularity: 60 * 60, // every 1 hour,
      allowedTiers: ['trial', 'plus', 'enterprise']
    })
  ]
])
```

filipgutica (changes requested):

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,287 @@
...
+      let start = 0
+      let end = 0
+
+      // If calendar currently shown, send start/end values
+      if (this.showCalendar) {
+        start = Math.floor(this.selectimedTimeRange.start / 1000)
+        end = Math.floor(this.selectimedTimeRange.end / 1000)
+      } else {
+        end = getUnixTime(roundToNearestMinutes(Date.now()))
+        start = end - this.selectedTimeframe.timeframeLength
```
filipgutica, line 211: Return `start` and `end` as defined by the `selectedTimeframe` start and end methods (Timeframes.ts)

filipgutica, line 211: this.$emit('changed', selectedTimeframe)

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,287 @@
...
+    },
+    submitTimeFrame () {
+      console.warn('>>> submitTimeFrame')
+      let start = 0
+      let end = 0
+
+      // If calendar currently shown, send start/end values
+      if (this.showCalendar) {
+        start = Math.floor(this.selectimedTimeRange.start / 1000)
+        end = Math.floor(this.selectimedTimeRange.end / 1000)
```
filipgutica, line 208:
```
// Create anew timeframe object, calculate the timeframeLength, and compute a granularity based off of it
// emit new timeframe object

const customTimeframe = new Timeframe(...) 
this.$emit('changed', customTimeframe)

```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,287 @@
...
+      } else {
+        end = getUnixTime(roundToNearestMinutes(Date.now()))
+        start = end - this.selectedTimeframe.timeframeLength
+      }
+
+      console.log({ start, end })
+
+      this.$emit('changed', { start, end })
```
filipgutica, line 216: can remove this, since emiting in the above block

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,287 @@
...
+</template>
+
+<script>
+// import { DatePicker } from 'v-calendar'
+import KButton from '@kongponents/kbutton/KButton.vue'
+import KPop from '@kongponents/kpop/KPop.vue'
+import KSegmentedControl from '@kongponents/ksegmentedcontrol/KSegmentedControl.vue'
+import { roundToNearestMinutes, getUnixTime } from 'date-fns'
+
+const allowedTimePeriods = [
```
adamdehaven, line 81: Can we move this `allowedTimePeriods` into a `.ts` file within the `KDateTimePciker/` directory and just import it here? Same if you end up with any other config objects, etc.

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,287 @@
...
+    value: {
+      type: Object,
+      required: true
+    }
+  },
+
+  data () {
+    return {
+      allowedTimePeriods,
+      mode: this.mode || 'custom',
```
adamdehaven, line 162: What is `this.mode` referencing here? I don't think this is valid

packages/KDateTimePicker/KDateTimePicker.vue
adamdehaven: What is `Timeframes.ts` ?

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,287 @@
...
+<style lang="scss">
+@import '~@kongponents/styles/variables';
+
+$margin: .2rem;
+
+.analytics-timepicker {
+  // TODO: consider moving to a new button theme
+  .timepicker-input {
+    padding: var(--spacing-sm) var(--spacing-xl) var(--spacing-sm) 40px !important;
+    background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' viewBox='0 0 16 16' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill-rule='evenodd' clip-rule='evenodd' d='M14 14H2V2H4V4H6V2H10V4H12V2H14V14ZM14 0H2C0.895 0 0 0.896 0 2V14C0 15.104 0.895 16 2 16H14C15.104 16 16 15.104 16 14V2C16 0.896 15.104 0 14 0ZM7 12H9V10H7V12ZM10 9H12V7H10V9ZM7 9H9V7H7V9ZM4 9H6V7H4V9ZM4 12H6V10H4V12Z' fill='%236F7787'/%3E%3C/svg%3E%0A");
```
adamdehaven, line 232: Do we need this custom calendar icon? My preference would be to add a new calendar icon to `KIcon` and then use that, if possible.

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,287 @@
...
+      .vc-time-icon {
+        display: none;
+      }
+      // TODO: current day highligh should be `blue-500` as well
+      // .vc-day-content.vc-focusable
+
+      // Time range highlight
+      .vc-day .vc-highlights.vc-day-layer .vc-highlight {
+        background-color: var(--blue-100);
+      }
```
adamdehaven, line 266: If these are all `v-calendar` overrides, would it make sense to put them in a separate block? This would allow us to keep them separate in case they change over time.

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,287 @@
...
+}
+
+</script>
+
+<style lang="scss">
+@import '~@kongponents/styles/variables';
+
+$margin: .2rem;
+
+.analytics-timepicker {
```
adamdehaven, line 228: The wrapper class (outer-most class on the Kongponent) should follow our `k-*` syntax, so can we change this and it's usage at the top of the component file to what I have here? 
```suggestion
.k-datetime-picker {
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,287 @@
+<template>
+  <KPop
+    class="analytics-timepicker"
```
adamdehaven, line 3: See my comment in the style block
```suggestion
    class="k-datetime-picker"
```

adamdehaven (changes requested):

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: It's where we stored the display text / actual timestamp value mapping:
https://github.com/Kong/khcp-ui/pull/2479/files#diff-017f16348af343f9855ca2f6c0e29e2bffed3e08cb17b9052dcec6af3a83d733

Thanks to @filipgutica work in that PR, we now have extended functionality that also dynamically determines `start` and `end` values (to support the newly requested **Previous** and **Current** time frames).

adamdehaven: So keep in mind you don't have access to that here in Kongponents, so is the thought that `Timeframes.ts` becomes something that lives inside Kongponents, or..? 

If yes, it wouldn't be something you all (i.e. Analytics team) could be changing all the time on the fly. Once it's in Kongponents, it's more of an uphill fight to make changes since this repo is open-source.

If it's something that is really specific to Konnect, I'd recommend allowing whatever it configures to be passed in via a prop or something.

filipgutica (changes requested):

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,311 @@
...
+        v-if="showCalendar"
+        :attributes="attrs"
+        :model-config="modelConfig"
+        v-model="selectedCalendarRange"
+        mode="dateTime"
+        is-range
+        is-expanded
+      />
+      <div
+        v-else-if="!showCalendar && hasRelativeTimeframes"
```
Leopoldthecoder, line 36: `v-else-if` already rules out `!showCalendar`, so we can just
```suggestion
        v-else-if="hasRelativeTimeframes"
```

Leopoldthecoder:
In the [figma mockup](https://www.figma.com/file/OMIETLMAU9mOXlpKUoKCKt/Analytics?node-id=1587%3A11287) the `displayedRange` is placed below `KSegmentedControl`:

image

But now we have `displayedRange` inside `KButton`:

image

Has the design changed?

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,311 @@
...
+  components: {
+    KButton,
+    KPop,
+    KSegmentedControl,
+    DatePicker: () => import('v-calendar/lib/components/date-picker.umd')
+  },
+  props: {
+    defaultMessage: {
+      type: String,
+      required: true
```
Leopoldthecoder, line 104: Instead of requiring users to pass a `defaultMessage`, we could have a default value for this prop so users don't have to pass it every time:

```js
defaultMessage: {
  type: String,
  required: false,
  default: 'Please select a date or time range'
},
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,311 @@
...
+    },
+    defaultValue: {
+      type: Object,
+      required: true
```
Leopoldthecoder, line 108: Should we really make this prop `required`?

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,135 @@
...
+
+<KCard>
+  <template v-slot:body>
+    <div class="mt-3">Emitted value: <pre>{{ currentValue1 }}</pre></div>
+    <KDateTimePicker
+      @changed="newVal => currentValue1 = newVal"
+      defaultMessage="Please select a date or time range"
+      :timePeriods="exampleTimeFrames"
+      :defaultValue="defaultRangeWeek"
+      v-model="currentValue1"
```
Leopoldthecoder, line 19: `v-model` is a sytax sugar for `value` prop and `input` event. Since we have none inside `KDateTimePicker`, I think adding `v-model` here has no effect.

Leopoldthecoder:
It would be awesome if the two panels (`Custom` and `Relative`) automatically synchronizes with each other. For example, when `Last 7 days` is picked from the `Relative` panel, and if we switch to the `Custom` panel, the calendar highlights the last week.

mihai-peteu:
RE: latest mockup - this should be it, as far as I am aware
https://www.figma.com/file/OMIETLMAU9mOXlpKUoKCKt/Analytics?node-id=1511%3A9092

RE: keeping `Custom` and `Relative` panels in sync - I have inquired about this early on with Product, and they said it's not necessary. Our mockups change... very often :)

docs/components/datetimepicker.md
mihai-peteu: `currentValue1` and `currentValue2` are using `v-model` for demo purposes in that specific markdown files, so I can reflect what values would get emitted to the parent in the `<pre>` tag above the component.

adamdehaven:
> In the [figma mockup](https://www.figma.com/file/OMIETLMAU9mOXlpKUoKCKt/Analytics?node-id=1587%3A11287) the displayedRange is placed below KSegmentedControl:

I personally don't see the need for displaying the value twice

mihai-peteu:
Spoke to Christian, will conditionalize the "readout" to only show when we're in Relative mode:
Screen Shot 2022-09-01 at 2 38 47 PM

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: I totally missed this suggestion.  This solves my need to import `date-fns` directly into the markdown file. Stay tuned!

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: moved to using `defaultRelative` for any potential relative time frame default, and `defaultCustom` for a calendar-specific value.  Neither is required.

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: Cleaned this up in my latest commit.

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: It's not finalized yet :/ I'll check and get back to you on that.

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: yup! I've broken things out into two separate (unscoped) SCSS sections

Everything owned / created by us
```
.k-datetime-picker { 
  .timepicker-input { ... }
  .k-popover { ... }
```

V-calendar overrides
```
.k-datetime-picker {
  .vc-container { ... }
}
```

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: You make an excellent point here 🤣

docs/components/datetimepicker.md
Leopoldthecoder:
```
@changed="newVal => currentValue1 = newVal"
```
This line already guarantees the emitted value is passed to `currentValue1`.

Using `v-model` on `KDateTimePicker` will only work if `KDateTimePicker` itself has a `value` prop and emits `input` events.

You can try removing `v-model="currentValue1"` and the example should still work as expected.

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,190 @@
...
+    {
+      section: 'Last',
+      values: [
+        { timeframeText: '15 minutes', timeframeLength: 60*15 },
+        ...
+      ]
+    },{
+      section: 'Previous',
+      values: [
+        { timeframeText: '15 minutes', timeframeLength: 60*15 },
```
filipgutica, line 40: A previous timeframe needs a custom `start()` and `end()` can't just use a timeframeLength to figure it out

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,190 @@
...
+            { timeframeText: '7 days', timeframeLength: 60*60*24*7 },
+            { timeframeText: '30 days', timeframeLength: 60*60*24*30 },
+            { timeframeText: '90 days', timeframeLength: 60*60*24*90 }
+          ]
+        },
+        {
+          section: 'Previous',
+          values: [
+            { timeframeText: 'Week', timeframeLength: 60*60*24*14 },
+            // { timeframeText: 'Quarter', timeframeLength: 60*60*24*90 }
```
filipgutica, line 175: Same here, previous timeframes need a `start()` and `end()` 
EG: Prevous week should be last, last monday -> last monday

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,391 @@
...
+     */
+    changeRelativeTimeframe (timeframe) {
+      this.selectedTimeframe = timeframe
+
+      // Update input field text
+      this.abbreviatedDisplay = this.selectedTimeframe.timeframeText
+
+      // Format the start/end values as human readable date
+      const end = getUnixTime(roundToNearestMinutes(Date.now()))
+      const start = end - this.selectedTimeframe.timeframeLength
```
filipgutica, line 223: As per my original comment, this needs to be updated.
Cannot just do `now - timeframeLength` for all relative timeframes.

```
const end = this.selectedTimeframe.end()
const start = this.selectedTimeframe.start()
```

filipgutica (changes requested):
Need to be able to pass `start()` and `end()` functions as the timeframes for relative, then simply use those functions to get the range.

filipgutica (changes requested):
Also we should probably deal in milliseconds for a generic component like this. 

Since we on the konnect side needs seconds, should be up to us after we consume the component to convert the start/end timestamps to seconds as expected by the queries for example

filipgutica (approved):

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,245 @@
+# DateTimePicker
```
adamdehaven, line 1: Title should be updated to render properly in the sidebar
```suggestion
# DateTime Picker
```

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,245 @@
...
+        end: new Date().getTime()
+      },
+      minDate: 'Tue Aug 31 2023 17:09:43 GMT-0400 (Eastern Daylight Time)',
+      maxDate: 'Mon Apr 4 2021 10:09:43 GMT-0400 (Eastern Daylight Time)',
+      exampleTimeFrames
+    }
+  },
+  computed: {
+    defaultRelativeTimeframe () {
+      console.log(TimePeriods.get(TimeframeKeys.ONE_DAY))
```
adamdehaven, line 231: Some logging left behind
```suggestion
```

docs/.vuepress/utils/KDatePickerMockData.js
```diff
@@ -0,0 +1,189 @@
+const CHART_DATA_REQUEST_TRIM = 10
```
adamdehaven, line 1: nit: can you add a comment at the top of this file that explains this is for the docs site examples only?

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,245 @@
...
+      calendarMode="date"
+      :isRange="false"
+    />
+    <div class="mt-6">Emitted value: <pre>{{ JSON.stringify(currentValue1) }}</pre></div>
+  </template>
+</KCard>
+
+### Single time
+
+<KCard>
```
adamdehaven, line 25: None of the examples should be inside of a `KCard` component -- they should just appear under the example heading.

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,245 @@
...
+/>
+```
+
+## Props
+
+### calendarMode
+
+- Allowed values are: `date`, `time`, `dateTime`
+
+### isRange
```
adamdehaven, line 139: We want to move away from prefixing prop names with `is-` and `has-` so can we change this (and its usage in the component) to just `range`
```suggestion
### range
```

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,245 @@
...
+  :defaultCustom="{
+    start: new Date().getTime(),
+    end: new Date().getTime() - (2*24*60*60*1000)
+  }"
+/>
+```
+
+## Props
+
+### calendarMode
```
adamdehaven, line 135: Can this prop just be named `mode`? 
```suggestion
### mode
```

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,245 @@
...
+
+- Allowed values are: `date`, `time`, `dateTime`
+
+### isRange
+
+Along with the `calendarMode` prop, this determines the whether the calendar allows selection of a single date or time, as opposed to a range of start and end values.
```
adamdehaven, line 141:
```suggestion
Along with the `mode` prop, this determines the whether the calendar allows selection of a single date or time, as opposed to a range of start and end values.
```

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,245 @@
...
+
+### timePeriods
+
+An array of time frame values to be displayed as buttons in the "Relative" section of the popover, eg:
+
+```json
```
adamdehaven, line 147: If you change the type here to `js` you won't get the formatting error.
```suggestion
```js
```

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,245 @@
...
+  {
+    section: 'Previous',
+    values: [
+      { timeframeText: 'Week', start: function(), end: function() }
+    ]
+  }
+]
+```
+
+### value
```
adamdehaven, line 166: question: does the `value` prop get mapped to the `v-model`? 

If yes, let's go ahead and rename the prop to `modelValue` to match the Vue 3 spec so we don't have to do it later.

docs/components/datetimepicker.md
```diff
@@ -0,0 +1,245 @@
...
+
+The timestamp that the component should default to
+
+### defaultMessage
```
adamdehaven, line 170: To provide a more obvious prop name, could this be `placeholder` instead?
```suggestion
### placeholder
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+      #content>
+      <KSegmentedControl
+        v-if="hasRelativeTimeframes && hasCalendar"
+        v-model="tabName"
+        :options="[
+          { label: 'Relative', value: 'relative' },
+          { label: 'Custom', value: 'custom' }
+        ]"
+        class="w-100 mb-5"
+        data-testid="analytics-time-toggle"
```
adamdehaven, line 29: We don't want to reference `konnect` or `analytics` in any of the components
```suggestion
        data-testid="k-datetime-picker-time-toggle"
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+      <div
+        v-else-if="hasRelativeTimeframes && hasTimePeriods"
+        class="d-flex flex-column"
+      >
+        <div
+          v-for="(item, index) in timePeriods"
+          :key="`section-${index}`"
+          class="timeframe-section d-flex flex-column"
+        >
+          <h4>{{ item.section }}</h4>
```
adamdehaven, line 50: question: does this need a `v-if` in case the user doesn't provide a string for the `item.section` to prevent the empty tag from rendering?
```suggestion
          <h4 v-if="item.section">{{ item.section }}</h4>
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+          <div class="timeframe-buttons d-flex flex-row justify-content-start">
+            <KButton
+              v-for="(timeFrame, index) in item.values"
+              :key="`time-${index}`"
+              :is-rounded="false"
+              :class="{'selected-option': timeFrame.timeframeText === selectedTimeframe.timeframeText}"
+              :data-testid="'select-timeframe-' + timeFrame.timeframeLength()"
```
adamdehaven, line 57: question: is there ever a chance `timeFrame.timeframeLength()` could eval to `undefined`?

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+export default {
+  name: 'KTimePicker',
+  components: {
+    KButton,
+    KPop,
+    KSegmentedControl,
+    DatePicker: () => import('v-calendar/lib/components/date-picker.umd')
+  },
+  props: {
+    defaultMessage: {
```
adamdehaven, line 109: You'll need to update the usage throughout the component as well.
```suggestion
    placeholder: {
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+      type: String,
+      required: false,
+      default: 'Select a time range'
+    },
+    calendarMode: {
```
adamdehaven, line 114: You'll need to update the usage throughout the component as well.
```suggestion
    mode: {
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+      type: String,
+      required: false,
+      default: '',
```
adamdehaven, line 117: I think there should be a default mode..?

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+      type: Object,
+      required: false,
+      default: () => {}
+    },
+    defaultCustom: {
+      type: Object,
+      required: false,
+      default: () => ({start: '', end: ''})
+    },
+    isRange: {
```
adamdehaven, line 132: You'll need to update the usage throughout the component as well.
```suggestion
    range: {
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+        return valuesArray.every((item) => {
+          return Array.isArray(item.values) && item.values.every((timeframe) => {
+            // Check validity of each timeframe
+            return typeof timeframe.timeframeText === 'string' &&
+              timeframe.timeframeLength !== undefined &&
+              timeframe.start !== undefined &&
+              timeframe.end !== undefined
+          })
+        })
+      }
```
adamdehaven, line 151: Is there a time when `timePeriods` value is required based on other component props/settings? If yes, that should be accounted for in the `validator`

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+    }
+  },
+
+  data () {
+    return {
+      hidePopover: false,
+      tabName: this.getDefaultTabName(),
```
adamdehaven, line 158: It's bad practice to call a method in the `data` definition. You'd be better to set initially to an empty string and then set the value in the `beforeMount` function.

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+      abbreviatedDisplay: this.defaultMessage,
+      fullRangeDisplay: '',
+      selectedCalendarRange: this.defaultCustom,
+      selectedTimeframe: this.timePeriods[0],
```
adamdehaven, line 162: `this.timePeriods` is not a required prop, meaning that the first object in the array may very well not exist. At the least you should provide a fallback value here
```suggestion
      selectedTimeframe: this.timePeriods[0] || '', // or whatever a valid fallback value would be
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+          end: { class: 'vcal-day-end' },
+          day: { class: 'vcal-day-preselect' }
+        }
+      }
+    }
+  },
+
+  computed: {
+    hasCalendar () {
+      return this.calendarMode !== ''
```
adamdehaven, line 180:
```suggestion
      return this.mode !== ''
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+      return this.timePeriods && this.timePeriods.length
+    },
+    showCalendar () {
+      return this.tabName === 'custom' || !this.hasTimePeriods
+    },
+    hasRelativeTimeframes () {
+      return this.timePeriods.length > 0
+    },
+    submitDisabled () {
+      return this.isRange || this.hasRelativeTimeframes
```
adamdehaven, line 192:
```suggestion
      return this.range || this.hasRelativeTimeframes
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+      return (this.defaultRelative !== undefined)
+        ? this.defaultRelative
+        : null
+    }
+  },
+
+  watch: {
+    // Updates input field's "human" date whenever v-calendar value is touched
+    selectedCalendarRange (newVal) {
+      if (newVal) {
```
adamdehaven, line 209: question: do you also need to check here to only evaluate when `newVal !== oldVal` ?

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+        let start = ''
+        let end = ''
+
+        // If value is an object, this is a time range. Else, a single date or time value.
+        if (newVal.hasOwnProperty('start')) {
```
adamdehaven, line 214:
```suggestion
        if (newVal.start !== undefined) {
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+      this.abbreviatedDisplay = this.formatDisplayDate(range)
+    } else if (this.defaultTimeframe) {
+      this.changeRelativeTimeframe(this.defaultTimeframe)
+    }
+  },
+
+  methods: {
+    getText (tf) {
+      return tf.timeframeText
+    },
```
adamdehaven, line 260: I don't believe this is being used; can likely be deleted.
```suggestion
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+        endISO: new Date(end).toISOString(),
+        timeframeText: this.selectedTimeframe.timeframeText
+      }
+
+      this.fullRangeDisplay = this.formatDisplayDate({ start, end })
+    },
+
+    clearSelection () {
+      this.selectedCalendarRange = null
+      this.abbreviatedDisplay = this.defaultMessage
```
adamdehaven, line 291:
```suggestion
      this.abbreviatedDisplay = this.placeholder
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+      if (this.calendarMode === 'date') {
+        fmtStr = 'PP'
+      } else if (this.calendarMode === 'time') {
+        fmtStr = 'PP hh:mm a'
+      } else if (this.calendarMode === 'dateTime') {
+        fmtStr = 'PP hh:mm a'
+      }
+
+      // isRange determines whether an end date/time should be shown
+      if (this.isRange || this.hasRelativeTimeframes) {
```
adamdehaven, line 313:
```suggestion
      if (this.range || this.hasRelativeTimeframes) {
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+        fmtRange = `${format(start, fmtStr)} - ${format(end, fmtStr)}`
+      } else {
+        fmtRange = `${format(start, fmtStr)}`
+      }
+
+      return fmtRange
+    },
+
+    getDefaultTabName () {
+      return (this.hasDefaultCustomValue ? 'custom' : 'relative')
```
adamdehaven, line 323: Don't need parenthesis
```suggestion
      return this.hasDefaultCustomValue ? 'custom' : 'relative'
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,502 @@
...
+      let fmtRange = ''
+
+      // calendarMode determines format string
+      if (this.calendarMode === 'date') {
+        fmtStr = 'PP'
+      } else if (this.calendarMode === 'time') {
+        fmtStr = 'PP hh:mm a'
+      } else if (this.calendarMode === 'dateTime') {
+        fmtStr = 'PP hh:mm a'
+      }
```
adamdehaven, line 310:
```suggestion
      if (this.mode === 'date') {
        fmtStr = 'PP'
      } else if (this.mode === 'time') {
        fmtStr = 'PP hh:mm a'
      } else if (this.mode === 'dateTime') {
        fmtStr = 'PP hh:mm a'
      }
```

adamdehaven (changes requested):
1. All of the examples need to be modified to include the HTML/vue code example under them to easily see how the example was configured, like this one (however even this one needs to show what the `exampleTimeFrames` and `defaultRelativeTimeframe` prop values are):
    image
1. There is a lot of extra spacing between the popup and the date field; this should be decreased to match how the dropdowns appear over an input in `KSelect`. It looks like you just need to remove the top margin on `.k-popover` or something?
    image
1. There is a lot of white space above the "Clear" and "Apply" buttons -- can we get rid of some of it? 
1. Is there a way (assuming with `v-calendar`) to restrict the user from picking a date in the future? We wouldn't want to enable this by default; however, I would imagine for analytics you would not want to allow the user to pick a future date/time.
1. If I click the "Clear" button, since it clears the selected value, shouldn't it emit event data as well (even if all of the properties are empty) to signal that the user cleared the value.
1. **Single date selection**: Picking a single date in the picker updates the display value even if I don't click the "Apply" button (the value doesn't get emitted until you click apply).
1. When any one picker on the page emits a value, it updates the `Emitted value` display section under every example with the same value. The box should only change if the related example is modified.
1. If you're just selecting a time, there should not be a date associated with it (for display purposes) or in the emitted value `{"start":1662447900000,"end":"","startISO":"2022-09-06T07:05:00.000Z","endISO":""}` as selecting a time by itself does not have a date part.
    image
1. Placeholder text for "Single date and time" example is wrong and reads, `Please select a time`
1. The "Custom ranges - show only the calendar" example doesn't actually allow picking a range.

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: Actually, i needed a simple method to concatenate the prefix (eg: Current, Last, etc) with the timeframeText, so I renamed it `getTimeframeText()` and used it in a couple of places :)

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: Haters gon' hate 🤠 

Call me old school, but I think of these parantheses as a soft, snuggly blanket that provides security and comfort to an otherwise floaty (and lonely) piece of logic.

🤣

adamdehaven: 😴 <- this is cozy Mihai 😂

mihai-peteu:
@adamdehaven Thanks for taking the time to review my massive, sprawling PR.

I have handled requests **1 through 4** from your list, plus **9 and 10** (I think 10 was an duplicated example instance, i dropped it).

RE: **5, 6, 7, and 8**

**5.** If "Clear" emits the value, then the user would get a blank charts, and/or a blank Metric card - I'm staying in "Analytics world" here :) 
**6.** This actually happens across the board - we don't emit anything until "Apply" is clicked. I've done this mostly to save us extra network requests (since the parent `khcp-ui` datepicker wrapper will assume that was the intent behind the emit)
**7.** Most likely a bug, because I copy/pasted in a hurry.  Should be fixed now.
**8.** We could potentially hack `v-calendar` to do this, but just a heads up that it chooses to display a (static) date and returns a full ISO timestamp in `time` mode:  https://vcalendar.io/datepicker.html#time

I'm okay with changing things if we can reach a consensus quickly. Otherwise, I say we move the tougher decisions into https://konghq.atlassian.net/browse/MA-896 to be resolved later.

cc @filipgutica @ryanmoore

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,527 @@
...
+  },
+  props: {
+    placeholder: {
+      type: String,
+      required: false,
+      default: 'Select a time range'
+    },
+    mode: {
+      type: String,
+      required: false,
```
adamdehaven, line 119: I'm not sure I understand why `mode` isn't a required prop? Is it ever valid if it's an empty string?

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,527 @@
...
+      default: '',
+      validator: (value) => {
+        return [ '', 'date', 'time', 'dateTime' ].indexOf(value) !== -1
```
adamdehaven, line 122: nit: simpler
```suggestion
        return [ '', 'date', 'time', 'dateTime' ].includes(value)
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -0,0 +1,527 @@
...
+    maxDate: {
+      type: Number,
+      required: false,
+      default: null
+    },
+    minDate: {
+      type: Number,
+      required: false,
+      default: null
+    },
```
adamdehaven, line 149: Is `Number` the right prop type here for minDate and maxDate?

adamdehaven:
> 5. If "Clear" emits the value, then the user would get a blank charts, and/or a blank Metric card - I'm staying in "Analytics world" here :)

Maybe then instead of a "Clear" button, we change this to "Cancel" and functionally it just resets the value to whatever it was before the user edited the picker (so whatever the previous state was) and closes the popup? 

If the Clear button doesn't emit anything, the UI for the component will show that nothing is selected (ie the display value will be cleared); however, the consuming app (eg Konnect) will have no idea the user cleared the value. 

> 6. 

The display date shouldn't change until the user clicks apply, same as the emitted event then. Otherwise, similar to my comments above, it will look to the user (and to the datepicker component) that user chose a new date; however, the app (Konnect) will have no idea anything changed. We can't update the display text without also emitting the value, so it has to be one or the other.

mihai-peteu:
@adamdehaven sorry i forgot to answer #3 in your list - i can remove the existing padding above `k-popover-footer`, but I think this is what you were referring to: 

Screen Shot 2022-09-06 at 4 09 36 PM

All that empty space is reserved for the next (or previous) month's extra week.

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: I guess passing in `relative` for `mode` when we want a timeframe-only instance should do the trick (we just have to make sure we never pass that to `v-calendar` or it will self-implode).  See my latest commit plz:

https://github.com/Kong/kongponents/pull/768/commits/03a802aaf22fcf522ce5b5d8ef52cff96d2fd168?w=1#diff-fa2c62a1ce67b2096a11c57505e380b633881e8a3ff68009f96fba3f8676074cR122

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: I can't tell if this is a trick question :)

`new Date().getTime()` seems to return a number, so...

adamdehaven:
The problem with 5… here’s a scenario:

1. User picks a date, and the value is emitted to the consuming app.
1. Consuming app takes the emitted value and does something assuming the user picked a date (e.g. on a form, it allows submission)
1. The user then clicks the “Clear” button which, from the user’s perspective, clears the selected value, which also clears the date display text in the DateTimePicker component (so visually, the user thinks they have cleared the value.
1. The consuming app (i.e. Konnect) has no knowledge the user cleared the date and continues on with the value selected in step 1. 

You now have a POLA violation as the user and the UI have a different understanding of what is selected in the DateTimePicker

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: Yup! i just pushed that now :)

docs/components/datetimepicker.md
mihai-peteu: updated in new PR

PR #774, opened 2022-08-31
fix(ktable): client sorting with number values

Leopoldthecoder:
# Summary

Client sorting by a column of number values will have a wrong order if that column contains zeros.

Reproduction link: https://codesandbox.io/s/mystifying-stallman-zzlfmu

I'm not sure why we need to check `if (a && b)` in the first place. IMO if we need to check anything it should be the type of `b`. The code down below guarantees `a` is a number, but leaves `b` unchecked (although in most cases they should be of the same type).

So I changed `if (a && b)` to `if (typeof b === 'number')`. Let me know your opinions. @adamdehaven @kaiarrowood 

---

## PR Checklist

### DIFF
--- a/src/composables/useUtilities.ts
+++ b/src/composables/useUtilities.ts
@@ -60,8 +60,8 @@ export default function useUtilities() {
   const clientSideSorter = (key: string, previousKey: string, sortOrder: string, items: []) => {
     let comparator = null
 
-    const numberComparator = (a: number, b: number) => {
-      if (a && b) {
+    const numberComparator = (a: number, b: number | string) => {
+      if (typeof b === 'number' && !Number.isNaN(a) && !Number.isNaN(b)) {
         return a - b
       }
 
--- a/src/composables/useUtilities.spec.ts
+++ b/src/composables/useUtilities.spec.ts
@@ -46,7 +46,7 @@ describe('Client-side sorting (deprecated in favor of server-side sorting)', () 
         id: '410ecd35-696e-4e7a-ad35-c69bd4e14cbb',
         username: 'henry',
       }, {
-        custom_id: 145,
+        custom_id: 0,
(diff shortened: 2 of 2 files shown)


src/composables/useUtilities.ts
```diff
@@ -60,8 +60,8 @@ export default function useUtilities() {
   const clientSideSorter = (key: string, previousKey: string, sortOrder: string, items: []) => {
     let comparator = null
 
-    const numberComparator = (a: number, b: number) => {
-      if (a && b) {
+    const numberComparator = (a: number, b: number | string) => {
+      if (typeof b === 'number') {
```
adamdehaven, line 64: I'm not sure I completely understand the previous logic either; however, to be safe, should we do something like this:

```ts
if (a && typeof b === 'number') {
```

Leopoldthecoder: This will make zeros fail the check, so we still cannot get the right order:

image

adamdehaven: Then what about something like this?
```ts
if (a !== undefined && typeof b === 'number') {
```

Leopoldthecoder: Yeah, this looks fine but we already guarantee `a` is a number when calling `numberComparator `:
```js
switch (typeof newValA) {
  case 'number':
    return numberComparator(newValA, newValB)
```

Considering `typeof NaN` is also a `number`, maybe we should do something like this:
```js
if (!Number.isNaN(a) && !Number.isNaN(b)) {
```

Leopoldthecoder: So I updated the code. Now we are sure `a` and `b` are both numbers and not `NaN`, so `a - b` won't give us any crazy result.

adamdehaven (approved):
lgtm

PR #776, opened 2022-09-01
fix(kselect): selected style when using v-model

Leopoldthecoder:
# Summary

When `v-model` is used in `KSelect`, after the user selects an item, the selected style in the dropdown is missing:

Kapture 2022-09-01 at 13 25 23

This is because in the newly added `setter` and `watcher`, we are filtering `props.items` and passing one of them to `handleItemSelect`. `props.items` do not have the `key` field, so in `handleItemSelect` every item's `selected` field to set to `false`.

The `main` branch is fine, so we don't need to fix there.

@adamdehaven @kaiarrowood 

---

## PR Checklist


kaiarrowood:
This looks good to me, I'd like @aanchalm01 to take a look and see if this addressing an issue she was seeing in the `KongFormGenerator` with `KSelect`

kaiarrowood (approved):
LGTM! 🚀

PR #777, opened 2022-09-01
docs(kselect): add missing documentation slots [KHCP-4631]

Leopoldthecoder:
# Summary

This PR adds `loading` and `empty` slots docs.

I noticed we have another `items` slot that replaces all the `KSelectItem`s. It seems if the user decides to use this slot, they'll have to deal with `handleItemSelect` logic themselves, which is not easy. Do you have any idea how should we document this slot? @adamdehaven 

## PR Checklist


adamdehaven:
> I noticed we have another `items` slot that replaces all the `KSelectItem`s. It seems if the user decides to use this slot, they'll have to deal with `handleItemSelect` logic themselves, which is not easy. Do you have any idea how we should document this slot? 

Well I already approved the `main` branch PR 😬 

Hmm, I'm not sure really. @kaiarrowood any ideas? Do we even use the `items` slot anywhere?

kaiarrowood:
> > I noticed we have another `items` slot that replaces all the `KSelectItem`s. It seems if the user decides to use this slot, they'll have to deal with `handleItemSelect` logic themselves, which is not easy. Do you have any idea how we should document this slot?
> 
> Well I already approved the `main` branch PR 😬
> 
> Hmm, I'm not sure really. @kaiarrowood any ideas? Do we even use the `items` slot anywhere?

I don't believe we do... it may be better to yank if we aren't actually documenting the slot and no easy way to handle the selection... Otherwise we need to handle with logic similar to how we do for the `selectionMenu` appearance in `KDropdownMenu`: https://kongponents.konghq.com/components/dropdown-menu.html#appearance

The `items` slot needs to return the `handleSelection` function for the user.

image

Leopoldthecoder:
So do we have a conclusion about the `items` slot? Is there anything I can do in this PR?

kaiarrowood:
> So do we have a conclusion about the `items` slot? Is there anything I can do in this PR?

Can you investigate how easy it would be to follow the pattern we are using in KDropdownMenu? I wouldn't spend more than a half day on it. If it's too difficult to add the support, lets just yank the slot.

adamdehaven:
@kaiarrowood if we're going to yank the `items` slot on `main` branch, I'd say let's remove it completely from `beta` as well.  We (Kong) aren't using it anywhere, and if we ever need it in the future, we can always add it back.

We have the `item-template` slot already, so I think that satisfies our needs.

@Leopoldthecoder would you want to remove the `items` slot on both `main` and `beta`? (This PR we will approve and merge, and you can do the work in [KHCP-4668](https://konghq.atlassian.net/browse/KHCP-4668))

Leopoldthecoder:
No problem, will do.

Leopoldthecoder:
Hi, I removed the `items` slot in #782 and #783.

PR #780, opened 2022-09-02
fix: not being able to import style.css

kleinfreund:
# Summary

Fixes an issue with not being able to import the style.css file using a bare module specifier (i.e. `import '@kong/kongponents/dist/style.css'`) due to the absence of the specifier definition from the package.json file’s `exports` field. A workaround for this issue is using `import '../node_modules/@kong/kongponents/dist/style.css'`.

---

## PR Checklist

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -87,7 +87,7 @@ import { createApp } from 'vue'
 import Kongponents from '@kong/kongponents'
 
 // Import Kongponents styles (path may vary)
-import '../node_modules/@kong/kongponents/dist/style.css'
+import '@kong/kongponents/dist/style.css'
 
 const app = createApp(App)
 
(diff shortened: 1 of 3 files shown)


kleinfreund:
I understand why the linter is failing, but I would rather not remove the relevant information from my commit message. I don’t think commit messages should have any sort of limit.

adamdehaven (approved):
This looks good; appreciate the submission 🚀 

Funny, I just found a note that I needed to fix the exports in the details of a ticket that was previously closed; I guess it was lost.

---

> I understand why the linter is failing, but I would rather not remove the relevant information from my commit message. I don’t think commit messages should have any sort of limit.

Enforcing Conventional Commits is tied to our `semantic-release` workflow; however, I agree limiting the body text can be limiting. I'll bump the commit body limit for future commits 👍🏼

PR #784, opened 2022-09-05
refactor(kselect): cannot read properties of undefined (reading replace) [KHCP-4680]

LukeSwierlik:
Task in Jira: https://konghq.atlassian.net/browse/KHCP-4680

# Summary
- **Fix** - Add an additional condition.
- **Refactor** - Removal of redundant `replace`.

## Vue 3

---

## PR Checklist

### DIFF
--- a/packages/KSelect/KSelect.vue
+++ b/packages/KSelect/KSelect.vue
@@ -432,7 +432,7 @@ export default {
           if (this.selectItems[i].label) {
             this.selectItems[i].key = `${this.selectItems[i].label.replace(' ', '-')}-${i}`
           } else {
-            this.selectItems[i].key = `test-label-${i}`
+            this.selectItems[i].key = `k-select-item-label-${i}`
           }
 
           if (this.selectItems[i].value === this.value || this.selectItems[i].selected) {
--- a/packages/KSelect/KSelect.vue
+++ b/packages/KSelect/KSelect.vue
@@ -429,7 +429,10 @@ export default {
             this.selectItems[i].selected = false
           }
 
-          this.selectItems[i].key = `${this.selectItems[i].label.replace(' ', '-').replace(/[^a-z0-9-_]/gi, '')}-${i}`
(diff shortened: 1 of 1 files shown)


packages/KSelect/KSelect.vue
```diff
@@ -435,7 +435,10 @@ export default {
             this.selectItems[i].selected = false
           }
 
-          this.selectItems[i].key = `${this.selectItems[i].label.replace(' ', '-').replace(/[^a-z0-9-_]/gi, '')}-${i}`
+          if (this.selectItems[i].label) {
+            this.selectItems[i].key = `${this.selectItems[i].label.replace(' ', '-')}-${i}`
+          }
```
adamdehaven, line 440: Each select item will need a unique key, so you'll need to add an `else` block or something here so that `key` is always defined. 

We do not want to utilize UUID or something similar as a fallback as this could cause issues with snapshot testing.

LukeSwierlik: Thanks for your comments! I have added generating a unique `key` in the easiest way 👍

packages/KSelect/KSelect.vue
```diff
@@ -435,7 +435,12 @@ export default {
             this.selectItems[i].selected = false
           }
 
-          this.selectItems[i].key = `${this.selectItems[i].label.replace(' ', '-').replace(/[^a-z0-9-_]/gi, '')}-${i}`
+          if (this.selectItems[i].label) {
+            this.selectItems[i].key = `${this.selectItems[i].label.replace(' ', '-')}-${i}`
+          } else {
+            this.selectItems[i].key = `test-label-${i}`
```
adamdehaven, line 441: Since this key could potentially appear even when not during testing, let's rename it 
```suggestion
            this.selectItems[i].key = `k-select-item-label-${i}`
```

adamdehaven (approved):
LGTM - @LukeSwierlik does this need to be ported to the `beta` branch as well?

PR #800, opened 2022-09-14
MA-896: Datetimepicker round 2 updates

mihai-peteu:
# Summary
https://konghq.atlassian.net/browse/MA-896

Handling a few of the items in MA-896:
- Use actual Calendar icon inside timepicker input instead of inline SVG
- Specify font size and weight for readout text instead of `h4` heading (which ended up looking totally different on khcp-ui)
- add prop that allows us to hide the calendar icon if needed (Analytics plans to do this on the Reports creation page)
- ensure flex layout of timeframe buttons expands to 100% of parent width
- only sets the `min-width` attribute on the input and popover if `TimePeriods` are sent to the Kongponent (eg: this is the Analytics case, where we want the input to be the same width as the popover); otherwise, keeps the auto width defaults
- general CSS fixes

Direct preview link (since Kongponent is hidden in sidebar):
**https://deploy-preview-800--kongponents.netlify.app/components/datetimepicker.html**

  
## Vue 3

- [x] **No**, because:

We are waiting for the usage of this kongponent in `khcp-ui` to "solidify" - once there are no more requests, we can port it over to the `beta` branch and start working on specs.  

---

## PR Checklist


packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -1,11 +1,11 @@
 <template>
   <KPop
     :hide-popover="hidePopover"
+    v-bind="position"
```
adamdehaven, line 4: Why would we ever want this set to false?

instead of using `v-bind` just use the KPop prop. 

```suggestion
    :position-fixed="position"
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -122,6 +131,11 @@ export default defineComponent({
     DatePicker: () => import('v-calendar/lib/components/date-picker.umd')
   },
   props: {
+    calendarIcon: {
```
adamdehaven, line 134: Can we just name this icon since it’s the only one? 

```suggestion
    icon: {
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -177,6 +191,10 @@ export default defineComponent({
       required: false,
       default: 'Select a time range'
     },
+    position: {
```
adamdehaven, line 194: Why is this an object?

packages/KIcon/icons/index.js
```diff
@@ -5,6 +5,8 @@ import arrowUp from './icn-arrow-up.svg'
 import back from './icn-back.svg'
 import book from './icn-book.svg'
 import brain from './icn-brain.svg'
+import calendar from './icn-calendar.svg'
```
adamdehaven, line 8: Why are we adding 2 different calendar icons? 

For the sake of not having a mismatch all over Konnect, can we stick with one?

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: yes, we can :)

packages/KIcon/icons/index.js
mihai-peteu: We will eventually have just one, i wanted to demo what the two look like to PM/UX

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: I was playing around with passing position fixed to both KPop, and to the DatePicker itself... 
V-calendar expects an object: https://vcalendar.io/api/v1.0/datepicker.html#popover-positionfixed

In this case, yes, it's incorrect, since KPop would just take a Boolean, eg: `:position-fixed="true"`

Reason for this exploration: UX requested that we allow the timepicker to scroll up or down with the page, they didn't like the stickiness. So I figured at the very least I should make this behavior optional.

I now realize that `position: fixed` is not what controls that behavior...

mihai-peteu: I think this is what we're aiming for:
https://kongponents.konghq.com/components/popover.html#positionfixed

https://user-images.githubusercontent.com/103061463/190220350-8f1c3b77-09bb-42bc-8f69-068fe380353b.mov

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: Ended up leaving this alone, we probably want to default to position fixed no matter what.

mihai-peteu:
@kaiarrowood (and @adamdehaven since he's not really respecting his own PTO) - i have updated the flex calculation to be more precise. `&:nth-child(3n)` was the golden ticket.

docs/components/datetimepicker.md
```diff
@@ -267,6 +267,12 @@ currentValue = {
 }
 ```
 
+### icon
+
+A `Boolean` which controls the display of the calendar icon.
```
adamdehaven, line 272:
```suggestion
A `Boolean` to show/hide the calendar icon displayed within the display field. 
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -457,19 +471,15 @@ export default defineComponent({
 </script>
 
 <style lang="scss">
-$margin: .2rem;
-$timepicker-min-width: 24rem;
+$timepicker-min-width: 22rem;
+$margin: 0.2rem;
```
adamdehaven, line 475: I didn't catch this in the original PR, but you need to remove **ALL** `rem` units from the component and replace with pixels or a relative unit (e.g. `%`)

Since we cannot control the default font size of the root document element in a consuming app, we cannot assume `rem` units will display as desired. 

Please change all `rem` units throughout the component and styles that were added.

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -489,30 +499,33 @@ $timepicker-min-width: 24rem;
...
         .timeframe-buttons {
           flex-wrap: wrap;
+
           .timeframe-btn {
             font-size: var(--type-sm);
-            flex-basis: calc(32% - $margin);
             font-weight: 400;
-            margin: $margin;
+            // Only two columns of the 3 will have a right margin
+            flex: 0 calc(33% - ($margin/2));
```
adamdehaven, line 516: You won't be able to use the slash `/` for division when we upgrade Sass or on the `beta` branch. You need to use `sass:math` or just use another variable that is explicitly set to half of the value of `$margin`

adamdehaven (commented):
The `rem` units change is required.

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: Good to know!

adamdehaven: I guess I'm actually asking to change it here if you don't mind

mihai-peteu:
@kaiarrowood re: broken markdown file - this is a "known issue", which is why @jillztom and Adam decided to simply hide the DateTimePicker from the side nav on my original PR. 

It breaks other things (eg: if you view DateTimepicker demo page, you'd have to hard refresh next time you attempt to view another kongponents' demo page).

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -457,29 +477,34 @@ export default defineComponent({
...
       box-shadow: none !important;
     }
   }
+
   .k-popover {
     max-height: 90vh;
-    min-width: $timepicker-min-width;
+    width: 100% !important;
     overflow: hidden;
     padding: var(--spacing-lg) var(--spacing-md);
```
jillztom, line 507: The only suggestion from the look and feel would be to use: `padding: var(--spacing-sm);`. Everything else looks good!

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -1,109 +1,124 @@
...
-    @closed="handleClose"
   >
-    <KButton
-      :is-rounded="false"
-      size="large"
-      class="timepicker-input"
-      data-testid="k-datetimepicker-display"
+    <KPop
+      :hide-popover="hidePopover"
+      class=""
```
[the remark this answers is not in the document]
mihai-peteu: whoopsie daisies

mihai-peteu:
@kaiarrowood I think it's the outer `<div v-if="hasMounted">`...  Notice that `popover.md` suffers from the same visible script tag issue:
https://github.com/Kong/kongponents/blob/main/docs/components/popover.md

Screen Shot 2022-09-15 at 12 18 41 PM

packages/KDateTimePicker/KDateTimePicker.vue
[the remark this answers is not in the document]
mihai-peteu: 🏖️

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -1,109 +1,124 @@
...
+        <!-- Custom | Relative toggle -->
+        <KSegmentedControl
+          v-if="hasTimePeriods && hasCalendar"
+          v-model="tabName"
+          :options="[
+            { label: 'Relative', value: 'relative' },
+            { label: 'Custom', value: 'custom' }
+          ]"
+          class="w-100 mb-4"
+          data-testid="analytics-time-toggle"
```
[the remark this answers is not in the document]
mihai-peteu: 🏖️🏖️

packages/KIcon/icons/index.js
```diff
@@ -5,6 +5,7 @@ import arrowUp from './icn-arrow-up.svg'
...
```
adamdehaven, line 8: This is **super-nit**...

If you compare the new `calendar` icon to the `contactSupport` icon (or any others), the border on the calendar icon looks much thicker with a weird border-radius. Is it possible to get an icon that looks more similar to the others?

image

mihai-peteu: I agree, it's a little off.

I think we need a one file that contains artboards (in Illustrator terms) of all the files, so we can have exact view boxes, and be able to tell how line thickness compares from one icon to another. If this file already exists, then you can ignore me :)

CC @jillztom @helen-tang27

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -1,109 +1,123 @@
 <template>
-  <KPop
-    :hide-popover="hidePopover"
+  <div
+    :class="{ 'set-min-width': hasTimePeriods }"
```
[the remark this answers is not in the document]
mihai-peteu: Agree. I ran into this issue on the `khcp-ui` PR where we are integrating the timepicker, and I have to override things with unscoped CSS to make that happen:
https://github.com/Kong/khcp-ui/pull/2729

Shall I pass in a prop named inputWidth and the options are `auto` | `full` ?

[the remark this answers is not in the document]
mihai-peteu: Or we make it a boolean, (`fullWidth`) and we can just pass in `full-width`.  Depends on how many width options we want to support in the future.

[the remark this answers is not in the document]
mihai-peteu: Here ya go 
https://deploy-preview-800--kongponents.netlify.app/components/datetimepicker.html#single-date-full-width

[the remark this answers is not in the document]
adamdehaven: If you go this route, I believe the default width value would be `auto`

packages/KIcon/icons/index.js
mihai-peteu: Punting on this, future pass ;)

packages/KDateTimePicker/KDateTimePicker.vue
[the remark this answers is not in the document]
mihai-peteu: Pushed a couple of commits, see if that suffices plz

docs/components/datetimepicker.md
```diff
@@ -36,26 +36,32 @@ Set the `v-model` to [Single date time picker](#single-date-time-picker-v-model)
...
   @change="newVal => emitVal = newVal"
-  placeholder="Please select a time"
-  mode="time"
+  placeholder="Please select a date"
+  mode="date"
+  width="300"
   :range="false"
 />
-``` -->
+```
```
adamdehaven, line 64: TBH I don't know if we need a repetitive example just showing how to customize the width. Instead, maybe just customize the width on the existing example above and add a note?

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -1,109 +1,125 @@
...
+        <!-- Custom | Relative toggle -->
+        <KSegmentedControl
+          v-if="hasTimePeriods && hasCalendar"
+          v-model="tabName"
+          :options="[
+            { label: 'Relative', value: 'relative' },
+            { label: 'Custom', value: 'custom' }
+          ]"
+          class="w-100 mb-4"
+          data-testid="datetimepicker-toggle"
```
adamdehaven, line 46:
```suggestion
          data-testid="k-datetime-picker-toggle"
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -1,109 +1,125 @@
...
-    </template>
-  </KPop>
+      </template>
+      <template
+        v-if="!hidePopover"
+        #footer>
+        <div class="d-flex justify-content-end">
+          <KButton
+            :is-rounded="false"
+            data-testid="k-datetimepicker-clear"
```
adamdehaven, line 100:
```suggestion
            data-testid="k-datetime-picker-clear"
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -1,109 +1,125 @@
...
+            size="medium"
+            appearance="btn-link"
+            @click="clearSelection()"
+          >
+            Clear
+          </KButton>
+          <KButton
+            :disabled="submitDisabled"
+            :is-rounded="false"
+            data-testid="k-datetimepicker-submit"
```
adamdehaven, line 111:
```suggestion
            data-testid="k-datetime-picker-submit"
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -1,109 +1,125 @@
...
-      <div
-        v-else-if="hasTimePeriods"
-        class="d-flex flex-column"
+      <KButton
+        :class="[{ 'set-min-width': hasTimePeriods }]"
+        :is-rounded="false"
+        :style="widthStyle"
+        size="large"
+        class="timepicker-input"
+        data-testid="k-datetimepicker-display"
```
adamdehaven, line 21:
```suggestion
        data-testid="k-datetime-picker-display"
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -241,6 +270,11 @@ export default defineComponent({
         ? !state.selectedRange.start || !state.selectedRange.end
         : !state.selectedRange.start
     })
+    const widthStyle = computed(() => {
+      return {
+        width: props.width === 'auto' || props.width.endsWith('%') || props.width.endsWith('px') ? props.width : props.width + 'px'
+      }
+    })
```
adamdehaven, line 277: We have a utility function that should handle this for you. Here's what you'll need (please test though)

```ts
import useUtilities from '@/composables/useUtilities'

const { getSizeFromString } = useUtilities()

const widthStyle = computed(() => {
  return getSizeFromString(props.width)
})
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -457,29 +492,34 @@ export default defineComponent({
...
-    min-width: $timepicker-min-width;
-    padding: var(--spacing-sm) var(--spacing-xl) var(--spacing-sm) 40px !important;
-    background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' viewBox='0 0 16 16' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill-rule='evenodd' clip-rule='evenodd' d='M14 14H2V2H4V4H6V2H10V4H12V2H14V14ZM14 0H2C0.895 0 0 0.896 0 2V14C0 15.104 0.895 16 2 16H14C15.104 16 16 15.104 16 14V2C16 0.896 15.104 0 14 0ZM7 12H9V10H7V12ZM10 9H12V7H10V9ZM7 9H9V7H7V9ZM4 9H6V7H4V9ZM4 12H6V10H4V12Z' fill='%236F7787'/%3E%3C/svg%3E%0A");
-    background-repeat: no-repeat;
-    background-position: 12px 50%;
+    padding: var(--spacing-sm) var(--spacing-sm) !important;
     color: var(--grey-600) !important;
     font-weight: 500;
-
+    &.set-min-width {
```
adamdehaven, line 510: To handle the small screen overflow, I believe you can add the `max-width` rule here

```suggestion
    max-width: 100%; // Prevent overflowing the container
    
    &.set-min-width {
```

adamdehaven (commented):
1. Is it possible to have the text in the input wrap "cleaner" on small screens?
    image
2. On small screens, the new `width` property forces the input off screen if set to larger than the viewport. We need to enforce `max-width: 100%` on either the whole component or the input field (or something) to prevent overflowing the container. If you add `max-width: 90%` on the `.timepicker-input` element (the button) I believe this is fixed.

    image

docs/components/datetimepicker.md
mihai-peteu: oh, duh. I don't know why I didn't think of that 🤦

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: Only available on beta branch :)

But yes, I will make sure to use this on my _other_ branch 👍

adamdehaven: oops 😢

mihai-peteu:
> 1. Is it possible to have the text in the input wrap "cleaner" on small screens?
>    image
> 2. On small screens, the new `width` property forces the input off screen if set to larger than the viewport. We need to enforce `max-width: 100%` on either the whole component or the input field (or something) to prevent overflowing the container. If you add `max-width: 90%` on the `.timepicker-input` element (the button) I believe this is fixed.
>    image

I'm thinking wrap start and end dates in two separate spans, and then i can do some sort of `white-space: nowrap` on their contents.  But the 2nd span will be allowed to break to next line.

docs/components/datetimepicker.md
```diff
@@ -12,15 +12,17 @@
 
 ### Single date
 
-Create a single date selection calendar
+Create a single date selection calendar, passing in a fixed pixel width for the input field.
```
adamdehaven, line 15:
```suggestion
Create a single date selection calendar, with an example of passing in a fixed pixel width for the input field.
```

packages/KDateTimePicker/KDateTimePicker.vue
adamdehaven: @mihai-peteu you need this `max-width: 100%` here to prevent it overflowing the page

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -457,29 +492,47 @@ export default defineComponent({
...
+      min-width: $timepicker-min-width;
+    }
     &:focus {
       box-shadow: none !important;
     }
+    .timepicker-display {
+      flex-wrap: wrap;
```
adamdehaven, line 519: Add a little line-height here for when it wraps to multiple lines.
```suggestion
      flex-wrap: wrap;
      line-height: 1.3;
```

packages/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -457,29 +492,47 @@ export default defineComponent({
...
+      div {
+        width: auto;
+        text-align: left;
+        padding: 0;
+        margin: 0;
+        line-height: var(--type-md, type(md));
```
adamdehaven, line 525: Update to be relative so that it spaces properly
```suggestion
        line-height: 1.3;
```

mihai-peteu: looks good
Screen Shot 2022-09-22 at 9 18 36 AM


Note: I moved the `&nbsp;` into the first div, otherwise the two stacked divs wont't left align cleanly in some cases.

packages/KDateTimePicker/KDateTimePicker.vue
mihai-peteu: I feel like this request should be outdated... 🤔  Github bug maybe

adamdehaven (approved):
LGTM 🌮

PR #819, opened 2022-10-07
feat(kcodeblock): adds KCodeBlock component

kleinfreund:
# Summary

Adds the KCodeBlock component. It’s a new component which features search and filter functionalities, built-in line numbers, and a copy button.

## Features

* **Search**: Optionally, a code block can be made searchable. The search highlights lines that contain matches to the query. The search supports two modes: default (exact matches, case-insensitive) and regular expressions (matches by regular expression, case-insensitive).
* **Filter mode**: Search results can be filtered grep-style instead of just highlight lines in potentially large code blocks. Matched terms will be highlighted. Both default and regular expression mode are supported in filter mode as well.
* **Shortcuts**: A couple of shortcuts for navigating through search results and toggling filter and regular expression mode are provided.
* **Copy button**: A button to copy the raw code of the code block.
* **Line numbers**: All code block show line numbers. Line numbers automatically function as anchors (of the form `id + '-L' + lineNumber`, e.g. `code-block-L137`). Optionally, they can also become links.

### Syntax highlighting is external

This component **does not** handle syntax highlighting. This is mainly motivated by wanting to avoid the addition of a syntax highlighting library like PrismJS to Kongponents. This allows consuming projects to make their own decision as to which library they want to use (if any). Also, in the case of libraries like PrismJS, it avoids needing to supply support for **all** languages any consuming project could possibly want.

---

## PR Checklist


docs/components/codeblock.md
```diff
@@ -0,0 +1,268 @@
...
+* **Default**: `''`
+
+Used as the initial value of the search input. Can be used to initialize a code block with a query which was read from client storage.
+
+<KCodeBlock
+  id="code-block-query"
+  :code="code"
+  language="json"
+  is-searchable
+  query="key"
```
adamdehaven, line 105: suggestion: let's change the initial query to a term that is present, e.g. `vite`

docs/components/codeblock.md
```diff
@@ -0,0 +1,268 @@
...
+/>
+```
+
+### showCopyButton
```
adamdehaven, line 118: nit: Can we rename this prop to just `copy` ? We're trying to get away from prefixed props, e.g. `is*`, `has*`, `show*`

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
+<template>
+  <div
+    :id="props.id"
+    ref="codeBlock"
+    class="k-code-block"
+    :class="{ 'k-code-block--has-line-numbers': props.showLineNumbers }"
```
adamdehaven, line 6: We do not use BEM CSS (or other `--` syntax). Please just use a single `-` in class names; this applies for the entire file.

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+const debouncedHandleSearchInputValue = debounce(handleSearchInputValue, 150)
+
+const props = defineProps({
+  /**
+   * ID value used for form elements like the search input and its label.
+   */
+  id: {
+    type: String,
+    required: false,
+    default: 'code-block',
```
adamdehaven, line 249: The default here should be an empty string, and then in the template, you should set the attribute to `undefined` (so it does not render) if the prop equates to an empty string.

Otherwise, having 2 `KCodeBlock` elements on the same page with no `id` prop provided will result in invalid HTML (2 elements with the same `id` on a page).

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+    required: true,
+  },
+
+  /**
+   * The syntax language of `props.code` (e.g. `'json'`).
+   */
+  language: {
+    type: String,
+    required: true,
+  },
```
adamdehaven, line 266: question: should we do any sort of validation on the supported languages here, if any?

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+
+  /**
+   * Shows an actions bar with a search input and related action buttons.
+   */
+  isSearchable: {
```
adamdehaven, line 271: Let's rename this prop to `searchable` as we're trying to get away from prefixed props, e.g. `is*`

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+    type: Boolean,
+    required: false,
+    default: false,
+  },
+
+  /**
+   * Allows controlling the processing state from outside the component. This allows a parent component to show the processing icon when it’s, for example, currently syntax highlighting the code.
+   */
+  isProcessing: {
```
adamdehaven, line 280: Let's rename this prop to `processing` as we're trying to get away from prefixed props, e.g. `is*`

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+  query: {
+    type: String,
+    required: false,
+    default: '',
+  },
+
+  /**
+   * Controls whether to show a copy button. It copies the original code (i.e. the value of `props.code`) to the clipboard.
+   */
+  showCopyButton: {
```
adamdehaven, line 298: Let's rename this prop to `copy` as we're trying to get away from prefixed props, e.g. `show*`

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+    type: Boolean,
+    required: false,
+    default: true,
+  },
+
+  /**
+   * Controls whether to show line numbers.
+   */
+  showLineNumbers: {
```
adamdehaven, line 307: Let's rename this prop to `lineNumbers` as we're trying to get away from prefixed props, e.g. `show*`

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+const numberOfMatches = ref(0)
+const matchingLineNumbers = ref<number[]>([])
+const currentLineIndex = ref<null | number>(null)
+
+const totalLines = computed(() => Array.from({ length: props.code.split('\n').length }, (_, index) => index + 1))
+const linePrefix = computed(() => props.id.toLowerCase().replace(/\s+/g, '-'))
+const isProcessing = computed(() => props.isProcessing || isProcessingInternally.value)
+const isShowingFilteredCode = computed(() => isFilterMode.value && filteredCode.value !== '')
+const regExpButtonAppearance = computed(() => regExpError.value !== null ? 'danger' : isRegExpMode.value ? 'secondary' : 'outline')
+const filteredCode = computed(function() {
```
adamdehaven, line 335: Can you add return types to these?

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+        v-if="props.showLineNumbers"
+        class="k-line-number-rows"
+      >
+        <a
+          v-for="line in matchingLineNumbers"
+          :id="`${linePrefix}-L${line}`"
+          :key="line"
+          :href="`#${linePrefix}-L${line}`"
+        >{{ line }}</a>
+      </span><code v-html="filteredCode" /></pre>
```
adamdehaven, line 161: I'm not sure of a way around this, but I don't love outputting `html` here, XSS vulnerability. Same comment for other instances of `v-html`

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+            title="Previous match (Shift+Enter)"
+            color="currentColor"
+          />
+        </template>
+
+        <span class="k-visually-hidden">Previous match</span>
+      </KButton>
+
+      <KButton
+        class="k-next-match-button"
```
adamdehaven, line 122: The classes and test ids in the file:

Instead of `k-{random}` can we prefix with `k-code-block-*`?

This allows for dynamic selector matching namespaced to the component.

Comment applies to all random classes/testid attributes that are using `k-*`

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+  let totalMatchingLineNumbers: number[] = []
+
+  if (shouldPerformSearch) {
+    try {
+      totalMatchingLineNumbers = getMatchingLineNumbers(props.code.toLowerCase(), query.value.toLowerCase(), isRegExpMode.value)
+    } catch (error) {
+      if (error instanceof Error) {
+        regExpError.value = error
+      } else {
+        throw error
```
adamdehaven, line 439: If this isn't an instance of `Error` should we instead do this? (Haven't run the code)

```
throw new Error(error)
```

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+  /**
+   * Disables the shortcut dynamically.
+   */
+  isDisabled?: () => boolean
+
+  shouldPreventDefaultAction?: boolean
+}
+
+const commands: Record<string, Command> = {
+  // TODO: This doesn’t make sense with multiple code blocks in one document.
```
adamdehaven, line 543: I came here to say this. Also, we'd need to document all of the keybindings if we ship this with the component.

Due to it possibly breaking if there are multiple, unsure if this should be included.

Also, do these only fire if the element is focused?

src/utils/debounce.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Returns a function, that, as long as it continues to be invoked, will not be triggered. The initial function will be called after the debounced function stops being called for a certain number of milliseconds.
+ *
+ * @param initialFunction Initial function to debounce
+ * @param delay Time to wait for recurring bounces
+ * @returns the debounced function.
+ */
+export function debounce(initialFunction: (...args: any) => any, delay: number): (...args: any) => void {
```
adamdehaven, line 8: Could you not use the existing debounce function in `useDebounce`?

adamdehaven (commented):
Left some initial comments. Ping me again when you're further along if you'd like 👍🏼

docs/components/codeblock.md
kleinfreund: Done. That was an artifact of changing the example code. I might change it again to be a bit more generic.

docs/components/codeblock.md
kleinfreund: > We're trying to get away from prefixed props, e.g. is*, has*, show*

I’m curious to hear a rational for this as I don’t agree with it. Names like `copy`, especially when they’re (also) verbs, are very ambiguous and it’s nice to know when something is a boolean. IDE support for extracting prop type information is generally poor (though this is drastically improving with the advent of composition API and script setup authored components) so seeing at a glance that a prop `is-something` is expected allows you to infer that you’re to provide a boolean. When I face a prop `copy`, my first intuition would be to provide my own copy handler which would be incorrect in this case.

I really don’t want to change this.

docs/components/codeblock.md
```diff
@@ -0,0 +1,268 @@
...
+```vue
+<KCodeBlock
+  id="code-block-show-copy-button"
+  :code="code"
+  language="json"
+  :show-copy-button="false"
+/>
+```
+
+### showLineNumbers
```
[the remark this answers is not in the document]
kleinfreund: Deferring to conversation in https://github.com/Kong/kongponents/pull/819#discussion_r990214347.

docs/components/codeblock.md
```diff
@@ -0,0 +1,268 @@
...
+  id="code-block-show-line-numbers"
+  :code="code"
+  language="json"
+  :show-line-numbers="false"
+/>
+```
+
+## Slots
+
+None.
```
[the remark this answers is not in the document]
kleinfreund: This section merely contains None because I’m not done writing it and because I wasn’t sure the component wouldn’t have any slots. I’ve removed it in the meantime.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: I don’t feel particularly strongly about this so I’m changing it.

That being said, I don’t think like project should prohibit the use of BEM-style selectors just because they look like BEM-style selectors. It’s of no negative consequence to anyone making use of them because they’re in no way special while they may aid those used to and familiar to BEM-style selectors in identifying intent without referring to markup (while reading styles) or styles (while reading markup).

In any case, as I wrote above, I’m changing it, so consider this merely a comment.

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
+<template>
+  <div
+    :id="props.id"
+    ref="codeBlock"
+    class="k-code-block"
+    :class="{ 'k-code-block--has-line-numbers': props.showLineNumbers }"
+    tabindex="0"
+    :style="`--max-line-number-width: ${totalLines[totalLines.length - 1].toString().length}ch`"
```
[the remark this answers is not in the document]
kleinfreund: I agree, it would be nicer to have the value be a computed property. I changed it.

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,829 @@
...
+      </KClipboardProvider>
+    </div>
+  </div>
+</template>
+
+<script lang="ts" setup>
+import { computed, nextTick, onBeforeUnmount, onMounted, ref, watch } from 'vue'
+
+import KButton from '@/components/KButton/KButton.vue'
+import KClipboardProvider from '@/components/KClipboardProvider/index'
```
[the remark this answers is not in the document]
kleinfreund: I don’t *need* it, no, but it’s almost unambiguous *what* I’m importing here. Importing an index.(ts|js) file by omitting it entirely from the path isn’t (e.g. it’s unclear whether ones imports a `@/components/KClipboardProvider.(ts|js)` or as `@/components/KClipboardProvider/index.(ts|js)` file). If I could, I would also add the .ts extension which TypeScript unfortunately disallows. I’m not changing this.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: That’s in my opinion a worse developer experience because it prevents you from using a single instance of `KCodeBlock` in a document without specifying `props.id`. Both when not providing a default value and when providing one will it be necessary for two `KCodeBlock` components to have distinct IDs so I’d consider the approach of having a default value the one with less friction because it helps prevent developer error.

That being said, perhaps `props.id` should actually be required. While this wouldn’t allow for using one code block while omitting its ID it would always alert on a missing one when using two code blocks which currently could silently cause errors (though it’s very likely as the code block is very unlikely to participate in a form context which is the prime candidate for ID conflict related errors).

For now I’m changing `props.id` to be required.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: In the version I implemented in Kuma GUI, I used a more specific type based on the languages I’ve loaded via PrismJS. This isn’t necessary here since this implementation intentionally excludes the responsibility for syntax highlighting. Otherwise, I would’ve needed to introduce PrismJS as a dependency of Kongponents.

In other words: we can’t know which language identifiers a project might end up using and have loaded (projects will certainly have differents needs with regards to which languages they need to support). So for now, `props.language` isn’t actually used _inside_ the component for anything other than emitting it again. It could possibly be used to add the language to the UI.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: Let’s have this conversation in https://github.com/Kong/kongponents/pull/819#discussion_r990214347.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: Let’s have this conversation in https://github.com/Kong/kongponents/pull/819#discussion_r990214347.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: Let’s have this conversation in https://github.com/Kong/kongponents/pull/819#discussion_r990214347.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: I can, but I shouldn’t. Computed properties aren’t functions (which should always be type-annotated) and in the presence of correctly typed sources, a computed property will always automatically infer its correct type already. All computed properties as they’re currently used in the code block’s implementation correctly infer there type with the type of `filteredCode` being `string`.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: The way around this would be introducing a sanitizing library and setting `codeElement.innerHTML` with its result. This _could_ be viable because DOM-based libraries like PrismJS wouldn’t conflict with it: Prism itself will set `code.innerHTML` so any already set HTML from `<code v-html>` would be overwritten. And this can be generalized: This component expects raw code in `props.code` and not HTML content that should be rendered. This could change should this component take up syntax highlighting responsibilities but that’s another story.

I’ll have a look what the cost of adding such a library would be.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: Sure, I changed them.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: I don’t think so. Presently, this code is not _expected_ to fail with anything but an instance of `Error`. However, should a non-Error throw be introduced, whether accidentally or intentionally, this style of catching ensures that the error isn’t accidentally hidden. The `else` branch here behaves as if we hadn’t caught the error to begin with so I think it’s best to just re-throw whatever was raised.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: That’s part of the work that I simply haven’t done yet as part of the scope of this PR.

The shortcuts certainly need to be documented. I lean towards not including the `/` shortcut. The others can be included as there is logic in place for scoping them to the respective code block. All of the other commands are only triggered if they respective event is fired while focus is on an element inside (or on) the code block. As such, they’re not in conflict with other code blocks on the page.

I’ve added a shortcuts section to the docs now.

src/utils/debounce.ts
kleinfreund: I’m not sure of this style of a debounce utility as a composable involving its own reactive state. It is *very* implementation-specific. While it happens to be search-specific and the use case for debouncing in the code block happens to *also* be search-specific, this is not true for all use cases of debouncing (think scroll and resize event listeners).

I would need to architecture the implementation of the code block around the implementation details of `useDebounce`. I would need to rewrite the actual search input handler that I already have in place. That’s not the case for a pure debounce function which allows me to use its result _exactly_ like its input with the _only_ difference being that its execution is debounced.

So, for me, this is not how I want to manage a debounced search event listener.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: Do you think it would be useful or necessary to allow the key map to be customizable via props?

kleinfreund: I’ve decided that this is useful enough to add it. I realized that you’d otherwise have no way of using the search features via shortcuts *while* avoiding conflicts with any shortcuts that you might already use in your project. Now, users can customize shortcuts via `props.keyMap` and also unbind the default ones.

docs/components/codeblock.md
```diff
@@ -0,0 +1,323 @@
...
+```vue
+<KCodeBlock
+  id="code-block-show-copy-button"
+  :code="code"
+  language="json"
+  :show-copy-button="false"
+/>
+```
+
+### showLineNumbers
```
kleinfreund, line 141: **Question** (cc @adamdehaven): What do you think about dropping `props.showLineNumbers` entirely? Perhaps there *are* cases where one wants to display a code block without line numbers, sure, but aren’t those typically cases where a plain `<pre><code></code></pre>` would be sufficient (i.e. they don’t need search)?

I’m asking because this would cut down on the complexity of the code block implementation. Getting hiding line numbers to work well with still showing highlighted line numbers gets tricky and I’m currently resorting to some, well, let’s call it more arcane techniques.

src/components/KCodeBlock/KCodeBlock.vue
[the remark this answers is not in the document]
adamdehaven: I'm not sure your reasoning makes sense... the `index.ts` by default exports the component you're importing, and the syntax you're using here conflicts with every other import in the project. Please update to remove `/index`

docs/components/codeblock.md
adamdehaven: Then as an alternative I'd suggest `copyButton` as the prop name. `true` indicates it's showing, set to `false` to hide.

src/components/KCodeBlock/KCodeBlock.vue
adamdehaven: Appreciate the change. It's more about consistency. I'm working to clean up a lot of the inconsistencies in some of our repositories and having different "flavors" of CSS within the projects isn't ideal.

Another example is `camelCase` versus `kebab-case` (preferred) selectors.

src/components/KCodeBlock/KCodeBlock.vue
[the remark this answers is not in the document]
kleinfreund: I’m not sure I’m following.

Yes, `src/components/KClipboardProvider/index.ts` has a default export which is the functional `KClipboardProvider` component, but there is no conflict “with every other import in the project”. What is that conflict?

My argument is that it’s ambiguous (in the sense of a module resolution strategy) whether `@/components/KClipboardProvider` refers to `@/components/KClipboardProvider.ts` or @/components/KClipboardProvider/index.ts` (and technically both variants with a .js extension, too). Adding `/index` removes this ambiguity. A module resolution strategy will inadvertently pick whatever it is defined to pick first. So, in short, my argument is that omitting `/index` makes it unclear what’s actually being imported. I don’t understand why that’s controversial.

In any case, I don’t feel too strongly about it to insist, but my reasoning not making sense is not true.

src/components/KCodeBlock/KCodeBlock.vue
adamdehaven: If you make it required and remove the default, I think that's better; we'll just need to outline in the docs that `id` must be unique (in case it isn't obvious).

kleinfreund: Good point; it’s not immediately obvious that a `props.id` is going to be basis for a form control `id` attribute. I’ll add a note to the props.id` section in the docs.

src/components/KCodeBlock/KCodeBlock.vue
adamdehaven: > Do you think it would be useful or necessary to allow the key map to be customizable via props?

No, I don't think we want to add in all of that complexity. Again, the primary user for this component would likely be internal and I don't think we would want our developers customizing the keybindings. This could easily lead to code blocks in multiple Kong apps having different key bindings depending on the developer that implemented them.

If we're going to have key bindings in the first place, I don't believe they should be customizable so that we have the same UX throughout all of our apps.

src/utils/debounce.ts
adamdehaven: I'm fine with the separate `debounce` my only follow-up request is if we can name it something more specific to the code block component so that it's evident we won't want to utilize this function in other existing debounces.

docs/components/codeblock.md
adamdehaven: I'm fine with line numbers always showing. We can add an enhancement in the future if we decide we need to be able to hide them.

src/components/KCodeBlock/KCodeBlock.vue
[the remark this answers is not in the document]
adamdehaven: > but there is no conflict “with every other import in the project”. What is that conflict?

i.e. we never do `path/to/import/index`

docs/components/codeblock.md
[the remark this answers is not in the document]
kleinfreund: Superseded as per https://github.com/Kong/kongponents/pull/819#discussion_r991377515.

adamdehaven:
One other thought related to syntax highlighting:

I like your train of thought about it not being included here as a dependency, allowing a consuming app to handle the dependency, supported languages, etc.; however, if we move forward with the assumption that Konnect and other Kong apps would likely utilize Prism as the library of choice, I think it would be nice to provide (on the docs page) an implementation example of how to integrate syntax highlighting with Prism.

If it's too much code to include on the page itself, I can spin up a Codepen/Code Sandbox linked to the site to demo the implementation with. I think it would be a great addition 🚀

kleinfreund:
> if we move forward with the assumption that Konnect and other Kong apps would likely utilize Prism as the library of choice, I think it would be nice to provide (on the docs page) an implementation example of how to integrate syntax highlighting with Prism.

@adamdehaven I’ve added a minimal integration example for PrismJS now. It’s pretty much exactly what Kuma GUI will likely use. I’ve just tried it with the current state of `KCodeBlock` and it’s working well.

> If it's too much code to include on the page itself, I can spin up a Codepen/Code Sandbox linked to the site to demo the implementation with. I think it would be a great addition

This turned out to be not too much at all. I think it really pays off to detach the functionality of the code block with syntax highlighting.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: I’m fine with removing this again. The argument to not promote a deviation in used shortcuts is reasonable. We do lose the ability to avoid conflicts with other tools, but that can be considered a minor issue at most.

src/utils/debounce.ts
kleinfreund: I don’t understand why this is needed. The debounce function I added is as general as it gets and not in the slightest specific to anything in the code block besides its one purpose: debouncing.

Why wouldn’t you want to utilize this function in other scenarios?

src/components/KCodeBlock/KCodeBlock.vue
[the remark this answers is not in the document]
kleinfreund: Alright. While “we never do that” is certainly not an argument I’m not interested in having a prolonged argument over it.

In my opinion it’s bad style to omit `/index(.js|.ts)` from import paths and if anything, that’s something to move away from, but it doesn’t matter for the scope of this PR.

src/utils/debounce.ts
adamdehaven: I typed too fast without flushing out what I wanted to say 😬 

What I'm really meaning here is can we move this debounce function into the `useUtilities` composable so that it can live alongside our other utility functions?

The current `src/utils/` folder houses code that isn't actually used within any Kongponents, but rather helpers for the docs mostly. I'm not sure why that's the case, but nevertheless, can we move this debounce function into the `useUtilities` function?

src/components/KCodeBlock/KCodeBlock.vue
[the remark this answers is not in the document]
adamdehaven: An import path can't actually end with `.ts` -- but yeah, let's just stick with the default

src/components/KCodeBlock/KCodeBlock.vue
adamdehaven: Makes sense; I'm good with this.

kleinfreund:
**Comment**: I noticed that in the deployment preview (and only on Netlify, not when I run `yarn run docs:preview` locally), the _rendered_ code blocks are somehow duplicated but ... incompletely? I have no idea what’s causing this.

adamdehaven:
> **Comment**: I noticed that in the deployment preview (and only on Netlify, not when I run `yarn run docs:preview` locally), the _rendered_ code blocks are somehow duplicated but ... incompletely? I have no idea what’s causing this.

It only appears to happen if you hard-refresh on the Code Block docs page. If you start on another page and navigate there, it doesn't happen. We've resolved something similar in the past by wrapping the docs page itself in a `v-if` that checks if the page has mounted before rendering the content. See `KSelect` here for an example if you want to try: https://github.com/Kong/kongponents/blob/ed5f8a6028b26259f22ea7f6d497e03965bd1769/docs/components/select.md?plain=1#L3

src/utils/debounce.ts
adamdehaven: > The debounce function is not a composable and shouldn’t be one

Agree, and to be honest, the current "composable" is a bit of a stretch in terms of how I would have done it 😬 

Maybe the `KDatePickerMockData.js` should actually be moved out of `/utils/` to somewhere more purposeful since it's only used in the docs site.

kleinfreund: I think it should live with the non-docs code because it _is_ being used by a test and it being used in the docs is pretty much just convenience.

For mock data structures, I like having a `test-data` directory somewhere because it’s not uncommon (at least in application projects; perhaps not so much for component libraries) for a couple of tests to make use of the same test data. We could do something similar. I think Cypress has a concept for this (there it’s called fixtures, same thing), so we could use that.

Then again, it’s also nice to have everything related in the same location (e.g. have `KDatePickerMockData.js` live in `src/components/KDateTimePicker/KDateTimePicker.spec.ts`).

kleinfreund: I’ve just moved the mock data file near its related test. What do you think about that? That makes `src/utils` contain only code-related utility functions now.

adamdehaven: We dont' want `KDatePickerMockData.js` in the `KDateTimePicker/` directory because then it will be packaged in the `dist/` which is not desired. Please revert that change; it's out of scope.

For the `/utils` directory:

The folder should have an `index.ts` file that exports all of the functions within it like:
```ts
// utils/index.ts

export { default as debounce } from './debounce'
// ... more
```

For the `copyTextToClipboard` new util, why can't you utilize the existing `KClipboardProvider` component in `KCodeBlock` instead of implementing a new function?

kleinfreund: > We dont' want KDatePickerMockData.js in the KDateTimePicker/ directory because then it will be packaged in the dist/ which is not desired. Please revert that change; it's out of scope.

Alright. That’s not the case, but I’ll revert it.

> For the copyTextToClipboard new util, why can't you utilize the existing KClipboardProvider component in KCodeBlock instead of implementing a new function?

It’s not possible to call the copy functionality of the KClipboardProvider component which I have to in my component so I needed to externalize this.

kleinfreund:
**Update**:

* I’ve added a search icon at the start of the search input.
* I’ve added a clear query button at the end of the search input.
* I’ve added a note to the top of the docs page indicating the component to be experimental and that design changes should be expected.

src/utils/debounce.ts
kleinfreund: I’ve added an `index.ts` file to the utilities directory re-exporting its functions.

docs/components/codeblock.md
```diff
@@ -0,0 +1,357 @@
...
+Shows an actions bar with a search input and related action buttons.
+
+<KCodeBlock
+  id="code-block-is-searchable"
+  :code="code"
+  language="json"
+  is-searchable
+/>
+
+```vue
```
adamdehaven, line 67:
```suggestion
```html
```

docs/components/codeblock.md
```diff
@@ -0,0 +1,357 @@
...
+
+<KCodeBlock
+  id="code-block-is-processing"
+  :code="code"
+  language="json"
+  is-searchable
+  is-processing
+/>
+
+```vue
```
adamdehaven, line 92:
```suggestion
```html
```

docs/components/codeblock.md
```diff
@@ -0,0 +1,357 @@
...
+
+<KCodeBlock
+  id="code-block-query"
+  :code="code"
+  language="json"
+  is-searchable
+  query="no"
+/>
+
+```vue
```
adamdehaven, line 117:
```suggestion
```html
```

docs/components/codeblock.md
```diff
@@ -0,0 +1,357 @@
...
+Controls whether to show a copy button which copies the original code (i.e. the value of `props.code`) to the clipboard.
+
+<KCodeBlock
+  id="code-block-show-copy-button"
+  :code="code"
+  language="json"
+  :show-copy-button="false"
+/>
+
+```vue
```
adamdehaven, line 142:
```suggestion
```html
```

docs/components/codeblock.md
```diff
@@ -0,0 +1,357 @@
...
+
+<KCodeBlock
+  id="code-block-show-line-number-links"
+  :code="code"
+  language="json"
+  :show-line-number-links="true"
+  style="--KCodeBlockMaxHeight: 200px"
+/>
+
+```vue
```
adamdehaven, line 167:
```suggestion
```html
```

docs/components/codeblock.md
```diff
@@ -0,0 +1,357 @@
...
+
+## Syntax highlighting
+
+The `KCodeBlock` component does not have syntax highlighting built in.
+
+### PrismJS
+
+Below is an integration example for the popular syntax highlighting library [PrismJS](https://prismjs.com/).
+
+```vue
```
adamdehaven, line 259:
```suggestion
```html
```

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,968 @@
...
+
+  // Ensures that no wasted debouncing takes place when the search input value being emitted is the same as the previous query. This also avoids unnecessarily showing the processing icon.
+  if (input.value !== query.value) {
+    isProcessingInternally.value = true
+    debouncedHandleSearchInputValue(input.value)
+  }
+}
+
+function handleSearchInputValue(inputValue: string): void {
+  query.value = inputValue
```
adamdehaven, line 514: question: Would it make sense to `trim` the search string here?

docs/components/codeblock.md
```diff
@@ -0,0 +1,357 @@
...
+/>
+```
+
+### showLineNumberLinks
+
+* **Type**: `boolean`
+* **Required**: no
+* **Default**: `false`
+
+Controls whether to add links to the current line number.
```
adamdehaven, line 157: This one confuses me a bit. If you're on a URL that already has a hash (e.g. the VuePress docs site) these links aren't functional. Ideas?

docs/components/codeblock.md
```diff
@@ -0,0 +1,357 @@
...
+  code: string
+  language: string
+  query: string
+  matchingLineNumbers: number[]
+}
+```
+
+## Syntax highlighting
+
+The `KCodeBlock` component does not have syntax highlighting built in.
```
adamdehaven, line 253:
```suggestion
The `KCodeBlock` component does not have syntax highlighting built in; however, your project can provide its own implementation.
```

src/utilities/copyTextToClipboard.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copies a piece of text to the clipboard. Adapted from [StackOverflow][1].
+ *
+ * [1]: https://stackoverflow.com/a/30810322
```
adamdehaven, line 4: I'm not sure why we had this reference, but let's just remove it
```suggestion
 * Copies a piece of text to the clipboard.
 *
```

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,968 @@
...
+<style lang="scss" scoped>
+.k-code-block {
+  --KCodeBlockBorderRadius: 5px;
+  --KCodeBlockFocusColor: var(--blue-500);
+  --KCodeBlockMatchHighlightColor: var(--blue-500);
+  --KCodeBlockColor: var(--black-85, color(black-85));
+  --KCodeBlockBackgroundColor: var(--grey-100);
+  --KCodeBlockFontSize: var(--type-xs);
+  --KCodeBlockFontFamilyMono: var(--font-family-mono);
+  --KCodeBlockTabSize: 2;
```
adamdehaven, line 746: If you define these variables, a user can never override them. Instead, you have to reference these CSS variables where they are used but then provide the values you have them set to as a fallback.

**You'll need to update all CSS variable usage like my example below**. Let me know if I can help

As an example:

```scss
.k-code-block pre,
.k-code-block code {
  tab-size: var(--KCodeBlockTabSize, 2);
  font-size: var(--KCodeBlockFontSize, var(--type-xs, 12px);
  font-family: var(--KCodeBlockFontFamilyMono, var(--font-family-mono, "Fira Mono", Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace));
  color: var(--KCodeBlockColor, var(--black-85, color(black-85)));
}
```

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: In other search contexts, I would’ve considered trimming the input, but since this will often involve code search, I rather prioritize the flexibility of searching for string with leading and/or trailing space because there are actually use cases for it (e.g. spotting a malformed configuration by searching for `": " value` where the space before `value` is erroneous).

docs/components/codeblock.md
kleinfreund: This is ironically why I introduced the prop in the first place: Currently (and I plan to change this), Kuma GUI uses vue router’s hash history mode meaning that virtually all URLs have a fragment identifier in the URL so there’s no way to use anchors like this.

On the other hand, for sites where this isn’t the case, you can get similar behavior to what GitHub does in code views (e.g. link to specific lines). This is naturally in conflict with pages that have many anchors for other use cases like documentation pages, but I think that’s fine: it’s not a severe conflict. Anchor conflicts are easily avoided using a strategic value for `props.id` and these anchors don’t prevent other anchors from working.

Is there a specific concern you want to address?

src/utilities/copyTextToClipboard.ts
kleinfreund: Done.

docs/components/codeblock.md
kleinfreund: Applied.

docs/components/codeblock.md
kleinfreund: Done.

docs/components/codeblock.md
```diff
@@ -0,0 +1,357 @@
...
+:::
+
+<KCodeBlock
+  id="code-block-default"
+  :code="code"
+  language="json"
+  is-searchable
+/>
+
+```vue
```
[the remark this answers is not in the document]
kleinfreund: Done.

docs/components/codeblock.md
```diff
@@ -0,0 +1,357 @@
...
+An ID used to identify form elements like the search input and its label.
+
+**Note**: If you use multiple instances of `KCodeBlock` in a document, you need to provide unique ID values.
+
+### code
+
+* **Type**: `string`
+* **Required**: yes
+
+The code that will be rendered as a text node inside of a `<pre><code></code></pre>` fragment.
```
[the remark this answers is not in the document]
kleinfreund: Done. I’ve added a new sentence detailing this a bit more.

docs/components/codeblock.md
```diff
@@ -0,0 +1,357 @@
+# CodeBlock
+
+**KCodeBlock** is a code block component with built-in search and filter functionalities.
+
+Searching highlights matching lines within the code while filtering shows only matching line (i.e. like the Unix grep command does). Searching and filtering can happen by exact matches (the default) and by matching regular expressions. The search-related UI controls can be interacted with [using keyboard shortcuts](#default-shortcuts).
```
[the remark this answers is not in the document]
kleinfreund: Applied.

docs/components/codeblock.md
adamdehaven: Matches what I was saying.

Can we add a note under this prop that outlines a use-case when it needs to be disabled, e.g. if the page utilizes hash navigation, etc.?

src/utils/debounce.ts
adamdehaven: I like how you externalized and then modified KClipboardProvider 👍🏼  nice solution

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: TL;DR: I’ve changed this as suggested.

---

That deserves a bit more nuance.

It’s not true that set custom properties can’t be overridden. They can be overridden like any regular CSS property:

```css
.k-code-block.k-code-block {
  --KCodeBlockTabSize: 2;
}

.k-code-block {
  /* the computed value for `var(--KCodeBlockTabSize)` for elements matching the selector will be `2`. */
  --KCodeBlockTabSize: 4;
}
```

What _is_ true is that setting them here will prevent inheritance from rules that were defined earlier or which have selectors with lower specificity (somewhat likely because all Kongponents components unfortunately use scoped style blocks).

```css
:root {
  --KCodeBlockTabSize: 2;
}

.k-code-block {
   /* the computed value for `var(--KCodeBlockTabSize)` for elements matching the selector will be `4`. */
  --KCodeBlockTabSize: 4;
}
```

With the strategy of _not_ setting component-specific properties, you allow for globally defined default styles/themes which is good indeed. However, that comes at a cost: you can no longer specify a component’s default styles without getting very verbose like you suggested. Those defaults also can’t readily be controlled from one central place. Overall, I do think it’s a bigger benefit to allow for globally defined themes and not requiring to specifically set the theme on each component’s main selector (e.g. `.k-code-block { ... }`.

docs/components/codeblock.md
kleinfreund: Yes, I’ve added a sentence for this now.

src/components/KCodeBlock/KCodeBlock.vue
adamdehaven: > What _is_ true is that setting them here will prevent inheritance from rules that were defined earlier or which have selectors with lower specificity (somewhat likely because all Kongponents components unfortunately use scoped style blocks).

This is what I was referring to. A consuming app should be able to set these variables as is standard at the `:root {}` level instead of having to define them within the scope of the component classes.

> However, that comes at a cost: you can no longer specify a component’s default styles without getting very verbose like you suggested

If I understand your meaning, you still can. Instead of defining CSS variables within the component's `style` block, define them as Sass variables. Then you can map to both while still being concise:

```scss
$KCodeBlockBorderRadius: 5px;
$KCodeBlockFocusColor: var(--blue-500);
$KCodeBlockMatchHighlightColor: var(--blue-500);
$KCodeBlockColor: var(--black-85, color(black-85));
$KCodeBlockBackgroundColor: var(--grey-100);
$KCodeBlockFontSize: var(--type-xs);
$KCodeBlockFontFamilyMono: var(--font-family-mono);
$KCodeBlockTabSize: 2;

.k-code-block pre,
.k-code-block code {
  tab-size: var(--KCodeBlockTabSize, $KCodeBlockTabSize);
  font-size: var(--KCodeBlockFontSize, $KCodeBlockFontSize);
  font-family: var(--KCodeBlockFontFamilyMono, $KCodeBlockFontFamilyMono);
  color: var(--KCodeBlockColor, $KCodeBlockColor);
}
```

kleinfreund: That’s a good idea. I’ve defined the component’s own defaults like that now which reduces the verbosity a bit.

I’ve also noticed the missing imports:

```scss
@import '@/styles/variables';
@import '@/styles/functions';
```

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -0,0 +1,985 @@
...
+    }, 1500)
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/styles/variables';
+@import '@/styles/functions';
+
+$borderRadius: 5px;
```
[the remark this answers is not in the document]
kleinfreund: Done.

PR #820, opened 2022-10-10
chore: clean up linter setup

kleinfreund:
# Summary

**chore: split off global properties types**

Splits off the global properties type augmentation (see https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties) because defining it next to the one in vue.d.ts actually breaks its effect of TypeScript no longer complaining about .vue imports.

**chore(deps): cleans up linter setup**

Removes `@vue/eslint-config-standard` and `@vue/eslint-config-typescript`. These packages were used by Vue CLI which this project no longer uses.

Replaces eslint-plugin-node and eslint-plugin-standard with eslint-config-standard and eslint-plugin-n respectively which brings the linter setup in line with how standard is supposed to be configured (see also the deprecation note at https://www.npmjs.com/package/eslint-plugin-standard).

Tunes the ESLint configuration. Specifically, the correct parser for TypeScript is now enabled. Declares the plugins that are in use. Removes unnecessary configuration extensions (e.g. essential is unnecessary when using recommended). Moves TypeScript specific rules (and correspondingly deactivated JavaScript rules) into a file extension-specific override so that they don’t apply to plain JavaScript files.

---

## PR Checklist


adamdehaven (approved):
lgtm -- going to run the `lint --fix` right after merging

PR #821, opened 2022-10-10
fix(kbutton): not considering visually-hidden text

kleinfreund:
# Summary

Replaces the `padding-right` declaration on a button’s icon with a `gap` property on the button itself which allows adding visually-hidden text to an icon button (this would otherwise have an incorrect gap to the right). This works because visually-hidden text won’t participate as a flex or grid item so no gap will be rendered.

---

## Example usage

```vue
<KButton>
  <template #icon>
    <KIcon icon="filter" />
  </template>

  <span class="k-visually-hidden">Filter mode enabled</span>
</KButton>
```

## PR Checklist

### DIFF
--- a/src/components/KButton/KButton.vue
+++ b/src/components/KButton/KButton.vue
@@ -204,6 +204,7 @@ export default defineComponent({
   position: relative;
   display: inline-flex;
   align-items: center;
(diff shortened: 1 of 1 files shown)


adamdehaven (approved):
I thought I already approved this 😬 - looks good

PR #834, opened 2022-10-19
fix(attrs): bind classes to outermost element

adamdehaven:
# Summary

Bind classes to the outermost wrapper element instead of the child element(s). 

As an example, this allows for doing `<KInput class="w-100 mr-4" />` and the classes to be applied correctly to the wrapper element.

If merged, Konnect will need to updated to add a class of `w-100` to most `KTextArea` components.

---

## PR Checklist

### DIFF
--- a/src/components/KSelect/KSelect.vue
+++ b/src/components/KSelect/KSelect.vue
@@ -75,7 +75,7 @@
               :style="widthStyle"
               show-caret
               :is-rounded="false"
-              v-bind="$attrs"
+              v-bind="modifiedAttrs"
               appearance="btn-link"
               @keyup="evt => triggerFocus(evt, isToggled)"
             >
@@ -117,7 +117,7 @@
             />
             <KInput
               :id="selectTextId"
-              v-bind="$attrs"
+              v-bind="modifiedAttrs"
               :model-value="filterStr"
               :label="label && overlayLabel ? label : undefined"
               :overlay-label="overlayLabel"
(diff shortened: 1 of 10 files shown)


src/components/KSelect/KSelect.vue
```diff
@@ -2,6 +2,7 @@
   <div
     :style="widthStyle"
     class="k-select"
+    :class="[$attrs.class]"
```
kaiarrowood, line 5: Do we need to do a `modifiedAttrs` computed like you've done in the other files and adjust the `v-bind` on the `KInput` or `KButton` down below?

src/styles/forms/_inputs.scss
```diff
@@ -51,6 +51,7 @@
 .form-control {
   &:not([type="checkbox"]):not([type="radio"]) {
     display: block;
+    width: 100%;
```
kaiarrowood, line 54: Just checking this is the right place to add it, we intend not to do this for checkbox and radio?

src/components/KSelect/KSelect.vue
adamdehaven: Yep, pushed.

kaiarrowood (approved):
LGTM! 🐉 🔥

jillztom:
I'm curious of this is going to break any classes we are passing in Konnect with the assumption that the classes were meant to be applied to `input` element for example?

adamdehaven:
> I'm curious of this is going to break any classes we are passing in Konnect with the assumption that the classes were meant to be applied to `input` element for example?

We'll definitely need to do some spot-checking before merging the changes into Konnect; however, the worst-case would be some styling changes; nothing should functionally break.

PR #835, opened 2022-10-21
fix(kpop): do not remove on destroy [KHCP-5139]

adamdehaven:
# Summary

`KPop` is throwing an error because of the `removeOnDestroy` setting on the Popperjs element. On the first `popper.destroy()` call the parent exists and the `removeChild` call works successfully; however, subsequent calls fail because the parentNode is no longer defined. I believe this is because of changes in Vue 3 and how the virtual DOM is rendered.

---

## PR Checklist


aanchalm01 (approved):
🚢

PR #846, opened 2022-10-24
fix(kcodeblock): a couple of issues

kleinfreund:
# Summary

fix(kcodeblock): handling small resolutions poorly:

Addresses a few issues with the rendering of `KCodeBlock` in smaller resolutions. The actions bar items are now wrapping appropriately and the search input no longer extends beyond the available size.

fix(kcodeblock): shortcuts using Alt not working on macOS:

Fixes a bug where the default shortcuts involving the Alt key weren’t working on macOS because Options+letter commonly results in a differently typed character. The shortcut mechanism now checks the `event.code` property instead of `event.key` to do this more reliably.

fix(kcodeblock): not preventing default action for some shortcuts:

Fixes an issue with not correctly preventing the default action for some shortcuts (e.g. `Alt+F` opening a Chrome menu by default).

## PR Checklist

### DIFF
--- a/src/components/KCodeBlock/KCodeBlock.spec.ts
+++ b/src/components/KCodeBlock/KCodeBlock.spec.ts
@@ -129,25 +129,40 @@ describe('KCodeBlock', () => {
   it('can be interacted with using default shortcuts', () => {
     const id = 'code-block'
     renderComponent({ id, isSearchable: true, query: 'key' })
+    const codeBlock = cy.get('.k-code-block')
 
     // Tests that scoped shortcuts don’t work when focus is not within the code block.
     cy.document().trigger('keydown', { code: 'F3' })
     cy.get('.k-line-is-highlighted-match').should('not.exist')
 
-    cy.get('.k-code-block').trigger('keydown', { code: 'F3' })
+    codeBlock.trigger('keydown', { code: 'F3' })
     cy.get('.k-line-is-highlighted-match .k-line-anchor').invoke('attr', 'id').should('equal', `${id}-L2`)
 
-    cy.get('.k-code-block').trigger('keydown', { code: 'F3' })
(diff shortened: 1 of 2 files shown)


adamdehaven (commented):
image

When on a smaller screen, the search result count appears at the top of the component when search text is present. This is a little jarring as a user because the entire layout shifts up/down depending on if there is search text. Would it be possible to move the count to the left of the code block "nav buttons" so that the layout doesn't shift?

Everything else lgtm

kleinfreund:
> When on a smaller screen, the search result count appears at the top of the component when search text is present. This is a little jarring as a user because the entire layout shifts up/down depending on if there is search text.

@adamdehaven I agree that this isn’t ideal, but there isn’t much gained from moving it next to the action buttons because that space is also not enough in all cases. As a result, this would avoid a layout shift for a sizing range of about 100 pixels (roughly the difference in width between the action buttons and the input block).

Additionally, this isn’t easy to implement: without broad support for container queries, there isn’t a good CSS-only way to express when this should happen. More explicitly: when would one switch to a different flex box item ordering or select a differently named grid item to accomplish this? Viewport-relational media queries aren’t appropriate here because they don’t allow me to infer anything about the current sizing constraints of the code block.

Let me consider this for a minute.

kleinfreund:
@adamdehaven I’ve reworked the layout of the actions bar to fine-tune layout shifts across all sizes of the code block in https://github.com/Kong/kongponents/pull/846/commits/333964d323a1b534fb12cceaf81a78e885639423.

I’ve taken some inspiration from VS Code’s in-editor search bar to decide what to show where and how to deal with its sizing needs and settled on an approach that always shows _something_ in the place where the number of results are displayed.

adamdehaven (approved):
LGTM 🚀

PR #859, opened 2022-10-28
feat(kmultiselect): new component [khcp-5147]

kaiarrowood:
# Summary

New `KMultiselect` component.
Addresses requirements in [KHCP-53314](https://konghq.atlassian.net/browse/KHCP-3314).
Completes subtask [KHCP-5147](https://konghq.atlassian.net/browse/KHCP-5147).

Updates to `KBadge`:
- dismiss action emits `dismissed`
- minor style updates
- add conditional display of tooltip if badge is truncated
- adds associated props `truncationTooltip` and `alwaysShowTooltip`
- adds `hidden` prop for `KBadge` usage in `KMultiselect` staging area

## PR Checklist

### DIFF
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -203,21 +203,16 @@ export default defineComponent({
 ### autosuggest
 
 Add the `autosuggest` prop to trigger a query to an API with the filter keyword, and then update `items` asynchronously as suggestions as the user types.
-Loading and empty state content can be configured using the `loading` and `empty` slots.
+Empty state content can be configured using the `empty` slot.
 
 <KMultiselect autosuggest
   :items="itemsForAutosuggest"
   :loading="loading"
-  width="300px"
-  appearance="select"
   @query-change="onQueryChange"
 >
   <template v-slot:item-template="{ item }">
     <div class="select-item-label">{{ item.label }}</div>
     <div class="select-item-desc">{{ item.description }}</div>
-  </template>
-  <template v-slot:loading>
-    <div>Loading...</div>
   </template>
   <template v-slot:empty>
     <div>No results found</div>
@@ -229,16 +224,11 @@ Loading and empty state content can be configured using the `loading` and `empty
   autosuggest
   :items="items"
   :loading="loading"
-  width="300px"
-  appearance="select"
   @query-change="onQueryChange"
 >
   <template v-slot:item-template="{ item }">
     <div class="select-item-label">{{ item.label }}</div>
     <div class="select-item-desc">{{ item.label }}</div>
-  </template>
-  <template v-slot:loading>
-    <div>Loading...</div>
   </template>
   <template v-slot:empty>
     <div>No results found</div>
@@ -303,24 +293,25 @@ If you need to send API requests in the `query-change` event handler, you may wa
 The following is an example:
 :::
 
-<KMultiselect autosuggest
-  :items="itemsForDebouncedAutosuggest"
-  :loading="loadingForDebounced"
-  width="300px"
-  @query-change="onQueryChangeForDebounced"
->
-  <template v-slot:item-template="{ item }">
-    <div class="select-item-label">{{ item.label }}</div>
-    <div class="select-item-desc">{{ item.label }}</div>
-  </template>
-</KMultiselect>
+<div>
+  <KMultiselect
+    autosuggest
+    :items="itemsForDebouncedAutosuggest"
+    :loading="loadingForDebounced"
+    @query-change="onQueryChangeForDebounced"
+  >
+    <template v-slot:item-template="{ item }">
+      <div class="select-item-label">{{ item.label }}</div>
+      <div class="select-item-desc">{{ item.label }}</div>
+    </template>
+  </KMultiselect>
+</div>
 
 ```html
 <KMultiselect
   autosuggest
   :items="items"
   :loading="loading"
-  width="300px"
   @query-change="onQueryChange"
 >
   <template v-slot:item-template="{ item }">
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -123,7 +123,7 @@ Use this prop to customize the number of rows of selections to display when the 
 
 ### slimDisplay
 
-By default, we display KMultiselect in `slim` mode. This means that when KMultiselect only takes up a single line when it doesn't have focus, and when focused, if the selected entries would display beyond the `selectedRowCount` we collapse them into the additional count badge. You can set `slimDisplay` to `false` if you want to continue to see the selections even when KMultiselect doesn't have focus. Instead of collapsing additional selections into the additional count badge we will instead allow you to scroll through all of your selections.
+By default, we display KMultiselect in `slim` mode. This means that KMultiselect only takes up a single line when it doesn't have focus, and when focused, if the selected entries would display beyond the `selectedRowCount` we collapse them into the additional count badge. You can set `slimDisplay` to `false` if you want to continue to see the selections even when KMultiselect doesn't have focus. Instead of collapsing additional selections into the additional count badge we will allow you to scroll through all of your selections.
 
 <div>
   <KMultiselect :slim-display="false" :items="deepClone(defaultItemsManySelected)" />
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -102,11 +102,11 @@ Because we are controlling the widths of multiple elements, we recommend using t
 :::
 
 <div>
-  <KMultiselect width="350" :items="deepClone(defaultItemsUnselect)" />
+  <KMultiselect width="80%" :items="deepClone(defaultItemsUnselect)" />
 </div>
 
 ```html
-<KMultiselect width="350" :items="items" />
+<KMultiselect width="80%" :items="items" />
 ```
 
 ### selectedRowCount
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -466,7 +466,7 @@ You can use the `empty` slot to customize the look of the dropdown list when the
 | :--------      | :------------------ |
 | `selected`     | array of selected item objects |
 | `input`        | array of selected item values |
-| `change`       | last selected/deselected Object or null |
+| `change`       | last item selected/deselected Object or null |
 | `query-change` | `query` String |
 
 </div>
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -121,16 +121,16 @@ Use this prop to customize the number of rows of selections to display when the 
 <KMultiselect :selected-row-count="1" :items="items" />
 ```
 
-### slimDisplay
+### expandSelected
 
-By default, we display KMultiselect in `slim` mode. This means that KMultiselect only takes up a single line when it doesn't have focus, and when focused, if the selected entries would display beyond the `selectedRowCount` we collapse them into the additional count badge. You can set `slimDisplay` to `false` if you want to continue to see the selections even when KMultiselect doesn't have focus. Instead of collapsing additional selections into the additional count badge we will allow you to scroll through all of your selections.
+By default, we try to keep the KMultiselect display slim. This means that KMultiselect only takes up a single line when it doesn't have focus, and when focused, if the selected entries would display beyond the `selectedRowCount` we collapse them into the additional count badge. You can set `expandSelected` to `true` if you want to continue to see the selections even when KMultiselect doesn't have focus. Instead of collapsing additional selections into the additional count badge we will allow you to scroll through all of your selections.
 
 <div>
-  <KMultiselect :slim-display="false" :items="deepClone(defaultItemsManySelected)" />
+  <KMultiselect expand-selected :items="deepClone(defaultItemsManySelected)" />
 </div>
 
 ```html
-<KMultiselect :slim-display="false" :items="items" />
+<KMultiselect expand-selected :items="items" />
 ```
 
 ### dropdownMaxHeight
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -121,6 +121,18 @@ Use this prop to customize the number of rows of selections to display when the 
 <KMultiselect :selected-row-count="1" :items="items" />
 ```
 
+### slimDisplay
+
+By default, we display KMultiselect in `slim` mode. This means that when KMultiselect only takes up a single line when it doesn't have focus, and when focused, if the selected entries would display beyond the `selectedRowCount` we collapse them into the additional count badge. You can set `slimDisplay` to `false` if you want to continue to see the selections even when KMultiselect doesn't have focus. Instead of collapsing additional selections into the additional count badge we will instead allow you to scroll through all of your selections.
+
+<div>
+  <KMultiselect :slim-display="false" :items="deepClone(defaultItemsCollapse)" />
+</div>
+
+```html
+<KMultiselect :slim-display="false" :items="items" />
+```
+
 ### dropdownMaxHeight
 
 You can pass a `dropdownMaxHeight` string for the dropdown. By default, the `dropdownMaxHeight` is `300px`. This is the maximum height of the `KMultiselect` dropdown when open. You can pass a number (will be converted to `px`), `auto`, percentages, or `vh` units.
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -192,6 +192,31 @@ export default defineComponent({
   data() {
     return {
       myVal: ['cats', 'bunnies'],
+      items: [{
+        label: 'Cats',
+        value: 'cats',
+        selected: true
+      }, {
+        label: 'Dogs',
+        value: 'dogs'
+      }, {
+        label: 'Bunnies',
+        value: 'bunnies',
+        selected: true
+      },
+      {
+        label: 'Lions',
+        value: 'lions'
+      }, {
+        label: 'Tigers',
+        value: 'tigers'
+      }, {
+        label: 'Bears',
+        value: 'bears'
+      }, {
+        label: 'A long & truncated item',
+        value: 'long'
+      }]
     }
   },
   methods: {
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -1,0 +1,635 @@
+# Multiselect
+
+<div v-if="hasMounted">
+
+**Multiselect** - Multiselect component
+
+<div>
+  <KMultiselect label="Pick Something:" :items="deepClone(defaultItemsUnselect)" />
+</div>
+
+```html
+<KMultiselect label="Pick Something:" :items="items" />
+```
+
+## Props
+
+### items
+
+An array of items containing a `label` and `value`. You may also specify that a certain item is `selected` by default.
+
+<div>
+  <KMultiselect :items="deepClone(defaultItems)" />
+</div>
+
+```html
+<KMultiselect :items="[{
+    label: 'Cats',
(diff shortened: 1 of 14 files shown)


src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+     * The width of the multiselect and popover's min-width
+     */
+    width: {
+      type: String,
+      default: '',
+    },
+    /**
+     * Number of rows of selections to show when focused
+     */
+    selectionRowCount: {
```
adamdehaven, line 306: nit: Since we call items `selected` or not, can we change this to `selectedRowCount` ?

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+     * {
+     *  label: 'Item 1',
+     *  value: 'item1'
+     * }
+     */
+    items: {
+      type: Array as PropType<MultiselectItem[]>,
+      default: () => [],
+      // Items must have a label & value
+      validator: (items: MultiselectItem[]) => !items.length || items.some(i => i.label !== undefined && i.value !== undefined),
```
adamdehaven, line 321: This validator logic is a bit flawed (and it's also incorrect inside `KSelect`)

The `items.some` logic is saying that the prop value is valid if at least one of the array items has both a label and a value.

Instead, it needs to be `items.every` like here:
```suggestion
      validator: (items: MultiselectItem[]) => !items.length || items.every(i => i.label !== undefined && i.value !== undefined),
```

Can you make the exact same update to the `KSelect` `items` validator as well?

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+        if (selectedItems.value.length === 1) {
+          return `${selectedItems.value.length} item selected`
+        }
+        return `${selectedItems.value.length} items selected`
+      }
+
+      if (props.placeholder) {
+        return props.placeholder
+      } else if (attrs.placeholder) {
+        return attrs.placeholder as string
```
adamdehaven, line 458: nit
```suggestion
        return String(attrs.placeholder || '')
```

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+        if (invisibleSelectedItemsStaging.value.length) {
+          const item = invisibleSelectedItemsStaging.value.pop()
+          if (item) {
+            visibleSelectedItemsStaging.value.push(item)
+          }
+        }
+      } else { // newly selected item
+        selectedItem.selected = true
+        selectedItem.key = selectedItem?.key?.includes('-selected') ? selectedItem.key : `${selectedItem.key}-selected`
+        selectedItem.key += '-selected'
```
adamdehaven, line 531: Something in the selected logic is flawed, as I can get selected items into a state that looks like this _(unsure if this is the right place to leave the comment)_

It looks like line `530` adds the `-selected` string in some cases, but then you're adding it again on the following line.

image

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+      invisibleSelectedItemsStaging.value = []
+      filterStr.value = ''
+      stageSelections()
+
+      emit('selected', [])
+      // this 'input' event must be emitted for v-model binding to work properly
+      emit('input', [])
+      emit('change', null)
+      emit('update:modelValue', [])
+      emit('query-change', '')
```
adamdehaven, line 570: nit: can these be put in the same order as the other emit block just for readability?

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+
+    const triggerFocus = (evt: any, isToggled: Ref<boolean>):void => {
+      // Ignore `esc` key
+      if (evt.keyCode === 27) {
+        isToggled.value = false
+        return
+      }
+
+      const inputElem = document.getElementById(multiselectTextId.value)
+      if (!isToggled.value && inputElem) { // simulate click to trigger dropdown open
```
adamdehaven, line 586: nit: readability
```suggestion
      if (!isToggled.value && inputElem) { 
        // simulate click to trigger dropdown open
```

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+              invisibleSelectedItemsStaging.value.push(item)
+            }
+            stagingKey.value++
+          } else {
+            visibleSelectedItems.value = cloneDeep(visibleSelectedItemsStaging.value)
+            invisibleSelectedItems.value = cloneDeep(invisibleSelectedItemsStaging.value)
+            key.value++
+          }
+        }
+      }, 0)
```
adamdehaven, line 628: question: still unsure what a `0` timeout is doing here?

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+      right: 10px;
+    }
+  }
+
+  .k-multiselect-trigger {
+    display: inline-block;
+    position: relative;
+    // mimic input's box shadow styling
+    background-color: var(--KInputBackground, var(--white));
+    box-shadow: inset 0 0 0 1px var(--KInputBorder, var(--grey-300)) !important;
```
adamdehaven, line 808: I think you could use the existing mixin for this?

https://github.com/Kong/kongponents/blob/6afe443a3c8e083a38cd7dc6fee753929c2d9783/src/styles/_mixins.scss#L22-L26

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+    border-radius: 3px;
+
+    &.hovered {
+      box-shadow: inset 0 0 0 1px var(--KInputHover, var(--blue-200)) !important;
+    }
```
adamdehaven, line 813: Could use existing mixin: https://github.com/Kong/kongponents/blob/6afe443a3c8e083a38cd7dc6fee753929c2d9783/src/styles/_mixins.scss#L34-L37

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+
+    &.focused {
+      box-shadow: inset 0 0 0 1px var(--KInputFocus, var(--blue-400)) !important;
+    }
```
adamdehaven, line 817: Could use existing mixin: https://github.com/Kong/kongponents/blob/6afe443a3c8e083a38cd7dc6fee753929c2d9783/src/styles/_mixins.scss#L28-L32

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+    .k-multiselect-input {
+      &.prevent-pointer-events {
+        pointer-events: none;
+      }
+
+      &.input-placeholder-dark::placeholder {
+        color: var(--KInputColor, var(--black-70, rgba(0, 0, 0, 0.7))) !important;
+      }
+
+      input.k-input:not([type=checkbox]):not([type=radio]) {
```
adamdehaven, line 849: question: since `KMultiselect` is always text, could you use that instead of the `:not` rules here?

Also, if you keep this, you should use the quotes in the selectors, e.g.
```
[type="checkbox"]
[type="radio"]
```

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+        height: 100%;
+        // slightly smaller than container so we can see
+        // the container's box-shadow
+        width: 99%;
```
adamdehaven, line 853: suggestion: using a relative width here could get a little dicey in some scenarios. Instead, could you just do something like
```scss
width: calc(100% - 3px); 
```

Where `3px` is the actual width of the box-shadow? (actual value may vary)

src/components/KMultiselect/KMultiselectItem.vue
```diff
@@ -0,0 +1,145 @@
+<template>
+  <div
+    :key="item.key"
+    class="k-multiselect-item"
+    :data-testid="`k-multiselect-item-${item.value}`"
+    @click="handleClick"
```
adamdehaven, line 6: suggestion: Since the `@click` event is actually on this element, could you make this outer element the `<button>` and then the interior `<button>` is just a div since it isn't actually focusable/clickable?

src/components/KMultiselect/KMultiselectItem.vue
```diff
@@ -0,0 +1,145 @@
...
+<script lang="ts">
+import { defineComponent } from 'vue'
+import KIcon from '@/components/KIcon/KIcon.vue'
+
+export default defineComponent({
+  name: 'MultiselectItem',
+  components: { KIcon },
+  props: {
+    item: {
+      type: Object,
```
adamdehaven, line 41: suggestion: Can you type this item as `MultiselectItem` via `PropType`?

If yes, you're going to have to move the `MultiselectItem` type interface to a separate file that then gets imported into `KMultiselect.vue` and `KMultiselectItem.vue` to avoid a circular dependency.

adamdehaven (changes requested):
@kaiarrowood I'm only submitting this before we review together because my Mac is forcing me to update and restart -- I set up time in the morning so we can go through the comments and remove anything that isn't needed 🙌🏼

adamdehaven (commented):
When I type in the `autosuggest` example, and then take focus back to the page, my inputted text still displays until I open the select again 
image

adamdehaven:
Also in the `autosuggest` example:

1. Select `Item 3` in the list
2. Focus away from the select
3. Focus again
4. Type `item 3` to filter the list
5. Select `Item 3` again

You now have the same item selected twice and can continue to select again and again.

image

I think we have to do some sort of deduplication based on the `item.value` as no other field matters.

docs/components/multiselect.md
```diff
@@ -0,0 +1,675 @@
...
+
+<script lang="ts">
+import { defineComponent } from 'vue'
+
+export default defineComponent({
+  data() {
+    return {
+      myVal: ['cats', 'bunnies'],
+    }
+  },
```
adamdehaven, line 193: Can you put the example `items` array in the data so it's clear?

docs/components/multiselect.md
```diff
@@ -0,0 +1,675 @@
...
+
+You can use the `empty` slot to customize the look of the dropdown list when there is no options. See [autosuggest](#autosuggest) for an example of this slot.
+
+## Events
+
+| Event          | returns             |
+| :--------      | :------------------ |
+| `selected`     | array of selected item objects |
+| `input`        | array of selected item values |
+| `change`       | last selected/deselected Object or null |
```
adamdehaven, line 469:
```suggestion
| `change`       | last item selected/deselected Object or null |
```

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+              :id="multiselectSelectedItemsId"
+              :key="key"
+              :style="widthStyle"
+              class="k-multiselect-selections"
+              data-testid="k-multiselect-selections"
+              @click.stop
+            >
+              <KBadge
+                v-for="item, idx in visibleSelectedItems"
+                :key="`${item ? item.key : idx}-badge`"
```
adamdehaven, line 64:
```suggestion
                :key="`${item.key ? item.key : idx}-badge`"
```

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+        selectedItems.value.push(selectedItem)
+        visibleSelectedItemsStaging.value.push(selectedItem)
+      }
+
+      stageSelections()
+      const selectedVals = selectedItems.value.map(anItem => anItem.value)
+
+      emit('selected', selectedItems.value)
+      emit('update:modelValue', selectedVals)
+      // this 'input' event must be emitted for v-model binding to work properly
```
adamdehaven, line 541:
```suggestion
      // this 'update:modelValue' event must be emitted for v-model binding to work properly
```

adamdehaven: Let's try removing `@input` events in the whole component, and just use `update:modelValue`

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+        anItem.key = anItem?.key?.replace(/-selected/gi, '')
+      })
+      selectedItems.value = []
+      visibleSelectedItemsStaging.value = []
+      invisibleSelectedItemsStaging.value = []
+      filterStr.value = ''
+      stageSelections()
+
+      emit('selected', [])
+      // this 'input' event must be emitted for v-model binding to work properly
```
adamdehaven, line 566:
```suggestion
      // this 'update:modelValue' event must be emitted for v-model binding to work properly
```

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+})
+</script>
+
+<style lang="scss" scoped>
+@import '@/styles/variables';
+@import '@/styles/functions';
+
+.k-multiselect {
+  width: fit-content; // necessary for correct placement of popup
+
```
adamdehaven, line 749: Let's add position: relative to keep the `.staging-area` confined.
```suggestion
  position: relative;

```

src/components/KBadge/KBadge.vue
```diff
@@ -54,11 +77,36 @@ export default defineComponent({
...
+     */
+    truncationTooltip: {
+      type: String,
+      default: '',
+    },
+    /**
+     * Use this prop if you always want to show the tooltip whether
+     * or not the badge text is truncated.
+     */
+    alwaysShowTooltip: {
```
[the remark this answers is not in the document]
kaiarrowood: Fixed

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+                evt.stopPropagation()
+              }"
+              @mouseenter="() => isHovered = true"
+              @mouseleave="() => isHovered = false"
+              @blur="() => isFocused = false"
+              @focus="isFocused = true"
+            >
+              <KMultiselectItem
+                v-for="item in sortedItems"
+                :key="item.key"
```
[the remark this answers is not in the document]
kaiarrowood: Fixed

src/components/KMultiselect/KMultiselect.vue
kaiarrowood: Fixed in both

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -0,0 +1,911 @@
...
+              />
+            </div>
+          </div>
+          <template #content>
+            <div
+              class="k-multiselect-list ma-0 pa-0"
+              @click="evt => {
+                // don't close drop down when selecting/deselecting items
+                evt.stopPropagation()
+              }"
```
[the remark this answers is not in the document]
kaiarrowood: Yep, it worked 👍🏻

src/components/KMultiselect/KMultiselect.vue
kaiarrowood: I needed to change this to `1px` so it looked right, not sure why we had `1.5px` when all the other styles are `1px`.

src/components/KMultiselect/KMultiselectItem.vue
kaiarrowood: Struggled with this a little. Ended up leaving the structure the same but moving the `@click` handler onto the `<button>`.

kaiarrowood:
> When I type in the `autosuggest` example, and then take focus back to the page, my inputted text still displays until I open the select again image

Fixed

kaiarrowood:
> In the slots example, if an item is selected, it shows the input text as:
> 
> image 623
> 
> But if I also try to filter and remove the focus, then it changes to what I typed in the input field:
> 
> image 626
> 
> image 627
> 
> Is that expected ?

Nope, fixed to clear the filter string when we close the dropdown now

kaiarrowood:
> Also in the `autosuggest` example:
> 
> 1. Select `Item 3` in the list
> 2. Focus away from the select
> 3. Focus again
> 4. Type `item 3` to filter the list
> 5. Select `Item 3` again
> 
> You now have the same item selected twice and can continue to select again and again.
> 
> image
> 
> I think we have to do some sort of de-duplication based on the `item.value` as no other field matters.

This was actually really tricky. I ended up need to add a note that if you're using `autosuggest` you have to also use `v-model` so we can track the selections between filtering. Here's the commit: https://github.com/Kong/kongponents/pull/859/commits/b64d377716ac89916303b4e832b0d22a5be079b5

(it also does a check of what is already in `selectedItems` before adding it to that array now, so we address the deduping you suggested.

docs/components/multiselect.md
```diff
@@ -0,0 +1,719 @@
...
+    :items="defaultItemsUnselect"
+  />
+</div>
+
+```html
+<KMultiselect
+  label="Name"
+  :label-attributes="{
+    help: 'I use the KLabel `help` prop',
+    'data-testid': 'test'
```
Leopoldthecoder, line 89: It seems setting `data-testid` has no effect. Maybe we should move `v-bind` below `data-testid` [here](https://github.com/Kong/kongponents/blob/da50fc4ab70a063cfb866c62eaccc058dbc267ab/src/components/KMultiselect/KMultiselect.vue#L10-L11).

Leopoldthecoder:
```vue
<KMultiselect
  :selected-row-count="1"
  width="80%"
  :items="deepClone(defaultItemsCollapse)" 
/>
```

Setting `width` to a percentage value causes two issues:
1. `KBadge`s are hidden except the one indicating `invisibleSelectedItems.length`
2. `KPop` has a more narrow width than the input box

Screenshot:

image

src/components/KMultiselect/KMultiselectItem.vue
```diff
@@ -0,0 +1,145 @@
+<template>
+  <div
```
Leopoldthecoder, line 2: Should we use `<li>` instead (also change its wrapper to `<ul>` in `KMultiselect.vue`)? I think `KSelect` is using `<ul>`/`<li>` elements.

adamdehaven: This PR also switches `KSelect` away from the `ul/li` structure. We keep facing issues when slotting and adding additional elements (e.g. a divider) that using `ul/li` makes very limited. By switching everything to `div` we get around the HTML syntax issues we've been seeing.

We could add `role="list"` and `role="listitem"` on the elements though; @kaiarrowood this might be a good addition.

docs/components/multiselect.md
kaiarrowood: Oh, good catch! Fixed this.

kaiarrowood:
> ```
> <KMultiselect
>   :selected-row-count="1"
>   width="80%"
>   :items="deepClone(defaultItemsCollapse)" 
> />
> ```
> 
> Setting `width` to a percentage value causes two issues:
> 
> 1. `KBadge`s are hidden except the one indicating `invisibleSelectedItems.length`
> 2. `KPop` has a more narrow width than the input box
> 
> Screenshot:
> 
> image

Thanks for catching this! I adjusted all but the top width styles to use a numeric width calculated off the top element.

adamdehaven (approved):
LGTM 🚀  nice job @kaiarrowood !

PR #892, opened 2022-11-10
feat(kselect): support disabled item

Leopoldthecoder:
# Summary

This PR adds disabled items support for `KSelect`.

## PR Checklist

### DIFF
--- a/src/components/KSelect/KSelectItem.vue
+++ b/src/components/KSelect/KSelectItem.vue
@@ -10,7 +10,8 @@
       class="d-block"
     >
       <button
-        :class="{ disabled, selected: item.selected }"
+        :class="{ selected: item.selected }"
+        :disabled="item.disabled"
         :value="item.value"
       >
         <span class="k-select-item-label mr-2">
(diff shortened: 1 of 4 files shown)


src/components/KSelect/KSelectItem.vue
```diff
@@ -10,7 +10,8 @@
       class="d-block"
     >
       <button
-        :class="{ disabled, selected: item.selected }"
+        :class="{ selected: item.selected }"
+        :disabled="item.disabled"
```
adamdehaven, line 14: I think we need to update this logic a bit so that we don't end up with `disabled="false"` on the element.
```suggestion
        :disabled="item.disabled === true ? true : undefined"
```

adamdehaven (commented):
@Leopoldthecoder I believe you also need to update the `item` interface here: https://github.com/Kong/kongponents/blob/d364c4f618b7bc248cb3276c99906f79f78828b4/src/components/KSelect/KSelect.vue#L209-L214

Leopoldthecoder:
@adamdehaven Updated.

adamdehaven (approved):
LGTM

PR #902, opened 2022-11-15
fix(kcodeblock): code-block-render event being emitted too soon

kleinfreund:
# Summary

Fixes a small issue with the `code-block-render` event being emitted before re-rendering is actually done. This is only an issue **when changing the `props.code`** as the event is already delayed by one Vue tick when turning off filter mode and doesn’t need to be delayed when emitting the event as part of the `mounted` life cycle hook.

## PR Checklist


adamdehaven (approved):
lgtm

PR #911, opened 2022-11-17
feat(ktreeelist): add component [khcp-5382]

kaiarrowood:
# Summary

## PR Checklist


src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+})
+
+const hasChildren = (item: TreeListItem) => {
+  return !!internalList.value.filter(anItem => anItem.id === item.id)?.[0].children?.length
+}
+
+const indentStyle = computed(() => {
+  const level1offset = props.level === 1 ? 6 : 0
+  return {
+    marginLeft: 8 * props.level + level1offset + 'px',
```
jillztom, line 169: Couldn't we do this through css pseudo elements? `:fist-child` and `:last-child`?

src/components/KTreeList/KTreeItem.vue
```diff
@@ -0,0 +1,107 @@
+<template>
+  <div
```
jillztom, line 2: Should we make this an anchor tag so you can tab through the items?

adamdehaven: `<a>` with `role="button"` -- it will also need `href="#" and `@click.prevent="handleClick"`

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+          </KTreeList>
+        </div>
+      </template>
+    </draggable>
+  </div>
+</template>
+
+<script lang="ts" setup>
+import { defineProps, defineEmits, computed, ref, watch, onMounted, PropType } from 'vue'
+import draggable from 'vuedraggable'
```
jillztom, line 85: [vue-draggable-next](https://www.npmjs.com/package/vue-draggable-next) has a good nested draggable solution.

docs/components/tree-list.md
```diff
@@ -0,0 +1,330 @@
+# Tree List
+
+**KTreeList** - A drag-n-drop reorderable list component.
+
+::: tip
+It can be tricky to determine if an item will be places below an item as a sibling or as a child when dropping. Items to be dropped as children should be dropped near the bottom border line of the parent. If this will be the first child of an item, you will notice the margin between the two items disappearing if you are hovering in the right place.
```
jillztom, line 6: For this problem, I think increasing the drop zone height could give more space to get the feeling that the element will be placed there?

aanchalm01, line 6: nit - replace `an item will be places` with `an item will be placed`

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+              </slot>
+            </template>
+          </KTreeItem>
+          <KTreeList
+            :key="`tree-item-children-${key}`"
+            v-model="element.children"
+            :level="level + 1"
+            :max-level="maxLevel"
+            :disable-drag="disableDrag"
+            :parent-id="element.id"
```
adamdehaven, line 59: question: I'm a little confused here; how/when does `element.id` ever get set?

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+  set(newValue: TreeListItem[]): void {
+    internalList.value = newValue
+  },
+})
+
+const hasChildren = (item: TreeListItem) => {
+  return !!internalList.value.filter(anItem => anItem.id === item.id)?.[0].children?.length
+}
+
+const indentStyle = computed(() => {
```
adamdehaven, line 166:
```suggestion
const indentStyle = computed((): Record<string, any> => {
```

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+  const level1offset = props.level === 1 ? 6 : 0
+  return {
+    marginLeft: 8 * props.level + level1offset + 'px',
+  }
+})
+
+const iconSecondaryColor = (item: TreeListItem) => {
```
adamdehaven, line 173:
```suggestion
const iconSecondaryColor = (item: TreeListItem): string | undefined => {
```

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+  if (item.icon === 'treeDoc' || !item.icon) {
+    return item.selected ? 'var(--teal-200)' : 'var(--grey-200)'
+  }
+
+  return undefined
+}
+
+const handleSelection = (itemToSelect: TreeListItem, list?: TreeListItem[]) => {
```
adamdehaven, line 181:
```suggestion
const handleSelection = (itemToSelect: TreeListItem, list?: TreeListItem[]): void => {
```

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+      item.selected = false
+      if (item.children?.length) {
+        handleSelection(itemToSelect, item.children)
+      }
+    })
+  }
+  emit('selected', itemToSelect)
+}
+
+const handleSetParent = (item: TreeListItem, parent: TreeListItem) => {
```
adamdehaven, line 222: Can you add a return type here?

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+  if (originalParent) {
+    // remove the item from the original parent
+    if (!originalParent.children) {
+      originalParent.children = []
+    }
+    originalParent.children = originalParent.children.filter((child: TreeListItem) => child.id !== item.id)
+  }
+}
+
+const handleChangeEvent = () => {
```
adamdehaven, line 257:
```suggestion
const handleChangeEvent = (): void => {
```

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+      children: internalList.value,
+    })
+  } else {
+    emit('change', {
+      items: internalList.value,
+    })
+  }
+}
+
+const maxLevelReached = computed(() => {
```
adamdehaven, line 270:
```suggestion
const maxLevelReached = computed((): boolean => {
```

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+
+/**
+     * Recursive check to get the maximum depth of an object.
+     * Documented here: https://stackoverflow.com/a/48505969
+     */
+const getMaximumDepth = ({ children = [] }): number => {
+  return children.length === 0 ? 0 : 1 + Math.max(...children.map(getMaximumDepth))
+}
+
+const checkMove = (target: any) => {
```
adamdehaven, line 282: I think this one is right?
```suggestion
const checkMove = (target: any): boolean => {
```

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+  // Only trigger the watcher if items actually change
+  if (JSON.stringify(newValue) === JSON.stringify(oldValue)) {
+    internalList.value = newValue
+  }
+})
+
+onMounted(() => {
+  if (props.items) {
+    internalList.value = props.items
+  } else if (props.modelValue) {
```
adamdehaven, line 346: question: Wouldn't `v-model` or `modelValue` take precedence over `props.items`?

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+  }
+}
+</style>
+
+<style lang="scss">
+// override cursor as grabbing when an item is being dragged
+.k-tree-list-grabbing * {
+  cursor: move !important; /* fallback: no `url()` support or images disabled */
+  cursor: -webkit-grabbing !important; /* Chrome 1-21, Safari 4+ */
+  cursor: -moz-grabbing !important; /* Firefox 1.5-26 */
```
adamdehaven, line 416: You shouldn't need the vendor prefixes here: https://caniuse.com/?search=grabbing
```suggestion
```

src/components/index.ts
```diff
@@ -33,6 +33,7 @@ export { default as KMenu } from './KMenu/KMenu.vue'
 export { default as KMenuItem } from './KMenu/KMenuItem.vue'
 export { default as KModal } from './KModal/KModal.vue'
 export { default as KModalFullscreen } from './KModalFullscreen/KModalFullscreen.vue'
+export { default as KMultiselect } from './KMultiselect/KMultiselect.vue'
```
adamdehaven, line 36: Alphabetical 👍🏼

docs/components/tree-list.md
```diff
@@ -0,0 +1,330 @@
...
+
+Use this prop to customize the maximum supported depth of the tree. We default to a max level of `2` or support for parents, children, and grandchildren.
+
+<KTreeList :items="maxLevelItems" :max-level="1" />
+
+```html
+<KTreeList :items="items" :max-level="1" />
+```
+
+### v-model
```
adamdehaven, line 76: If `v-model` usage is preferred, I think it should be listed above the `items` prop

docs/components/tree-list.md
```diff
@@ -0,0 +1,330 @@
...
+})
+</script>
+```
+
+## KTreeItem
+
+### Properties
+
+- `name` (required) - text displayed as the label for the item
+- `id` (required) - a unique value used to identify the item
```
adamdehaven, line 151:
```suggestion
- `id` (required) - a unique `string` used to identify the item
```

docs/components/tree-list.md
```diff
@@ -0,0 +1,330 @@
...
+- `selected` - boolean to indicate whether the current item is selected or not
```
adamdehaven, line 152: I don't understand `selected` -- how is an item "selected" in the list..?

docs/components/tree-list.md
```diff
@@ -0,0 +1,330 @@
...
+  </template>
+</KTreeList>
+```
+
+## Events
+
+- `@change` - Emitted when there is a change to the root level items
+  - returns `items` - an array of `KTreeItem`s
+- `@child-change` - Emitted when an item is added or removed at the non-root level
+  - returns `parent` - id of the parent item; `children` - an array of `KTreeItem`s
```
adamdehaven, line 199: question: what if the parent doesn't have an `id`?

docs/components/tree-list.md
```diff
@@ -0,0 +1,330 @@
...
+  - **Note:** two separate `child-change` events will fire if an item is moved from one parent to another
```
adamdehaven, line 200: This is confusing to me. How will the consuming app know which event is correct? Is there a way to emit another value with the payload like `add` or `remove`?

docs/components/tree-list.md
```diff
@@ -0,0 +1,330 @@
...
+  }
+})
+</script>
+```
+
+## KTreeItem
+
+### Properties
+
+- `name` (required) - text displayed as the label for the item
```
adamdehaven, line 150: I think these properties would go in the `items` prop documentation.

The `KTreeItem` component shouldn't be documented since its internal and cannot be used by a developer individually.

docs/components/tree-list.md
```diff
@@ -0,0 +1,330 @@
...
+- `id` (required) - a unique value used to identify the item
+- `selected` - boolean to indicate whether the current item is selected or not
+- `icon` - `KIcon` name to be displayed to the left of the item `name` (defaults to `treeDoc`, specify `none` to not display any icon)
+- `children` - an array of `KTreeItem`'s that will be styled as children of the current item
```
adamdehaven, line 154: When you move this under the `items` prop, this would be changed to `an array of items that will be styled...`

docs/components/tree-list.md
```diff
@@ -0,0 +1,330 @@
...
+
+### Item Slots
+
+- `item-icon` - slot for content displayed to the left of the item name
+- `item-label` - slot for the main content of an item (default's to display the `name` of the item)
+
+See the [Slots section](#Slots) for an example.
```
adamdehaven, line 161: I don't think these would be documented anymore.

adamdehaven (dismissed):
Functional questions/issues:

1. I can't seem to be able to drag a L1 item to the bottom of the list without making it a child of the last L1 item.
    image
2. The only way it appears that you can move a L2 item to the top-level (to become L1) is to drag it to the top of the list
3. There doesn't appear to be a visual way of telling where I am dropping an item into the list. The "blue line" appears but doesn't seem to indicate if I am moving an item to be L1 or a child element. I'm not sure what the fix is, but as a user I definitely would want a way to tell where exactly I'm moving the item to.
4. Are the `teal-100` and `teal-200` colors what we want to be using here? They don't match anything else I can find in Kongponents 😬

src/components/KIcon/icons/index.ts
```diff
@@ -87,6 +87,7 @@ export { default as table } from './icn-table.svg?raw'
 export { default as team } from './icn-team.svg?raw'
 export { default as teamMember } from './icn-team-member.svg?raw'
 export { default as trash } from './icn-trash.svg?raw'
+export { default as treeDoc } from './icn-tree-doc.svg?raw'
```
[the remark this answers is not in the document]
jillztom: > one line in the middle darker than all the rest

I think that's intentional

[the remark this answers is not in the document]
adamdehaven: But should it be? It seems like there are too many lines in the icon; you can't even really tell what it's supposed to be. Here's a similar icon
image

adamdehaven:
It looks like every item is also generating a nested `k-tree-list` even if it doesn't have children
image

docs/components/tree-list.md
kaiarrowood: If I update that it adds more margin between the items making it stop matching the design.

src/components/KTreeList/KTreeList.vue
kaiarrowood: We are using `vue-draggable@next` is that not the right package? I got confused by this. Should it be `vue-draggable-next` or `vue-draggable@next`?

src/components/KTreeList/KTreeList.vue
kaiarrowood: We could do this if we didn't need to calculate the total left margin based off the level, but since we most multiple by the level to get the left margin it doesn't make sense to do some of the calculation in CSS and some of it in JS.

docs/components/tree-list.md
kaiarrowood: All items are required to have a `name` and `id`, so it will.

docs/components/tree-list.md
kaiarrowood: They are both correct, you are removing the item from one parent and adding it to another. The consuming app would need to update both `parent` items with the `children` arrays we provide on the event in order for the list to still be accurate. I tried simplifying this down to `add` / `remove` events but because we are piggybacking off the `draggable` events we are very limited in our options. I couldn't get that to work so instead I just provide you with the `id` of the item that changed and the updated `children` array.

src/components/KIcon/KIcon.vue
```diff
@@ -183,6 +183,8 @@ export default defineComponent({
 
       if (!hasPreservedColor && fillValue && isSecondary && props.secondaryColor) {
         el.setAttribute('fill', props.secondaryColor)
+      } else if (!hasPreservedColor && strokeValue && isSecondary && props.secondaryColor) {
+        el.setAttribute('stroke', props.secondaryColor)
```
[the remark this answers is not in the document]
kaiarrowood: As I called out in the PR description, it adds support for setting the `stroke` of a `secondaryColor` on the icon. Previously, we only allowed you to set the `fill` of a `secondaryColor`.

src/components/KTreeList/KTreeItem.vue
```diff
@@ -0,0 +1,107 @@
...
+
+ .k-tree-item {
+  color: var(--black-500);
+  border: 1px solid var(--grey-200);
+  border-radius: 8px;
+  padding: 4px;
+
+  .k-tree-item-icon {
+    height: 20px;
+  }
```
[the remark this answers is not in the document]
kaiarrowood: I fiddled with this for about 20 minutes, I don't know why, but if I don't explicitly set the height the stupid thing ends up with a height of `30px` and I can't see any added margin or padding that would cause this

src/components/KTreeList/KTreeItem.vue
```diff
@@ -0,0 +1,107 @@
...
+    border-color: var(--teal-200);
+  }
+
+  &:hover {
+    cursor: grab;
+  }
+
+  &.not-draggable {
+    cursor: default;
+  }
```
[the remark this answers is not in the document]
kaiarrowood: That is the `selected` state. This is completely unrelated to drag-n-drop support

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+      <template #item="{ element }">
+        <div
+          :style="indentStyle"
+          :class="{
+            'non-root': level !== 0
+          }"
+          class="k-tree-item-container"
+        >
+          <KTreeItem
+            :key="`tree-item-${key}`"
```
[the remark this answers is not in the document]
kaiarrowood: This is used to trigger manual redraw of the element after a drag-n-drop event. I will update it to also use the item's `id` so they are unique.

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+}
+
+const props = defineProps({
+  modelValue: {
+    type: Array as PropType<TreeListItem[]>,
+    default: () => [],
+  },
+  items: {
+    type: Array as PropType<TreeListItem[]>,
+    default: null,
```
adamdehaven, line 104: Since you're saying `item.id` is required (along with the other required params) you need to add a validator here to ensure the properties are set accordingly.

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+            }"
+            @selected="handleSelection"
+          >
+            <template #item-icon>
+              <slot
+                :item="element"
+                name="item-icon"
+              >
+                <KIcon
+                  v-if="element.icon !== 'none'"
```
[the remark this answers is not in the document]
kaiarrowood: No, we don't have a use case for this, but I figured we should offer it. Additionally, we can't use a prop because this field is at the item level and it could be different for each item (we may want to show an icon for some but not others).

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+                  :icon="element.icon ? element.icon : 'treeDoc'"
+                  :secondary-color="iconSecondaryColor(element)"
+                  size="24"
```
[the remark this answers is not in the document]
kaiarrowood: Explained above

docs/components/tree-list.md
adamdehaven: I understand what you're saying, but again this seems really confusing to me without a fully working example in the docs.

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+            :key="`tree-item-children-${key}`"
+            v-model="element.children"
+            :level="level + 1"
+            :max-level="maxLevel"
+            :disable-drag="disableDrag"
+            :parent-id="element.id"
+            @change="handleChangeEvent"
+            @selected="handleSelection"
+          >
+            <!-- if TS errors appear on these lines switch to `v-slot:` notation and Save to let linter clear -->
```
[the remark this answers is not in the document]
kaiarrowood: We probably can, but I don't know if we should. As soon as someone causes a TS error in this file, the TS compiler errors on this line and you have to do the fix I describe in order for it to go away again.

src/components/KTreeList/KTreeItem.vue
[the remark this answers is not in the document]
adamdehaven: Maybe try adjusting the line-height to `1` ?

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,419 @@
...
+<script lang="ts" setup>
+import { defineProps, defineEmits, computed, ref, watch, onMounted, PropType } from 'vue'
+import draggable from 'vuedraggable'
+import KTreeItem, { TreeListItem } from '@/components/KTreeList/KTreeItem.vue'
+
+export interface ChangeEvent {
+  items: TreeListItem[]
+}
+
+export interface ChildChangeEvent {
```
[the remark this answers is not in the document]
kaiarrowood: If I don't have `export` I get errors trying to reference them in the `defineEmits`
image

src/components/KTreeList/KTreeItem.vue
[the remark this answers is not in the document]
adamdehaven: What does "selecting" an item do...? I see it's emitted from `KTreeItem` but since it's not emitted from the parent `KTreeList` it doesn't seem like the user can actually listen for something being selected?

src/components/KTreeList/KTreeList.vue
kaiarrowood: They should be doing one or the other, but sure, I can switch the ordering

src/components/KTreeList/KTreeList.vue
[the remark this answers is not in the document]
adamdehaven: I'll also say I think this may only be an issue with your local machine; I never get the type error

src/components/KTreeList/KTreeItem.vue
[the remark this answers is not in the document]
kaiarrowood: Oops, I need to be emiting from `KTreeList`.

Here is the mock:

image

The `selected` state will be used to give an indication of which item you are displaying the content of in the main section. I'll update the code `KTreeList` to emit it as well.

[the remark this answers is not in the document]
kaiarrowood: https://github.com/Kong/kongponents/pull/911/files#diff-bb325a6b7c3cee3e1bf290610bf221ce08bd05ea9cf7e03f83e615be208aaaecR219

Here's where we are doing the emit in `KTreeList`

src/components/KTreeList/KTreeList.vue
kaiarrowood: Ok, I thought I could get it for free since it's required in the type interface 😞

docs/components/tree-list.md
kaiarrowood: Added an example

src/components/KTreeList/KTreeList.vue
jillztom: I would opt for `vue-draggable-next`. It's more recently updated and has got better grouping, nesting and transitions

src/components/KTreeList/KTreeList.vue
jillztom: I don't understand why we need to calculate the `left-margin` based off of the level? Shouldn't that automatically be taken care when we use the pseudo selector?

jillztom: In your case, applying `margin-left: 8px` to the `.k-tree-item-container` should do the trick and we don't have to calculate margin-left for all children. Also to avoid the margin in the most parent, you can have a `margin-left: -8px` in the `.k-tree-list` class.

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,456 @@
...
+      :sort="false"
+      @start="onStartDrag"
+      @end="onStopDrag"
+      @change="handleChangeEvent"
+    >
+      <template #item="{ element }">
+        <div
+          :style="indentStyle"
+          :class="{
+            'non-root': level !== 0
```
jillztom, line 23: Curious why we need this `non-root` class?

kaiarrowood: So that we don't get the connector lines that child items have to their parents

jillztom: You can target the immediate children using the `>` operator and do something like this: 
`.k-tree-select > .k-tree-item` and set `display: none` for any pseudo selectors.

kaiarrowood: We can't use this because we create an instance of `KTreeList` inside of `KTreeList` and those are the nested children. This means that there will be `.k-tree-list ... .k-tree-list` instances that we sometimes need to apply to.

docs/components/tree-list.md
kaiarrowood: 10px might be a big much, we have 4px right now. Here is how it would look with 10px.
image

src/components/KTreeList/KTreeList.vue
kaiarrowood: This solution won't work for nesting multiple levels deep. If we have children 1, 2, and 3 levels deep we are going to have a different margin for each level... maybe I'm not understanding what you are suggesting.

src/components/KTreeList/KTreeList.vue
jillztom: We should be able to have a wrapper element for the entire KTree. May be giving a class name to that element will give us an option to only select the L1 items. So you can do:
```
.k-tree-wrapper > .[immediate-child-class-name]
```

If you want a quick reference: https://codesandbox.io/s/eloquent-shamir-op101f

src/components/KTreeList/KTreeList.vue
kaiarrowood: I will try doing a straight switch over from `vuedraggable@next` to `vue-draggable-next` but if that doesn't work I'd prefer to stick with `vuedraggable@next` since I've spent a large number of hours getting it to work already.

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,458 @@
...
+    return item.selected ? 'var(--teal-200)' : 'var(--grey-200)'
+  }
+
+  return undefined
+}
+
+const handleSelection = (itemToSelect: TreeListItem, list?: TreeListItem[]): void => {
+  if (!list) { // root level
+    // select the item
+    const selectedItem = internalList.value.filter((item: TreeListItem) => item.id === itemToSelect.id)?.[0]
```
jillztom, line 201: Quick one: Isn't `TreeListItem` what we need here? Could we avoid the filtering in that case?

src/components/KTreeList/KTreeList.vue
kaiarrowood: Hooray! We got lucky and the switch over was super easy. We are now using `vue-draggable-next`

src/components/KTreeList/KTreeList.vue
jillztom: It will work for any number of levels. Sorry if it's not making clear. May be refer to the codesandbox link I provided earlier? It has multiple levels and works without any issues.

kaiarrowood:
> * I can't seem to be able to drag a L1 item to the bottom of the list without making it a child of the last L1 item.

I THINK this is related to how small the drop zone is, so if we increase the drop zone maybe this will become possible.

>   image
> * The only way it appears that you can move a L2 item to the top-level (to become L1) is to drag it to the top of the list

Same comment about drop zone size

> * There doesn't appear to be a visual way of telling where I am dropping an item into the list. The "blue line" appears but doesn't seem to indicate if I am moving an item to be L1 or a child element. I'm not sure what the fix is, but as a user I definitely would want a way to tell where exactly I'm moving the item to.

I had some logic to set a grey background to try to differentiate this but Josh asked me to remove it, so we may just have to make do.

> * Are the `teal-100` and `teal-200` colors what we want to be using here? They don't match anything else I can find in Kongponents 😬

It's what Josh chose in the design and we already offer these colors as vars so I guess this component gets to be the first instance of it.

adamdehaven:
> I had some logic to set a grey background to try to differentiate this but Josh asked me to remove it, so we may just have to make do.

I think we should push back on this, but I know a lot of changes have gone in since I left my review so maybe we can revisit? A user needs to be able to tell where the item will be inserted before they release their mouse.

docs/components/tree-list.md
```diff
@@ -0,0 +1,429 @@
...
+
+<KTreeList disable-drag :items="disableItems" />
+
+```html
+<KTreeList disable-drag :items="items" />
+```
+
+### maxLevel
+
+Use this prop to customize the maximum supported depth of the tree. We default to a max level of `2` or support for parents, children, and grandchildren.
```
adamdehaven, line 131: > We default to a max level of `2` or support for parents, children, and grandchildren.

This sentence doesn't make sense to me? Typo maybe?

docs/components/tree-list.md
```diff
@@ -0,0 +1,429 @@
...
+
+<KTreeList :items="maxLevelItems" :max-level="1" />
+
+```html
+<KTreeList :items="items" :max-level="1" />
+```
+
+## Slots
+
+`KTreeList` allows you to customize individual tree items via the item slots. These slots return the current `item` data as a slot param.
```
adamdehaven, line 141:
```suggestion
`KTreeList` allows you to customize individual tree items via the item slots. The slots provide the current `item` data as a slot param.
```

docs/components/tree-list.md
```diff
@@ -0,0 +1,429 @@
...
+
+- `item-icon` - slot for content displayed to the left of the item name
```
adamdehaven, line 143:
```suggestion
- `item-icon` - slot for content displayed to the left of the item name in place of the default icon
```

docs/components/tree-list.md
```diff
@@ -0,0 +1,429 @@
...
+- `item-label` - slot for the main content of an item (default's to display the `name` of the item)
```
adamdehaven, line 144:
```suggestion
- `item-label` - slot for the main content of an item (defaults to the `name` of the item)
```

docs/components/tree-list.md
```diff
@@ -0,0 +1,429 @@
...
+
+<KTreeList :items="slotItems">
+  <template #item-icon="{ item }">
+    {{ item.id === 'cats' ? '😸' : item.id === 'bunnies' ? '🐰' : '🐶' }}
+  </template>
```
adamdehaven, line 149: You may need to adjust the line-height a little for these emojis

docs/components/tree-list.md
```diff
@@ -0,0 +1,429 @@
...
+  <template #item-label="{ item }">
+    <span class="color-purple-400">
+    {{ item.name }}
+    </span>
```
adamdehaven, line 153: suggestion: To show that you're slotting something here besides just a color change, what if you added a prefix to the name or something as well?
```suggestion
    <span class="color-purple-400">
    Animal: {{ item.name }}
    </span>
```

(Don't forget to update the example too)

docs/components/tree-list.md
adamdehaven, line 143: super-nit: When you documented the events, the sentence after the "dash" starts with a capital letter. 

All instances

docs/components/tree-list.md
```diff
@@ -0,0 +1,429 @@
...
+```
+
+## Events
+
+- `@change` - Emitted when there is a change to the root level items
+  - returns `items` - an array of tree items
+- `@child-change` - Emitted when an item is added or removed at the non-root level
+  - returns `parent` - id of the parent item; `children` - an array of tree items
+  - **Note:** two separate `child-change` events will fire if an item is moved from one parent to another
+- `@selected` - Emitted when you click (and don't drag-n-drop) an item; returns the selected item's data
```
adamdehaven, line 177: What if you click and then drag? Does it only fire on click and release?

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,170 @@
...
+const props = defineProps({
+  modelValue: {
+    type: Array as PropType<TreeListItem[]>,
+    default: null,
+    validator: (items: TreeListItem[]) => !items.length || items.every(i => i.name !== undefined && i.id !== undefined),
+  },
+  items: {
+    type: Array as PropType<TreeListItem[]>,
+    default: null,
+    validator: (items: TreeListItem[]) => !items.length || items.every(i => i.name !== undefined && i.id !== undefined),
```
adamdehaven, line 56: A couple issues:

In the validator (here on `modelValue` as well as the `items` validator), you also need to loop through all items and ensure all `id` properties are unique. 

Additionally, this validator doesn't account for validating children recursively, meaning the L1 items may pass validation, but all children (infinite levels) would never be evaluated. 

Since you're using the same validation logic on both props, I would extract the logic into a function to call within the validator that returns a boolean. Then you can use the same function wherever the validator is needed.

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,170 @@
...
+  },
+  disableDrag: {
+    type: Boolean,
+    default: false,
+  },
+  maxLevel: {
+    type: Number,
+    default: 2,
+  },
```
adamdehaven, line 65: suggestion: we should add a reasonable maximum to the `maxLevel` prop and enforce it via a prop validator. 

Could we check with the backend teams to see if they have any restrictions? Otherwise I'd say something like `6`

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,170 @@
...
+  items: {
+    type: Array as PropType<TreeListItem[]>,
+    default: null,
+    validator: (items: TreeListItem[]) => !items.length || items.every(i => i.name !== undefined && i.id !== undefined),
+  },
+  disableDrag: {
+    type: Boolean,
+    default: false,
+  },
+  maxLevel: {
```
adamdehaven, line 62: question: I see that we enforce `maxLevel` by disallowing the user to create a new set of children; however, how is this enforced if the user passes in a `v-model` or `items` value that already exceeds the maximum?

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,170 @@
...
+  },
+})
+
+const emit = defineEmits<{
+  (event: 'change', data: ChangeEvent): void,
+  (event: 'child-change', data: ChildChangeEvent): void,
+  (event: 'selected', item: TreeListItem): void
+}>()
+
+// const key = ref(0)
```
adamdehaven, line 74: nit: can we remove?

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,170 @@
...
+    })
+  }
+})
+
+onMounted(() => {
+  if (props.modelValue) {
+    internalList.value = props.modelValue
+  } else if (props.items) {
+    internalList.value = props.items
+  }
```
adamdehaven, line 141: Should we `console.warn` if the user defines both `v-model` (or `modelValue`) along with `items`? 

Everything would still work as you have it, but we could say something like:

```ts
console.warn('KTreeList: You should not provide both v-model (or props.modelValue) along with props.item')
```

adamdehaven (changes requested):
@kaiarrowood I submitted the code review; I haven't functionally tested. 
 
@jillztom I played around on the deploy preview and noticed that the component, while it displays fine, is essentially unusable on a touch device (e.g. my phone). Is there anything we can do?

src/components/KTreeList/KTreeList.vue
adamdehaven, line 62: Kind of a nit: I think this prop name should be plural
```suggestion
  maxLevels: {
```

src/components/KTreeList/KTreeDraggable.vue
```diff
@@ -0,0 +1,345 @@
...
+  items: {
+    type: Array as PropType<TreeListItem[]>,
+    required: true,
+    validator: (items: TreeListItem[]) => !items.length || items.every(i => i.name !== undefined && i.id !== undefined),
+  },
+  disableDrag: {
+    type: Boolean,
+    default: false,
+  },
+  maxLevel: {
```
adamdehaven, line 112: See suggestion in other file
```suggestion
  maxLevels: {
```

src/components/KTreeList/KTreeDraggable.vue
```diff
@@ -0,0 +1,345 @@
...
+    })
+  }
+}
+
+const handleSelectionEvent = (item: TreeListItem) => {
+  emit('selected', item)
+}
+
+const maxLevelReached = computed((): boolean => {
+  return props.level > props.maxLevel
```
adamdehaven, line 188: I'm not exactly sure, but is this correct or should it be `>=` ?

src/components/KTreeList/KTreeDraggable.vue
```diff
@@ -0,0 +1,345 @@
...
+})
+
+/**
+ * Recursive check to get the maximum depth of an object.
+ * Documented here: https://stackoverflow.com/a/48505969
```
adamdehaven, line 193: nit: can we remove the stack overflow reference? If we need more comments that's fine, but let's not leave this in the codebase

src/components/KTreeList/KTreeDraggable.vue
```diff
@@ -0,0 +1,345 @@
...
+  dragging.value = false
+  setDragCursor(false)
+
+  key.value++
+}
+
+// override cursor when dragging
+const setDragCursor = (value: boolean) => {
+  // must be on html element to keep style applied no matter where they drag
+  const html = document.getElementsByTagName('html').item(0)
```
adamdehaven, line 234: Just in case this gets called before mount somehow (unlikely)
```suggestion
  const html = document?.getElementsByTagName('html').item(0)
```

src/components/KTreeList/KTreeItem.vue
```diff
@@ -0,0 +1,106 @@
...
+  selected?: boolean
+  icon?: string
+  children?: TreeListItem[]
+}
+
+const props = defineProps({
+  item: {
+    type: Object as PropType<TreeListItem>,
+    required: true,
+    validator: (item: TreeListItem) => item.name !== undefined && item.id !== undefined,
```
adamdehaven, line 47: Same comments on validator issues here.

src/components/KTreeList/KTreeList.cy.ts
```diff
@@ -0,0 +1,8 @@
+import { mount } from 'cypress/vue'
+import KTreeList from '@/components/KTreeList/KTreeList.vue'
+
+describe('KTreeList', () => {
```
adamdehaven, line 4: We definitely need tests before merge

src/components/KTreeList/KTreeList.vue
```diff
@@ -0,0 +1,170 @@
...
+    </KTreeDraggable>
+  </div>
+</template>
+
+<script lang="ts" setup>
+/**
+ *  Note: if TS errors appear on the
+ *  <template #[itemIcon]="slotProps"> or
+ *  <template #[itemLabel]="slotProps">
+ * lines switch to `v-slot:` notation and Save to let linter clear -->
```
adamdehaven, line 32:
```suggestion
 *  lines switch to `v-slot:` notation and Save to let linter clear -->
```

src/components/KTreeList/KTreeList.cy.ts
kaiarrowood: Agree, but I don't know how to test drag-n-drop and that is 90% of the code. I can test some of the props and slots, but looking for suggestions on the drag-n-drop

docs/components/tree-list.md
adamdehaven: I'm still confused... if `maxLevel = 2` I would think that means parent and child (two levels)

But you're saying that allows you to have parent, child, and grandchild? (three levels)

src/components/KTreeList/KTreeList.cy.ts
adamdehaven: Yeah maybe for starters you test everything that isn't related to drag-n-drop functionality? That will probably be it's own nested `describe` block in the tests file anyway

src/components/KTreeList/KTreeDraggable.vue
kaiarrowood: It works with the current implementation so I'm not going to change this.

src/components/KTreeList/KTreeDraggable.vue
kaiarrowood: We have references to stackoverflow in `khcp-ui`. We can remove it but it's a straight copy/paste of someone else's solution, are you sure we should?

adamdehaven: Yes... there's no copyright on stack overflow solutions lol

src/components/KTreeList/KTreeList.vue
kaiarrowood: I added a console.warn for this scenario.

adamdehaven: but does the component break? e.g. if they already have too many levels, can they still do anything, or...?

docs/components/tree-list.md
kaiarrowood: Fixed as we discussed on Friday

adamdehaven (approved):
LOOKS. GREAT. TO. ME. 🚀 

Great stuff @kaiarrowood 🎉

PR #914, opened 2022-11-20
fix(kselect): bg color issue

aanchalm01:
Addresses the issue that is part of:
[https://konghq.atlassian.net/browse/KHCP-5396](https://konghq.atlassian.net/browse/KHCP-5396)
# Summary
**Before:**
image 772

**After:**
image 770

## PR Checklist

### DIFF
--- a/src/components/KSelect/KSelect.vue
+++ b/src/components/KSelect/KSelect.vue
@@ -186,7 +186,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, ref, Ref, computed, watch, PropType, nextTick } from 'vue'
+import { defineComponent, ref, Ref, computed, watch, PropType, nextTick, onMounted } from 'vue'
 import { v1 as uuidv1 } from 'uuid'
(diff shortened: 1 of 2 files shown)


src/components/KSelect/KSelect.vue
```diff
@@ -600,6 +600,16 @@ export default defineComponent({
       }
     }
 
+    // If the select is readonly with chevron, need to add the bg-color to the parent container
+    const addReadonlyStyle = () => {
+      const all = document.querySelector('.k-select-input.no-filter.select-input-container .k-input[readonly]')as HTMLInputElement
```
adamdehaven, line 605: I don’t think this is a viable solution. If there are multiple KSelect elements on the page, this code would only ever style the first one. 

Also, I don’t think we want to resort to using JavaScript to apply styles

adamdehaven:
@aanchalm01 you can likely close this one in favor of #922

aanchalm01:
This issue is fixed in this PR:
https://github.com/Kong/kongponents/pull/922

PR #918, opened 2022-11-22
fix(kalert): add additional slot [khcp-5372]

aanchalm01:
# Summary
Addresses:
[https://konghq.atlassian.net/browse/KHCP-5372](https://konghq.atlassian.net/browse/KHCP-5372)

## PR Checklist

### DIFF
--- a/docs/components/alert.md
+++ b/docs/components/alert.md
@@ -25,7 +25,7 @@ The additional message for the alert.
     :is-showing="dismissTypeIcon"
     @closed="dismissTypeIcon = false"
   >
-    <template v-slot:alertSecondaryMessage>
+    <template #alertSecondaryMessage>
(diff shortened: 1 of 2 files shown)


docs/components/alert.md
```diff
@@ -14,6 +14,38 @@
 
 The main content of the alert.
 
+### alertSecondaryMessage
```
adamdehaven, line 17: Instead of adding another prop, could we just change KAlert to support line breaks in the `alertMessage` prop?

adamdehaven: Alternatively, couldn't this be resolved by just slotting the message instead of putting the message in the prop?

aanchalm01:
This change will be handled in khcp-ui instead of kongponent.

PR #921, opened 2022-11-22
fix(kprompt): kprompt is closed when prevent-proceed-on-enter is set [KHCP-5444]

LukeSwierlik:
Task in Jira: https://konghq.atlassian.net/browse/KHCP-5444

# Summary
- **Fix** - Events on `KPropmt` have been removed, because they already exist when the component is mounted and are additionally validated.

## PR Checklist


adamdehaven (approved):
I haven't been able to test every scenario, but we can revert if we find an issue

adamdehaven (approved):

PR #937, opened 2022-11-29
feat(ksegmentedcontrol): updates [khcp-5421]

kaiarrowood:
# Summary

Adds theming and slotting support to `KSegmentedControl` in order to support new designs for analytics.
Addresses [KHCP-5421](https://konghq.atlassian.net/browse/KHCP-5421).

image

image

## PR Checklist

### DIFF
--- a/docs/components/segmented-control.md
+++ b/docs/components/segmented-control.md
@@ -185,12 +185,12 @@ export default defineComponent({
 
 | Variable | Purpose
 |:-------- |:-------
-| `--KSegmentText` | Option text color
-| `--KSegmentSelectedBackground` | Option background color when selected
-| `--KSegmentSelectedBorder` | Option border color when selected
-| `--KSegmentUnselectedBackground` | Option background color when not selected
-| `--KSegmentUnselectedBorder` | Option border color when not selected
-| `--KSegmentGap` | Gap between the options
+| `--KSegmentedControlText` | Option text color
+| `--KSegmentedControlSelectedBackground` | Option background color when selected
+| `--KSegmentedControlSelectedBorder` | Option border color when selected
+| `--KSegmentedControlUnselectedBackground` | Option background color when not selected
+| `--KSegmentedControlUnselectedBorder` | Option border color when not selected
+| `--KSegmentedControlGap` | Gap between the options
 
 An Example of changing the KSegmentedControl to a purple theme instead of blue might look like:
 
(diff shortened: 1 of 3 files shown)


docs/components/segmented-control.md
```diff
@@ -118,3 +180,58 @@ export default defineComponent({
 })
 </script>
 ```
+
+## Theming
+
+| Variable | Purpose
+|:-------- |:-------
+| `--KSegmentTextColor` | Option text color
+| `--KSegmentSelectedBackground` | Option background color when selected
```
adamdehaven, line 189: Should we allow for customizing the unselected background color as well? `--KSegmentBackground`

docs/components/segmented-control.md
```diff
@@ -118,3 +180,58 @@ export default defineComponent({
 })
 </script>
 ```
+
+## Theming
+
+| Variable | Purpose
+|:-------- |:-------
+| `--KSegmentTextColor` | Option text color
```
adamdehaven, line 188: question: since we don't use the `Color` suffix in any of the other variables, should this just be `--KSegmentText` instead?

If not, then as the inverse, I'd recommend adding a suffix of `*Color` to all the variables below.

src/components/KSegmentedControl/KSegmentedControl.vue
```diff
@@ -90,48 +95,54 @@ export default defineComponent({
...
-  border-radius: 0;
-  margin-left: -1px;
-  flex: 1;
-
-  &.primary {
-    z-index: 1;
-    border-color: var(--blue-500);
-  }
+.segmented-control {
+  column-gap: var(--KSegmentGap, 0px);
```
adamdehaven, line 99: I believe instead of `column-gap` this should just be `gap` in case the segments wrap to a new line (unsure if that's even possible)

src/components/KSegmentedControl/KSegmentedControl.vue
```diff
@@ -90,48 +95,54 @@ export default defineComponent({
...
+
+  .k-button {
```
adamdehaven, line 101: Since this is within `scoped` styles, don't you need `:deep(.k-button) {` for this selector?

docs/components/segmented-control.md
kaiarrowood: I thought about this but wasn't sure. Under the covers we are using `KButton`'s `outline` appearance which doesn't allow you to customize the background of the button. We could easily hijack a different appearance if we want to support this though.

src/components/KSegmentedControl/KSegmentedControl.vue
kaiarrowood: Added, but it looks like the styles somehow work without it which makes no sense...

docs/components/segmented-control.md
kaiarrowood: Added support for this 👍🏻

docs/components/segmented-control.md
```diff
@@ -118,3 +180,61 @@ export default defineComponent({
...
+</template>
+
+<style>
+.purple-segment {
+  --KSegmentText: var(--purple-400);
+  --KSegmentSelectedBackground: var(--white);
+  --KSegmentSelectedBorder: var(--purple-300);
+  --KSegmentUnselectedBackground: var(--blue-100);
+  --KSegmentUnselectedBorder: var(--purple-200);
+  --KSegmentGap: 12px;
```
adamdehaven, line 225: So I thought about this a bit more... the rest of our components utilize CSS variables that are the name of the component, e.g. `--KButtonPrimaryHover`

Can we update the variables used in this PR to be prefixed with `--KSegmentedControl*` ?

Everything else lgtm

src/components/KSegmentedControl/KSegmentedControl.vue
```diff
@@ -90,48 +95,64 @@ export default defineComponent({
...
-  border-radius: 0;
-  margin-left: -1px;
-  flex: 1;
-
-  &.primary {
-    z-index: 1;
-    border-color: var(--blue-500);
-  }
+.segmented-control {
+  gap: var(--KSegmentGap, 0px);
```
adamdehaven, line 99: Missed one
```suggestion
  gap: var(--KSegmentedControlGap, 0px);
```

adamdehaven (approved):
LGTM

PR #940, opened 2022-12-01
feat(kmodal): add focus trap support for better a11y

Leopoldthecoder:
# Summary

This PR traps focus inside `<KModal>` for the sake of a11y, which fixes [a vpat violation](https://docs.google.com/spreadsheets/d/1JinlVfNFFvxCB2qug3YYc-wGhg8ptpF3/edit#gid=684765038&range=51:51):

image

I checked [vuetify](https://github.com/vuetifyjs/vuetify/blob/bb9412cb3efefd4852b3220705cc41ecfb4d438f/packages/vuetify/src/components/VDialog/VDialog.ts#L216-L217) and [element](https://github.com/element-plus/element-plus/blob/33ca7ee4f06febfd9bcec33bb66e62e27f0102c1/packages/components/dialog/src/dialog.vue#L28-L36), and they both trap focus inside modals, so I think this is something we should do in kongponents.

This PR introduces two new dependencies, resulting in a 5KB~ package size increase:
image

Let me know what you think.

## PR Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,8 @@
     "axios": "^0.27.2",
     "date-fns": "^2.29.3",
     "date-fns-tz": "^1.3.7",
+    "focus-trap": "^7.1.0",
+    "focus-trap-vue": "^3.3.1",
     "popper.js": "^1.15.0",
     "swrv": "^1.0.0",
     "uuid": "^8.3.2",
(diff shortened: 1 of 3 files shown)


adamdehaven (approved):
LGTM

PR #942, opened 2022-12-01
fix(kbutton): btn-link styles [khcp-5475]

kaiarrowood:
# Summary

Remove the padding on `btn-link` appearance buttons and apply outline for `:focus-visible`.
Addresses [KHCP-5475](https://konghq.atlassian.net/browse/KHCP-5475).

## PR Checklist

### DIFF
--- a/src/components/KButton/KButton.vue
+++ b/src/components/KButton/KButton.vue
@@ -405,7 +405,7 @@ export default defineComponent({
     &:focus-visible:disabled {
       border-radius: unset;
       text-decoration: none;
-      outline: Highlight auto 1px;
+      outline: auto 1px black;
       outline: -webkit-focus-ring-color auto 1px;
       outline-offset: 3px;
       transition: none;
--- a/src/components/KButton/KButton.vue
+++ b/src/components/KButton/KButton.vue
@@ -407,7 +407,8 @@ export default defineComponent({
       text-decoration: none;
       outline: Highlight auto 1px;
       outline: -webkit-focus-ring-color auto 1px;
-      outline-offset: 2px;
+      outline-offset: 3px;
+      transition: none;
     }
 
     &:disabled,
--- a/src/components/KButton/KButton.vue
+++ b/src/components/KButton/KButton.vue
@@ -387,15 +387,29 @@ export default defineComponent({
       color: var(--grey-400) !important;
     }
   }
+
   &.btn-link {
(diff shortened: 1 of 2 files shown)


src/components/KButton/KButton.vue
```diff
@@ -387,15 +387,29 @@ export default defineComponent({
...
+      text-decoration: underline;
+    }
+
+    &:focus-visible,
+    &:focus-visible:disabled {
+      border-radius: unset;
+      text-decoration: none;
+      outline: Highlight auto 1px;
+      outline: -webkit-focus-ring-color auto 1px;
+      outline-offset: 2px;
```
adamdehaven, line 410: I think we should bump this to `3px`

src/components/KButton/KButton.vue
```diff
@@ -387,15 +387,29 @@ export default defineComponent({
...
     &:focus {
-      @include boxShadow(var(--KButtonOutlineBorder, var(--blue-500, color(blue-500))), 0, 2px);
+      text-decoration: underline;
+    }
+
+    &:focus-visible,
+    &:focus-visible:disabled {
+      border-radius: unset;
+      text-decoration: none;
+      outline: Highlight auto 1px;
```
adamdehaven, line 408: I don't think this `Highlight` property is valid and/or doing anything.

Just `auto 1px` should be sufficient I believe.

src/components/KButton/KButton.vue
```diff
@@ -387,15 +387,29 @@ export default defineComponent({
...
       text-decoration: underline;
     }
+
     &:focus {
-      @include boxShadow(var(--KButtonOutlineBorder, var(--blue-500, color(blue-500))), 0, 2px);
+      text-decoration: underline;
+    }
+
+    &:focus-visible,
+    &:focus-visible:disabled {
```
adamdehaven, line 405: You also need `transition: none` here so the outline doesn't have a weird animation look.

adamdehaven (changes requested):

src/components/KButton/KButton.vue
adamdehaven: But the order is wrong (`Highlight` has to go at the end)... however, if you put it at the end, the color then turns lighter and isn't ideal 😬

kaiarrowood: changed to default to `auto 1px black`. I think that's right?

adamdehaven (approved):
lgtm

PR #956, opened 2022-12-06
fix(ksegmentedcontrol): disallow slot pointer events by default [MA-1189]

adamdehaven:
# Summary

Adds an `allow-pointer-events` slot to `KSegmentedControl` (`false` by default) which disallows pointer events on the button slot content.

## PR Checklist


mihai-peteu (approved):
works as promised :)

PR #992, opened 2022-12-19
fix(file-upload): update style if label is absent [khcp-5654]

aanchalm01:
# Summary

## PR Checklist

### DIFF
--- a/src/components/KFileUpload/KFileUpload.vue
+++ b/src/components/KFileUpload/KFileUpload.vue
@@ -290,7 +290,7 @@ export default defineComponent({
     pointer-events: none;
 
     .label-tooltip span.kong-icon.kong-icon-help {
-      pointer-events: visible;
+      pointer-events: auto;
     }
   }
 
--- a/src/components/KFileUpload/KFileUpload.vue
+++ b/src/components/KFileUpload/KFileUpload.vue
@@ -5,7 +5,6 @@
     <KLabel
       v-if="label"
       v-bind="labelAttributes"
-      class="cursor-pointer"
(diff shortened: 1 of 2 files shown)


src/components/KFileUpload/KFileUpload.vue
```diff
@@ -285,22 +286,38 @@ export default defineComponent({
 .k-file-upload {
   position: relative;
 
+  :deep(.k-input-label) {
+    pointer-events: none;
```
adamdehaven, line 290: question: I don't think we want to block pointer-events on the label, is there a reason this is needed?

src/components/KFileUpload/KFileUpload.vue
```diff
@@ -285,22 +286,38 @@ export default defineComponent({
...
+
+    .label-tooltip span.kong-icon.kong-icon-help {
+      pointer-events: visible;
```
adamdehaven, line 293: If you keep this, instead of `visible` just set to `auto`

aanchalm01: Updated.

src/components/KFileUpload/KFileUpload.vue
adamdehaven: The label can be clickable (this is standard web behavior) I think the issue was just the `.help` container was also clickable. 

I'd suggest removing the pointer-events rules you added, and then I think everything else LGTM

aanchalm01: Updated.

adamdehaven (approved):
LGTM

PR #994, opened 2022-12-19
feat(kbadge): add borderColor prop

dustinryerson:
# Summary

Adds new `borderColor` prop to allow customizing badge border color.

I decided against adding a `--KBadgeBorderColor` variable because it seemed difficult to determine a good default without knowing more about the theme of the badge.

## Screenshots
Screen Shot 2022-12-19 at 3 49 51 PM
Screen Shot 2022-12-19 at 3 49 36 PM

## PR Checklist

### DIFF
--- a/docs/components/badge.md
+++ b/docs/components/badge.md
@@ -43,7 +43,15 @@ Use the `isBordered` prop for bordered badges. The border color matches the text
 <KBadge appearance="warning" is-bordered class="mr-2">WARNING</KBadge>
 <KBadge appearance="danger" is-bordered class="mr-2">DANGER</KBadge>
 <KBadge appearance="info" is-bordered class="mr-2">INFO</KBadge>
-<KBadge is-bordered>DEFAULT</KBadge>
+<KBadge is-bordered class="mr-2">DEFAULT</KBadge>
+<KBadge
+  background-color="var(--purple-100)"
+  border-color="var(--purple-400)"
+  color="var(--purple-400)"
+  is-bordered
+>
+  CUSTOM
+</KBadge>
 
 ```html
 <KBadge appearance="success" is-bordered>SUCCESS</KBadge>
@@ -51,6 +59,14 @@ Use the `isBordered` prop for bordered badges. The border color matches the text
 <KBadge appearance="danger" is-bordered>DANGER</KBadge>
 <KBadge appearance="info" is-bordered>INFO</KBadge>
 <KBadge is-bordered>DEFAULT</KBadge>
+<KBadge
+  background-color="var(--purple-100)"
+  border-color="var(--purple-400)"
+  color="var(--purple-400)"
+  is-bordered
+>
+  CUSTOM
+</KBadge>
 ```
 
 ### shape
@@ -68,7 +84,7 @@ The Badge has two shapes that can be changed with a `shape` property.
 <KBadge appearance="warning" shape="rectangular">Rectangular</KBadge>
 ```
 
-### color, background-color
+### color, background-color, border-color
 
 Using the `custom` appearance in conjunction with `color` and `background-color`:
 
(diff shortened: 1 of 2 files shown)


kaiarrowood:
question: Are we sure we need this? We already having theming vars for the border color of each badge type:
`--KBadgeSuccessBorder`
`--KBadgeWarningBorder`
`--KBadgeInfoBorder`
`--KBadgeDangerBorder`

adamdehaven:
> question: Are we sure we need this? We already having theming vars for the border color of each badge type: `--KBadgeSuccessBorder` `--KBadgeWarningBorder` `--KBadgeInfoBorder` `--KBadgeDangerBorder` `--KBadgeDefaultBorder`
> 
> By default the border color is mapped to the `color` of the badge if `is-bordered` is used.

I'm wondering if it makes sense to have a single CSS variable to override the border color instead of overriding each style independently. I feel like this is simpler and then each badge even with the same appearance can be overridden with a class. Thoughts?

dustinryerson:
> question: Are we sure we need this? We already having theming vars for the border color of each badge type: `--KBadgeSuccessBorder` `--KBadgeWarningBorder` `--KBadgeInfoBorder` `--KBadgeDangerBorder` `--KBadgeDefaultBorder`
> 
> By default the border color is mapped to the `color` of the badge if `is-bordered` is used.

@kaiarrowood the problem here is that I need to use a custom badge. Looks like we do have `appearance="custom"` as an option, but we have no corresponding variable, i.e. `--KBadgeCustomBorder`.

IMHO we should have some default options like you mentioned above to make it easy to implement the most common use cases, but we should also have the option to fully customize the appearance of the badge for cases that are less common but still required.

> I'm wondering if it makes sense to have a single CSS variable to override the border color instead of overriding each style independently. I feel like this is simpler and then each badge even with the same appearance can be overridden with a class. Thoughts?

@adamdehaven would this cover the use case of having multiple appearance types on the same page like we do on our route configuration form?

src/components/KBadge/KBadge.vue
```diff
@@ -156,6 +162,28 @@ const scrollWidth = ref(0)
 const truncationCalculated = ref(false)
 const isTruncated = computed(() => offsetWidth.value < scrollWidth.value)
 
+const badgeCustomStyles = computed(() => {
+  const styles = {} as {
+    backgroundColor: string
+    borderColor: string
+    color: string
```
adamdehaven, line 169: Should these be marked as optional?

dustinryerson:
For posterity, here's where we landed with this PR after some internal discussion:
- Add new`border-color` prop
- Add new `hover-color` prop
- Deprecate all CSS variables related to `color` (listed below):
  - --KBadgeSuccessColor
  - --KBadgeSuccessButtonHoverColor
  - --KBadgeSuccessBorder
  - --KBadgeSuccessBackground
  - --KBadgeWarningColor
  - --KBadgeWarningButtonHoverColor
  - --KBadgeWarningBorder
  - --KBadgeWarningBackground
  - --KBadgeInfoColor
  - --KBadgeInfoButtonHoverColor
  - --KBadgeInfoBorder
  - --KBadgeInfoBackground
  - --KBadgeDangerColor
  - --KBadgeDangerButtonHoverColor
  - --KBadgeDangerBorder
  - --KBadgeDangerBackground

dustinryerson:
@adamdehaven this is ready for another round of review

docs/components/badge.md
```diff
@@ -85,6 +103,61 @@ Using the `custom` appearance in conjunction with `color` and `background-color`
 <KBadge color="#dfe6e9" background-color="#636e72">Something</KBadge>
 <KBadge color="var(--red-500)" background-color="var(--red-300)">Long Badge 236bfb09-fe79-4cc9-99be-9361d6b1db64 aa07575b-bcd3-4bb2-bfd7-998224e3d31e 364b78fc-dba3-4b94-9134-388515496de5</KBadge>
 ```
+### border-color
```
kaiarrowood, line 106:
```suggestion
### borderColor
```

docs/components/badge.md
```diff
@@ -85,6 +103,61 @@ Using the `custom` appearance in conjunction with `color` and `background-color`
...
+  background-color="var(--purple-100)"
+  border-color="var(--purple-400)"
+  color="var(--purple-400)"
+  is-bordered
+>
+  Organization Admin
+</KBadge>
+```
+
+### hover-color
```
kaiarrowood, line 132:
```suggestion
### hoverColor
```

kaiarrowood (approved):
LGTM! 🚀

PR #1007, opened 2022-12-22
fix(code-block): remove extra z-index [khcp-5683]

aanchalm01:
# Summary
Addresses:
[https://konghq.atlassian.net/browse/KHCP-5683](https://konghq.atlassian.net/browse/KHCP-5683)

Before:
image 839

After:
image 840
## PR Checklist

### DIFF
--- a/src/components/KCodeBlock/KCodeBlock.vue
+++ b/src/components/KCodeBlock/KCodeBlock.vue
@@ -1012,6 +1012,7 @@ $dark-focusColor: var(--green-500, color(green-500));
   top: var(--spacing-xs, spacing(xs));
   right: var(--spacing-md, spacing(md));
   display: block;
+  z-index: 1;
 
   &.k-button {
     background-color: transparent;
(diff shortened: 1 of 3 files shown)


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -1011,7 +1011,6 @@ $dark-focusColor: var(--green-500, color(green-500));
   position: absolute;
   top: var(--spacing-xs, spacing(xs));
   right: var(--spacing-md, spacing(md));
-  z-index: 2;
```
adamdehaven, line 1014: I don't think this should be removed; instead the `.k-modal-fullscreen-header` likely just needs a higher `z-index`

Does adding `z-index` there cause issues?

adamdehaven (changes requested):

src/components/KCodeBlock/KCodeBlock.vue
aanchalm01: What's the reasoning behind `z-index: 2` here? I checked the stack order and the tab sequence and its same without it.

adamdehaven: TBH I'm not 100%. @kleinfreund created the component; I'm guessing it's to keep the copy button over the code.

What if we just drop it to `1` instead? I still think we should bump the z-index of the modal header

aanchalm01: Updated.

adamdehaven (approved):

PR #1027, opened 2023-01-03
feat(codeBlock): add useSingleLine prop

dustinryerson:
# Summary

- Adds a new `useSingleLine` prop, which will display the code passed into the CodeBlock component on a single line. We have a number of use cases and requirements for this in Konnect.
- Adds docs for the new prop
- Adds a test that verifies the new class is added

**Notes:**
- Line numbers will not be displayed and code will not be searchable when `useSingleLine` is enabled
- I'd thought about adding a prop to allow toggling wether or not the content on the single line gets truncated after a certain length, but decided to truncate by default for now. Happy to change this if anyone thinks truncation should be able to be toggled by default.

### Screenshots
**Docs addition**
Screenshot 2023-01-03 at 3 30 33 PM

**With copy button**
Screenshot 2023-01-03 at 3 37 34 PM

**Without copy button**
Screenshot 2023-01-03 at 3 37 51 PM

## PR Checklist

### DIFF
--- a/docs/components/codeblock.md
+++ b/docs/components/codeblock.md
@@ -217,7 +217,7 @@ Sets the display theme of the component.
 />
 ```
 
-### useSingleLine
+### isSingleLine
 
 * **Type**: `boolean`
 * **Required**: no
@@ -231,21 +231,21 @@ Line numbers will not be displayed and the code will not be searchable when this
 
 <ClientOnly>
   <KCodeBlock
-    id="code-block-use-single-line"
+    id="code-block-is-single-line"
     :code="cert"
     language="plaintext"
     theme="dark"
-    use-single-line
+    is-single-line
   />
 </ClientOnly>
 
 ```html
 <KCodeBlock
-  id="code-block-use-single-line"
+  id="code-block-is-single-line"
   :code="cert"
   language="plaintext"
   theme="dark"
-  use-single-line
+  is-single-line
 />
 ```
 
--- a/docs/components/codeblock.md
+++ b/docs/components/codeblock.md
@@ -217,6 +217,38 @@ Sets the display theme of the component.
 />
 ```
 
+### useSingleLine
+
+* **Type**: `boolean`
+* **Required**: no
+* **Default**: `false`
+
+The code passed into the component will be displayed on a single line.
+
+::: tip NOTE
+Line numbers will not be displayed and the code will not be searchable when this prop is enabled.
+:::
+
+<ClientOnly>
+  <KCodeBlock
+    id="code-block-use-single-line"
+    :code="cert"
+    language="plaintext"
+    theme="dark"
+    use-single-line
+  />
+</ClientOnly>
+
+```html
+<KCodeBlock
+  id="code-block-use-single-line"
+  :code="cert"
+  language="plaintext"
+  theme="dark"
+  use-single-line
+/>
+```
+
 ## Events
 
 ### code-block-render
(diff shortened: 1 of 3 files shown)


adamdehaven:
Some of your CSS changes appears to have changed the default copy button color
image

adamdehaven:
> Adds a new useSingleLine prop, which will display the code passed into the CodeBlock component on a single line. We have a number of use cases and requirements for this in Konnect.

@dustinryerson curious, why is this prop needed? Shouldn't the code be formatted by the consuming application format the code as desired before passing to this component? e.g. condensing down to a single line. I wouldn't think this component should do any formatting of the code input on its own.

dustinryerson:
@adamdehaven @kaiarrowood this is ready for another round of review whenever you have some time

docs/components/codeblock.md
```diff
@@ -217,6 +217,40 @@ Sets the display theme of the component.
...
+* **Required**: no
+* **Default**: `false`
+
+The code passed into the component will be displayed on a single line.
+
+::: tip NOTE
+Please keep the following in mind when using `isSingleLine`:
+- Line numbers will not be displayed
+- Code will not be searchable
+- Long code snippets will be truncated to allow for responsive design
```
adamdehaven, line 232: I'm still wondering if we should instead allow the code to overflow:auto horizontally and scroll... I'll think about this when the deploy preview builds so I can play with it 😬

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -192,41 +192,57 @@
...
-      </span><code v-html="filteredCode" /></pre>
+        <code v-html="filteredCode" />
+      </pre>
 
       <pre
         v-else
         class="k-highlighted-code-block"
+        :class="{
+          'is-single-line': isSingleLine,
+          'show-copy-button': showCopyButton
```
adamdehaven, line 220: nit/suggestion: if this class is only responsible for truncating the code, should the class be something like `truncate-single-line-code` instead so it's more clear?

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -388,8 +413,19 @@ const codeBlockCopyButton = ref<typeof KButton | null>(null)
 const numberOfMatches = ref(0)
 const matchingLineNumbers = ref<number[]>([])
 const currentLineIndex = ref<null | number>(null)
+const codeRef = toRef(props, 'code')
```
adamdehaven, line 416: Since you created a ref from the prop value, I don't think you want it to be changeable, right? 

If that's the case, I'd recommend creating `codeRef` as a `computed` variable since all you need is the value down below.
```suggestion
const codeRef = computed(() => props.code)
```

docs/components/codeblock.md
dustinryerson: Definitely a good option to consider, let me know what you think and I can make any changes

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -412,6 +448,7 @@ const filteredCode = computed(function() {
     })
     .join('\n')
 })
+const finalCode = computed(() => props.isSingleLine ? codeRef.value?.replace('\n', '') : props.code)
```
adamdehaven, line 451: I don't think you actually need the `codeRef` variable or `toRefs` up above where it's created since you're not mutating the value.

You're simply grabbing the prop value and replacing the spaces, so instead I think you can still just point to the prop value

```suggestion
const finalCode = computed(() => props.isSingleLine ? props.code.replace('\n', '') : props.code)
```

docs/components/codeblock.md
```diff
@@ -217,6 +217,40 @@ Sets the display theme of the component.
 />
 ```
 
+### isSingleLine
```
adamdehaven, line 220: nit: Can we move this prop above the `theme` prop in the docs?

dustinryerson: Fixed

src/components/KCodeBlock/KCodeBlock.vue
dustinryerson: Fixed

adamdehaven:
Can we adjust the CSS a bit when the `isSingleLine` prop is `true` to prevent the scrollbar from looking like shown below?

scrollbar

```scss
pre.is-single-line {
  padding-bottom: 0;
  padding-left: 0;

  code {
    padding-bottom: var(--spacing-xs, spacing(xs)); // Use something different if you don't think this one looks right
    padding-left: var(--spacing-sm, spacing(sm));
  }
}
```

image

adamdehaven (approved):
lgtm

PR #1043, opened 2023-01-10
fix(ksegmentedcontrol): add option validation

kaiarrowood:
# Summary

Add validation and normalize code for `options`.

## PR Checklist

### DIFF
--- a/src/components/KSegmentedControl/KSegmentedControl.vue
+++ b/src/components/KSegmentedControl/KSegmentedControl.vue
@@ -5,7 +5,7 @@
     :class="{ 'allow-pointer-events': allowPointerEvents }"
   >
     <KButton
-      v-for="option in nOptions"
+      v-for="option in normalizedOptions"
       :key="`${option.value}-option`"
       :appearance="getAppearance(option)"
       class="justify-content-center"
@@ -34,6 +34,10 @@ export interface SegmentedControlOption {
   disabled?: boolean
 }
 
+const itemsHaveRequiredProps = (items: SegmentedControlOption[]): boolean => {
+  return items.every(i => i.value !== undefined)
+}
+
 // functions used in prop validators
 const getValues = (items: SegmentedControlOption[]) => {
   const vals:string[] = []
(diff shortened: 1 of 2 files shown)


src/components/KSegmentedControl/KSegmentedControl.vue
```diff
@@ -34,6 +34,37 @@ export interface SegmentedControlOption {
   disabled?: boolean
 }
 
+// functions used in prop validators
+const getValues = (items: SegmentedControlOption[]) => {
+  const vals:string[] = []
+  items.forEach((item: SegmentedControlOption) => vals.push(item.value + ''))
```
adamdehaven, line 40: question: are you appending an empty string to ensure `item.value` is a string?

What if `item.value` is `null` or `undefined`? Do we still want to store an empty string?

src/components/KSegmentedControl/KSegmentedControl.vue
```diff
@@ -34,6 +34,37 @@ export interface SegmentedControlOption {
...
+  const uniqueValues = new Set(vals)
+
+  return vals.length === uniqueValues.size
+}
+
+const validateItems = (items: SegmentedControlOption[] | string[]): boolean => {
+  const nItems = normalizeItems(items)
+
+  return itemValuesAreUnique(nItems)
+}
```
adamdehaven, line 56: You need to move this function definition below the `normalizeItems` function so that you are not calling it before it is defined.

src/components/KSegmentedControl/KSegmentedControl.vue
kaiarrowood: `value` is required, so it should fail the validator in that scenario.

src/components/KSegmentedControl/KSegmentedControl.vue
```diff
@@ -1,24 +1,24 @@
 
 <template>
   <div
-    class="segmented-control d-flex"
-    :class="{ 'allow-pointer-events': allowPointerEvents}"
+    class="k-segmented-control segmented-control d-flex"
```
portikM, line 4: @kaiarrowood out of curiosity, why do we need both `k-segmented-control` and `segmented-control` classes here?

adamdehaven: @kaiarrowood this is a good question since this doesn't follow the `.k-*` syntax. I know it's out of scope, but would you mind updating all instances of `.segmented-control` with `.k-segmented-control`? (I think it's just here and in the tests)

portikM: Oh, good call 🤔
I don't have any objections if we leave that class there, since it doesn't hurt us. Thoughts @adamdehaven?

adamdehaven: @kaiarrowood let's delete it; all of our classes are documented as `k-*` anyway

kaiarrowood: Done and ready for final review 🚀

adamdehaven (approved):
lgtm

PR #1056, opened 2023-01-17
fix(kdatetimepicker): update styling according to mocks [KHCP-5697]

portikM:
# Summary

Ticket: https://konghq.atlassian.net/browse/KHCP-5697

**Changes:**

1. Adjust Month and Year switcher to font size 16px
2. Adjust the spacing between elements within the date-picker according to the mock
3. Adjusting the default color of the calendar section date text to use grey-600
4. Adjusting the color for the calendar date within a range to use the default state color grey-600
6. The selected state of the date picker input field should be the same as a regular input field, where there should be one outline with color blue-500
7. When the date picker input field is activated, the icon color should remain unchanged as grey-500, while the filled text color should be using the regular text input field text color black-70

Before:
Screen Shot 2023-01-18 at 2 44 55 PM

After:
Screen Shot 2023-01-18 at 2 45 08 PM

9. In time selector, once selected a time, the selector should be at default state, currently, it’s having a white background until clicked elsewhere

Before:
Screen Shot 2023-01-18 at 2 35 27 PM

After:
Screen Shot 2023-01-18 at 2 35 17 PM

## PR Checklist

### DIFF
--- a/src/components/KDateTimePicker/KDateTimePicker.vue
+++ b/src/components/KDateTimePicker/KDateTimePicker.vue
@@ -843,6 +843,7 @@ $margin: 6px;
 
       // AM / PM highlights
       .vc-am-pm {
+        background-color: color(grey-200);
         color: $text-color-darker;
         button {
           &:active,
--- a/src/components/KDateTimePicker/KDateTimePicker.vue
+++ b/src/components/KDateTimePicker/KDateTimePicker.vue
@@ -691,6 +691,7 @@ $margin: 6px;
 .k-datetime-picker {
   $highlight-color: color(blue-200);
   $selected-color: color(blue-500);
+  $text-color-lighter: color(grey-200);
   $text-color: color(grey-500);
   $text-color-darker: color(grey-600);
 
@@ -819,15 +820,15 @@ $margin: 6px;
     .vc-time-picker {
       // Time Range
       .vc-select select {
-        background-color: color(grey-200);
+        background-color: $text-color-lighter;
         color: $text-color-darker;
         &:hover {
           color: $text-color-darker;
         }
         &:focus {
           color: $text-color-darker;
-          background-color: color(grey-200);
-          border: 2px solid color(grey-200);
+          background-color: $text-color-lighter;
+          border: 2px solid $text-color-lighter;
 
           + .vc-select-arrow {
             color: color(grey-500);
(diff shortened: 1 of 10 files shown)


src/components/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -705,16 +703,23 @@ $margin: 6px;
       border: 0;
     }
 
+    // disabled day
+    .vc-day-content.is-disabled {
+      pointer-events: none;
+    }
```
adamdehaven, line 709: question: is this desired over a `not-allowed` cursor?

portikM: If you think it's better to add `cursor: not-allowed` I'm fine with that, but then I suggest that I add it to both disabled days and disabled months

src/components/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -804,33 +811,39 @@ $margin: 6px;
...
       // Time Range
       .vc-select select {
+        background-color: color(grey-200);
         color: $text-color-darker;
         &:hover {
           color: $text-color-darker;
         }
         &:focus {
           color: $text-color-darker;
-          background-color: color(gray-200);
```
mihai-peteu, line 821: whoops.  thanks @portikM :)

src/components/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -804,33 +811,39 @@ $margin: 6px;
...
+        color: $text-color-darker;
       }
       .vc-month, .vc-day {
-        color: $text-color;
+        color: $text-color-darker;
       }
 
       // AM / PM highlights
       .vc-am-pm {
+        background-color: color(grey-200);
```
mihai-peteu, line 846: Since we're using `color(grey-200)` in four instances, can you please create a `$text-color-lighter: color(grey-200);` variable above `$text-color: color(grey-500);` and use that variable throughout?

It's slightly cleaner, and helps us push back if UX wants yet another shade of gray in the future. We can just say "we have 3 to choose from, pick one" ;)

mihai-peteu (dismissed):
Thanks for handling these styling changes! Left one request.

src/components/KDateTimePicker/KDateTimePicker.vue
portikM: Good idea @mihai-peteu! Added those changes, please take a look :)

kaiarrowood (approved):
LGTM! 🚀

PR #1059, opened 2023-01-19
feat(kradio): add support for card style component [KHCP-5660]

portikM:
# Summary

Ticket: https://konghq.atlassian.net/browse/KHCP-5660

Figma mocks: https://www.figma.com/file/kY9vZ7xhA3pHwHpBxbVP9w/Kongponents-Rebuild?node-id=314%3A4052&t=fgPCwD6ipS1XoeGI-1

Default:
Screen Shot 2023-01-20 at 12 13 50 PM

Hover:
Screen Shot 2023-01-20 at 12 14 05 PM

Checked:
Screen Shot 2023-01-20 at 12 14 18 PM

Disabled:
Screen Shot 2023-01-20 at 12 14 47 PM

Checked & disabled:
Screen Shot 2023-01-20 at 12 15 05 PM

## PR Checklist


src/components/KRadio/KRadio.vue
```diff
@@ -10,106 +10,196 @@
...
+    type: [String, Number, Boolean, Object],
+    required: true,
+  },
+  type: {
+    type: String,
+    default: 'default',
+    validator: (value: string): boolean => {
+      return ['default', 'card'].includes(value)
     },
   },
```
portikM, line 87: **Note for PR review:** besides this prop, all changes in `<script>` tag are due to switch to setup syntax.

docs/components/radio.md
```diff
@@ -68,9 +72,59 @@ Will place description text under the radio label (required). Can also be [slott
 
 <KRadio :selected-value="true" v-model="radio" label="Label Example" description="Some subheader text" />
 
-### selectedValue
+### type
 
-The value of the `KRadio` option that will be emitted by the `change` and `update:modelValue` events.
+Controlls apperance of radio input element. Accpted values: `default`(default) and `card`.
```
adamdehaven, line 77: nit: typo
```suggestion
Controls apperance of radio input element. Accepted values: `default`(default) and `card`.
```

docs/components/radio.md
```diff
@@ -68,9 +72,59 @@ Will place description text under the radio label (required). Can also be [slott
...
+  <KRadio type="card" label="Foo" description="This subheader" selected-value="foo" v-model="cardRadio">
+    <template v-slot:before>
+      🦍
+    </template>
+  </KRadio>
+  <KRadio type="card" label="Bar" description="That subheader" selected-value="barv-model="cardRadio">
+    <template v-slot:before>
+      🦍
+    </template>
+  </KRadio>
```
adamdehaven, line 108: Unsure why this second radio is showing the icon upside-down?

image

docs/components/radio.md
```diff
@@ -106,6 +160,36 @@ Any valid attribute will be added to the input. You can read more about `$attrs`
 </KRadio>
 ```
 
+-  `before` - Available only when `type` prop is `card`. Will place content above the title and description (if provided).
+-  `after` - Available only when `type` prop is `card`. Will place content below the title and description (if provided).
```
adamdehaven, line 164: suggestion: Can we rename these slots? You'll need to update the usage as well. I think it makes sense to name the slots accordingly since they only correspond to the specific `type`

`cardTop` and `cardBottom`

src/components/KRadio/KRadio.vue
```diff
@@ -10,106 +10,196 @@
       type="radio"
       @click="handleClick"
     >
-    <span
+    <div
+      v-if="type === 'card' && $slots.before"
```
adamdehaven, line 14: suggestion: Instead of checking for `type === 'card` or `type === 'default` in so many places, can you map this to a computed variable and do something like this instead? Then you can test the computed in the `template` section.

```ts
const isDefaultType = computed((): boolean => props.type === 'default`)
```

src/components/KRadio/KRadio.vue
```diff
@@ -10,106 +10,196 @@
...
+      <template v-if="type === 'default'">
+        <span class="k-radio-label">
+          <slot>{{ label }}</slot>
+        </span>
+      </template>
+      <template v-if="type === 'card'">
+        <span class="k-radio-label">
+          {{ label }}
+        </span>
+      </template>
```
adamdehaven, line 31: question: Why does the HTML here need to be different based on the `type` property? Can't they utilize the same structure either way?

src/components/KRadio/KRadio.vue
```diff
@@ -10,106 +10,196 @@
...
+})
 
-    const isSelected = computed((): boolean => props.selectedValue === props.modelValue)
+const slots = useSlots()
 
-    const handleClick = (): void => {
-      emit('change', props.selectedValue)
-      emit('update:modelValue', props.selectedValue)
-    }
+const hasLabel = computed((): boolean => !!(props.label || slots.default))
```
adamdehaven, line 92: Since you're mapping the hasLabel slot to a computed, can you do the same for the new slots?

adamdehaven (dismissed):

src/components/KRadio/KRadio.vue
```diff
@@ -10,106 +10,196 @@
...
+  }
+
+  .k-radio-content-after {
+    margin-top:  var(--spacing-sm);
+  }
+
+  &:has(.k-input:disabled) {
+    background-color: $background-color-card-disabled;
+    cursor: not-allowed;
+    opacity: 0.6;
```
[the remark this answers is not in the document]
portikM: Yeah shouldn't be a problem even if the bug is there - since this is for `disabled` state only.
Regarding the `pointer-events: none;` - it isn't needed since it's handled by the input element when `disabled` attribute is passed to it.

src/components/KRadio/KRadio.vue
portikM: Yeah you're right, I think tried to achieve something else but forgot to change it back

docs/components/radio.md
portikM: How about `header` and `footer` @adamdehaven?

adamdehaven: I saw your comment earlier about how you can’t use the single slot because of the `description` prop but I’m still not sure I understand how that prevents the single slot solution?

If the `type` is `card` it could essentially just “override” all of the other slots and use the single-slot solution you were originally going with along with the new classes. Would this work?

portikM: You know what - I haven't tried that tbh. I was trying to hit two birds with one stone and make the component flexible so that in both default radio input and card they could choose between using props or slots for label - but that approach adds too many complications. "Overriding" the default slot when `type` is `card` like you suggested is gonna be much cleaner solution - only potential downside is that they will only be able to use default slot, but not the `label` or `description` props, is that okay?

docs/components/radio.md
```diff
@@ -68,9 +72,59 @@ Will place description text under the radio label (required). Can also be [slott
...
 
-The value of the `KRadio` option that will be emitted by the `change` and `update:modelValue` events.
+Controlls apperance of radio input element. Accpted values: `default`(default) and `card`.
+
+<KCard>
+  <template v-slot:body>
+    <div class="d-flex">
+      <KRadio type="card" label="Foo" description="This subheader" selected-value="foo" v-model="cardRadio">
+        <template v-slot:before>
+          <img src="/img/kong-logomark.png" alt="Kong logo" />
```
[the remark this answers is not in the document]
portikM: I used logo image so that it batches the Figma design as closely as possible. Is there Kong logo `KIcon`?

docs/components/radio.md
adamdehaven: Yes I think that's fine as long as you document that only the default slot is used in the description of the `type` prop when it is set to `card`

docs/components/radio.md
```diff
@@ -68,9 +72,59 @@ Will place description text under the radio label (required). Can also be [slott
...
+        </template>
+      </KRadio>
+      <KRadio type="card" label="Bar" description="That subheader" selected-value="bar" v-model="cardRadio">
+        <template v-slot:before>
+          <img src="/img/kong-logomark.png" style="transform: rotate(180deg);" alt="Kong logo" />
```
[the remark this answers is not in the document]
portikM: See above

docs/components/radio.md
```diff
@@ -68,9 +72,59 @@ Will place description text under the radio label (required). Can also be [slott
...
+  </KRadio>
+  <KRadio type="card" label="Bar" description="That subheader" selected-value="barv-model="cardRadio">
+    <template v-slot:before>
+      🦍
+    </template>
+  </KRadio>
+  <div class="mt-3">Selected: {{ cardRadio }}</div>
+</template>
+
+<script lang="ts">
```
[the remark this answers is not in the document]
portikM: Will skip this for now, but will include this in another ticket related to documentation

src/components/KRadio/KRadio.vue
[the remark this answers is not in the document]
portikM: We can't handle both `label` prop and slot because label class should either be outside or inside the slot, it can't be both

src/components/KRadio/KRadio.vue
```diff
@@ -1,7 +1,8 @@
 <template>
   <label
+    :checked="isSelected"
     class="k-radio"
-    :class="$attrs.class"
+    :class="`${isTypeDefault ? 'k-radio-default' : `k-radio-${type}`} ${$attrs.class ? $attrs.class : ''}`"
```
adamdehaven, line 5: suggestion: I think this logic can be simplified. 

Untested, so please make sure this looks correct
```suggestion
    :class="[isTypeDefault ? 'k-radio-default' : `k-radio-${type}`, ...$attrs.class]"
```

src/components/KRadio/KRadio.vue
```diff
@@ -11,105 +12,219 @@
...
-.k-radio-label {
-  font-size: var(--type-sm, type(sm));
+$text-color-default: var(--black-45, rgba(0, 0, 0, 0.45));
+$text-color-card: color(black-500);
+$border-color-card: color(grey-300);
+$background-color-card-checked: color(blue-100);
+$border-color-card-checked: color(blue-300);
+$background-color-card-disabled: color(grey-200);
+
+@mixin disabled {
```
adamdehaven, line 122: suggestion: we have some other mixins. Can we update the name of this local mixin to something more specific, like `@mixin kRadioDisabled` or something similar?

The same comment applies to the other local mixins (e.g. `checked`)

docs/components/radio.md
```diff
@@ -60,17 +64,76 @@ Will place label text to the right of the radio. Can also be [slotted](#slots).
...
 ```
 
 <KRadio :selected-value="true" v-model="radio" label="Label Example" description="Some subheader text" />
 
-### selectedValue
+### type
 
-The value of the `KRadio` option that will be emitted by the `change` and `update:modelValue` events.
+Controls appearance of radio input element. Accepted values:
+- `card`
```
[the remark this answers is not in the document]
portikM: I see your point, totally valid concern but I thought this approach might leave some room for adding more types of card in the future - maybe a color selector or different style of regular radio input.

Screen Shot 2023-01-25 at 9 12 03 AM
Screen Shot 2023-01-25 at 9 12 29 AM

docs/components/radio.md
```diff
@@ -60,17 +64,76 @@ Will place label text to the right of the radio. Can also be [slotted](#slots).
...
+  </KRadio>
+  <KRadio type="card" selected-value="bar" v-model="cardRadio">
+    <img class="mb-2" src="/img/kong-logo.png" alt="Kong logo" />
+    <div class="k-radio-label">Bar</div>
+    <div class="k-radio-description">That subheader</div>
+  </KRadio>
+  <div class="mt-3">Selected: {{ cardRadio }}</div>
+</template>
+
+<script lang="ts">
```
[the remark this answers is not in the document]
portikM: I feel like it might be weird to have only example in vue 3 syntax while all other examples on this page in vue 2. Will add this to ticket about updating examples in docs.

docs/components/radio.md
```diff
@@ -81,28 +144,53 @@ Any valid attribute will be added to the input. You can read more about `$attrs`
...
-    <KRadio v-model="radioState" disabled>Disabled radio</KRadio>
+  <template #body>
+    <KRadio v-model="radioState" :selected-value="true" disabled>Disabled radio</KRadio>
   </template>
 </KCard>
 
 ## Slots
 
-- `default` - Anything passed in to the default slot will replace the label prop text
+- `default` - Anything passed in to the default slot will replace the `label` prop text.
```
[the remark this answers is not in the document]
portikM: I feel like one example is enough - there isn't too many different configurations in the first place

src/components/KRadio/KRadio.vue
portikM: Sure, done

src/components/KRadio/KRadio.vue
portikM: Iterator won't work, but making the value an array rather than string does make it a little bit cleaner, thanks

docs/components/radio.md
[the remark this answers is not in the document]
portikM: Hmm yeah I guess I agree - pushed changes.
Fyi @kaiarrowood - I liked your idea of keeping `''` for default, but I agree with Adam that it's better to have more descriptive values in case we'll have more types in future.

docs/components/radio.md
```diff
@@ -24,21 +24,21 @@
...
-      objB: { name: 'b' },
-      radio: true,
-    })
-
-    return {
-      ...toRefs(data),
-    }
-  }
-})
+  import { defineComponent, reactive, toRefs } from "vue";
```
adamdehaven, line 27: I think you may have an issue with your ESLint settings -- these shouldn't be indented

docs/components/radio.md
```diff
@@ -131,14 +230,14 @@ An Example of changing the background color of KRadio to lime might look like:
...
-    <KRadio v-model="selected" :selected-value="true" />
+    <KRadio v-model="isStateOn" :selected-value="true" />
   </div>
 </template>
 
 <style>
-.KRadio-wrapper {
-  --KRadioPrimary: lime;
-}
+  .KRadio-wrapper {
```
adamdehaven, line 238: Same comment here on indention

src/components/KRadio/KRadio.vue
```diff
@@ -1,7 +1,8 @@
 <template>
   <label
+    :checked="isSelected"
     class="k-radio"
-    :class="$attrs.class"
+    :class="[isTypeDefault ? 'k-radio-default' : `k-radio-${type}`, $attrs.class ? $attrs.class : '']"
```
adamdehaven, line 5: question: can't you just append `$attrs.class` regardless of whether it is empty?

src/components/KRadio/KRadio.vue
```diff
@@ -11,105 +12,219 @@
...
     </span>
     <div
-      v-if="hasLabel && (description || $slots.description)"
+      v-if="isTypeDefault && hasLabel && hasDescription"
       class="k-radio-description"
     >
       <slot name="description">{{ description }}</slot>
     </div>
+    <div v-if="!isTypeDefault && hasLabel">
+      <slot />
```
adamdehaven, line 27: nit: can you add the default name here so it's obvious?
```suggestion
      <slot name="default" />
```

docs/components/radio.md
portikM: Oops, stupid Prettier

src/components/KRadio/KRadio.vue
portikM: If it's empty it was rendering as `class="undefined"` which isn't pretty, that's why I added this check

src/components/KRadio/KRadio.vue
portikM: Sure, done

adamdehaven (approved):
LGTM 

@portikM I'm not sure if all Kai's comments are resolved, but if yes, I think you're good; nice job 🎉

docs/components/radio.md
```diff
@@ -24,21 +24,21 @@
...
-import { defineComponent, reactive, toRefs } from 'vue'
+import { defineComponent, reactive, toRefs } from "vue";
 
 export default defineComponent({
   setup() {
     const data = reactive({
-      objA: { name: 'a' },
-      objB: { name: 'b' },
+      objA: { name: "a" },
+      objB: { name: "b" },
```
adamdehaven, line 33:
```suggestion
      objA: { name: 'a' },
      objB: { name: 'b' },
```

adamdehaven (approved):

docs/components/radio.md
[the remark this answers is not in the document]
kaiarrowood: Yes, but this solution is fine.

image

kaiarrowood (approved):
Awesome job! 🚀

PR #1066, opened 2023-01-23
fix(kpop): stop document click event propagation only when event is about to hide popper

tkajtoch:
# Summary

This fixes the issue where other document click event listeners were not being fired because the stopPropagation() was always being called no matter what was the actual action of the click.

Most UI libraries use document-bound click listeners for execution optimization reasons, and this behavior caused them to not work anymore.

stopPropagation() should only be called when the click event is hiding the popper.

## PR Checklist


adamdehaven (approved):
lgtm

PR #1067, opened 2023-01-23
feat(ktable): add styling for multi-row layout [KHCP-5819]

portikM:
# Summary

https://konghq.atlassian.net/browse/KHCP-5819

## PR Checklist


docs/components/select.md
```diff
@@ -499,7 +499,9 @@ You can pass any input attribute and it will get properly bound to the element.
 ### Item Template
 You can use the `item-template` slot to customize the look and feel of your items. Use slots to gain access to the `item` data.
 
+::: tip NOTE
```
adamdehaven, line 502:
```suggestion
::: tip TIP
```

docs/components/table.md
```diff
@@ -1028,6 +1028,116 @@ export default {
 </script>
 ```
 
+Here is another example of how to use cell slot to completely customize content inside the cell.
+
+::: tip NOTE
```
adamdehaven, line 1033:
```suggestion
::: tip TIP
```

docs/components/table.md
```diff
@@ -1028,6 +1028,116 @@ export default {
...
+    }
+  },
+  methods: {
+    fetcher() {
+      return {
+        data: [
+          {
+            name: 'Basic Auth',
+            img: {
+              src: '🦍',
```
adamdehaven, line 1093: I don't understand the gorilla emoji in the examples? We're rendering a KIcon in the table..?

docs/components/table.md
```diff
@@ -1028,6 +1028,116 @@ export default {
...
+
+<script>
+export default {
+  data() {
+    return {
+      headers: [
+        { label: 'Name', key: 'name', sortable: true },
+        { label: 'ID', key: 'id' },
+        { label: 'Enabled', key: 'enabled' },
+        { key: 'actions', hideLabel: true },
```
adamdehaven, line 1082: I think you need to remove this actions column since you're not actually utilizing it

docs/components/table.md
```diff
@@ -1449,6 +1559,12 @@ export default defineComponent({
         { label: 'Version', key: 'version', sortable: false },
         { label: 'Connected', key: 'connected', sortable: false },
         { label: 'Last Seen', key: 'last_seen', sortable: false }
+      ],
+      cellSlotHeaders: [
+        { label: 'Name', key: 'name', sortable: true },
+        { label: 'ID', key: 'id' },
+        { label: 'Enabled', key: 'enabled' },
+        { key: 'actions', hideLabel: true }
```
adamdehaven, line 1567: Again, remove actions if they aren't being used

docs/components/table.md
```diff
@@ -1028,6 +1028,116 @@ export default {
...
+    :enableClientSort="true"
+    :fetcher="fetcher"
+    :headers="headers"
+    hidePaginationWhenOptional
+  >
+    <template #name="{row}">
+      <img :alt="row.img.alt" class="mr-2" :src="row.img.src">
+      <div class="d-flex flex-column">
+        <span class="k-table-cell-title">{{ row.name }}</span>
+        <span class="k-table-cell-description">{{ row.description }}</span>
```
adamdehaven, line 1068: In your example, can these be `div` instead of `span` since that will be the typical usage?
```suggestion
        <div class="k-table-cell-title">{{ row.name }}</div>
        <div class="k-table-cell-description">{{ row.description }}</div>
```

docs/components/table.md
```diff
@@ -1028,6 +1028,116 @@ export default {
...
+    :fetcher="cellSlotFetcher"
+    :cellAttrs="cellSlotAttrs"
+    :enableClientSort="true"
+    hidePaginationWhenOptional
+    >
+    <template v-slot:name="{row}">
+      <img class="mr-2" src="/img/kong-logomark.png" :alt="row.img.alt">
+      <div class="d-flex flex-column">
+        <span class="k-table-cell-title">{{row.name}}</span>
+        <span class="k-table-cell-description truncate">{{row.description}}</span>
```
adamdehaven, line 1049: In your example, can these be `div` instead of `span` since that will be the typical usage?
```suggestion
        <div class="k-table-cell-title">{{row.name}}</div>
        <div class="k-table-cell-description truncate">{{row.description}}</div>
```

src/components/KTable/KTable.vue
```diff
@@ -1063,6 +1063,16 @@ export default defineComponent({
         margin-top: calc(-1 * var(--KButtonPaddingY, var(--spacing-xs)));
         margin-bottom: calc(-1 * var(--KButtonPaddingY, var(--spacing-xs)));
       }
+      .k-table-cell-title {
+        color: color(grey-600);
+        font-weight: 600;
+        font-size: var(--type-md)
+      }
+      .k-table-cell-description {
+        color: color(grey-600);
```
adamdehaven, line 1072: issue: You have both the title and description using the same color; however, in the Figma, the description text is a lighter color `grey-500`

adamdehaven (changes requested):

docs/components/table.md
portikM: It's just here as a placeholder. If you think it's better that I replace it with something looking more like actual value I can do that

src/components/KTable/KTable.vue
portikM: Good catch, thanks @adamdehaven! Addressed this and other comments

docs/components/table.md
portikM: In the table we are rendering an `` tag so I used an emoji here instead of using a dummy value for `src`

adamdehaven: Typically we make the examples pretty close to what we're actually doing. So if you switch the rendered example to put the emoji in the table, then that works (Instead of the icon)

portikM: Okay, let me update that

docs/components/table.md
```diff
@@ -1028,6 +1028,115 @@ export default {
 </script>
 ```
 
+Here is another example of how to use cell slot to completely customize content inside the cell.
```
adamdehaven, line 1031:
```suggestion
Here is another example of how to use slot to completely customize content inside the cell.
```

docs/components/table.md
```diff
@@ -1028,6 +1028,115 @@ export default {
 </script>
 ```
 
+Here is another example of how to use slot to completely customize content inside the cell.
```
adamdehaven, line 1031: One more suggestion: can we add a subheader here?
```suggestion
#### Built-in cell styles

Here is another example of how to use slot to completely customize content inside the cell.
```

adamdehaven (approved):
lgtm

PR #1071, opened 2023-01-24
feat(KCollapse): add support to display tags content [KHCP-5895]

lukewhchen:
# Summary
Ticket: https://konghq.atlassian.net/browse/KHCP-5895

Screenshot 2023-01-24 at 11 27 01 AM
Screenshot 2023-01-24 at 11 27 19 AM
Screenshot 2023-01-24 at 11 27 50 AM

## PR Checklist

### DIFF
--- a/docs/components/collapse.md
+++ b/docs/components/collapse.md
@@ -143,6 +143,128 @@ To set the default state (collapsed or expanded) without binding to v-model you 
   I am expanded by default
 </KCollapse>
 ```
+### type
+
+Values: `tags` or `default`. Use this optional prop to specific render the tags(`KBadge`) content.  The trigger icon will automatically appear when the content width is longer than the parent container. If type is setted to tags, only tags related props: `tags`, `tagsAppearance`, and `noCollapse` will take effect.
+
+<KCard title="Tags">
+  <template v-slot:body>
+    <KCollapse
+      type="tags"
+      :tags="complexTags"
+      >
+    </KCollapse>
+  </template>
+</KCard>
+
+```html
+<KCollapse
+  type="tags"
+  :tags="['tag-001', 'tag-002', 'tag-003', 
+  'tag-004', 'tag-005','tag-006', 
+  'BasicAuth', 'WebsiteDesktop', 'AndroidApp', 
+  'iOSApp', 'LinuxVirtualMachine']"
+  >
+</KCollapse>
+```
+
+### tags
+
+Use this prop to provide the tags content of an array.
+
+<KCard title="Tags">
+  <template v-slot:body>
+    <KCollapse
+      type="tags"
+      :tags="['tag-001', 'tag-002', 'tag-003']"
+      >
+    </KCollapse>
+  </template>
+</KCard>
+
+```html
+<KCollapse
+  type="tags"
+  :tags="['tag-001', 'tag-002', 'tag-003']"
+  >
+</KCollapse>
+```
+
+### tagsAppearance
+
+Values: `light` or `default`. Use this optional prop to set the display appearance of the tags.
+
+<KCard title="Tags">
+  <template v-slot:body>
+    <KCollapse
+      type="tags"
+      :tags="complexTags"
+      tags-appearance="light"
+      >
+    </KCollapse>
+  </template>
+</KCard>
+
+```html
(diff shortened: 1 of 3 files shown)


docs/components/collapse.md
```diff
@@ -143,6 +143,128 @@ To set the default state (collapsed or expanded) without binding to v-model you
   I am expanded by default
 </KCollapse>
 ```
+### type
+
+Values: `tags` or `default`. Use this optional prop to specific render the tags(`KBadge`) content.  The trigger icon will automatically appear when the content width is longer than the parent container. If type is setted to tags, only tags related props: `tags`, `tagsAppearance`, and `noCollapse` will take effect.
```
adamdehaven, line 148: question: I don't understand what this means:
```
Values: `tags` or `default`
```

What is default? What are tags, an array?

docs/components/collapse.md
```diff
@@ -143,6 +143,128 @@ To set the default state (collapsed or expanded) without binding to v-model you
...
+<KTable :fetcher="tableOptionsFetcher" :headers="tableOptionsHeaders" >
+  <template v-slot:tags="{rowValue}">
+    <KCollapse
+      type="tags"
+      :tags="rowValue"
+      >
+    </KCollapse>
+  </template>
+</KTable>
+```
```
adamdehaven, line 265: We don't typically have "Examples" sections, but rather the examples are rendered inline with the props they correspond to.

src/components/KCollapse/KCollapse.vue
```diff
@@ -1,6 +1,14 @@
 <template>
   <div class="k-collapse w-100">
+    <div v-if="type === 'tags'">
+      <KCollapseItem
+        :no-collapse="noCollapse"
+        :tags="tags"
+        :tags-appearance="tagsAppearance"
+      />
+    </div>
```
adamdehaven, line 9: The `v-if` should be able to go directly on the `<KCollapseItem>` component without the need for the wrapper `div`

src/components/KCollapse/KCollapse.vue
```diff
@@ -111,6 +121,25 @@ export default defineComponent({
...
+    tagsAppearance: {
+      type: String as PropType<'default' | 'light'>,
+      required: false,
+      default: 'default',
+    },
+    noCollapse: {
+      type: Boolean,
+      required: false,
+      default: false,
+    },
```
adamdehaven, line 142: The point of this component is to collapse; I don't think it makes sense to allow disabling.

src/components/KCollapse/KCollapse.vue
```diff
@@ -111,6 +121,25 @@ export default defineComponent({
...
+    tags: {
+      type: Array,
+      default: () => [],
+    },
+    tagsAppearance: {
+      type: String as PropType<'default' | 'light'>,
+      required: false,
+      default: 'default',
+    },
```
adamdehaven, line 137: The tag appearance should be integrated into the `tags` prop if it's going to be a thing; there's no need for a separate prop. `tags` can be an array of objects, similar to:
```ts
[
  {
    appearance: '',
    text: '',
  }
]
```

src/components/KCollapse/KCollapseItem.vue
```diff
@@ -0,0 +1,176 @@
+<template>
```
adamdehaven, line 1: issue: Again, this new child component is adding a lot of complexity; I think this should be handled by providing generic slots whereby the host app can inject whatever content is needed in the visible/hidden slots instead of such a rigid layout.

adamdehaven (changes requested):
I left some general comments; however, @jillztom I think we need to rethink the implementation here as this Kongponent has become extremely opinionated on the layout of the visible/hidden content, and what is allowed.

@lukewhchen really appreciate you putting up the PR, but I think we need to rethink this component entirely so that these type of changes aren't necessary.

If we think ahead, what happens if someone wants to show something besides a Badge (or "Tag")? Are we going to keep extending?

lukewhchen:
After talked with @jillztom and @kaiarrowood, we decided to close this PR. At the same time, we will create a new Kongponent (https://konghq.atlassian.net/browse/KHCP-5895) to better handle this issue.

PR #1073, opened 2023-01-26
feat(*): add external link component [KHCP-5745]

portikM:
# Summary

Ticket: https://konghq.atlassian.net/browse/KHCP-5745

## PR Checklist

### DIFF
--- a/docs/components/external-link.md
+++ b/docs/components/external-link.md
@@ -48,12 +48,12 @@ If true, icon won't be rendered.
 
 <KCard>
     <template v-slot:body>
-      <KExternalLink href="https://kongponents.konghq.com/" hideIcon>Kongponents</KExternalLink>
+      <KExternalLink hide-icon href="https://kongponents.konghq.com/">Kongponents</KExternalLink>
     </template>
 </KCard>
 
 ```html
-<KExternalLink hideIcon href="https://kongponents.konghq.com/">
+<KExternalLink hide-icon href="https://kongponents.konghq.com/">
   Kongponents
 </KExternalLink>
 ```
@@ -64,12 +64,12 @@ If true, icon won't be rendered.
 
 ## Theming
 
-| Variable                    | Purpose                                     |
-| :-------------------------- | :------------------------------------------ |
-| `--KExternalLinkColor`      | KExternalLinkColor text color               |
-| `--KExternalLinkColorHover` | KExternalLinkColor text color on mouse over |
+| Variable                    | Purpose                           |
+| :-------------------------- | :-------------------------------- |
+| `--KExternalLinkColor`      | Text and icon color               |
+| `--KExternalLinkColorHover` | Text and icon color on mouse over |
 
-An Example of changing the text color of KExternalLink to gold might look
+An Example of changing the text color of KExternalLink to red might look
 like:
 
 <KCard>
@@ -91,15 +91,15 @@ like:
 
 <style lang="scss">
 .KExternalLink-wrapper {
-  --KExternalLinkColor: gold;
-  --KExternalLinkColorHover: yellow;
+  --KExternalLinkColor: red;
+  --KExternalLinkColorHover: brown;
 }
 </style>
 ```
 
 <style lang="scss">
 .KExternalLink-wrapper {
-  --KExternalLinkColor: gold;
-  --KExternalLinkColorHover: yellow;
+  --KExternalLinkColor: red;
+  --KExternalLinkColorHover: brown;
 }
 </style>
--- a/docs/components/external-link.md
+++ b/docs/components/external-link.md
@@ -61,3 +61,45 @@ If true, icon won't be rendered.
 ## Slots
 
 - `default` - link text
+
+## Theming
+
+| Variable                    | Purpose                                     |
+| :-------------------------- | :------------------------------------------ |
+| `--KExternalLinkColor`      | KExternalLinkColor text color               |
+| `--KExternalLinkColorHover` | KExternalLinkColor text color on mouse over |
+
+An Example of changing the text color of KExternalLink to gold might look
+like:
+
+<KCard>
+  <template v-slot:body>
+    <div class="KExternalLink-wrapper">
+      <KExternalLink href="https://kongponents.konghq.com/">Kongponents</KExternalLink>
+    </div>
+  </template>
+</KCard>
+
+```html
+<template>
+  <div class="KExternalLink-wrapper">
+    <KExternalLink href="https://kongponents.konghq.com/">
+      Kongponents
+    </KExternalLink>
+  </div>
+</template>
+
+<style lang="scss">
+.KExternalLink-wrapper {
+  --KExternalLinkColor: gold;
+  --KExternalLinkColorHover: yellow;
+}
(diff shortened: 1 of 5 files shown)


src/components/KExternalLink/KExternalLink.vue
```diff
@@ -0,0 +1,57 @@
...
+</template>
+
+<script setup lang="ts">
+defineProps({
+  href: {
+    type: String,
+    required: true,
+    validator: function(value:string) {
+      const urlPattern = new RegExp('^(https?:\\/\\/)?' + // protocol
+    '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' + // domain name
```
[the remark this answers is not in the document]
portikM: This regex is carried over from: https://github.com/Kong/public-ui-components/blob/main/packages/core/app-layout/src/components/helpers/isValidUrl.ts

src/components/KExternalLink/KExternalLink.vue
```diff
@@ -0,0 +1,57 @@
...
+    '((\\d{1,3}\\.){3}\\d{1,3}))' + // OR ip (v4) address
+    '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port and path
+    '(\\?[;&a-z\\d%_.~+=-]*)?' + // query string
+    '(\\#[-a-z\\d_]*)?$', 'i') // fragment locator
+
+      return !!urlPattern.test(value)
+    },
```
adamdehaven, line 33: suggestion: Instead of using this RegEx (unsure where it came from originally) we could instead try constructing a URL with a try/catch within the validator. Something like this:
```suggestion
    validator: (url: string): boolean => {
      try {
        const linkUrl = new URL(url)
        
        // No error was thrown, so return true (we could check some of the URL pieces here, but I'm not sure it's necessary)        
        return true
      } catch (err) {
        // Invalid URL
        return false
      }
    },
```

docs/components/external-link.md
```diff
@@ -0,0 +1,47 @@
+# External Link
+
+**KExternalLink** - a link with an icon
```
adamdehaven, line 3:
```suggestion
**KExternalLink** - a component to render an anchor tag (that opens in a new window) with an icon
```

src/components/KExternalLink/KExternalLink.vue
```diff
@@ -0,0 +1,57 @@
...
+})
+</script>
+
+<style lang="scss" scoped>
+@import '@/styles/variables';
+@import '@/styles/functions';
+
+.k-external-link {
+  align-items: center;
+  color: color(blue-500);
```
adamdehaven, line 48: suggestion: we may want to allow users to customize the text and icon colors

You should be able to do something similar to this (you'll need to document the CSS variables in a "Theming" section on the docs page; see existing examples).
```suggestion
  color: var(--KExternalLinkColor, color(blue-500));
```

adamdehaven (dismissed):
image

issue: the icon should always appear directly next to the link, meaning the entire component should likely have `display: inline-flex`. With this CSS change, you'll likely need some default margin between the text and the icon.

src/components/KExternalLink/KExternalLink.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <a
+    v-if="href"
```
adamdehaven, line 3: question: this logic will still render the content in the DOM if a user provides a string for `href` even if it's not valid; should we map the `v-if` to a function that checks for the validity of the href so that if an invalid string is returned, nothing renders?

If yes, you can utilize the same function in the validator that I suggested as long as you set it to a `const` before defining the props.

portikM: Good idea!

src/components/KExternalLink/KExternalLink.vue
```diff
@@ -0,0 +1,77 @@
...
+      size="12"
+    />
+  </a>
+</template>
+
+<script lang="ts">
+const isValidUrl = (url: string): boolean => {
+  try {
+    // eslint-disable-next-line no-new
+    new URL(url)
```
adamdehaven, line 25: You shouldn't need the eslint comment if you set it to a variable

src/components/KExternalLink/KExternalLink.vue
```diff
@@ -0,0 +1,77 @@
...
+
+<script setup lang="ts">
+// eslint-disable-next-line import/first
+import { computed, ref } from 'vue'
+
+const props = defineProps({
+  href: {
+    type: String,
+    required: true,
+    validator: (value: string) => { return isValidUrl(value) },
```
adamdehaven, line 44:
```suggestion
    validator: (value: string): boolean => !!isValidUrl(value),
```

adamdehaven (commented):
A couple small suggestions

src/components/KExternalLink/KExternalLink.vue
portikM: Then I get the unused variable error lol. Either way need eslint comment

docs/components/external-link.md
```diff
@@ -0,0 +1,105 @@
...
+If true, icon won't be rendered.
+
+<KCard>
+    <template v-slot:body>
+      <KExternalLink href="https://kongponents.konghq.com/" hideIcon>Kongponents</KExternalLink>
+    </template>
+</KCard>
+
+```html
+<KExternalLink hideIcon href="https://kongponents.konghq.com/">
```
adamdehaven, line 56: nit: prefer kebab-case props in the examples
```suggestion
<KExternalLink hide-icon href="https://kongponents.konghq.com/">
```

docs/components/external-link.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Slots
+
+- `default` - link text
+
+## Theming
+
+| Variable                    | Purpose                                     |
+| :-------------------------- | :------------------------------------------ |
+| `--KExternalLinkColor`      | KExternalLinkColor text color               |
```
adamdehaven, line 69:
```suggestion
| `--KExternalLinkColor`      | text and icon color               |
```

docs/components/external-link.md
```diff
@@ -0,0 +1,105 @@
...
+| `--KExternalLinkColorHover` | KExternalLinkColor text color on mouse over |
```
adamdehaven, line 70:
```suggestion
| `--KExternalLinkColorHover` | text and icon color on mouse over |
```

adamdehaven (commented):
just a couple super-small change requests (sorry)

One other request: the gold theming example is a little hard to see against the white background. Can you update to a color that’s a little easier on the eyes?

BDA75CC8-0244-4FD3-9DD0-9F44708C51B2

portikM:
> just a couple super-small change requests (sorry)
> 
> One other request: the gold theming example is a little hard to see against the white background. Can you update to a color that’s a little easier on the eyes?
> 
> BDA75CC8-0244-4FD3-9DD0-9F44708C51B2

Done @adamdehaven  😎

adamdehaven (approved):
lgtm

PR #1076, opened 2023-01-27
chore(stylelint): add stylelint-order

portikM:
# Summary

- added stylelint-order plugin (https://github.com/hudochenkov/stylelint-order)
- added stylelint command in pre-push hook
- stylelint:fix applied to all components

## PR Checklist

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -1,6 +1,11 @@
 # Reference:
 # https://github.com/evilmartians/lefthook/blob/master/docs/full_guide.md
 
+pre-commit:
+  commands:
+    stylelint:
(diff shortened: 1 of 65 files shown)


lefthook.yml
```diff
@@ -1,6 +1,11 @@
 # Reference:
 # https://github.com/evilmartians/lefthook/blob/master/docs/full_guide.md
 
+pre-commit:
+  commands:
+    stylelint:
+      run: yarn stylelint:fix && git add .
```
adamdehaven, line 7: We don't want to fix in a commit-hook; when we chat I can explain why.

adamdehaven (approved):
lgtm

PR #1077, opened 2023-01-27
feat(ktextarea): add prop to allow resizing of textarea

dustinryerson:
# Summary

Adds `is-resizable` prop to allow manually resizing of a textarea

Screenshot 2023-01-27 at 3 26 26 PM

## PR Checklist

### DIFF
--- a/src/components/KTextArea/KTextArea.vue
+++ b/src/components/KTextArea/KTextArea.vue
@@ -232,7 +232,7 @@ export default defineComponent({
     resize: none;
 
     &.is-resizable {
-      resize: both;
+      resize: vertical;
     }
 
     &:focus::placeholder {
--- a/src/components/KTextArea/KTextArea.vue
+++ b/src/components/KTextArea/KTextArea.vue
@@ -7,6 +7,7 @@
       v-if="!label"
       v-bind="modifiedAttrs"
       class="form-control k-input style-body-lg"
+      :class="[isResizable ? 'is-resizable' : undefined]"
       :cols="cols"
       :rows="rows"
       :value="getValue()"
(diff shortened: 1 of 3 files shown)


src/components/KTextArea/KTextArea.vue
```diff
@@ -224,6 +231,10 @@ export default defineComponent({
     font-family: var(--font-family-sans);
     resize: none;
 
+    &.is-resizable {
+      resize: both;
```
adamdehaven, line 235: question: should we change this property to `resize: vertical` so that the user cannot resize it beyond the width of its container?

adamdehaven: image
(Otherwise it lets me do this)

aanchalm01:
Quick question: would it make sense to have a min width/height ? so that an user cannot do this:

image 911

dustinryerson:
@adamdehaven I think that makes sense. The main goal for adding this is to make it easier for users to enter code into textarea inputs, so hopefully a vertical resize will be enough. We can always revisit if we get feedback that a horizontal resize would be helpful.

> Quick question: would it make sense to have a min width/height ? so that an user cannot do this:
> 
> image 911

@aanchalm01 great question. I noticed that simply by leveraging the `rows` and `cols` props the text box can be sized down to this:

Screenshot 2023-01-31 at 9 17 21 AM

What do you think about setting reasonable minimum values for the `rows` and `cols` props, and then using the size of the textarea at those minimum values as our `min-width` and `min-height` values?

aanchalm01:
@dustinryerson - Using the size of the textarea with minimum values of props as the min-width and min-height values sounds good!

adamdehaven (approved):
lgtm

PR #1085, opened 2023-01-31
fix(kcodeblock): minor code block issues

kleinfreund:
# Summary

* **fix(kcodeblock): not replacing all new lines in single line mode**

  Fixes an issue with the code block component not replacing all new line characters when 
 `props.isSingleLine` is `true`.

  **Dev note**: With this change, it’s no longer necessary to set `white-space: nowrap` on the `code` element.

  **Dev note**: Replaces access of `props.code` via optional chaining with direct access as `props.code` is required and of type `String`. It is not expected to ever be `undefined` or `null`.

  **Dev note**: As an aside, it’s not required to special case the computation of `totalLines`; hence, the previous version of the computation was restored. It behaves correctly for the empty string, strings without new line characters, and strings with new line characters.
* **fix(kcodeblock): scrolling moving line numbers out of view**

  Fixes horizontal scrolling of code to cause the line numbers to be scrolled out of view.
* **refactor(kcodeblock): moves styles into scoped block**

  Moves some styles from the unscoped style block into the scoped style block which were unintentionally placed there.
* **refactor(kcodeblock): always accesses props using `props.`**

## Comment

I’d rather have favored an implementation of a single line mode that expects `props.code` to come preformatted (i.e. as a string without new line characters) as also suggested/asked about by Adam in https://github.com/Kong/kongponents/pull/1027#issuecomment-1370301538, but changing this now would be a breaking change.

## PR Checklist

### DIFF
--- a/src/components/KCodeBlock/KCodeBlock.vue
+++ b/src/components/KCodeBlock/KCodeBlock.vue
@@ -793,6 +793,23 @@ $dark-focusColor: var(--green-500, color(green-500));
   padding: var(--spacing-md, spacing(md)) 0 0 var(--spacing-sm, spacing(sm));
 }
 
+.k-code-block pre.is-single-line {
+  grid-template-columns: auto;
+  padding: var(--spacing-sm, spacing(sm)) var(--spacing-xxl, spacing(xxl)) 0 0;
+
+  code {
+    line-height: 29px;
+    margin-right: 20px;
+    overflow-x: auto;
+    padding-bottom: var(--spacing-xs, spacing(xs));
+    padding-left: var(--spacing-sm, spacing(sm));
+  }
+
+  + .k-code-block-copy-button {
+    top: var(--spacing-xs, spacing(xs));
+  }
+}
+
 .k-code-block.theme-dark pre {
   background-color: var(--KCodeBlockBackgroundColor, $dark-backgroundColor);
 }
@@ -1161,8 +1178,6 @@ $dark-focusColor: var(--green-500, color(green-500));
 </style>
 
 <style lang="scss">
-@import '@/styles/variables';
-
 .k-matched-term {
   color: var(--teal-500, color(teal-500));
   font-weight: 900;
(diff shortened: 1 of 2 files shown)


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -9,7 +9,7 @@
     tabindex="0"
   >
     <div
-      v-if="isSearchable && !isSingleLine"
+      v-if="isSearchable && !props.isSingleLine"
```
adamdehaven, line 12: All places where you're utilizing props in the `template` section: Since props are defined down in the `script setup` function, you don't need to reference them with `props.*` in the template (and actually, you shouldn't).

They are automatically exposed as the prop name, e.g. `isSingleLine`

adamdehaven (commented):
I reviewed before I realized this was still a draft PR. Just one comment about accessing `props` in the `template` tag -- everything else looks ok so far

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: Ha, interesting. I added those references specifically because I believe in the opposite: props should always be referred to using `props.` because that makes it unambiguous where they come from.

kleinfreund: I’ve removed all instances of this now. I still think this is worse because it’s now more ambiguous where symbols come from, but I’m fine enough with not doing it.

kleinfreund: I understand well _how_ this work and which things are exposed in this manner and which aren’t.

My point is that I think this is a unnecessarily ambiguous way of referencing symbols in the template. Personally, I think referring to `props.showCopyButton` is **far** more clear than just `showCopyButton`: In other words, I would actually _disallow_ referring to props without accessing the `props` object.

But as I said, I’m fine with this stance as well. :)

adamdehaven: Last comment for me 😬 

> Declared props are automatically exposed to the template.
> [Source](https://vuejs.org/guide/essentials/component-basics.html#passing-props)

But yes I don't disagree at all about it being more readable.

adamdehaven (approved):
lgtm

PR #1088, opened 2023-01-31
fix(kbadge): update hover behavior [KHCP-5973]

portikM:
# Summary

https://konghq.atlassian.net/browse/KHCP-5973

## PR Checklist


adamdehaven:
Might make sense to add a CSS transition to the hover state as well, something like `.2s ease-in-out`

portikM:
> Might make sense to add a CSS transition to the hover state as well, something like `.2s ease-in-out`

Will try tomorrow but have a feeling the two won't play nice together.

src/components/KBadge/KBadge.vue
```diff
@@ -338,6 +342,27 @@ watch(badgeText, () => {
 @import '@/styles/functions';
 
 .k-badge {
+  // default appearance colors local variables
```
kaiarrowood, line 345: ❤️

kaiarrowood (approved):
Looks great, thank you for cleaning up the styles with the CSS vars! 🚀

PR #1094, opened 2023-02-01
fix(kmultiselect): dismissable badge [khcp-5995]

aanchalm01:
# Summary
Addresses:
[https://konghq.atlassian.net/browse/KHCP-5995](https://konghq.atlassian.net/browse/KHCP-5995)

Updates KMultiselect so that user can have an option of whether KBadge for an item should be dismissable or not.
image 942
image 945
image 946

## PR Checklist

### DIFF
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -45,7 +45,7 @@ You can specify `disabledTooltipText` property to customize the disabled tooltip
     value: 'dogs',
     selected: true,
     disabled: true,
-    disabledTooltipText: 'This item is locked and can not be removed'
+    disabledTooltipText: 'This item is locked and cannot be removed'
   }, {
     label: 'Bunnies',
     value: 'bunnies',
@@ -743,7 +743,7 @@ export default defineComponent({
         value: 'dogs',
         selected: true,
         disabled: true,
-        disabledTooltipText: 'This item is locked and can not be removed'
+        disabledTooltipText: 'This item is locked and cannot be removed'
       }, {
         label: 'Bunnies',
         value: 'bunnies',
@@ -800,7 +800,7 @@ export default defineComponent({
         value: 'dogs',
         selected: true,
         disabled: true,
-        disabledTooltipText: 'This item is locked and can not be removed'
+        disabledTooltipText: 'This item is locked and cannot be removed'
       }, {
         label: 'Bunnies',
         value: 'bunnies',
@@ -830,7 +830,7 @@ export default defineComponent({
         value: 'dogs',
         selected: true,
         disabled: true,
-        disabledTooltipText: 'This item is locked and can not be removed'
+        disabledTooltipText: 'This item is locked and cannot be removed'
       }, {
         label: 'Bunnies',
         value: 'bunnies'
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -42,7 +42,8 @@ If an item is specified with `selected: true` and `disabled: true`, then the ite
     label: 'Dogs',
     value: 'dogs',
     selected: true,
-    disabled: true
+    disabled: true,
+    disabledTooltipText: 'This selection is dismissible'
   }, {
     label: 'Bunnies',
     value: 'bunnies',
@@ -739,7 +740,8 @@ export default defineComponent({
         label: 'Dogs',
         value: 'dogs',
         selected: true,
-        disabled: true
+        disabled: true,
+        disabledTooltipText: 'This selection is dismissible'
       }, {
         label: 'Bunnies',
         value: 'bunnies',
@@ -795,8 +797,8 @@ export default defineComponent({
         label: 'Dogs',
         value: 'dogs',
         selected: true,
-        selected: true,
-        disabled: true
+        disabled: true,
+        disabledTooltipText: 'This selection is dismissible'
       }, {
         label: 'Bunnies',
         value: 'bunnies',
@@ -825,7 +827,8 @@ export default defineComponent({
         label: 'Dogs',
         value: 'dogs',
         selected: true,
-        disabled: true
+        disabled: true,
+        disabledTooltipText: 'This selection is dismissible'
       }, {
         label: 'Bunnies',
         value: 'bunnies'
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -24,7 +24,10 @@ An array of items containing a `label` and `value`.
 You may also specify:
 - a certain item is `selected` by default
 - a certain item is `disabled`
-- if a certain item has `selected: true` and `disabled: true`, then KBadge for that item is not dismissable
+
+:::tip TIP
+If an item is specified with `selected: true` and `disabled: true`, then the item will be selected, disabled in the dropdown list, and the dismiss button will be removed, meaning a user cannot remove the selected item.
+:::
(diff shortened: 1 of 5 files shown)


docs/components/multiselect.md
```diff
@@ -24,6 +24,7 @@ An array of items containing a `label` and `value`.
 You may also specify:
 - a certain item is `selected` by default
 - a certain item is `disabled`
+- a certain item is readonly by using `isReadonly` property
```
adamdehaven, line 27: suggestion: can this be updated to `readonly` since the other properties do not use the `is*` prefix?

aanchalm01: I can update that, but since we have `attrs.readonly` as well `readonly class` used in the kongponent, I chose to keep `isReadonly` to avoid confusion. How does dismissable sound or should stick with `readonly` ?

adamdehaven: `attrs.readonly` shouldn't conflict since this is a property of an `item`, right?

I wouldn't go with `dismissable` since you're not really "dismissing" the item but removing it. 

You could do `removable`? 

---

I'm also wondering if you could instead just use the existing `disabled` property? 

- If `disabled: true` in the list, then you can't select it.
- If `disabled: true && selected: true` then you can't remove it.

Unsure if this logic makes sense though, but if yes, it would alleviate the need for another property.

aanchalm01: There is no conflict with attrs.readonly, just wanted to avoid any confusion. For the usecase that I am working with `disabled: true && selected: true` won't work because not every `disabled + selected item` needs to be `readonly`. `removable` sounds good.

adamdehaven:
image

We need to adjust some of the CSS Variables (I think) for a badge that does not have the dismiss button so that it renders at the same size.

adamdehaven:
image

Since `Dogs` is not allowed to be removed, I thought you were going to add some description text or something?

docs/components/multiselect.md
```diff
@@ -24,6 +24,7 @@ An array of items containing a `label` and `value`.
 You may also specify:
 - a certain item is `selected` by default
 - a certain item is `disabled`
+- if a certain item has `selected: true` and `disabled: true`, then KBadge for that item is not dismissable
```
adamdehaven, line 27: Let's change the formatting here a bit
```suggestion

:::tip TIP
If an item is specified with `selected: true` and `disabled: true`, then the item will be selected, disabled in the dropdown list, and the dismiss button will be removed, meaning a user cannot remove the selected item.
```

aanchalm01:
> image
> 
> Since `Dogs` is not allowed to be removed, I thought you were going to add some description text or something?

hmm I was thinking of doing that part through Konnect directly

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -211,7 +215,7 @@
           v-for="item, idx in visibleSelectedItemsStaging"
           :key="`${item.key ? item.key : idx}-badge`"
           class="mr-1 mt-2"
-          dismissable
+          :dismissable="!(item.selected && item.disabled)"
```
adamdehaven, line 218: nit(non-blocking): this seems more readable IMO
```suggestion
          :dismissable="item.selected && !item.disabled"
```

src/components/KMultiselect/KMultiselectItem.vue
```diff
@@ -16,11 +16,30 @@
         @click="handleClick"
       >
         <span class="k-multiselect-item-label mr-2">
-          <slot name="content">{{ item.label }}</slot>
+          <slot name="content">{{ item.label }}
+            <div
+              v-if="(item.selected && item.disabled)"
```
adamdehaven, line 21: nit: parenthesis aren't needed
```suggestion
              v-if="item.selected && item.disabled"
```

src/components/KMultiselect/KMultiselectItem.vue
```diff
@@ -41,6 +60,11 @@ const props = defineProps({
     // Items must have a label and value
     validator: (item: Record<string, string | number | boolean>): boolean => item.label !== undefined && item.value !== undefined,
   },
+  disabledTooltipText: {
+    type: String,
+    // Default tooltip text
+    default: 'This item is readonly',
+  },
```
adamdehaven, line 67: question: why are we adding this new prop? I thought we were just adding the icon?

aanchalm01: Fixed.

docs/components/multiselect.md
```diff
@@ -25,6 +25,12 @@ You may also specify:
 - a certain item is `selected` by default
 - a certain item is `disabled`
 
+:::tip TIP
+If an item is specified with `selected: true` and `disabled: true`, then the item will be selected, disabled in the dropdown list, and the dismiss button will be removed, meaning a user cannot remove the selected item.
+
+You can specify `disabledTooltipText` property to customize the disabled tooltip label. Defaults to `This item cannot be removed`.
```
adamdehaven, line 31:
```suggestion
You can specify `disabledTooltipText` property to customize the disabled tooltip that appears when hovering over the lock icon. Defaults to `This item cannot be removed`.
```

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -211,7 +215,7 @@
           v-for="item, idx in visibleSelectedItemsStaging"
           :key="`${item.key ? item.key : idx}-badge`"
           class="mr-1 mt-2"
-          dismissable
+          :dismissable="item.selected && !item.disabled"
```
adamdehaven, line 218: question: just making sure this logic looks correct?

You can remove the item if it IS selected, and NOT disabled. If this sounds good, just resolve this comment 😬

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -828,6 +833,10 @@ watch(() => props.items, (newValue, oldValue) => {
       unfilteredItems.value[i].selected = false
     }
 
+    if (unfilteredItems.value[i].selected === true && unfilteredItems.value[i].disabled === true && !unfilteredItems.value[i].disabledTooltipText) {
```
adamdehaven, line 836: suggestion: adding a comment
```suggestion
    // If an item is `selected` and `disabled`, provide fallback tooltip text if not provided 
    if (unfilteredItems.value[i].selected === true && unfilteredItems.value[i].disabled === true && !unfilteredItems.value[i].disabledTooltipText) {
```

src/components/KMultiselect/KMultiselect.vue
aanchalm01: That sounds right.

aanchalm01:
@adamdehaven - all feedback addressed.

docs/components/multiselect.md
```diff
@@ -36,7 +42,10 @@ You may also specify:
     selected: true
   }, {
     label: 'Dogs',
-    value: 'dogs'
+    value: 'dogs',
+    selected: true,
+    disabled: true,
+    disabledTooltipText: 'This item is non-dismissible'
```
adamdehaven, line 48: Can you change all the instances of `This item is non-dismissible` to `This item is locked and can not be removed`
```suggestion
    disabledTooltipText: 'This item is locked and can not be removed'
```

PR #1097, opened 2023-02-02
feat: add ktruncate component [KHCP-5895]

portikM:
# Summary

Screen Shot 2023-02-03 at 6 14 49 PM

Ticket: https://konghq.atlassian.net/browse/KHCP-5895

## How To Test

**Some tips on how you could try to break it:**

* [ ] feeding it too many items to see if it truncates them
* [ ] feeding it just enough items to fit in the number of rows you specified and resizing your screen to see if it will truncate them when the container gets too small and elements overflow
* [ ] performing same with text - give it just enough text so it doesn't overflow and then resize your screen so it does and see if truncates it
* [ ] specifying rows prop way greater than amount of content you provide to confirm it doesn't look/behave weird (for example, give it just two elements in default slot and specify number of rows as 10)
* [ ] performing the same with text content

## PR Checklist


src/components/KTruncate/KTruncate.vue
```diff
@@ -0,0 +1,281 @@
...
+    wrapperHeight.value = container.value.offsetHeight > targetWrapperHeight ? `${targetWrapperHeight}px` : 'auto'
+  }
+}
+
+/**
+ * Determines whether expand/collapse toggle need to be shown.
+ * In other words, determines whether the content overflows.
+ * If the height of the content is greater than height of the wrapper element - content does overflow.
+ */
+const updateToggleVisibility = () => {
```
adamdehaven, line 160:
```suggestion
const updateToggleVisibility = (): void => {
```

src/components/KTruncate/KTruncate.vue
```diff
@@ -0,0 +1,281 @@
...
+
+const excessElementsCount = ref<number>(0)
+
+/**
+ * Sets wrapper height.
+ * When rows prop === 1: sets wrapper height equal to height of the tallest child.
+ * When rows prop > 1: wrapper hight will be rows count multiplied by height of the tallest child plus gap (10px).
+ * For example if rows is 2 and all elements are equal height if 22px, wrapper height will be set to 54px (2 * 22 + 10).
+ */
+const setWrapperHeight = () => {
```
adamdehaven, line 138:
```suggestion
const setWrapperHeight = (): void => {
```

src/components/KTruncate/KTruncate.vue
```diff
@@ -0,0 +1,281 @@
...
+const props = defineProps({
+  rows: {
+    type: Number,
+    default: 1,
+    validator: (value: Number) => value > 0,
+  },
+  isTextContent: {
+    type: Boolean,
+    default: false,
+  },
```
adamdehaven, line 112: I'm a little confused on the prop name, but also on why this prop is necessary. 

You can slot a tag (e.g. `<p>`) with text content, so really HTML and text are valid.

src/components/KTruncate/KTruncate.vue
```diff
@@ -0,0 +1,281 @@
+<template>
+  <div
+    ref="wrapper"
+    class="k-truncate"
+    :class="[expanded ? 'expanded' : '', `k-truncate-${isTextContent ? 'text' : 'content'}`]"
+  >
+    <div
+      ref="container"
+      class="container"
```
adamdehaven, line 9: suggestion: `container` is a pretty generic class that could have potential style conflicts given a host app's other styles. 

Can you rename this class something more specific, or at least prefix it like `k-truncate-container`?

src/components/KTruncate/KTruncate.vue
```diff
@@ -0,0 +1,281 @@
...
+      v-if="isTextContent && (showToggle || expanded)"
+      ref="textToggleControls"
+      class="place-self-end mt-2"
+    >
+      <div
+        v-if="!expanded"
+        data-testid="expand-trigger-wrapper"
+      >
+        <slot
+
```
adamdehaven, line 69: nit: extra empty line
```suggestion
```

src/components/KTruncate/KTruncate.vue
```diff
@@ -0,0 +1,281 @@
...
+</template>
+
+<script setup lang="ts">
+import { onMounted, onBeforeUnmount, ref, nextTick } from 'vue'
+
+const props = defineProps({
+  rows: {
+    type: Number,
+    default: 1,
+    validator: (value: Number) => value > 0,
```
adamdehaven, line 107: nit
```suggestion
    validator: (value: Number): boolean => value > 0,
```

src/components/KTruncate/KTruncate.vue
```diff
@@ -0,0 +1,281 @@
...
+    }
+  }
+  &.k-truncate-text {
+    display: flex;
+    flex-direction: column;
+
+    .container {
+      -webkit-box-orient: vertical;
+      display: -webkit-box;
+      -webkit-line-clamp: v-bind('$props.rows');
```
adamdehaven, line 268: question: should this be `props.rows` since you are setting the const?

adamdehaven:
The trigger appears below if it's text content, but alongside the items if not text content. Is this desired or should the layout be similar?
image

adamdehaven:
The "Show more" trigger doesn't display like a link, so it's not entirely intuitive that it is clickable
image

adamdehaven:
The default trigger needs some sort of treatment to indicate its clickable to the user 
image

src/components/KTruncate/KTruncate.vue
portikM: This prop is necessary because by default the component treats the content as collection of HTML elements and applies more complex logic. When this prop is true, that logic is skipped, instead -webkit-line-clamp property is applied. What do you think would be a good name?

adamdehaven: Can you pass in multiple `<p>` tags and still allow truncation, or when this prop is true can there only be a single child? Regardless of the answer, I think we need to add more details in the documentation for the prop.

portikM: Yes, you can pass multiple `<p>` tags, doesn't have to be single child. I added another paragraph explaining this logic, plz take a look and let me know if you think we need anything else

portikM:
> The trigger appears below if it's text content, but alongside the items if not text content. Is this desired or should the layout be similar? image

Yes, this is the desired behavior confirmed with design

docs/components/truncate.md
```diff
@@ -0,0 +1,230 @@
...
+    </KBadge>
+  </KTruncate>
+</template>
+```
+
+### isTextContent
+
+By default the component treats anything passed through the `default` slot as collection of HTML elements. Use this prop if you want to truncate text.
+
+When this prop is set tu true, component applies different logic - truncation is achieved through `-webkit-line-clamp` CSS property, rather than using height of child elements to determine the row height.
```
adamdehaven, line 55: typo
```suggestion
When this prop is set to `true`, the component applies different logic; truncation is achieved via the `--webkit-line-clamp` CSS property rather than using the height of child elements to determine the row height.
```

docs/components/truncate.md
```diff
@@ -0,0 +1,230 @@
...
+    </KBadge>
+    <template #expand-trigger="{ excessElementsCount, expand }">
+      <span @click="expand">Show {{ excessElementsCount }} more</span>
+    </template>
+    <template #collapse-trigger="{ collapse }">
+      <span @click="collapse">Show less</span>
+    </template>
+  </KTruncate>
+</template>
+```
```
adamdehaven, line 171: On the example here, the trigger IMO is too close to the items 
image

Is it possible to add a little extra space between?

docs/components/truncate.md
```diff
@@ -0,0 +1,230 @@
...
+## Slots
+
+### default
+
+Slot for content.
+
+To ensure the best experience with the component, please pass elements of equal height in the default slot. The component will base its guess for the height of the element on the height of each individual sibling, so passing elements of equal height will result in a more accurate and consistent output.
+
+:::tip TIP
+The component is reactive to it's dimensions changes. To see it in effect, try resizing your browser window and see the behavior of the example below.
```
adamdehaven, line 114: typo
```suggestion
The component is reactive to its dimensions changes. To see this in action, try resizing your browser and notice the behavior of the example below.
```

docs/components/truncate.md
```diff
@@ -0,0 +1,230 @@
...
+      <KBadge v-for="n in 30" :key="n">
+        Item {{ n }}
+      </KBadge>
+    </KTruncate>
+  </template>
+</KCard>
+
+### expand-trigger
+
+Slot for custom expand trigger element. Slot props:
```
adamdehaven, line 129:
```suggestion
Slot for a custom expand trigger element. Slot props:
```

docs/components/truncate.md
```diff
@@ -0,0 +1,230 @@
...
+- `excessElementsCount` (type: `Number`) - Number of elements that overflow. **Note**: this slot prop is only available when not `isTextContent`
+- `expand` (type: `Function`) - Function to expand
+
+### collapse-trigger
+
+Slot for custom collapse trigger element. Slot props:
```
adamdehaven, line 135:
```suggestion
Slot for a custom collapse trigger element. Slot props:
```

docs/components/truncate.md
```diff
@@ -0,0 +1,230 @@
...
+        Item {{ n }}
+      </KBadge>
+    </KTruncate>
+  </template>
+</KCard>
+
+### expand-trigger
+
+Slot for custom expand trigger element. Slot props:
+- `excessElementsCount` (type: `Number`) - Number of elements that overflow. **Note**: this slot prop is only available when not `isTextContent`
```
adamdehaven, line 130: question: instead of `excessElementsCount` would something like `truncatedCount` be easier/make more sense?

docs/components/truncate.md
```diff
@@ -0,0 +1,230 @@
...
+- `expand` (type: `Function`) - Function to expand
+
+### collapse-trigger
+
+Slot for custom collapse trigger element. Slot props:
+
+- `collapse` (type: `Function`) - Function to collapse
+
+Example of using `expand-trigger` and `collapse-trigger` slots for creating custom toggle elements:
```
adamdehaven, line 139:
```suggestion
Example of using the `expand-trigger` and `collapse-trigger` slots for creating custom toggle elements:
```

src/components/KTruncate/KTruncate.vue
```diff
@@ -0,0 +1,280 @@
...
+/**
+ * Sets wrapper height.
+ * When rows prop === 1: sets wrapper height equal to height of the tallest child.
+ * When rows prop > 1: wrapper hight will be rows count multiplied by height of the tallest child plus gap (10px).
+ * For example if rows is 2 and all elements are equal height if 22px, wrapper height will be set to 54px (2 * 22 + 10).
+ */
+const setWrapperHeight = (): void => {
+  if (props.isTextContent) {
+    return
+  }
```
adamdehaven, line 140: question: since the prop values can change, instead of just returning here, should you reset the `wrapperHeight.value` back to the default `0px`?

src/components/KTruncate/KTruncate.vue
```diff
@@ -0,0 +1,280 @@
...
+@import '@/styles/variables';
+@import '@/styles/functions';
+
+.k-truncate {
+  align-items: flex-start;
+  display: flex;
+  overflow: hidden;
+  width: 100%;
+
+  .expand-trigger, .collapse-trigger {
```
adamdehaven, line 224: nit: move to new line
```suggestion
  .expand-trigger, 
  .collapse-trigger {
```

src/components/KTruncate/KTruncate.vue
```diff
@@ -0,0 +1,280 @@
...
+      display: -webkit-box;
+      -webkit-line-clamp: v-bind('props.rows');
+      overflow: hidden;
+    }
+    &.expanded {
+      .k-truncate-container {
+        display: block;
+      }
+    }
+    .expand-trigger, .collapse-trigger {
```
adamdehaven, line 275: nit: move to new line
```suggestion
    .expand-trigger, 
    .collapse-trigger {
```

src/components/KTruncate/KTruncate.vue
portikM: I'm not sure I understand what you mean. The return here is needed rather for optimization purposes - to avoid doing those calculations because they are not needed when `props.isTextContent === true`, the `wrapperHeight.value` is not taken into account in such case

docs/components/truncate.md
portikM: Good idea, I like that name better

portikM:
> The default trigger needs some sort of treatment to indicate its clickable to the user image

This is how it's presented in the design mocks. Besides, it's using the KButton (`btn-link` appearance) under the hood - users should be somewhat familiar with this visual element.

portikM:
@adamdehaven All feedback addressed :)

src/components/KTruncate/KTruncate.vue
adamdehaven: Right but what I'm asking is if a scenario like this is covered:

1. The user initializes the component with `isTextContent="false"` and so then the `wrapperHeight.value` is set to something that is not `0px`
2. In the host app, the slot content is changed dynamically and the `isTextContent` prop value is now switched to `true`

This means that the logic in this block would be ignored, but the `wrapperHeight.value` is still set to a value other than `0px`. Does this matter?

docs/components/truncate.md
```diff
@@ -0,0 +1,236 @@
+# Truncate
+
+**KTruncate** - A Kongponent that limits content to a specific number of lines and provides controls to expand or collapse it.
```
adamdehaven, line 3: nit
```suggestion
**KTruncate** - A Kongponent that limits content to a specific number of lines and provides controls to show or hide the truncated content.
```

docs/components/truncate.md
```diff
@@ -0,0 +1,236 @@
...
+    </KBadge>
+  </KTruncate>
+</template>
+```
+
+## Props
+
+### rows
+
+Number of rows to truncate content at. Default value is `1`.
```
adamdehaven, line 29:
```suggestion
Number of rows to truncate content. Default value is `1`.
```

docs/components/truncate.md
```diff
@@ -0,0 +1,236 @@
...
+    <KBadge v-for="n in 25" :key="n">
+      Item {{ n }}
+    </KBadge>
+  </KTruncate>
+</template>
+```
+
+### isTextContent
+
+By default the component treats anything passed through the `default` slot as collection of HTML elements. Use this prop if you want to truncate text.
```
adamdehaven, line 53:
```suggestion
By default the component treats anything passed through the `default` slot as collection of HTML elements. Use this prop if you want to truncate elements that only contain text, such as one or more paragraph `<p>` tags.
```

docs/components/truncate.md
```diff
@@ -0,0 +1,236 @@
...
+
+When this prop is set to `true`, the component applies different logic; truncation is achieved via the `-webkit-line-clamp` CSS property, rather than assessing the height of child elements to determine the row height.
```
adamdehaven, line 55:
```suggestion
When this prop is set to `true`, the component applies different logic; truncation is achieved via the `--webkit-line-clamp` CSS property, rather than assessing the height of child elements to determine the row height.
```

docs/components/truncate.md
```diff
@@ -0,0 +1,236 @@
...
+      organized and readable manner, allowing the user to quickly scan 
+      and understand the main points of each item.</p>
+      <p>The truncated text can also serve as a teaser, encouraging the user to click or tap to view the full content.</p>
+  </KTruncate>
+</template>
+```
+
+### isExpanded
+
+If passed `true`, the component will be rendered expanded by default.
```
adamdehaven, line 83: nit
```suggestion
When `true`, the component will not truncate the content and the collapse trigger will be visible.
```

docs/components/truncate.md
```diff
@@ -0,0 +1,236 @@
...
+    </KBadge>
+  </KTruncate>
+</template>
+```
+
+## Slots
+
+### default
+
+Slot for content.
```
adamdehaven, line 109: nit
```suggestion
The content to truncate.
```

docs/components/truncate.md
```diff
@@ -0,0 +1,236 @@
...
+
+To ensure the best experience with the component, please pass elements of equal height in the default slot. The component will base its guess for the height of the element on the height of each individual sibling, so passing elements of equal height will result in a more accurate and consistent output.
```
adamdehaven, line 111: The explanation here is a little misleading if the user is providing paragraph text as the content. Does it make sense to note that here?

docs/components/truncate.md
```diff
@@ -0,0 +1,236 @@
...
+        Item {{ n }}
+      </KBadge>
+    </KTruncate>
+  </template>
+</KCard>
+
+### expand-trigger
+
+Slot for a custom expand trigger element. Slot props:
+- `truncatedCount` (type: `Number`) - Number of elements that overflow. **Note**: this slot prop is only available when not `isTextContent`
```
adamdehaven, line 130: nit: need an empty line before a list
```suggestion

- `truncatedCount` (type: `Number`) - Number of elements that overflow. **Note**: this slot prop is only available when not `isTextContent`
```

docs/components/truncate.md
portikM: It's actually `-webkit-line-clamp`, not `--webkit-line-clamp`

adamdehaven: 😨 whoops

docs/components/truncate.md
portikM: Agree, maybe a little confusing -- added a note, specifying that this doesn't apply to text truncation scenarios

adamdehaven (approved):
LGTM

kaiarrowood (dismissed):
Looks great! 🚀

kaiarrowood (approved):
LGTM! Very nice ✨

PR #1111, opened 2023-02-10
fix(kcodeblock): escape html tags in code block

davidma415:
# Summary
Fixes an issue when you're rendering `HTML` inside of the code block. It would previously display the actual `div` if you entered a div tag for instance.

## PR Checklist


adamdehaven (approved):
LGTM - thanks

PR #1125, opened 2023-02-16
feat(kmodalfullscreen): add footer content slot [KHCP-6093]

portikM:
# Summary

Ticket: https://konghq.atlassian.net/browse/KHCP-6093

Screen Shot 2023-02-17 at 10 01 25 AM

## PR Checklist

### DIFF
--- a/docs/components/modal-fullscreen.md
+++ b/docs/components/modal-fullscreen.md
@@ -107,6 +107,7 @@ There are 6 designated slots you can use to display content in the fullscreen mo
 - `action-buttons` - contains action buttons which are right-aligned. If not used, provide default Cancel/Submit buttons
 - `body-header` - title to display in the body section
 - `body-header-description` - description text displayed beneath the `body-header`
+- `footer-content` - footer content. If provided, the footer will stick to the bottom of the fullscreen modal and the header behavior will change to static. The footer will contain duplicate of action buttons passed through the `action-buttons` slot. If no action buttons are provided, a default Submit button will be included (see example below).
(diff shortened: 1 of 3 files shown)


docs/components/modal-fullscreen.md
```diff
@@ -107,25 +107,25 @@ There are 6 designated slots you can use to display content in the fullscreen mo
 - `action-buttons` - contains action buttons which are right-aligned. If not used, provide default Cancel/Submit buttons
 - `body-header` - title to display in the body section
 - `body-header-description` - description text displayed beneath the `body-header`
+- `footer-content` - footer content. If provided, the footer will stick to the bottom of the fullscreen modal and the header behavior will change to static. The footer will contain duplicate of action buttons passed through the `action-buttons` slot. If no action buttons are provided, a default Submit button will be included (see example below).
```
jillztom, line 110: > a default Submit button will be included
^ I like making the header static if there's a footer bar present. Wondering if there should be a default "Submit" button if nothing is provided 🤔

jillztom (commented):
Everything else looks good to me

docs/components/modal-fullscreen.md
portikM: This is tailored to our use case - in the Figma mock we only have a Submit button in the footer and we are using the default ones. But also, from UX perspective I think this is okay because more often then not content inside of these fullscreen modals is not too long for user to scroll back up to header where the Cancel button is in case they changed their mind.

PR #1166, opened 2023-03-02
fix(ktruncate): a11y and interaction improvements [KHCP-6105]

portikM:
# Summary

Ticket: https://konghq.atlassian.net/browse/KHCP-6105

Improvements:
- expand trigger is the first element with component focus is set on
- truncated elements are not tabbable
- the outline of some items is no longer cut off by the height of the truncate container
- collaps trigger focus-within styles added (screenshot)
Screen Shot 2023-03-01 at 7 14 09 PM

## PR Checklist


adamdehaven (dismissed):
image

(non-blocking): Looks like some of the outline is cut off by the container; would it make sense to add padding or something? I remember we chatted about this previously but I don't remember where that thread ended up.

Everything else LGTM

portikM:
> image
> 
> (non-blocking): Looks like some of the outline is cut off by the container; would it make sense to add padding or something? I remember we chatted about this previously but I don't remember where that thread ended up.
> 
> Everything else LGTM

Fixed in latest commit

adamdehaven (approved):
lgtm

PR #1171, opened 2023-03-06
fix(kmultiselect): fix input element full width [KHCP-6460]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-6460

## PR Checklist

### DIFF
--- a/src/components/KMultiselect/KMultiselect.vue
+++ b/src/components/KMultiselect/KMultiselect.vue
@@ -962,6 +962,7 @@ onMounted(() => {
     position: absolute;
     right: 1px;
     top: 1px;
+    z-index: 1;
 
     .k-multiselect-chevron-icon {
       position: relative;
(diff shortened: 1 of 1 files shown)


src/components/KMultiselect/KMultiselect.vue
```diff
@@ -1002,6 +1003,7 @@ onMounted(() => {
     .k-multiselect-input {
       display: inline-block;
       position: relative;
+      width: 100%;
```
adamdehaven, line 1006: I'm unsure if this fixes in all scenarios. When I inspect the multiselect item template slot example, I see this rule taking precedence
image

portikM: The added `width` property is actually targeting the wrapper element - which is the one limiting input elements width. I tested the scenario when `item-template` slot is passed - it works as expected.

adamdehaven (approved):

PR #1172, opened 2023-03-06
docs(alert): responsiveness improvements

portikM:
# Summary

Before:
Screen Shot 2023-03-06 at 5 58 26 PM

After:
Screen Shot 2023-03-06 at 5 59 23 PM

## PR Checklist


src/components/KAlert/KAlert.vue
```diff
@@ -116,6 +116,7 @@
 
       <KButton
         v-if="dismissType === 'button'"
+        class="white-space-nowrap"
```
adamdehaven, line 119: question:
Shouldn't a KButton, by default, have `white-space: nowrap`?

suggestion: I think this rule should just be added directly to KButton instead?

Also, let's not keep adding utility classes. Again, as we move towards an actual design system, all of these utility classes will have to be refactored out of Kongponents (and utilize different, targeted classes) so let's not add more in the meantime if not absolutely necessary.

portikM: Yeah you're right, better handle that from KButton itself. Removed the newly added utility class too (keep forgetting we don't want to add new ones at all).

adamdehaven (approved):
LGTM -- you'll definitely want to manually pull this into `khcp-ui` and test locally to ensure the button CSS change isn't breaking

portikM:
> LGTM -- you'll definitely want to manually pull this into `khcp-ui` and test locally to ensure the button CSS change isn't breaking

Do you think we have any occurrences of KButton where text wraps in Konnect?

adamdehaven:
> Do you think we have any occurrences of KButton where text wraps in Konnect?

There are definitely places where they wrap _unintentionally_ -- so I think it will be fine; I'm just saying click around and spot-check 👍🏼

PR #1182, opened 2023-03-08
feat(kinput): add icon slot [KHCP-6455]

portikM:
# Summary

Ticket: https://konghq.atlassian.net/browse/KHCP-6455

Screen Shot 2023-03-09 at 4 36 02 PM
Screen Shot 2023-03-09 at 4 36 17 PM

## PR Checklist


docs/components/input.md
```diff
@@ -159,6 +159,24 @@ String to be displayed as error message if `hasError` prop is `true`.
 <KInput label="Large" size="large" hasError errorMessage="Service name should not contain '_'" :overlay-label="true" />
 ```
 
+### iconPosition
+
+Controls position of the icon provided through the [slot](#slots).
```
adamdehaven, line 164: suggestion: can you add a note here on the allowed values?

docs/components/input.md
```diff
@@ -221,6 +239,52 @@ export default defineComponent({
 </script>
 ```
 
+## Slots
+
+- `icon` - slot for icon on the left or right of the input (position can be controlled through [prop](#iconposition))
+
+:::tip TIP
+Whether you choose to use KIcon Kongponent or your own SVG, the component stylings will adjust the icon size to match the size of the component.
```
adamdehaven, line 247:
```suggestion
Whether you choose to use `KIcon` Kongponent or your own SVG, the component's styles will adjust the icon size to match the size of the component.
```

src/components/KInput/KInput.vue
```diff
@@ -144,6 +156,11 @@ export default defineComponent({
       // Ensure the characterLimit is greater than zero
       validator: (limit: number):boolean => limit > 0,
     },
+    iconPosition: {
+      type: String,
```
adamdehaven, line 160:
```suggestion
      type: String as PropType<'left' | 'right'>,
```

docs/components/input.md
```diff
@@ -272,18 +336,56 @@ Fired when the text starts or stops exceeding the limit, returns an object:
 </KComponent>
 ```
 
+### `icon:click`
```
adamdehaven, line 339: Two issues:

1. Can we make the emit `click:icon` instead? This way when it's bound to the component, it will look like `@click:icon` which seems more standard.
2. I don't see where you actually added the `emit` in the component.

portikM: 1. Respectfully disagree - we have similar pattern in `KTable` and we use `@row:click` and `@cell:click` there.
2. The component doesn't emit the event - it calls bound callback function directly (again, similar pattern we have in `KTable`) - basically a workaround I had to do because adding 'icon:click' to emits will remove it from attributes so isIconClickable.value always returns false. Let me know if this is confusing

docs/components/input.md
```diff
@@ -159,6 +159,24 @@ String to be displayed as error message if `hasError` prop is `true`.
 <KInput label="Large" size="large" hasError errorMessage="Service name should not contain '_'" :overlay-label="true" />
 ```
 
+### iconPosition
+
+Controls position of the icon provided through the [slot](#slots). Accepted values: `left` (default) and `right`.
```
kaiarrowood, line 164: suggestion: I'm not a big fan of `right`/`left` - if we ever get big enough to support a RTL language this will not translate well. Can we use `leading`/`trailing` instead?

docs/components/input.md
```diff
@@ -272,18 +336,56 @@ Fired when the text starts or stops exceeding the limit, returns an object:
 </KComponent>
 ```
 
+### `icon:click`
+
+You can make icon passed through the [slot](#slots) clickable by binding to `@icon:click` event.
```
kaiarrowood, line 341:
```suggestion
You can make the icon passed through the [slot](#slots) clickable by binding to the `@icon:click` event.
```

src/components/KInput/KInput.cy.ts
```diff
@@ -131,4 +132,35 @@ describe('KInput', () => {
...
+    cy.get('.input-icon').should('be.visible').should('contain.text', icon)
+  })
+
+  it('should render icon clickable when event listener is bound', () => {
+    const onIconClickSpy = cy.spy().as('onIconClickSpy')
+    mount(KInput, {
+      props: {
+        testMode: true,
+        // make it clickable
+        'onIcon:click': onIconClickSpy,
```
kaiarrowood, line 156: ✨

src/components/KInput/KInput.vue
```diff
@@ -20,7 +20,7 @@
           :id="inputId"
           :aria-invalid="hasError || charLimitExceeded ? 'true' : undefined"
           class="form-control k-input"
-          :class="`k-input-${size}`"
+          :class="[`k-input-${size}`, $slots['icon'] ? `has-icon icon-${iconPosition}` : '' ]"
```
kaiarrowood, line 23:
```suggestion
          :class="{ `k-input-${size}`: size, `has-icon icon-${iconPosition}`: $slots['icon'] }"
```
nit

src/components/KInput/KInput.vue
```diff
@@ -53,7 +53,7 @@
         :id="inputId"
         :aria-invalid="hasError || charLimitExceeded ? 'true' : undefined"
         class="form-control k-input"
-        :class="`k-input-${size}`"
+        :class="[`k-input-${size}`, $slots['icon'] ? `has-icon icon-${iconPosition}` : '' ]"
```
kaiarrowood, line 56:
```suggestion
        :class="{ `k-input-${size}`: size, `has-icon icon-${iconPosition}`: $slots['icon'] }"
```

src/components/KInput/KInput.vue
```diff
@@ -71,7 +71,7 @@
       v-bind="modifiedAttrs"
       :aria-invalid="hasError || charLimitExceeded ? 'true' : undefined"
       class="form-control k-input"
-      :class="`k-input-${size}`"
+      :class="[`k-input-${size}`, $slots['icon'] ? `has-icon icon-${iconPosition}` : '' ]"
```
kaiarrowood, line 74:
```suggestion
      :class="{ `k-input-${size}`: size, `has-icon icon-${iconPosition}`: $slots['icon'] }"
```

docs/components/input.md
adamdehaven: Even if we support RTL, `left/right` still makes sense as it's the position of the icon, not the "start text on the left/right"

portikM: I think so too ^

src/components/KInput/KInput.vue
portikM: I'm afraid this won't work
Screen Shot 2023-03-10 at 12 14 13 PM

src/components/KInput/KInput.vue
portikM: Same as above

src/components/KInput/KInput.vue
portikM: Same as above

docs/components/input.md
kaiarrowood: That's not how RTL rendering works. It's explained in detail [here](https://medium.com/@ahmed044/leading-trailing-vs-left-right-constraints-c315e6f92aa9#:~:text=In%20locales%20where%20the%20reading%20direction%20is%20right%20to%20left,will%20be%20the%20left%20side.), but I'll give a simple example.

I have a use case to have a search input with a clear button rendered at the end.

In LTR I would render that like this:

```html
<KInput icon-position="right">
  <template #icon>
    <KIcon icon="clear" />
  </template>
</KInput>
```

in RTL I would render it this way:

```html
<KInput icon-position="left">
  <template #icon>
    <KIcon icon="clear" />
  </template>
</KInput>
```

So if I want to support both languages now I have to introduce a conditional. But instead I am suggesting we use `leading`/`trailing` - in that scenario I don't need a conditional, this code does what I need in both scenarios.

```html
<KInput icon-position="trailing">
  <template #icon>
    <KIcon icon="clear" />
  </template>
</KInput>
```

src/components/KInput/KInput.vue
kaiarrowood: Updated the snippet

docs/components/input.md
portikM: We decided to go with `start`/`end`

adamdehaven: I'm still not sure this makes sense to me as the component still doesn't behave how you're suggesting when I manually change the direction, meaning there would have to be conditional logic regardless.

image

- The first input above has no modifications, and you can see the icon is displayed on the left.
- For the second input wrapper I have manually changed the `direction: rtl` and you can see the input text _does_ switch as expected; however, the icon position remains on the left.

portikM: That's true - we paired with Kai and I asked the same question. Basically what I got from him is that yes, to fully support RTL we would have to redefine bunch of styles (in this case too). However, if we supported RTL and still had `left` and `right` values accepted by this prop, one would have to leave them as is, which would result in host app having to introduce a condition (something like `:icon-position="isRtl? 'left' : 'right'"`). I agreed with Kai that that would be confusing. @kaiarrowood anything to add?

portikM: Don't mean to be rude but Adam should be enjoying his Friday off 😅

adamdehaven: We'd have to do a bunch of changes to support RTL anyway; whatever you all think is good by me. I was just noting that the behavior is still not there regardless of the prop name

portikM: Yeah, the point of changing the prop values was not to achieve certain behaviour, but rather to abstract towards more relative values

kaiarrowood: Yeah, my thought process is that the logic to correctly display an element in RTL should live in the component and not the consuming application. If we use left/right the host application is forced to conditionally apply the treatment because when speaking about RTL support, the vast majority of use cases want things positioned at the start/end of something rather than always displaying on the left or always displaying on the right. Just speaking from experience I haven't really been able to think of a use case and if you go to a website that supports RTL and switch to an RTL language you'll see the same thing - which is that the UI is basically mirrored, not just the text on the page.

kaiarrowood (commented):
This looks good to me! I'll let @adamdehaven give final approval since there is no rush on this ticket.

adamdehaven (approved):
If @kaiarrowood's good with it, I'm good with it

PR #1186, opened 2023-03-09
fix(kdropdownmenu): divider style

Leopoldthecoder:
# Summary

This PR fixes an expanded background bug when hovering on a `li` element with a nested `a.has-divider` element.

Kapture 2023-03-09 at 17 40 59

## PR Checklist


adamdehaven (commented):
I think this looks good, but @kaiarrowood I know you just modified these same lines yesterday. Can you take a quick look and just verify?

kaiarrowood (approved):
Confirmed this is actually better than the fix I put in, nice 🚀

PR #1193, opened 2023-03-10
feat(chore): added types for collapse, datapicker and dropdown [KHCP-6342]

LukeSwierlik:
Task in Jira: https://konghq.atlassian.net/browse/KHCP-6342

# Summary
- **Feat** - Added types for `KCollapse`
- **Feat** - Added types for `KDataTimePicker`
- **Feat** - Added types for `Dropdown`

## PR Checklist


src/components/KCollapse/KCollapse.vue
```diff
@@ -104,21 +110,21 @@ export default defineComponent({
       default: '',
     },
     triggerAlignment: {
-      type: String,
+      type: String as PropType<TriggerAlignment>,
       required: false,
-      default: 'trailing',
-      validator: (value: string): boolean => {
-        return ['leading', 'trailing'].includes(value)
+      default: triggerAlignmentRecord.trailing,
```
adamdehaven, line 115: When we move to script setup we won't be able to utilize variables as the prop defaults like this; I think you just need to hard-code it.

src/components/KDropdownMenu/KDropdownMenu.vue
```diff
@@ -105,9 +102,9 @@ export default defineComponent({
   },
   props: {
     appearance: {
-      type: String,
-      default: 'menu',
-      validator: (value: string) => ['menu', 'selectionMenu'].includes(value),
+      type: String as PropType<Appearance>,
+      default: appearanceRecord.menu,
```
adamdehaven, line 106: Same comment here on default value

src/types/index.ts
```diff
@@ -5,5 +5,8 @@ export * from './button'
 export * from './card'
 export * from './catalog'
 export * from './code-block'
+export * from './collapse'
+export * from './data-time-picker'
```
adamdehaven, line 9: This file is named incorrectly; it should be this
```suggestion
export * from './date-time-picker'
```

src/types/data-time-picker.ts
```diff
@@ -0,0 +1,37 @@
+export interface TimeRange {
```
adamdehaven, line 1: This filename should be `date-time-picker.ts`

adamdehaven (changes requested):

adamdehaven (approved):
lgtm

PR #1220, opened 2023-03-16
fix(*): update useDebounce to fix delays in KTable after clearing the search

sumimakito:
# Summary

After clearing the search input in a KTable, a perceptible delay is observed before KTable triggers the `fetcher` function.

This PR also refactored the `useDebounce` function to accept functions, which makes it more general.

Another option to fix this is to use the `defineExpose` to expose a function to clear the search input and trigger the `revalidate` function immediately.

KHCP-6651

## PR Checklist


src/components/KTable/KTable.vue
```diff
@@ -736,12 +736,20 @@ export default defineComponent({
 
       return `k-table_${Math.floor(Math.random() * 1000)}_${props.fetcherCacheKey}` as string
     })
-    const { query, search } = useDebounce('', 350)
-    const { revalidate } = useRequest(
+
+    const query = ref('')
+    const [searchDebounced, debounceSearch] = useDebounce((q: string) => { query.value = q }, 350)
```
adamdehaven, line 741: These variable names are a bit confusing to me; the names are essentially just flipped

src/components/KCatalog/KCatalog.vue
```diff
@@ -486,7 +486,9 @@ export default defineComponent({
       hasInitialized.value = true
     }
 
-    const { query, search } = useDebounce('', 350)
+    const query = ref('')
+    const [search] = useDebounce((q: string) => { query.value = q })
```
adamdehaven, line 490: This appears to be a breaking change since the return format is now an array instead of an object?

src/composables/useUtilities.ts
```diff
@@ -1,3 +1,5 @@
+/** @format */
```
adamdehaven, line 1: question: what is this `@format`?

src/composables/useUtilities.ts
```diff
@@ -1,3 +1,5 @@
...
+
 import { ref, Ref, computed, watchEffect } from 'vue'
```
sumimakito, line 3: Sorry, my editor left that.

```suggestion
import { ref, Ref, computed, watchEffect } from 'vue'
```

src/components/KCatalog/KCatalog.vue
sumimakito: Yes. We only have two components using this composable: KTable and KCatalog. I intended to make `useDebounce` return two functions: one is the denounced function, and the other one is a function that accepts a delay (other than the default one) and returns a new debounced function with that new delay (but which shares the same timeout handle). I tried to implement this composable to make it as general as possible for different circumstances.

This allows us to trigger the underlying function with a different delay. For example, if we want to call the function immediately and clear all pending calls, we can use the second function to generate a debounced function with no delay and trigger it.

src/components/KTable/KTable.vue
sumimakito: Would `debouncedSearch` and `generateDebouncedSearch` sounds better here?

src/components/KCatalog/KCatalog.vue
adamdehaven: Is it possible to refactor the composable to return an object instead of an array; one property for each function?

sumimakito: I refactored the `useDebounce` composable to make it returns an object instead of an array.

src/components/KTable/KTable.vue
```diff
@@ -845,11 +853,19 @@ export default defineComponent({
     }
 
     watch(() => props.searchInput, (newValue) => {
-      search(newValue)
+      if (newValue === '') {
+        search(newValue)
+      } else {
+        debouncedSearch(newValue)
+      }
```
adamdehaven, line 860: suggestion: Can you add some comments here to explain why we are using two different methods?

src/components/KTable/KTable.vue
```diff
@@ -845,11 +853,19 @@ export default defineComponent({
...
 
-    watch(() => [query.value, page.value, pageSize.value], () => {
-      revalidate()
+    watch(() => [query.value, page.value, pageSize.value], ([newQuery, , , oldQuery]) => {
+      if (newQuery === '' && newQuery !== oldQuery) {
+        revalidate()
+      } else {
+        debouncedRevalidate()
+      }
     }, { deep: true, immediate: true })
```
adamdehaven, line 869: This is a bit confusing with the empty array in the callback; could this be refactored or utilize `watchEffect` instead?

src/components/KTable/KTable.vue
sumimakito: Comments added

src/components/KTable/KTable.vue
sumimakito: Thanks. That's a good idea.

adamdehaven (approved):
@sumimakito I reviewed everything here and it appears to be solid.

After merge, can you pull it into Kong Manager on Wednesday and test/verify all of the table fetchers are still working as intended? It's hard to test every use-case here

Leopoldthecoder (approved):

PR #1225, opened 2023-03-17
fix(kdropdownitem): add hasDivider to KDropdownItem type [KHCP-6613]

mptap:
# Summary

Adds `hasDivider` optional boolean type to `DropdownItem` so that it can be specified with a divider in `KDropdownItem`

## PR Checklist


adamdehaven (approved):
good catch

PR #1250, opened 2023-03-29
fix(ktabs): using non-semantic anchor for tab items

kleinfreund:
# Summary

**chore(ktabs): expose types**

Moves types defined in `KTabs.vue` into a type definition file and exposes those types.

**fix(ktabs): using non-semantic anchor for tab items**

Replaces `a.tab-link` with `span.tab-link` to allow using `router-link`s for tab slots (nesting `a` is forbidden).

## PR Checklist


src/components/KTabs/KTabs.vue
```diff
@@ -44,10 +44,7 @@
 <script lang="ts">
 import { defineComponent, ref, watch, PropType } from 'vue'
 
-export interface Tab {
-  hash: string
-  title: string
-}
+import type { Tab } from '@/types/tabs'
```
adamdehaven, line 47: nit: import from the root
```suggestion
import type { Tab } from '@/types'
```

src/components/KTabs/KTabs.vue
```diff
@@ -18,9 +18,9 @@
         @keydown.enter.prevent="handleTabChange(tab.hash)"
         @keydown.space.prevent="handleTabChange(tab.hash)"
       >
-        <a class="tab-link">
+        <span class="tab-link">
```
adamdehaven, line 21: If we're already going to change the structure, I'm wondering if we should make this a `div` rather than a `span` to allow semantic HTML based on the notion that a user could literally slot any HTML element.

adamdehaven (changes requested):

src/components/KTabs/KTabs.vue
kleinfreund: Changed in https://github.com/Kong/kongponents/pull/1250/commits/3d667c17cf9cc0acdeaf7b4a56fa35cef093a177.

src/components/KTabs/KTabs.vue
kleinfreund: Changed in https://github.com/Kong/kongponents/pull/1250/commits/abbeb7e26993380252edd1686e999d8e2a46b3ab.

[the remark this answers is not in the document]
kleinfreund: This comment looks to be a duplicate of https://github.com/Kong/kongponents/pull/1250#discussion_r1151856871 so I’m resolving this.

adamdehaven (approved):
LGTM

@Leopoldthecoder & @sumimakito I looked through `khcp-ui` and `kong-admin` and don't see anything that should break because of this change, but let me know if something pops up

PR #1251, opened 2023-03-29
feat(ktabs): accept route to generate router-links

kleinfreund:
# Summary

**feat(ktabs): accept route to generate router-links**

Adds support for a new optional `route` property on the `props.tabs` elements. When provided, `KTabs` will use a `router-link` components instead of the `li` element to render actual links in the tabs. It will also automatically add the `router-view` component in the tab content containers.

## PR Checklist

### DIFF
--- a/src/components/KTabs/KTabs.vue
+++ b/src/components/KTabs/KTabs.vue
@@ -4,26 +4,45 @@
       aria-label="Tabs"
       role="tablist"
     >
-      <li
+      <template
         v-for="(tab, i) in tabs"
-        :id="`${tab.hash.replace('#','')}-tab`"
         :key="tab.hash"
-        :aria-controls="`panel-${i}`"
-        :aria-selected="activeTab === tab.hash ? 'true' : 'false'"
-        class="tab-item"
-        :class="{ active: activeTab === tab.hash }"
-        role="tab"
-        tabindex="0"
-        @click="handleTabChange(tab.hash)"
-        @keydown.enter.prevent="handleTabChange(tab.hash)"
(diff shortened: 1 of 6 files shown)


src/components/KTabs/KTabs.vue
```diff
@@ -36,7 +53,9 @@
       <slot
         v-if="activeTab === tab.hash"
         :name="tab.hash.replace('#','')"
-      />
+      >
+        <router-view v-if="tab.route !== undefined" />
```
kleinfreund, line 57: **Comment**: Technically, one `<router-view />` _replacing_ the tab content containers would be sufficient, but I decided to stick to the general approach of this component.

src/types/tabs.ts
```diff
@@ -1,4 +1,7 @@
+import type { RouteLocationNamedRaw } from 'vue-router'
+
 export interface Tab {
   hash: string
   title: string
+  route?: RouteLocationNamedRaw
```
kleinfreund, line 6: **Comment**: I generally use the named variant of `RouteLocationRaw` for such interfaces even though it limits the route styles you can use (i.e. disallowing path-based routes) because it makes working with the objects implementing `Tab` easier. They will always be able to access `tab.route.name` which is not possible with `RouteLocationRaw` (as it can be just a string or miss the `name` property from an object). I think this is fine because it’s best to keep routes addressed via path out of application code (i.e. paths should only be used in route definitions) as that allows for easier changes to an application’s routes (one only has to change them in the route definitions, all uses of `router.push`, etc. remain intact).

PR #1263, opened 2023-04-07
fix(inputs): support required state [khcp-6528]

kaiarrowood:
# Summary

**NOTE: This will require changes when adopting in `khcp-ui` to make sure we aren't showing duplicate asterisks anywhere!**
Adoption discussion in this [slack thread](https://kongstrong.slack.com/archives/CTAMPSTNX/p1681327020300829)

Add support for displaying an asterisk to indicate `required` state.
Addressed [KHCP-6528](https://konghq.atlassian.net/browse/KHCP-6528).

Support added for:
- KInput
- KLabel
- KMultiselect
- KSelect
- KTextarea

Doc updates:
- Added `Theming` section to `KLabel`
- Added `Attribute Bindings` section to `KTextarea`

image

## PR Checklist

[KHCP-6528]: https://konghq.atlassian.net/browse/KHCP-6528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/components/label.md
+++ b/docs/components/label.md
@@ -83,12 +83,12 @@ Use the `for` attribute to bind a label to an input element for accessibility.
 | :-------------------------------- | :------------------------------- |
 | `--KInputLabelColor`              | Label text color                 |
 | `--KLabelRequiredAsteriskColor`   | Label required '*' color         |
-| `--KInputLabelFont`               |                                  |
-| `--KInputLabelSize`               |                                  |
+| `--KInputLabelFont`               | Label font                       |
+| `--KInputLabelSize`               | Label text size                  |
 | `--KInputLabelWeight`             | Label font weight                |
 | `--KInputCheckboxLabel`           | Checkbox/radio label color       |
-| `--KInputCheckboxLabelFont`       |                                  |
-| `--KInputCheckboxLabelSize`       |                                  |
+| `--KInputCheckboxLabelFont`       | Checkbox/radio font              |
+| `--KInputCheckboxLabelSize`       | Checkbox/radio text size         |
 
 An example of theming the label might look like:
 
--- a/docs/components/label.md
+++ b/docs/components/label.md
@@ -38,6 +38,16 @@ Use the `info` prop to display information help text.
 <KLabel info="This is an example">Input Title</KLabel>
 ```
 
+### isRequired
+
+Use the `isRequired` prop to indicate requiredness of a field by displaying an `*` after the label.
+
+<KLabel is-required>Name</KLabel>
+
+```html
+<KLabel is-required>Name</KLabel>
+```
+
 ### tooltipAttributes
 
 Use the `tooltipAttributes` prop to configure the **KTooltip's** [props](/components/tooltip) if using the `help` or `info` props.
@@ -66,3 +76,42 @@ Use the `for` attribute to bind a label to an input element for accessibility.
 <KLabel for="service" help="A service is an API that you want to offer">Service Name</KLabel>
 <KInput id="service"/>
 ```
+
+## Theming
+
+| Variable                          | Purpose                          |
+| :-------------------------------- | :------------------------------- |
+| `--KInputLabelColor`              | Label text color                 |
+| `--KLabelRequiredAsteriskColor`   | Label required '*' color         |
+| `--KInputLabelFont`               |                                  |
+| `--KInputLabelSize`               |                                  |
+| `--KInputLabelWeight`             | Label font weight                |
+| `--KInputCheckboxLabel`           | Checkbox/radio label color       |
+| `--KInputCheckboxLabelFont`       |                                  |
+| `--KInputCheckboxLabelSize`       |                                  |
+
+An example of theming the label might look like:
+
+<KLabel for="service" is-required class=custom-label>Service Name</KLabel>
+<KInput id="service"/>
+
+```html
+<template>
+  <KLabel for="service" is-required class=custom-label>Service Name</KLabel>
+  <KInput id="service"/>
+</template>
+
+<style>
+.custom-label {
+  --KInputLabelColor: var(--purple-400);
+  --KLabelRequiredAsteriskColor: var(--red-500);
+}
+</style>
+```
+
+<style lang="scss">
+.custom-label {
+  --KInputLabelColor: var(--purple-400);
+  --KLabelRequiredAsteriskColor: var(--red-500);
+}
+</style>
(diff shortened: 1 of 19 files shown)


src/components/KInput/KInput.vue
```diff
@@ -218,9 +224,20 @@ export default defineComponent({
...
+      // default to length of currVal
+      let length = currValLength
+
+      // if there is a model value and it hasn't been modified yet, use that instead
+      if (!modelValueChanged.value && modelValLength) {
+        length = modelValLength
+      }
+
+      return !!props.characterLimit && length > props.characterLimit
+    })
```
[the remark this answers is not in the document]
kaiarrowood: It had a TS error so I expanded the logic since I couldn't figure out how to fix it when it was compressed into one line.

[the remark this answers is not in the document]
adamdehaven: sgtm

src/styles/forms/_inputs.scss
```diff
@@ -13,6 +13,13 @@
     transition: color 0.1s ease;
   }
 
+  .is-required {
+    color: var(--KLabelRequiredAsteriskColor, var(--KInputBorder));
```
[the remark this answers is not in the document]
kaiarrowood: I wasn't sure because the handling for inlined labels is different and lower in this file we are setting the label color to match the border color:

image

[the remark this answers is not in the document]
adamdehaven: Hmm... then maybe at least fallback to something like this?

```scss
var(--KLabelRequiredAsteriskColor, var(--KInputBorder, var(--grey-600)));
```

src/components/KLabel/KLabel.cy.ts
```diff
@@ -38,6 +38,20 @@ describe('KLabel', () => {
     cy.get('.k-input-label .label-tooltip').should('not.be.empty')
   })
 
+  it('renders an asterisk when `isRequired` is true', () => {
```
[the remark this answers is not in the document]
kaiarrowood: Covered in unit tests

docs/components/label.md
```diff
@@ -66,3 +76,42 @@ Use the `for` attribute to bind a label to an input element for accessibility.
...
+| Variable                          | Purpose                          |
+| :-------------------------------- | :------------------------------- |
+| `--KInputLabelColor`              | Label text color                 |
+| `--KLabelRequiredAsteriskColor`   | Label required '*' color         |
+| `--KInputLabelFont`               |                                  |
+| `--KInputLabelSize`               |                                  |
+| `--KInputLabelWeight`             | Label font weight                |
+| `--KInputCheckboxLabel`           | Checkbox/radio label color       |
+| `--KInputCheckboxLabelFont`       |                                  |
+| `--KInputCheckboxLabelSize`       |                                  |
```
adamdehaven, line 91: docs: Can you add "Purpose" text for all of the variables?

src/composables/useUtilities.ts
```diff
@@ -242,12 +242,33 @@ export default function useUtilities() {
...
+  /**
+   * For the label of a `required` input.
+   * If the field is required and the label ends with an '*', strip it out.
+   *
+   * @param label the label to strip
+   * @param required whether or not the field is required
+   * @returns the stripped label
+   */
+  const stripRequiredLabel = (label: string, required: boolean): string => {
+    if (!required) {
```
adamdehaven, line 254: question: do we need to just return an empty string if `label` is somehow undefined?

src/composables/useUtilities.ts
```diff
@@ -242,12 +242,33 @@ export default function useUtilities() {
...
+      return label
+    }
+
+    if (label.match(/( )?\*$/gi)) {
```
adamdehaven, line 258: You need to force a boolean here so that you don't pass with an empty array
```suggestion
    if (!!label.match(/( )?\*$/gi)) {
```

kaiarrowood: I tried it like this but I get a TS warning:

image

src/composables/useUtilities.ts
kaiarrowood: This is the kind of stuff where I get confused, if I don't make a param optional, wouldn't someone trying to pass me undefined get a TS error trying to pass a var that has the possibility of being undefined? Same question I think for the need to use the `!!` - if I explicitly type `label` to be `string` they couldn't pass an empty array in the first place.

How do you know when you need to guard with a conditional that verifies a required param was passed and that it was the correct type and when you can you rely on TS to do the guarding implicitly?

src/composables/useUtilities.ts
adamdehaven: Ah do this then instead
```
Boolean(label.match(/( )?\*$/gi))
```

src/composables/useUtilities.ts
adamdehaven: So yes, they would get a TS error at compile time (or type check) but we still want to prevent the app from erroring at runtime, just in case.

kaiarrowood: So, is it like, if you are writing logic that is in a component that can be consumed by another app, you need to explicitly guard yourself. But if you're the top level app, you can rely on TS guarding? Or do you always have to explicitly guard? And if you always have to explicitly guard what's the point of using TS 😅

src/composables/useUtilities.ts
kaiarrowood: heh, nope, can't get around it that way either:

image

src/composables/useUtilities.ts
adamdehaven: It's more of a case-by-case type of thing.

Example: you write a function that accepts a single `object` parameter, but it's used all over the place in lots of different ways: sometimes the object is dynamic, originating from API response data or user input. In this scenario, maybe you can't always ensure that there isn't a way for an invalid value to be passed in. In that case, you'd include a check in the function to exit or return early in case of an invalid parameter in order to prevent it from throwing an error, and to avoid having to wrap in a try/catch, etc.

src/composables/useUtilities.ts
adamdehaven: lol ok then let's do this:

```ts
/( )?\*$/gi.test(label)
```

kaiarrowood: image

adamdehaven (commented):
image

Should the asterisk be blue like the label text in the overlay label when focused?

adamdehaven (approved):
lgtm

adamdehaven (approved):
lgtm

Mierenga (approved):

PR #1284, opened 2023-04-13
feat(KEmptyState): switch to setup syntax [KHCP-6501]

DariaYeremina:
# Summary

https://konghq.atlassian.net/browse/KHCP-6501
Switch to setup syntax

## PR Checklist

### DIFF
--- a/src/components/KEmptyState/KEmptyState.vue
+++ b/src/components/KEmptyState/KEmptyState.vue
@@ -40,43 +40,39 @@
   </section>
 </template>
 
-<script lang="ts">
-import { defineComponent } from 'vue'
+<script lang="ts" setup>
 import KButton from '@/components/KButton/KButton.vue'
 import KIcon from '@/components/KIcon/KIcon.vue'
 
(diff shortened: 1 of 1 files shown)


src/components/KEmptyState/KEmptyState.vue
```diff
@@ -40,43 +40,39 @@
...
-    handleClick: {
-      type: Function,
-      default: null,
-    },
-    iconColor: {
-      type: String,
-      default: '',
-    },
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const props = defineProps({
```
adamdehaven, line 48: `defineProps` is a compiler macro, so if you don't need to access the props in the setup function, you don't need to set it to a const

```suggestion
defineProps({
```

DariaYeremina: fixed

adamdehaven (approved):
lgtm

kaiarrowood (approved):

PR #1286, opened 2023-04-14
feat(KSegmentedControl): switch to setup syntax [KHCP-6514]

DariaYeremina:
# Summary

https://konghq.atlassian.net/browse/KHCP-6514
Switch to setup syntax

## PR Checklist


src/components/KSegmentedControl/KSegmentedControl.vue
```diff
@@ -24,102 +24,52 @@
   </div>
 </template>
 
-<script lang="ts">
-import { defineComponent, ref, PropType } from 'vue'
+<script lang="ts" setup>
+import { ref, PropType } from 'vue'
 import KButton from '@/components/KButton/KButton.vue'
+import { SegmentedControlOption } from '@/types/segmented-control'
+// need to be imported as const from separate file due to an error "defineProps are referencing locally declared variables"
```
kaiarrowood, line 31: You can fix this by including the items in a separate `<script lang="ts">` block (no `setup`) like we are doing here: https://github.com/Kong/kongponents/blob/main/src/components/KMultiselect/KMultiselect.vue#L244

If you have the non-`setup` `script` block BEFORE the `setup` block you need to move all `import` statements into the non-`setup` block. If you have the non-`setup` block AFTER the `setup` block you only have to move the functions being used in `defineProps`.

DariaYeremina: then I got another error with duplicated import of type

DariaYeremina: updated

src/components/KSegmentedControl/KSegmentedControl.vue
```diff
@@ -71,55 +66,50 @@ const validateItems = (items: SegmentedControlOption[] | string[]): boolean => {
...
-    },
-    isDisabled: {
-      type: Boolean,
-      default: false,
-    },
-    allowPointerEvents: {
-      type: Boolean,
-      default: false,
-    },
+export default {}
```
kaiarrowood, line 69: nit: I think you can remove this line

kaiarrowood (approved):
Looking good 🚀

PR #1299, opened 2023-04-19
fix(KTable): loosen types passed to slots in KTable

adorack:
# Summary

Return a typeof `any` for `row` and `column` slot props to prevent having to explicitly typecast in host applications. Also adds a column slot name type interface to properly resolve the slots in the template along with the corresponding slot props.

## PR Checklist


kaiarrowood (approved):
LGTM! 🚀

PR #1312, opened 2023-04-21
feat(ktable): add options for table column header icons

dustinryerson:
# Summary

Adds 7 new options to add and customize icons in table column headers. I did my best to make the options descriptive and specific so that there was as little ambiguity as possible.
Screenshot 2023-04-24 at 11 01 33 AM

## Screenshots
Screenshot 2023-04-24 at 10 53 30 AM
Screenshot 2023-04-24 at 11 03 02 AM

## PR Checklist


adamdehaven (commented):
@dustinryerson would it be possible to just allow slotting in an icon into the table header instead of defining all these props, etc.? Then the icon would still be controlled via the host app, but would prevent deepening the dependence/integration of KIcons into KTable.

dustinryerson:
> @dustinryerson would it be possible to just allow slotting in an icon into the table header instead of defining all these props, etc.? Then the icon would still be controlled via the host app, but would prevent deepening the dependence/integration of KIcons into KTable.

@adamdehaven yep you're right, I think that's the best route. I'll close this. Thanks!

PR #1317, opened 2023-04-24
feat(KSelect, KStep, KSkelton, KSlideout): improve types [KHCP-6972]

DariaYeremina:
# Summary

https://konghq.atlassian.net/browse/KHCP-6972
Improve ts types

## PR Checklist


src/components/KStepper/KStepState.vue
```diff
@@ -1,33 +1,46 @@
...
+    case 'error':
+      c = KErrorState
+      break
+    case 'pending':
+      c = KPendingState
+      break
+  }
+
+  return c
+})
```
mptap, line 43: suggestion: could we use a more descriptive variable name instead of 1 letter?

src/types/skeleton.ts
```diff
@@ -0,0 +1,21 @@
...
+  'form',
+  'spinner',
+  'fullscreen-kong',
+  'fullscreen-generic',
+  '', // default Skeleton.vue
+] as const
+
+export type SkeletonType = AnyElementOf<typeof SkeletonTypeArray>
+
+export const SkeletonBoxWidthArray = ['1', '2', '5', '6', '10', '25', '50', '75', '100', ''] as const
```
mptap, line 15: Shouldn't the `SkeletonBoxWidthArray` NOT have empty string `''`? What is the implication when `SkeletonBoxWidth` is `''`?

src/types/stepper.ts
```diff
@@ -1,6 +1,10 @@
-export type StepperState = '' | 'active' | 'default' | 'pending' | 'completed' | 'error'
+import { AnyElementOf } from '@/types/utils'
+
+export const StepperStateArray = ['active', 'default', 'pending', 'completed', 'error'] as const
```
mptap, line 3: If we're following the pattern of the type array containing the empty string as an element, then we could have
`export const StepperStateArray = ['', 'active', 'default', 'pending', 'completed', 'error'] as const`
and `state?: StepperState`

src/components/KStepper/KStep.vue
```diff
@@ -37,9 +37,9 @@ const props = defineProps({
     required: true,
   },
   state: {
-    type: String as PropType<StepperState>,
+    type: String as PropType<StepperState | ''>,
```
mptap, line 40: If we have `''` empty string as an element in the `StepperStateArray` then we wouldn't have to specify it as the fallback here.

src/components/KStepper/KStepState.vue
```diff
@@ -1,33 +1,46 @@
...
-  components: { KActiveState, KCompletedState, KDefaultState, KErrorState, KPendingState },
-  props: {
-    state: {
-      type: String as PropType<StepperState>,
-      default: 'default',
-      validator: (value: StepperState) => ['default', 'active', 'pending', 'completed', 'error'].includes(value),
-    },
+const props = defineProps({
+  state: {
+    type: String as PropType<StepperState | ''>,
```
mptap, line 18: If we have `''` empty string as an element in the `StepperStateArray` then we wouldn't have to specify it as the fallback here.

src/types/select.ts
```diff
@@ -0,0 +1,25 @@
+import { AnyElementOf } from '@/types/utils'
+
+export interface SelectItem {
```
kaiarrowood, line 3: question: Can this interface extend `Record<string, any>`? Either that or can we have a `record` prop (optional) that can contain the full object data (we would also want this for MultiselectItem). Sometimes there is a need to reference a property that is being used for the `label` or the `value`.

src/components/KStepper/KStepState.vue
kaiarrowood: I don't think this is right, we don't support `''` as a value for `state` why do we need this? I would think they either pass a valid `StepperState` or leave it `undefined`

src/types/stepper.ts
kaiarrowood: Again I don't think we should support `''`, that isn't a valid state

src/types/stepper.ts
```diff
@@ -1,6 +1,10 @@
...
+
+export type StepperState = AnyElementOf<typeof StepperStateArray>
 
 export interface StepItem {
   label: string
-  state?: StepperState
+  state?: StepperState | ''
```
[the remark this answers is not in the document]
kaiarrowood: See above

src/types/skeleton.ts
kaiarrowood: Same question, I don't think we should support `''` as a value, we don't handle it.

src/types/select.ts
```diff
@@ -0,0 +1,25 @@
...
+  query: string
+}
+
+export interface SelectItemWithGroup extends SelectItem {
+  group: string
+}
+
+export type SelectDropdownFooterTextPosition = 'sticky' | 'static'
+
+export const SelectAppearanceArray = ['dropdown', 'select', 'button'] as const
```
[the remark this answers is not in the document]
DariaYeremina: `as const` at the end makes this variable readonly

src/components/KStepper/KStepState.vue
DariaYeremina: for me it was obvious, but I changed it

src/types/skeleton.ts
DariaYeremina: `''` was used in components for `width`, changed it

src/types/stepper.ts
DariaYeremina: removed

src/components/KStepper/KStep.vue
DariaYeremina: removed

src/types/stepper.ts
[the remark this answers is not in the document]
DariaYeremina: removed

kaiarrowood (approved):
LGTM! 🚀

PR #1359, opened 2023-05-08
fix(kpagination): change order so background color styling is applied

davidma415:
# Summary
Order needs to be adjusted so that the styling from the CSS var is actually being used. Otherwise in its current order, the default color overrides it even if it is set.

## PR Checklist


adamdehaven (approved):
LGTM

adamdehaven (approved):

PR #1380, opened 2023-05-16
feat(kcodeblock): adds secondary actions slot

kleinfreund:
# Summary

Adds an optional `secondary-actions` slot to `KCodeBlock`. It allows adding elements **after** the copy button (if it’s shown) or in its place (if it’s not shown).

## PR Checklist

### DIFF
--- a/docs/components/codeblock.md
+++ b/docs/components/codeblock.md
@@ -261,7 +261,6 @@ Allows adding elements after the location of the copy button.
     id="code-block-slots-secondary-actions"
     :code="code"
     language="json"
-    style="--KCodeBlockMaxHeight: 200px"
   >
     <template #secondary-actions>
       <KButton
--- a/docs/components/codeblock.md
+++ b/docs/components/codeblock.md
@@ -250,6 +250,65 @@ Sets the display theme of the component.
 />
 ```
 
+## Slots
+
+### secondary-actions
+
+Allows adding elements after the location of the copy button.
+
+<ClientOnly>
+  <KCodeBlock
+    id="code-block-slots-secondary-actions"
+    :code="code"
+    language="json"
+    style="--KCodeBlockMaxHeight: 200px"
+  >
+    <template #secondary-actions>
+      <KButton
+        appearance="outline"
+        size="small"
+        :is-rounded="false"
+        title="Copy as Kubernetes"
+      >
+        <KIcon
+          color="currentColor"
+          icon="copy"
+          size="18"
+          title="Copy as Kubernetes"
+        />
+        as k8s
+      </KButton>
+    </template>
+  </KCodeBlock>
+</ClientOnly>
+
+```html
+<KCodeBlock
+  id="code-block-slots-secondary-actions"
+  :code="code"
+  language="json"
+>
+  <template #secondary-actions>
+    <KButton
+      appearance="outline"
+      size="small"
+      :is-rounded="false"
+      title="Copy as Kubernetes"
+      @click="copyAsKubernetes"
+    >
+      <KIcon
+        color="currentColor"
+        icon="copy"
+        size="18"
+        title="Copy as Kubernetes"
+      />
+
+      as k8s
+    </KButton>
+  </template>
+</KCodeBlock>
+```
+
 ## Events
 
 ### code-block-render
--- a/src/components/KCodeBlock/KCodeBlock.vue
+++ b/src/components/KCodeBlock/KCodeBlock.vue
@@ -1148,67 +1148,69 @@ $dark-focusColor: var(--green-500, color(green-500));
 @import '@/styles/variables';
 @import '@/styles/functions';
 
-.k-matched-term {
-  color: var(--teal-500, color(teal-500));
-  font-weight: 900;
-}
-
-.theme-dark .k-matched-term {
-  color: var(--green-500, color(green-500));
-}
-
-.k-code-block .k-button.small {
-  padding-left: var(--spacing-xs, spacing(xs));
-  padding-right: var(--spacing-xs, spacing(xs));
-}
-
-.k-code-block .kong-icon {
-  align-items: center;
-  display: inline-flex;
-  justify-content: center;
-}
-
-// TODO: KButton theming should live in KButton.vue.
-// TODO: Fix these styles not always providing a solid background color for the copy button allowing content to clip through it.
-.k-code-block-secondary-actions .k-button:not(.increase-specificity) {
-  @media (min-width: $viewport-md) {
-    background-color: transparent;
-    border-color: transparent;
+.k-code-block {
+  .k-matched-term {
+    color: var(--teal-500, color(teal-500));
+    font-weight: 900;
   }
 
-  &:hover {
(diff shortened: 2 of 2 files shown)


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -1195,4 +1167,48 @@ $dark-focusColor: var(--green-500, color(green-500));
   display: inline-flex;
   justify-content: center;
 }
+
+// TODO: KButton theming should live in KButton.vue.
+// TODO: Fix these styles not always providing a solid background color for the copy button allowing content to clip through it.
+.k-code-block-secondary-actions .k-button:not(.increase-specificity) {
```
kleinfreund, line 1173: **Comment**: I’ve moved these styles out of the scoped style block so they affect the content put into the secondary-actions slot (which is likely going to be `KButton` instances).

kaiarrowood: suggestion: for readability, and easy to verify all unscoped styles are properly scoped could you put ALL styles in the unscoped block under `.k-code-block`?

ex.

```
<style lang="scss">
@import '@/styles/variables';
@import '@/styles/functions';

.k-code-block {
 // ... all styles
}
</style>
```

kaiarrowood (commented):
Just a few suggestions, otherwise this is looking good!

docs/components/codeblock.md
```diff
@@ -250,6 +250,65 @@ Sets the display theme of the component.
...
+### secondary-actions
+
+Allows adding elements after the location of the copy button.
+
+<ClientOnly>
+  <KCodeBlock
+    id="code-block-slots-secondary-actions"
+    :code="code"
+    language="json"
+    style="--KCodeBlockMaxHeight: 200px"
```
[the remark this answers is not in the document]
kleinfreund: This was accidentally copied from the example for `props.showLineNumberLinks` where it is already used. I set this property only on one code block to be able to see the difference between a code block with and without scrollbars in the docs site.

I’m removing it from the example I added.

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -1195,4 +1167,48 @@ $dark-focusColor: var(--green-500, color(green-500));
...
+  &:hover:active {
+    background-color: var(--steel-500, color(steel-500));
+    border-color: var(--steel-500, color(steel-500));
+    color: #fff;
+  }
+}
+
+// TODO: KButton theming should live in KButton.vue.
+// TODO: Fix these styles not always providing a solid background color for the copy button allowing content to clip through it.
+.theme-dark .k-code-block-secondary-actions .k-button:not(.increase-specificity) {
```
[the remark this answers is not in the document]
kleinfreund: I’m not sure I’m following your meaning here. I think it’s a misunderstanding, perhaps. These styles are related to the component’s `props.theme` support which I believe is used in KHCP UI and should definitely be part of Kongponents as it’s shipped and not the Kongponents docs site CSS.

kleinfreund:
> suggestion: for readability, and easy to verify all unscoped styles are properly scoped could you put ALL styles in the unscoped block under .k-code-block?

@kaiarrowood I’m not sure that’s appropriate to do as part of this PR.

Generally, components in Kongponents use scoped style blocks so that’s what I’m sticking to by default. It’s just that some styles in KCodeBlock can’t be scoped for various reasons (e.g. when applying styles to elements slotted in from the outside).

However, I do actually think that a component library should generally **not** use scoped styles and rather rely solely on using properly prefixed class names (like we’re already doing). This practice has the advantage of all selectors having lower specificity and are therefore easier to override in a consuming application:

```css
/* specificity 0 2 0 */
.k-code-block[v-scoped-style-attribute]

/* specificity 0 1 0 */
.k-code-block
```

The reason why I would avoid doing this in this PR is that because of the change to the selectors’ specificity, this is a potentially breaking change: Selectors that previously had a specificity of "0 2 0" will now have "0 1 0" and depending on a consuming project’s stylesheet order, a CSS rule set with a selector like `.k-button` would suddenly start overriding rule set in this component with the same selector.

src/components/KCodeBlock/KCodeBlock.vue
kleinfreund: > suggestion: for readability, and easy to verify all unscoped styles are properly scoped could you put ALL styles in the unscoped block under .k-code-block?

@kaiarrowood I’m not sure that’s appropriate to do as part of this PR.

Generally, components in Kongponents use scoped style blocks so that’s what I’m sticking to by default. It’s just that some styles in KCodeBlock can’t be scoped for various reasons (e.g. when applying styles to elements slotted in from the outside).

However, I do actually think that a component library should generally **not** use scoped styles and rather rely solely on using properly prefixed class names (like we’re already doing). This practice has the advantage of all selectors having lower specificity and are therefore easier to override in a consuming application:

```css
/* specificity 0 2 0 */
.k-code-block[v-scoped-style-attribute]

/* specificity 0 1 0 */
.k-code-block
```

The reason why I would avoid doing this in this PR is that because of the change to the selectors’ specificity, this is a potentially breaking change: Selectors that previously had a specificity of "0 2 0" will now have "0 1 0" and depending on a consuming project’s stylesheet order, a CSS rule set with a selector like `.k-button` would suddenly start overriding rule set in this component with the same selector.

src/components/KCodeBlock/KCodeBlock.vue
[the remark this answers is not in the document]
kaiarrowood: Oh! My bad, I got that confused with vitepress' dark mode. I completely forgot this component had it's own theming support.

src/components/KCodeBlock/KCodeBlock.vue
kaiarrowood: (Granted, I've had to reread your comment a few times to make sure I understand what you're saying here - specificity is a complex thing. So apologies if I misunderstood AGAIN). Are you saying you don't want to increase the specificity of the unscoped styles by adding another selector in? ie. `.k-code-block .k-code-block-secondary-actions` is harder to override than `.k-code-block-secondary-actions`

kleinfreund: Oh, of course, that makes more sense now. I misunderstood you. I’ve made this change now.

---

> Are you saying you don't want to increase the specificity of the unscoped styles by adding another selector in? ie. .k-code-block .k-code-block-secondary-actions is harder to override than .k-code-block-secondary-actions

Yes, that’s more or less accurate, but is only really of value in a wider sense where it’s library standard.

Ideally, those two things are true:

- Selectors are written in such a way that conflicts are unlikely (i.e. prefixed)
- Component styles use unscoped style blocks

Then, you have both robust styles (sure, not quite as robust as with scoping) _and_ as low as possible specificity which makes it easier to override unwanted styles in a consuming project (while surely everyone would like to avoid the need for that, it just happens).

That being said, I certainly don’t want to propose this as part of this PR. 😁

kaiarrowood (approved):
LGTM! Nice work 🚀

PR #1399, opened 2023-05-23
fix(kradio): using !important on inline-block

kleinfreund:
# Summary

Changes the `.k-radio` element to use `display: inline-block` instead of `display: inline-block !important` so it can be more easily overridden.

See also https://github.com/Kong/kongponents/pull/1381.
## PR Checklist


kleinfreund:
**Comment**: The "Publish Previews" step in the CI will fail whenever you need to re-run "Run Component Tests" because by the second run, there will already be a published version for the current commit. This step should really be skipped in that case.

adamdehaven (approved):

adamdehaven (approved):

PR #1400, opened 2023-05-23
ci: skips preview publishing if version already exists

kleinfreund:
# Summary

Skips trying to publish a preview package if a package with the same version already exists. This is necessary in order to be able to re-run flaky tests.

## Notes

Someone with better Shell foo should be looking over this

## PR Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,11 +73,7 @@ jobs:
 
           yarn version --prerelease --preid ${preid} --allow-branch ${{ github.head_ref }} --no-git-tag-version --yes --amend
 
-          package_version=$(cat package.json \
-            | grep version \
-            | head -1 \
-            | awk -F: '{ print $2 }' \
-            | sed 's/[", ]//g')
+          package_version=$(jq -r ".version" package.json)
           package=@kong/kongponents@"${package_version}"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -71,11 +71,24 @@ jobs:
...
+            | grep version \
+            | head -1 \
+            | awk -F: '{ print $2 }' \
+            | sed 's/[", ]//g')
+          package=@kong/kongponents@"${package_version}"
+
+          npm show "${package}" >/dev/null 2>&1 && npm_show_status=0 || npm_show_status=1
+          if [ $npm_show_status -eq 0 ]; then
+            echo "Package ${package} is already published. Skipping publishing."
+            exit 0
```
adamdehaven, line 86: We don't want to exit here without returning success as this will also prevent the tests from running.

kleinfreund: Are you sure? Shouldn’t this just exit this step with a successful status?

adamdehaven: Yep. 0. Ignore this 🤦🏼‍♂️

ValeryG (changes requested):

ValeryG (approved):

adamdehaven (approved):
lgtm

PR #1405, opened 2023-05-24
feat(kcatalog,ktable): emit state

kaiarrowood:
# Summary

In order to support caching correctly in Konnect, we have some cases where content that is external to `KTable` (or `KCatalog`) needs to know the table's state. To handle this scenario, I've added a new emit that will allow tracking the state.

image

## PR Checklist


adamdehaven:
> In order to support caching correctly in Konnect, we have some cases where content that is external to KTable (or KCatalog) needs to know the table's state

I’m trying to understand the context. Would you mind expanding with an example?

src/components/KTable/KTable.vue
```diff
@@ -501,7 +501,15 @@ const props = defineProps({
...
 
-const emit = defineEmits(['sort', 'ktable-error-cta-clicked', 'ktable-empty-state-cta-clicked', 'row-click', 'cell-click', 'update:table-preferences'])
+const emit = defineEmits<{
+  (e: 'cell-click', value: { data: any }): void
+  (e: 'row-click', value: { data: any }): void
+  (e: 'ktable-error-cta-clicked'): void
+  (e: 'ktable-empty-state-cta-clicked'): void
+  (e: 'update:table-preferences', preferences: TablePreferences): void
+  (e: 'sort', value: { prevKey: string, sortColumnKey: string, sortColumnOrder: string }): void
+  (e: 'state', value: { state: string, hasData: boolean }): void
```
adamdehaven, line 511: suggestion: if the strings state can emit are restricted, can we better type this instead of `string`?

adamdehaven (commented):
Since we’re already accessing the swrv state in the component, instead of emitting two separate properties for state and hasData, could we just emit the active swrv state which alone indicates both the state of the table as well as if there is data or not?

mptap (approved):
lgtm

adamdehaven:
> The primary example is the Service Hub page which shares the search input / action button between an instance of KCatalog and an instance of KTable.

Couldn't each component have it's own search input/action button? I'm assuming I'm still missing context but if I'm understanding, I wouldn't think we should modify the underlying component logic to comply with a layout implementation 🤔 

> I don't want to give people opportunities to associate the wrong swrvState with the loading state of the table. It's about more than providing hasData, we are also indicating the state of the table which has special logic around the swrvState to determine whether or not to set the table into loading.

So the host app just needs to know if the table is loading? Or the table has data _and_ is in a loading state?

adamdehaven:
_(Adding a comment from our discussion earlier)_

Since we're trying to cover a single use-case within Konnect, IMO it doesn't make sense to update the component logic/functionality to satisfy a single design decision

src/types/table.ts
```diff
@@ -27,7 +27,10 @@ export type TableColumnSlotName = `column-${string}`
 
 export type SwrvState = 'VALIDATING' | 'VALIDATING_HAS_DATA' | 'PENDING' | 'SUCCESS' | 'SUCCESS_HAS_DATA' | 'ERROR' | 'STALE_IF_ERROR'
 
+export type EmitState = 'loading' | 'error' | 'success'
```
adamdehaven, line 30: Can you please rename this to `TableState` so that it's scoped to the component? 

Also, don't you need to add types for KCatalog as well?

src/types/table.ts
```diff
@@ -27,7 +27,10 @@ export type TableColumnSlotName = `column-${string}`
...
+
 export interface SwrvStateData {
   hasData: boolean
-  state: SwrvState
+  state: SwrvState,
+  emitState: EmitState
```
adamdehaven, line 35: question: why is `emitState` now inside of this `SwrvStateData`? This state has nothing to do with `swrv`

src/components/KTable/KTable.vue
```diff
@@ -501,7 +501,15 @@ const props = defineProps({
...
 })
 
-const emit = defineEmits(['sort', 'ktable-error-cta-clicked', 'ktable-empty-state-cta-clicked', 'row-click', 'cell-click', 'update:table-preferences'])
+const emit = defineEmits<{
+  (e: 'cell-click', value: { data: any }): void
+  (e: 'row-click', value: { data: any }): void
+  (e: 'ktable-error-cta-clicked'): void
+  (e: 'ktable-empty-state-cta-clicked'): void
+  (e: 'update:table-preferences', preferences: TablePreferences): void
+  (e: 'sort', value: { prevKey: string, sortColumnKey: string, sortColumnOrder: string }): void
```
adamdehaven, line 510: These changes don't seem to be related to this PR?

src/components/KCatalog/KCatalog.vue
```diff
@@ -425,7 +425,12 @@ const props = defineProps({
   },
 })
 
-const emit = defineEmits(['kcatalog-error-cta-clicked', 'kcatalog-empty-state-cta-clicked', 'update:catalog-preferences'])
+const emit = defineEmits<{
+  (e: 'kcatalog-error-cta-clicked'): void
+  (e: 'kcatalog-empty-state-cta-clicked'): void
+  (e: 'update:catalog-preferences', preferences: CatalogPreferences): void
```
adamdehaven, line 431: These don't seem related

src/types/table.ts
kaiarrowood: I'll do the rename, but your second question is something I wanted to ask about. The way we do the imports from `/types` we aren't allowed to have any name duplication for the `type`s we define, since both `SwrvState` and `EmitState` are an exact match between the 2 components is there a need to define them twice with slightly different names?

adamdehaven: It depends if they always match really. Since we know we're going to refactor KTable at some point prob makes sense to name them `TableSwrvState` etc.

src/components/KCatalog/KCatalog.vue
kaiarrowood: They aren't, this improves the ts

src/types/table.ts
kaiarrowood: Switched to `TableState`

jillztom (approved):

PR #1418, opened 2023-05-29
feat(KToaster): switch to setup syntax [KHCP-6522]

DariaYeremina:
# Summary

https://konghq.atlassian.net/browse/KHCP-6522
Switch to setup syntax

## PR Checklist


src/types/toaster.ts
```diff
@@ -0,0 +1,12 @@
+import { AlertAppearance, AlertAppearanceRecord } from '@/types/alert'
+import { appearances } from '@/components/KAlert/KAlert.vue'
+
+export interface Toast {
+  key?: any // unique identifier of toaster
+  appearance?: AlertAppearance
+  message: string // Text to display in toaster
+  timer?: any
+  timeoutMilliseconds?: number
```
adamdehaven, line 9: Based on the usage [here](https://github.com/Kong/kongponents/blob/f29961b2f862926c71e9d7158e1dbf05a02e0284/src/components/KToaster/ToastManager.ts#L50-L67) shouldn't some of these properties be required? e.g. `timeoutMilliseconds` and `message`

DariaYeremina: `message` is required.
`timeoutMilliseconds ` - I don't think it should be required, because there's a default value for it 
```
timeoutMilliseconds: timeoutMilliseconds || this.timeout,
```

adamdehaven: Yep let's update the message prop -- we'll need to make sure that doesn't throw errors when imported into khcp-ui

PR #1423, opened 2023-05-30
feat(kdropdownmenu): add slot prop method to close dropdown [KCGW-84]

portikM:
# Summary

Ticket: https://konghq.atlassian.net/browse/KCGW-84

Adds slot prop method `closeDropdown` to close dropdown programmatically.

Screen Shot 2023-05-30 at 1 54 07 PM

## PR Checklist


src/components/KDropdownMenu/KDropdownMenu.vue
```diff
@@ -144,6 +146,8 @@ const emit = defineEmits<{
   (e: 'change', value: DropdownItem): void;
 }>()
 
+const hidePopover = ref<true | undefined>()
```
adamdehaven, line 149: question: why would we type to `true | undefined`? If it's not set to true, it's going to eval the same as false in the `KPop` component right?

src/components/KDropdownMenu/KDropdownMenu.vue
```diff
@@ -170,6 +174,16 @@ const handleSelection = (item: DropdownItem): void => {
...
 }
 
+const handleCloseDropdown = (): void => {
+  hidePopover.value = true
+
+  // reset the hidePopover value so it's ready for the next time the dropdown is opened
+  // need nextTick to ensure the popover is hidden before resetting the value
+  nextTick(() => {
+    hidePopover.value = undefined
+  })
```
adamdehaven, line 184: issue: This `nextTick` isn't doing anything unless you `await`

What's the purpose here?

adamdehaven (changes requested):

src/components/KDropdownMenu/KDropdownMenu.vue
portikM: Good call. The purpose is to reset it to initial value. Even though it clearly worked without `await`, still kept the `nextTick` and added the `await` to avid possibility of race condition.

docs/components/dropdown-menu.md
```diff
@@ -438,13 +481,43 @@ There are 3 primary item types:
...
-| `@click` | Fires when a `button` type menu item is clicked |
-| `@change` | Fires when items within a `selectionMenu` are clicked; returns the selected menu item object or `null` |
-| `@toggleDropdown` | Fires when the button to toggle the menu is clicked; returns `true` if the menu is open, or `false` |
+| Event             | Description                                                                                                              |
+| :---------------- | :----------------------------------------------------------------------------------------------------------------------- |
+| `@click`          | Fires when a `button` type menu item is clicked                                                                          |
+| `@change`         | Fires when items within a `selectionMenu` are clicked; returns the selected menu item object or `null`                   |
+| `@toggleDropdown` | Fires when the button to toggle the menu is clicked; returns `true` if the menu is open, or `false`                      |
+| `@action:click`   | Emitted only by by the `KDropdownItem` component. Fires when content passed through [`action`](#action) slot is clicked. |
```
adamdehaven, line 520: question: why emit a namespaced event when there's nothing else going on? Can't this just be `@click` or `@dropdown-item-click`?

src/components/KDropdownMenu/KDropdownMenu.vue
adamdehaven: I'm still not sure I understand. You're essentially creating what appears to be a boolean  behave like a non-boolean (non-native) type. I think a regular boolean would be more expected

docs/components/dropdown-menu.md
portikM: Because `@click` is already emitted by `KDropdownItem`. I think `action:click` is the most intuitive for click on `action` slot **specifically**.
```
<KDropdownItem
  @click="clickHandler"
  @action:click="closeDropdown"
>...
```

adamdehaven:
Also have a question on the related ticket here. 

image

What is the purpose of the trash icon? The screenshot shows a Multiselect component but we're editing KDropdown components in this PR?

src/components/KDropdownMenu/KDropdownMenu.vue
portikM: Think of it like this: boolean means `hidePopover.value` could be either `=== true` or `=== false`, meanwhile in our component we only want it to be `=== true` or undefined.

adamdehaven: I understand what you're saying; however, I don't understand why `false` is invalid here for either component.

portikM: Changed to `boolean`

portikM:
> Also have a question on the related ticket here.
> 
> image
> 
> What is the purpose of the trash icon? The screenshot shows a Multiselect component but we're editing KDropdown components in this PR?

Addressed in a separate thread.

src/components/KDropdownMenu/KDropdownItem.vue
```diff
@@ -158,11 +158,11 @@ const availableComponents = computed((): DropdownItemRenderedRecord => ({
 li.k-dropdown-item {
   align-items: center;
   display: flex;
-  font-size: 16px;
+  font-size: var(--type-md);
```
adamdehaven, line 161: Can we provide fallback values when using CSS variables?

src/components/KDropdownMenu/KDropdownMenu.vue
```diff
@@ -170,6 +174,16 @@ const handleSelection = (item: DropdownItem): void => {
   selectedItem.value = item
 }
 
+const handleCloseDropdown = async (): Promise<void> => {
+  hidePopover.value = true
+
+  // reset the hidePopover value so it's ready for the next time the dropdown is opened
+  // need nextTick to ensure the popover is hidden before resetting the value
+  await nextTick(() => {
```
adamdehaven, line 182: Even though you’re now awaiting this promise, when called from the event emit it will still not actually await before moving on. 

This may not matter if it appears to work, but pointing out that this won’t actually be called asynchronously.

portikM: It appears to work, should be fine

src/components/KDropdownMenu/KDropdownItem.vue
adamdehaven: 😢  CSS variables must _always_ have a fallback value. If you want to utilize the Sass equivalent, you can do this:
```scss
font-size: var(--type-md, type(md));
```

src/components/KDropdownMenu/KDropdownItem.vue
```diff
@@ -158,17 +158,17 @@ const availableComponents = computed((): DropdownItemRenderedRecord => ({
 li.k-dropdown-item {
   align-items: center;
   display: flex;
-  font-size: 16px;
+  font-size: var(--type-md, 16px);
   line-height: 1;
 
   &:not(:first-of-type).has-divider {
-    $k-dropdown-item-divider-container-height: 24px; // set to the same value as --spacing-lg without the units
+    $k-dropdown-item-divider-container-height: var(--spacing-lg, 24px); // set to the same value as --spacing-lg without the units
```
adamdehaven, line 165: Can you use the spacing Sass utils here?
```suggestion
    $k-dropdown-item-divider-container-height: var(--spacing-lg, spacing(lg); // set to the same value as --spacing-lg without the units
```

Same comment for others

portikM: Done

PR #1426, opened 2023-06-01
feat(ktabs): hide panels [KHCP-7542]

adamdehaven:
# Summary

Add a new `hasPanels` prop to `KTabs` in order to only render the tab controls, without the corresponding panels.

## PR Checklist


kleinfreund (dismissed):

kleinfreund (approved):

johncowen (approved):
I tried this out in the sandbox, in the docs and in place and it looks good to me, means we can remove a lot of custom CSS and just use native KTabs 🎉 

Thanks for sorting this out and going the extra few miles with all the docs etc 🙌

PR #1427, opened 2023-06-02
fix(kpagination): kpopAttrs not being defined but used

kleinfreund:
# Summary

**chore: ensures pagesVisible is list of number**

Changes `getVisiblePages` to return `[]` when `props.disablePageJump` is true instead of `0` allowing the function’s return type to be `number[]` instead of `number | number[]` which is likely unintentional seeing as `pageVisible` is used in a `v-for` directive. This silences the TypeScript error shown in the template related to this.

**fix(kpagination): kpopAttrs not being defined but used**

Adds `kpopAttrs` again which was lost in a previous refactoring. This ensures `placement: top` is passed to `KSelect` and silences the Vue warning appearing during development.

## PR Checklist

### DIFF
--- a/src/components/KPagination/KPagination.vue
+++ b/src/components/KPagination/KPagination.vue
@@ -207,9 +207,9 @@ const pageSizeOptions = props.pageSizes.map((size, i) => ({
 }))
 const pageSizeText = ref('')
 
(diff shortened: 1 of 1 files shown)


kaiarrowood (approved):
LGTM! 🚀  Verified KPagination and KTable continue to function as expected.

PR #1428, opened 2023-06-02
fix(kmultiselect): width calc

kaiarrowood:
# Summary

Make it reactive, so it sizes correctly if the multiselect starts off hidden with no width (ex. if it is inside a `KCollapse`).

## PR Checklist


src/components/KMultiselect/KMultiselect.vue
```diff
@@ -905,9 +905,13 @@ watch(() => props.items, (newValue, oldValue) => {
...
 const numericWidth = ref<number>(300)
-onMounted(() => {
-  // @ts-ignore
+const setNumericWidth = (): void => {
   numericWidth.value = multiselectRef.value?.clientWidth || 300
+}
+
+const resizeObserver = ref()
+onMounted(() => {
+  resizeObserver.value = new ResizeObserver(setNumericWidth).observe(multiselectRef.value as HTMLDivElement)
```
DariaYeremina, line 914: suggestion: It's better to add `unobserve` on `onUnmounted` to prevent listeners after component is onUnmounted.

PR #1439, opened 2023-06-07
feat(kmethodbadge): add kmethodbadge component [KHCP-7611]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-7611

Adds KMethod component

Screen Shot 2023-06-06 at 8 00 15 PM

Screen Shot 2023-06-07 at 9 32 09 AM

## PR Checklist


adamdehaven:
question: Should this component be `KMethodBadge` instead to match with the Figma and Design naming, as well as, well, what it actually is?

src/components/KMethod/KMethod.vue
```diff
@@ -0,0 +1,117 @@
...
+    },
+  }
+})
+</script>
+
+<style lang="scss" scoped>
+@import '@/styles/variables';
+@import '@/styles/functions';
+
+</style>
```
adamdehaven, line 117: suggestion: If you don't have anything in the style block, you can remove it

src/components/KMethod/KMethod.vue
```diff
@@ -0,0 +1,117 @@
...
+})
+
+const methodLabel = computed((): string => {
+  return props.label || props.method.toUpperCase()
+})
+
+interface MethodBadgeColors {
+  color: string
+  backgroundColor: string
+}
```
adamdehaven, line 65: suggestion: can you move this interface to the types file you created and import from there?

src/components/KMethod/KMethod.vue
```diff
@@ -0,0 +1,117 @@
...
+export const shapes: MethodAppearance[] = ['rounded', 'rectangular']
+</script>
+
+<script setup lang="ts">
+const props = defineProps({
+  method: {
+    type: String as PropType<Method>,
+    required: true,
+    validator: (value: string): boolean => {
+      return methods.includes(value as Method)
```
adamdehaven, line 29: suggestion: you can clean up the types here a bit
```suggestion
    validator: (value: Method): boolean => {
      return methods.includes(value)
```

src/components/KMethod/KMethod.vue
```diff
@@ -0,0 +1,117 @@
...
+  </KBadge>
+</template>
+
+<script lang="ts">
+import { PropType, computed } from 'vue'
+import type { Method, MethodAppearance } from '@/types'
+
+export const methods: Method[] = ['get', 'post', 'put', 'patch', 'delete', 'head', 'options', 'trace', 'connect', 'custom']
+
+export const shapes: MethodAppearance[] = ['rounded', 'rectangular']
```
adamdehaven, line 20: question: why are these defined here instead of the types file you created? We shouldn't be exporting constants from the component. 

There are existing examples of this pattern in the other components

src/components/KMethod/KMethod.vue
```diff
@@ -0,0 +1,117 @@
...
+  color: {
+    type: String,
+    required: false,
+    default: '',
+  },
+  backgroundColor: {
+    type: String,
+    required: false,
+    default: '',
+  },
```
adamdehaven, line 55: question: if the purpose of this component is to standardize the method badge colors, why would we provide props to allow overriding the colors? 

If these props are being used, they may as well just use the native KBadge component. 

Same question for the `appearance` prop

adamdehaven: > Well so even though we are building this component for our own use case, open source users might want to be able to configure something custom

Then wouldn't they just use `KBadge`? Literally the only difference in this component is the colors, right?

portikM:
> question: Should this component be `KMethodBadge` instead to match with the Figma and Design naming, as well as, well, what it is?

Good idea, rename the component @adamdehaven

src/components/KMethod/KMethod.vue
kaiarrowood: I disagree with this, we've had this discussion before. Whether or not it's pure, Konnect frequently drives our Kongponents. We have three use cases in Konnect for the method variants and they all follow this color scheme. We need a place where all of these components can be included and it doesn't make sense to have two places where we have to maintain these color mappings. If you don't think kongponents is the right repo, fair enough, but we need a component like this in some repo somewhere.

adamdehaven: I'll defer on the toggle; however, I still believe the color props should be removed.

src/components/KMethodBadge/KMethodBadge.vue
```diff
@@ -0,0 +1,167 @@
+<template>
+  <KBadge
+    :background-color="backgroundColor || methodBadgeColors[method].backgroundColor"
+    :border-color="borderColor || methodBadgeColors[method].borderColor"
+    class="k-method-badge"
+    :class="[ `method-${method}`, isToggle ? 'k-method-toggle' : '']"
```
kaiarrowood, line 6:
```suggestion
    :class="[ `method-${method}`, { 'k-method-toggle': isToggle }]"
```

nit

src/components/KMethodBadge/KMethodBadge.vue
```diff
@@ -0,0 +1,167 @@
...
+    :color="color || methodBadgeColors[method].color"
+    :is-bordered="isToggle"
+    max-width="auto"
+    :shape="shape === 'rectangular' ? 'rectangular' : 'rounded'"
+    v-on="isToggle ? { click: toggleValue } : {}"
```
kaiarrowood, line 11: question: why `v-on` vs `@click`?

src/components/KMethodBadge/KMethodBadge.vue
```diff
@@ -0,0 +1,167 @@
+<template>
+  <KBadge
+    :background-color="backgroundColor || methodBadgeColors[method].backgroundColor"
+    :border-color="borderColor || methodBadgeColors[method].borderColor"
+    class="k-method-badge"
+    :class="[ `method-${method}`, isToggle ? 'k-method-toggle' : '']"
+    :color="color || methodBadgeColors[method].color"
+    :is-bordered="isToggle"
+    max-width="auto"
+    :shape="shape === 'rectangular' ? 'rectangular' : 'rounded'"
```
kaiarrowood, line 10:
```suggestion
    :shape="shape"
```
question: Wouldn't this work?

src/components/KMethodBadge/KMethodBadge.vue
```diff
@@ -0,0 +1,167 @@
...
+    type: String,
+    required: false,
+    default: '',
+  },
+  isToggle: {
+    type: Boolean,
+    required: false,
+    default: false,
+  },
+  value: {
```
kaiarrowood, line 72: question: should this be `modelValue` ?

src/components/KMethodBadge/KMethodBadge.vue
```diff
@@ -0,0 +1,167 @@
...
+    type: Boolean,
+    required: false,
+    default: false,
+  },
+})
+
+const emit = defineEmits(['update:modelValue'])
+
+const modelValue = ref<boolean>(props.value)
```
kaiarrowood, line 81: It seems weird to set a ref to a prop

src/components/KMethodBadge/KMethodBadge.vue
portikM: Because if we have `@click` (even with undefined event handler function) KBadge automatically makes the element clickable (under the hood KBadge checks if `onClick` attribute is present on the element). So as a way to go around that, I have a `v-on` that only adds `click` event handler if it needs to.

src/components/KMethodBadge/KMethodBadge.vue
portikM: In my opinion very efficient way to set it right on component setup. Maybe needs a watcher though, to watch for prop changes - just added it.

src/components/KMethodBadge/KMethodBadge.vue
portikM: Fixed

src/components/KMethodBadge/KMethodBadge.vue
portikM: Fixed

src/components/KMethodBadge/KMethodBadge.vue
adamdehaven: Wouldn't this unbind _any_ event listener if `isToggle` is false? I'm not sure that's ideal.

kaiarrowood (dismissed):
LGTM! 🚀

src/components/KMethodBadge/KMethodBadge.vue
portikM: Done

src/components/KMethodBadge/KMethodBadge.vue
```diff
@@ -0,0 +1,161 @@
+<template>
+  <KBadge
+    :background-color="methodBadgeColors[method].backgroundColor"
+    :border-color="methodBadgeColors[method].borderColor"
+    class="k-method-badge"
+    :class="[ `method-${method}`, { 'k-method-toggle': isToggle }]"
+    :color="methodBadgeColors[method].color"
+    :is-bordered="isToggle"
+    max-width="auto"
```
adamdehaven, line 9: question: you have `max-width: auto` here; however, aren't all the method badges the same width? (looking at the new design [here](https://www.figma.com/file/Yze0SWXl5nKjR0rFdilljK/Components?type=design&node-id=467-10283&t=taJN6RPvzvewyenM-0))

src/components/KMethodBadge/KMethodBadge.vue
```diff
@@ -0,0 +1,161 @@
...
+    default: 'custom',
+  },
+  shape: {
+    type: String as PropType<MethodShape>,
+    required: false,
+    validator: (value: MethodShape): boolean => {
+      return ShapesArray.includes(value as MethodShape)
+    },
+    default: 'rounded',
+  },
```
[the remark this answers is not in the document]
portikM: I thought of that but I wanted this component to be as close as possible to KBdage (KBdage has same `shape` prop with same values)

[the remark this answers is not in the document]
adamdehaven: Since this is a new component, let's follow the `isRounded` pattern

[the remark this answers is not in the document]
adamdehaven: TBH this prop will probably go away with the redesign anyway

src/components/KMethodBadge/KMethodBadge.vue
portikM: `max-width="auto"` just prevents it from truncating. Now in the link you sent the square ones are all same width - do we think they all (both square and rounded) should be same width? In my opinion either both should be same width or none - we don't want round badges to be content width while square ones are fixed.

src/components/KMethodBadge/KMethodBadge.vue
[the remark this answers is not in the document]
portikM: Agreed

src/components/KMethodBadge/KMethodBadge.vue
adamdehaven: I'm good with whatever you think here; just wanted to bring it up

portikM: Okay, I guess I wanna leave it flexible in this component for now - if we need them fixed width we can either change it later or control that from the host app

kaiarrowood: +1 on leaving it until we implement the new designs everywhere

PR #1452, opened 2023-06-12
feat(KTabs, KTextArea, KToaster): add types [KHCP-6974]

DariaYeremina:
# Summary

https://konghq.atlassian.net/browse/KHCP-6974
Add types for KTabs, KTextArea, KToaster

## PR Checklist


src/components/KToaster/ToastManager.ts
```diff
@@ -43,18 +43,18 @@ export default class ToastManager {
     createApp(Toast).mount(`#${this.id}`)
   }
 
-  setTimer(key: any, timeout: number) {
+  setTimer(key: any, timeout: number): number {
```
adamdehaven, line 46: question: why is the return type `number` here?

src/types/text-area.ts
```diff
@@ -0,0 +1,6 @@
+export interface TextAreaLimitExceed {
+  value: string,
+  length: number,
+  characterLimit: number,
+  limitExceeded: boolean,
```
adamdehaven, line 5: nit: no commas

src/components/KToaster/ToastManager.ts
DariaYeremina: https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#return_value

src/types/text-area.ts
DariaYeremina: fixed

kaiarrowood (approved):
LGTM! 🚀

PR #1457, opened 2023-06-13
feat(KTable): add types [KHCP-6973]

DariaYeremina:
# Summary

https://konghq.atlassian.net/browse/KHCP-6973
Add types for KTable

## PR Checklist


src/components/KTable/KTable.vue
```diff
@@ -215,7 +215,18 @@ import type {
   TableState,
   PageChangedData,
   PageSizeChangedData,
+  SortColumnOrder,
+  TableOptions,
+  TableSortOrder,
+  TableSortPayload, TableStatePayload,
```
kaiarrowood, line 221:
```suggestion
  TableSortPayload, 
  TableStatePayload,
```

nit

src/components/KTable/KTable.vue
```diff
@@ -228,7 +239,7 @@ const props = defineProps({
    * @param {Array} options.data - Array of Objects defining column data
    */
   options: {
-    type: Object,
+    type: Object as PropType<TableOptions>,
```
kaiarrowood, line 242: I have concerns about breaking existin implementations by making the prop types too strict. We want to avoid making users typecast as much as possible in the host app. Having to do things like 
image
everywhere is not a great DX.

Can we keep the types loose in the prop defs and have the strict types on a local copy of the prop? We can just cast to the correct type like we do for our handling of the `headers` prop:

https://github.com/Kong/kongponents/blob/main/src/components/KTable/KTable.vue#L729

kaiarrowood (changes requested):
I don't think we want to make the prop types so tight, can we loosen them to avoid typecasting in the host app?

We've had to back out changes like this already: https://github.com/Kong/kongponents/pull/1299

adamdehaven:
@DariaYeremina what do we need to do to get this PR merged?

adamdehaven (commented):
> Will cause ts errors in `khcp-ui` and potentially `shared-ui-components` and `public-ui-components`

Can you expand on this a bit more? What will be the issues?

src/components/KTable/KTable.vue
DariaYeremina: tried this new PR in khcp-ui and shared-ui, no ts errors

DariaYeremina:
> > Will cause ts errors in `khcp-ui` and potentially `shared-ui-components` and `public-ui-components`
> 
> Can you expand on this a bit more? What will be the issues?

With today's fixes there's no ts errors in hosted apps

adamdehaven (approved):

adamdehaven (approved):

PR #1460, opened 2023-06-14
fix(kselect): value when `enableItemCreation` is true

Leopoldthecoder:
# Summary

`<KSelect>` with `enableItemCreation` should accept a previously created item as its value, otherwise there'll be a problem when editing an entity. Context: https://kongstrong.slack.com/archives/C04FLSMJ2ET/p1686726647993929.

## PR Checklist


portikM (approved):
LGTM 🚀

PR #1474, opened 2023-06-22
fix(kemptystate): unnecessary spaces and gaps

kleinfreund:
# Summary

Changes the empty state component to only render container divs if the associated slots which don’t have fallback content are slotted in (e.g. if `title` is not slotted in, the component no longer renders an empty `.k-empty-state-title-header` div).

Changes the margins used to space out the individual parts of the empty state component to only apply if they make sense (e.g. the `message` slot will no longer have a large bottom margin if the CTA is hidden).

## PR Checklist

### DIFF
--- a/src/components/KEmptyState/KEmptyState.vue
+++ b/src/components/KEmptyState/KEmptyState.vue
@@ -91,7 +91,7 @@ defineProps({
 .empty-state-wrapper {
   background-color: var(--KEmptyBackground, var(--white));
   border-radius: 4px;
-  padding: 48px 0;
+  padding: var(--spacing-xxl, spacing(xxl)) 0;
   text-align: center;
 
   > * + * {
--- a/src/components/KEmptyState/KEmptyState.vue
+++ b/src/components/KEmptyState/KEmptyState.vue
@@ -6,7 +6,7 @@
     <div class="empty-state-title">
       <div
         v-if="isError || icon"
-        class="k-empty-state-icon card-icon mb-3"
+        class="k-empty-state-icon card-icon"
         :class="{ 'warning-icon': isError }"
       >
         <KIcon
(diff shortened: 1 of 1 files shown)


src/components/KEmptyState/KEmptyState.vue
```diff
@@ -43,6 +51,9 @@
 <script lang="ts" setup>
 import KButton from '@/components/KButton/KButton.vue'
 import KIcon from '@/components/KIcon/KIcon.vue'
+import { useSlots } from 'vue'
+
+const slots = useSlots()
```
portikM, line 56: nit: since you're only using `slots` in the template, you can remove the const declaration here along with the import form `vue` and access it as a dollar sign variable in the template like so: `v-if="$slots.title"`.

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -86,24 +97,35 @@ defineProps({
   padding: 48px 0;
```
portikM, line 97:
```suggestion
  padding: var(--spacing-xxl, spacing(xxl)) 0;
```
While you're in this component

portikM (commented):
Thanks for adding this @kleinfreund! Just a few minor nits, otherwise everything looks good!

src/components/KEmptyState/KEmptyState.vue
kleinfreund: This was changed now.

src/components/KEmptyState/KEmptyState.vue
kleinfreund: Ah, nice. I didn’t see that one. There are a couple more things in here that could use custom properties (e.g. the border radius), but I’d rather keep the scope to spacing for the time being.

Changed.

portikM (approved):
LGTM 🚀

PR #1482, opened 2023-06-26
fix: using correct env token to make it less confusing [KHCP-7461]

ValeryG:
# Summary

cut & paste error fix. using correct token.


adamdehaven (approved):
LGTM

PR #1500, opened 2023-07-05
fix(deps): move @kong/design-tokens from devDeps to deps

johncowen:
# Summary

Moves `@kong/design-tokens` from devDeps to deps.

Files from `@kong/design-tokens` are required in production. Therefore if/when `@kong/kongponents` is installed as an indirect dependency, the current configuration (using `devDependencies`) means `@kong/design-tokens` is not available, therefore breaking the application.

## PR Checklist

* [ ] Does not introduce dependencies **_(It doesn't 'introduce' as such but makes dependencies available that unintentionally weren't previously available)_**
* [ ] **Tests pass:** check the output of yarn test. **_Tests don't run in CI due to what seems like a lack of token, locally theres failures but `main` also fails. Let me know if you need me to do anything._**

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "semantic-release": "semantic-release"
   },
   "dependencies": {
+    "@kong/design-tokens": "^1.2.0",
     "axios": "^0.27.2",
     "date-fns": "^2.29.3",
     "date-fns-tz": "^1.3.7",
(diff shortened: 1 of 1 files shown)


adamdehaven (changes requested):
holding this until we chat

johncowen:
Most probably superseded by https://github.com/Kong/kongponents/pull/1501

johncowen:
Superseded by https://github.com/Kong/kongponents/pull/1501

PR #1519, opened 2023-07-11
fix(KModal): fix focusTrap error [KHCP-7758]

DariaYeremina:
# Summary

https://konghq.atlassian.net/browse/KHCP-7758

Fix `Your focus-trap must have at least one container with at least one tabbable node in it at all times` DD error

## PR Checklist

### DIFF
--- a/src/components/KModal/KModal.vue
+++ b/src/components/KModal/KModal.vue
@@ -7,12 +7,14 @@
     role="dialog"
   >
     <div
+      ref="modalOuter"
       class="k-modal-backdrop modal-backdrop"
       @click="(evt: any) => close(false, evt)"
     >
       <FocusTrap
         ref="focusTrap"
         :active="false"
+        :fallback-focus="modalOuter?.$el"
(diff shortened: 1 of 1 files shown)


src/components/KModal/KModal.vue
```diff
@@ -7,12 +7,14 @@
...
   >
     <div
+      ref="modalOuter"
       class="k-modal-backdrop modal-backdrop"
       @click="(evt: any) => close(false, evt)"
     >
       <FocusTrap
         ref="focusTrap"
         :active="false"
+        :fallback-focus="modalOuter?.$el"
```
adamdehaven, line 17: if this can be undefined, how does using this element as a fallback resolve the issue?

DariaYeremina: it's null until component is mounted

adamdehaven: so then why is the optional chaining needed if it's there on mount?

adamdehaven (approved):

PR #1523, opened 2023-07-12
fix(KSelect): add pre-check presence for selectItems.value  [KHCP-7756]

DariaYeremina:
# Summary

https://konghq.atlassian.net/browse/KHCP-7756

Fix `TypeError: Cannot read properties of null (reading 'length')` error

## PR Checklist

### DIFF
--- a/src/components/KSelect/KSelect.vue
+++ b/src/components/KSelect/KSelect.vue
@@ -480,7 +480,7 @@ const value = computed({
     return props.modelValue
   },
   set(newValue: string | number): void {
-    const item = selectItems.value.filter((item: SelectItem) => item.value === newValue)
+    const item = selectItems.value?.filter((item: SelectItem) => item.value === newValue)
     if (item.length) {
       handleItemSelect(item[0])
     } else if (!newValue) {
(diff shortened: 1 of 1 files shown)


src/components/KSelect/KSelect.vue
```diff
@@ -480,7 +480,7 @@ const value = computed({
     return props.modelValue
   },
   set(newValue: string | number): void {
-    const item = selectItems.value.filter((item: SelectItem) => item.value === newValue)
+    const item = selectItems.value?.filter((item: SelectItem) => item.value === newValue)
     if (item.length) {
```
adamdehaven, line 484: if `item` can be undefined, then the if statement below needs to optionally chain checking for the `length`

src/components/KSelect/KSelect.vue
```diff
@@ -597,7 +597,7 @@ const handleAddItem = (): void => {
     return
   }
 
-  const pos = selectItems.value.length + 1
+  const pos = selectItems.value?.length + 1
```
adamdehaven, line 600: question: won't this error if undefined and then try to add 1?

src/components/KSelect/KSelect.vue
DariaYeremina: fixed

adamdehaven (approved):

PR #1539, opened 2023-07-17
fix(slideout): Header updates [khcp-7919]

aanchalm01:
# Summary
Addresses:
[https://konghq.atlassian.net/browse/KHCP-7919](https://konghq.atlassian.net/browse/KHCP-7919)

## PR Checklist


src/components/KSlideout/KSlideout.vue
```diff
@@ -77,18 +90,10 @@ onUnmounted(() => {
...
-    position: fixed;
-    right: 0;
-    top: 0;
-    z-index: 9999;
-  }
-
   .panel {
     background-color: var(--white, color(white));
+    display: flex;
+    flex-direction: column;
```
adamdehaven, line 96: question: would it make sense to add a new prop `topOffset<number>` that allows for modifying the CSS `top: 0` rule below? It could default to `0`, but would allow a host app to define the offset from the top of the page.

In Konnect UI, we could pass in the height of the navbar and display the sidebar under the nav
image

aanchalm01: Done

src/components/KSlideout/KSlideout.vue
```diff
@@ -48,6 +50,24 @@ const props = defineProps({
...
+      return ['start', 'end'].includes(value)
+    },
+  },
+  // enable/disable overlay to be able to interact with the background content while the slideout is expanded
+  overlayEnabled: {
+    type: Boolean,
+    default: false,
+  },
+  // allows a host app to define the offset from the top of the page
+  topOffsetNumeric: {
```
adamdehaven, line 67: can this just be named `offsetTop`?

Also it doesn't match the name of the prop in the docs

src/components/KSlideout/KSlideout.vue
```diff
@@ -48,6 +50,24 @@ const props = defineProps({
...
+    type: Number,
+    default: 40,
```
adamdehaven, line 69: question: why doesn't this default to zero?

src/components/KSlideout/KSlideout.vue
```diff
@@ -77,36 +103,43 @@ onUnmounted(() => {
...
-
   .panel {
     background-color: var(--white, color(white));
+    display: flex;
+    flex-direction: column;
     height: 100vh;
     max-width: 500px;
     position: fixed;
     right: 0;
-    top: 0;
```
adamdehaven, line 96: question: instead of removing this style and adding inline styles, cant you just bind the prop value here?

You may need to apply the offset to the overlay background as well, not sure.

docs/components/slideout.md
```diff
@@ -64,6 +64,71 @@ Below we demonstrate wrapping `KSlideout` in the [`KToggle`](/components/renderl
 
 Tells the component whether or not to render the open panel.
 
+### closeBtnAlignment
+
+Controls the close button alignment, defaults to `start`.
+
+### topOffsetStyle
```
adamdehaven, line 71: this doesn't match the prop name in the component

src/components/KSlideout/KSlideout.vue
```diff
@@ -48,6 +50,24 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+  // controls close button alignment
+  closeBtnAlignment: {
```
adamdehaven, line 54: suggestion: don't abbreviate prop names
```suggestion
  closeButtonAlignment: {
```

src/components/KSlideout/KSlideout.vue
aanchalm01: typo. fixed

src/components/KSlideout/KSlideout.vue
aanchalm01: Fixed.

adamdehaven:
The `overlayEnabled` example doesn't actually show an overlay

aanchalm01:
> The `overlayEnabled` example doesn't actually show an overlay

Is there anything else that's not working out?

aanchalm01:
@filipgutica , @mihai-peteu - Can you verify if this PR needs need any other change regarding KSlideout, or is it good to merge now ?

filipgutica:
In our mockup for the api request log detail slideout, the "title" of the slide, is on the same line with the "X" close button.

Is there a possibility to include a slot where we could place information or a title in the header of the slide out (on the same line with the "X") 

See the mock:
image

filipgutica:
> @filipgutica , @mihai-peteu - Can you verify if this PR needs need any other change regarding KSlideout, or is it good to merge now ?

had one question

adamdehaven:
> Is there a possibility to include a slot where we could place information or a title in the header of the slide out (on the same line with the "X")

IMO the header shouldn't be on the same line as the close X -- that's going to inevitably lead to needed to truncating the title; I think the mock should be updated to match the existing component

aanchalm01:
@filipgutica - I agree with Adam's comment, placing the header info in same line with 'X' will require customization like truncating the text.. Let me know what do you think ?

filipgutica:
Fine by me, will relay that to @Helen and @sichvoge if they ask 🙂

filipgutica:
> > Is there a possibility to include a slot where we could place information or a title in the header of the slide out (on the same line with the "X")
> 
> 
> 
> IMO the header shouldn't be on the same line as the close X -- that's going to inevitably lead to needed to truncating the title; I think the mock should be updated to match the existing component

I have a feeling design may be quite adamant about having the "slide title" on the same line as the "X" I think this is a pretty standard pattern seen across the industry 

See this screenshot from dd for example 

I think this may warrant further discussion with design. 

image

docs/components/slideout.md
```diff
@@ -64,9 +64,214 @@ Below we demonstrate wrapping `KSlideout` in the [`KToggle`](/components/renderl
...
+
+### iconType
+
+Determines the icon type that is placed next to Title, defaults to `copy`.
+
+### iconColor
+
+Determines the icon color that is placed next to Title, defaults to `blue`.
+
+### iconSize
```
kaiarrowood, line 214: issue: We are actually trying to move away from this type of approach with too many props customizing optional parts of the design. It would be much cleaner to simply offer a single slot following the title maybe call it `title-after` and let the consuming app be responsible for rendering the badge/icon if it's needed.

docs/components/slideout.md
```diff
@@ -77,6 +282,12 @@ There is one default slot which is used to place content into the slideout panel
 </KSlideout>
 ```
 
+- `badgeContent` - used to customize `badge` inside the header
```
kaiarrowood, line 285: suggestion to use three slots: `title-before`, `title`, and `title-after` as discussed above

docs/components/slideout.md
```diff
@@ -77,6 +282,12 @@ There is one default slot which is used to place content into the slideout panel
...
+- `titleContent` - used to customize `title` inside the header
+- `iconContent` - used to customize `icon` inside the header
+
 ## Events
 
-- `@close` - Emitted when the close button is clicked, anything outside the panel is clicked, or the `esc` key is pressed.
+- `close` - Emitted when the close button is clicked, anything outside the panel is clicked, or the `esc` key is pressed.
+
+- `kclipboard-icon-clicked` - Emitted when the `icon` next to `title` is clicked.
```
kaiarrowood, line 293: suggestion: if the icon content is slottable we should use a more generic event (if we even need it) - if we allow the consuming app to completely control the content via slots, we no longer need to handle/emit the click event.

src/components/KSlideout/KSlideout.vue
```diff
@@ -3,7 +3,7 @@
     <transition name="fade">
       <div
         v-if="isVisible"
-        class="panel-background"
+        :class="{ 'panel-background': overlayEnabled }"
         @click="(event: any) => handleClose(event, true)"
```
kaiarrowood, line 7: question: does this actually work like we want? I don't think it's enough to conditionally apply a class to the overlay div because we are still catching all click events and using it to toggle the visibility of the slideout which is not what we want. If overlay is disabled, there should be no overlay div and clicks on background content should actually be applied to the background content.

src/components/KSlideout/KSlideout.vue
```diff
@@ -13,10 +13,65 @@
       <div
         v-if="isVisible"
         class="panel"
-        :class="{ isVisible: 'is-visible' }"
+        :class="{ 'is-visible': isVisible, 'border-styles': !overlayEnabled }"
       >
+        <!-- If there is title prop, display the header content -->
+        <div
+          v-if="title"
```
kaiarrowood, line 20: this conditional also needs to check for slot content - `title` could be left empty and only the `titleContent` slot is used which we should still allow.

src/components/KSlideout/KSlideout.vue
```diff
@@ -13,10 +13,65 @@
...
+
+          <button
+            class="close-button-with-title"
+            @click="(event: any) => handleClose(event, true)"
+          >
+            <KIcon
+              color="var(--grey-600)"
+              icon="close"
+              size="24"
+              view-box="0 0 24 24"
```
kaiarrowood, line 67: question: do we need the viewbox/size? Aren't those the defaults?

src/components/KSlideout/KSlideout.vue
```diff
@@ -38,27 +93,96 @@
...
+  // controls close button alignment
+  closeButtonAlignment: {
+    type: String,
+    default: 'start',
+    validator: (value: string): boolean => {
+      return ['start', 'end'].includes(value)
+    },
+  },
+  // enable/disable overlay to be able to interact with the background content while the slideout is expanded
+  overlayEnabled: {
```
kaiarrowood, line 117: suggestion: can we change to just `overlay` or `has-overlay`?

src/components/KSlideout/KSlideout.vue
```diff
@@ -38,27 +93,96 @@
...
+    type: Boolean,
+    default: true,
+  },
+  // allows a host app to define the offset from the top of the page
+  offsetTop: {
```
kaiarrowood, line 122: Can we change this to take a string, so users can provide things like percentages or `vh`?

src/components/KSlideout/KSlideout.vue
```diff
@@ -76,40 +200,82 @@ onUnmounted(() => {
...
-    top: 0;
-    z-index: 9999;
+  .k-slideout-header-content {
+    display: flex;
+    .k-slideout-title {
+      color: var(--black-400, #3C4557);
+      flex:1;
+      font-size: 16px;
+      font-weight: 500;
+      line-height: 24px;
```
kaiarrowood, line 210: nit: can use tokens

src/components/KSlideout/KSlideout.vue
```diff
@@ -76,40 +200,82 @@ onUnmounted(() => {
...
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+      &.title-spacing {
+        margin-left: 12px;
```
kaiarrowood, line 215: nit: can use tokens

src/components/KSlideout/KSlideout.vue
```diff
@@ -76,40 +200,82 @@ onUnmounted(() => {
...
+    .close-button-start {
+      align-self: flex-start;
       background: none;
       border: none;
       cursor: pointer;
-      height: 16px;
-      left:  16px;
+      display: flex;
+      height: auto;
+      margin-left: 16px;
```
kaiarrowood, line 238: nit: can use tokens

src/components/KSlideout/KSlideout.vue
```diff
@@ -76,40 +200,82 @@ onUnmounted(() => {
...
       outline: inherit;
-      padding: 0;
+      padding-top: 16px;
```
kaiarrowood, line 240: nit: can use tokens

src/components/KSlideout/KSlideout.vue
```diff
@@ -76,40 +200,82 @@ onUnmounted(() => {
...
+      margin-right: 6px;
+    }
+
+    .icon-with-title {
+      align-self: center;
+      cursor: pointer;
+    }
+
+    .close-button-with-title {
+      margin-left: 16px;
```
kaiarrowood, line 274: nit: can use tokens

docs/components/slideout.md
```diff
@@ -64,9 +64,214 @@ Below we demonstrate wrapping `KSlideout` in the [`KToggle`](/components/renderl
...
+        <p>Anim officia eiusmod duis est consequat nulla tempor ad non magna Lorem ullamco nostrud amet. Occaecat voluptate dolor enim eiusmod do qui nulla pariatur enim. Et elit elit consequat do do duis enim est ullamco id sunt sunt amet eiusmod. Do minim mollit irure ea sunt officia minim sint eiusmod enim amet. Quis exercitation in ullamco quis aliqua.</p>
+        <p>Anim officia eiusmod duis est consequat nulla tempor ad non magna Lorem ullamco nostrud amet. Occaecat voluptate dolor enim eiusmod do qui nulla pariatur enim. Et elit elit consequat do do duis enim est ullamco id sunt sunt amet eiusmod. Do minim mollit irure ea sunt officia minim sint eiusmod enim amet. Quis exercitation in ullamco quis aliqua.</p>
+        <p>Anim officia eiusmod duis est consequat nulla tempor ad non magna Lorem ullamco nostrud amet. Occaecat voluptate dolor enim eiusmod do qui nulla pariatur enim. Et elit elit consequat do do duis enim est ullamco id sunt sunt amet eiusmod. Do minim mollit irure ea sunt officia minim sint eiusmod enim amet. Quis exercitation in ullamco quis aliqua.</p>
+      </div>
+      <KButton appearance="primary">Buttons</KButton>
+      <KCard class="mt-2">
```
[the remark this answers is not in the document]
aanchalm01: I have already updated this class for the token work, will be updated here once rebased.

src/components/KSlideout/KSlideout.vue
```diff
@@ -66,10 +123,51 @@ const props = defineProps({
...
+    default: 'copy',
+  },
+  iconColor: {
+    type: String,
+    default: 'var(--blue-500)',
+  },
+  iconSize: {
+    type: String,
+    default: '16',
+  },
```
filipgutica: This seems a little over-complicated. 

Rather than having all of these extra props and logic with badges and icons and what not, why not just provide a "header" or "title" `<slot>` where the consuming app can pass in whatever arbitrary html they want into the "header" or "title" section (in our case the kbadge + text + clipboard copy icon). Keeps this component more simple, and offloads any "custom" header logic to the consuming app. 

Keep the "title" prop, and by default the "title" <slot> will just display that title with default truncating logic

filipgutica (dismissed):
one comment

src/components/KSlideout/KSlideout.vue
adamdehaven: I agree here. 

Alternatively, the title could be a prop string, and then provide slots for the badge and a separate slot for the icon? 

It doesn’t make sense to add a prop for the icon since we are soon moving to a model where all icons are slotted

docs/components/slideout.md
aanchalm01: Yes, removed this.

docs/components/slideout.md
```diff
@@ -64,9 +64,214 @@ Below we demonstrate wrapping `KSlideout` in the [`KToggle`](/components/renderl
...
+        appearance="info"
+        alert-message="Or even an alert!"
+        class="mt-2" />
+    </KSlideout>
+  </div>
+</KToggle>
+```
+### hasOnlyTitle
+
+Determines whether the header will display only `Title` or has other attributes like `Badge`, `Icon` etc, defaults to `false`.
```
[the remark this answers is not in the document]
aanchalm01: removed this prop.

src/components/KSlideout/KSlideout.vue
```diff
@@ -13,16 +13,41 @@
...
+              class="k-slideout-title"
+            >
+              {{ title }}
+            </p>
+            <button
+              class="close-button-with-title"
+              @click="(event: any) => handleClose(event, true)"
+            >
+              <KIcon
+                color="var(--grey-600)"
```
portikM, line 34: Could you provide a token fallback here?

src/components/KSlideout/KSlideout.vue
```diff
@@ -13,16 +13,41 @@
...
+        </div>
+
         <button
-          class="close-btn"
+          v-else
+          :class="closeButtonAlignment === 'start' ? 'close-button-start' : 'close-button-end'"
           @click="(event: any) => handleClose(event, true)"
         >
           <KIcon
+            color="var(--grey-600)"
```
portikM, line 48: And here

aanchalm01:
@kaiarrowood , @filipgutica , @mihai-peteu - all the feedback has been addressed.

src/components/KSlideout/KSlideout.vue
adamdehaven: TBH I don't think relative values make sense here -- you wouldn't offset the slideout based on `vh` or percentage for something like a navbar, etc.

src/components/KSlideout/KSlideout.vue
```diff
@@ -78,49 +142,109 @@ onUnmounted(() => {
...
+    left: 0;
+    position: fixed;
+    right: 0;
+    top: v-bind('offsetTopValue');
+    z-index: -1;
+  }
+
+  .border-styles {
+    border-left: var(--kui-border-width-10, $kui-border-width-10) solid var(--grey-300, var(--kui-color-border-neutral-weak, $kui-color-border-neutral-weak));
+    box-shadow: #0000000d -2px 0px 5px;
```
adamdehaven, line 246: question: why the non-standard hex value here? Can you use `rgba` instead?

src/components/KSlideout/KSlideout.vue
adamdehaven: I don't think they would ever be used, so not sure support is needed.

e.g. if you have the top of the slideout set to a value of `10%` or `10vh` if the user resizes their browser window, the slideout will move from its position

aanchalm01: Okay, makes sense, I can revert this one.
@kaiarrowood - do you have a specific reason to support this prop as String instead of Number before I revert this change ?

docs/components/slider.md
kaiarrowood, line 1: I think this is an unintentional file addition?

src/components/KSlideout/KSlideout.cy.ts
```diff
@@ -20,18 +20,6 @@ describe('KSlideout', () => {
     cy.get('p').should('be.visible')
   })
 
-  it('emits close when panel-background is clicked', () => {
```
kaiarrowood, line 23: Why did we remove this test?

src/components/KSlideout/KSlideout.vue
```diff
@@ -3,28 +3,50 @@
...
+              {{ title }}
+            </p>
+          </slot>
+        </div>
+
+        <!-- header custom slot -->
+        <div v-if="hasHeader">
+          <slot name="header" />
+        </div>
+
```
kaiarrowood: I don't think you want this to be two different slots? You don't mention the `title` slot in the docs and I would assume using the `header` slot would completely override display of the `title`/`title` slot.

aanchalm01: I will ping when its ready for review, I am still making changes.

src/components/KSlideout/KSlideout.vue
kaiarrowood: If you're going to force it as `numeric` I would just explicitly say in the docs that it takes a `number` and that `number` will use `px` as the unit. However you want to explain it so there is no confusion

docs/components/slider.md
aanchalm01: It doesnt exist, was added when Jilson was helping with a bug.

tsconfig.json
```diff
@@ -22,7 +22,8 @@
     "types": [
       "vite/client",
       "cypress",
-      "./cypress/support"
+      "./cypress/support",
+      "web-bluetooth"
```
kaiarrowood, line 26: is this an intentional change?

aanchalm01: yes.

mihai-peteu (dismissed):

src/components/KSlideout/KSlideout.cy.ts
aanchalm01: I removed it because now with the use of `onClickOutside`, even though the Close is emitted, Cypress doesn't gets that emitted value and it fails.

src/components/KSlideout/KSlideout.vue
```diff
@@ -3,28 +3,55 @@
...
+          >
+            <slot name="before-title" />
+          </div>
+
+          <!-- title -->
+          <div class="k-slideout-main-title">
+            <p
+              class="k-slideout-title"
+              data-testid="k-slideout-title"
+            >
```
kaiarrowood, line 29:
```suggestion
            <p
              class="k-slideout-title"
              data-testid="k-slideout-title"
              :title="title"
            >
```

issue: always offer the full text on hover of truncated strings

docs/components/slideout.md
```diff
@@ -64,9 +62,113 @@ Below we demonstrate wrapping `KSlideout` in the [`KToggle`](/components/renderl
 
 Tells the component whether or not to render the open panel.
 
+### closeButtonAlignment
+
+Controls the close button alignment, can be `start` (default) or `end`.
+
+### offsetTop
+
+Allows a host app to define the offset from the top of the page, it takes a `number` and that number uses `px` as the unit, defaults to `0px`.
```
kaiarrowood, line 71:
```suggestion
Allows a host app to define the offset from the top of the page, expects a `number` and that will be converted to `px`, defaults to `0px`.
```

suggestion

kaiarrowood (approved):
LGTM! 🚀 

Nice work @aanchalm01 !

adamdehaven (approved):

PR #1540, opened 2023-07-17
feat(kselect): introduce kui tokens [KHCP-7722]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-7722

Changes:
- introduces `kui-` design tokens in `KSelect` component
- removes any usage of Kongponents utility classes in `KSelect` component template

## PR Checklist

### DIFF
--- a/src/components/KSelect/KSelect.vue
+++ b/src/components/KSelect/KSelect.vue
@@ -45,9 +45,9 @@
           @keyup.enter="clearSelection"
         >
           <KIcon
-            color="var(--blue-200)"
+            :color="`var(--blue-200, ${KUI_COLOR_TEXT_PRIMARY})`"
             icon="clear"
-          /><!-- TODO: token needed -->
+          />
         </button>
       </div>
       <KToggle v-slot="{ toggle, isToggled }">
@@ -269,7 +269,7 @@ import {
   SelectAppearance,
   SelectAppearanceArray,
 } from '@/types'
-import { KUI_COLOR_TEXT_NEUTRAL, KUI_ICON_SIZE_30 } from '@kong/design-tokens'
+import { KUI_COLOR_TEXT_NEUTRAL, KUI_ICON_SIZE_30, KUI_COLOR_TEXT_PRIMARY } from '@kong/design-tokens'
 
 export default {
   inheritAttrs: false,
--- a/src/components/KSelect/KSelect.vue
+++ b/src/components/KSelect/KSelect.vue
@@ -47,7 +47,7 @@
           <KIcon
             color="var(--blue-200)"
             icon="clear"
-          />
+          /><!-- TODO: token needed -->
         </button>
       </div>
       <KToggle v-slot="{ toggle, isToggled }">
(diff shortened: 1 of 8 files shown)


src/components/KSelect/KSelect.vue
```diff
@@ -40,14 +40,14 @@
...
-          class="clear-selection-icon cursor-pointer non-visual-button"
+          class="clear-selection-icon"
           @click="clearSelection"
           @keyup.enter="clearSelection"
         >
           <KIcon
             color="var(--blue-200)"
             icon="clear"
-          />
+          /><!-- TODO: token needed -->
```
portikM, line 50: @adamdehaven any idea what can we do here? Appropriate `color-text-primary-...` token variant doesn't exist and `$tmp-` SCSS won't work in template (unless used with interpolation?)

adamdehaven: You're saying there's not a token for `blue-200`, correct?

If that's the case, I would pick the _closest existing new `kui` token_ and use that, with its fallback, instead. 

Changing the icon color a little isn't going to break anything.

portikM: There is no close value color text token (closest one is `color-text-primary` which is noticeably different). There are close tokens for background color, but we decided against using background color tokens for icons color. I would add a temp SCSS token for this so we can decide later, but like I said it won't work.

adamdehaven: Just use the closest and put a screenshot in the PR

portikM: Here are before and after screenshots
Screen Shot 2023-07-18 at 11 13 25 AM
Screen Shot 2023-07-18 at 11 16 17 AM

adamdehaven: I think the easiest thing for this particular scenario is to hard-code the icon color to the hex value and just drop CSS variable support. I don’t believe we’re customizing he select icon color here anyway

src/components/KSelect/KSelect.vue
```diff
@@ -1059,14 +1070,19 @@ $chevronDownIconMargin: 10px;
...
-      top: 0;
+      top: var(--kui-space-0, $kui-space-0);
     }
 
     .k-popover-content {
       @include kSelectPopoverMaxHeight;
 
+      .k-select-list {
+        margin: var(--kui-space-0, kui-space-0) !important;
+        padding: var(--kui-space-0, kui-space-0) !important;
```
adamdehaven, line 1083: issue: missing `$` here

src/components/KSelect/KSelectItem.vue
```diff
@@ -99,60 +104,62 @@ const handleClick = (e: MouseEvent): void => {
...
 
     .k-select-item-label {
-      color: var(--grey-600);
-      font-size: 14px;
-      font-weight: 500;
-      line-height: 20px;
-      padding: 8px;
+      color: var(--grey-600, var(--kui-color-text-primary-strong, $kui-color-text-primary-strong));
+      font-size: var(--kui-font-size-30, $kui-font-size-30);
+      font-weight: var(--kui-font-weight-regular, $kui-font-weight-regular);
```
adamdehaven, line 109: issue: regular font weight is `400` -- you need `500` here

adamdehaven (changes requested):

src/components/KSelect/KSelectItem.vue
portikM: Fixed

src/components/KSelect/KSelect.vue
portikM: Fixed

adamdehaven (approved):

adamdehaven (approved):

PR #1542, opened 2023-07-18
feat(kpop): introduce kui tokens [KHCP-7718]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-7718

Changes:
- introduces `kui-` design tokens in `KPop` component
- removes any usage of Kongponents utility classes in `KPop` component template

## PR Checklist

### DIFF
--- a/src/components/KPop/KPop.vue
+++ b/src/components/KPop/KPop.vue
@@ -667,7 +667,7 @@ export default defineComponent({
 .fadeIn,
 .fade-leave-active,
 .fadeOut {
-  animation-duration: $tmp-animation-timing-3;
+  animation-duration: $tmp-animation-timing-2;
   animation-fill-mode: both;
 }
 .fade-enter-active,
--- a/src/components/KPop/KPop.vue
+++ b/src/components/KPop/KPop.vue
@@ -34,7 +34,7 @@
         >
           <div
             v-if="$slots.title || title || $slots.actions"
-            class="k-popover-header d-flex"
+            class="k-popover-header"
           >
             <div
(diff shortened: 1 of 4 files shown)


src/components/KPop/KPop.vue
```diff
@@ -506,24 +509,24 @@ export default defineComponent({
 
     &:after {
       border-color: rgba(255, 255, 255, 0);
-      border-bottom-color: var(--KPopBackground, var(--white, color(white)));
+      border-bottom-color: var(--KPopBackground, var(--white, $tmp-color-white));
```
adamdehaven, line 512: question: why is this a temp token? Same question for all usage of `$tmp-color-white`

src/components/KPop/KPop.vue
```diff
@@ -661,7 +667,7 @@ $transition-speed: .3s !default;
 .fadeIn,
 .fade-leave-active,
 .fadeOut {
-  animation-duration: $transition-speed;
+  animation-duration: $tmp-animation-timing-3;
```
adamdehaven, line 670: question: can we just use the existing `$tmp-animation-timing-2` here and just standardize animation durations to .2s?

src/components/KPop/KPop.vue
portikM: Because we don't have a border color token for white and we can't go with a closest value we have token for.

src/components/KPop/KPop.vue
portikM: Okay, fixed now

PR #1543, opened 2023-07-18
feat(kcollapse): replace utility classes

aanchalm01:
# Summary
Addresses:
[https://konghq.atlassian.net/browse/KHCP-7701](https://konghq.atlassian.net/browse/KHCP-7701)

## PR Checklist

### DIFF
--- a/docs/components/collapse.md
+++ b/docs/components/collapse.md
@@ -97,7 +97,7 @@ If no label is provided, only a caret will be displayed.
 
 KCollapse can be controlled with `v-model`.
 
-<div class="collapse-parent-wrapper">
+<div class="collapse-wrapper">
   <div>
     <KLabel>Is Collapsed:&nbsp;</KLabel>
     {{ myIsCollapsed }}
@@ -113,7 +113,7 @@ KCollapse can be controlled with `v-model`.
 </KCard>
 
 ```html
-<div class="collapse-parent-wrapper">
+<div class="collapse-wrapper">
   <div>
     <KLabel>Is Collapsed:&nbsp;</KLabel>
     {{ myIsCollapsed }}
@@ -161,7 +161,7 @@ To set the default state (collapsed or expanded) without binding to v-model you 
     <KCollapse title="Look Mah!">
       <template #trigger-content>
         <div class="trigger-wrapper">
-          <KIcon icon="help" :size="KUI_ICON_SIZE_30" color="var(--blue-500, var(--kui-color-text-primary, $kui-color-text-primary))" class="icon-style" />
+          <KIcon icon="help" :size="KUI_ICON_SIZE_30" class="icon-style" />
           Toggle
         </div>
       </template>
@@ -177,7 +177,7 @@ To set the default state (collapsed or expanded) without binding to v-model you 
 <KCollapse title="Look Mah!">
   <template #trigger-content>
     <div class="trigger-wrapper">
-      <KIcon icon="help" color="var(--blue-500, var(--kui-color-text-primary, $kui-color-text-primary))" class="icon-style" />
+      <KIcon icon="help" class="icon-style" />
       Toggle
     </div>
   </template>
@@ -251,7 +251,7 @@ An example of theming the collapse:
 
 <style lang="scss">
 .k-collapse-wrapper {
-  --KCollapseTriggerColor: var(--red-500, var(--kui-color-border-danger-weak, $kui-color-border-danger-weak));
+  --KCollapseTriggerColor: red;
 }
 </style>
 ```
@@ -270,33 +270,22 @@ export default defineComponent({
 </script>
 
 <style lang="scss" scoped>
-@mixin margin-spacing {
+.collapse-wrapper {
   margin-bottom: var(--kui-space-50, $kui-space-50) !important;
 }
 
-@mixin icon-spacing {
-  margin-right: var(--kui-space-40, $kui-space-40) !important;
-}
-
-@mixin display-type {
-  display: flex !important;
-}
-
-.collapse-parent-wrapper {
-  @include margin-spacing;
-}
-
 .icon-style {
-  @include icon-spacing;
+  margin-right: var(--kui-space-40, $kui-space-40) !important;
+  color: var(--blue-500, var(--kui-color-text-primary, $kui-color-text-primary))
 }
 
 .trigger-wrapper {
-  @include display-type;
+  display: flex !important;
 }
 </style>
 
 <style lang="scss">
 .k-collapse-wrapper {
-  --KCollapseTriggerColor: var(--red-500, var(--kui-color-border-danger-weak, $kui-color-border-danger-weak));
+  --KCollapseTriggerColor: red;
 }
 </style>
--- a/docs/components/collapse.md
+++ b/docs/components/collapse.md
@@ -97,7 +97,7 @@ If no label is provided, only a caret will be displayed.
 
 KCollapse can be controlled with `v-model`.
 
-<div class="mb-3">
+<div class="collapse-margin-bottom">
   <div>
     <KLabel>Is Collapsed:&nbsp;</KLabel>
     {{ myIsCollapsed }}
@@ -113,7 +113,7 @@ KCollapse can be controlled with `v-model`.
 </KCard>
 
 ```html
-<div class="mb-3">
+<div class="collapse-margin-bottom">
   <div>
     <KLabel>Is Collapsed:&nbsp;</KLabel>
     {{ myIsCollapsed }}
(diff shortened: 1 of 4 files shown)


src/components/KCollapse/KCollapse.vue
```diff
@@ -1,30 +1,28 @@
 <template>
-  <div class="k-collapse w-100">
+  <div class="k-collapse">
     <div
-      class="k-collapse-heading mb-3"
-      :class="trailingTrigger ? 'd-flex' : 'd-block'"
+      class="k-collapse-heading"
+      :class="trailingTrigger ? 'k-collapse-heading-flex' : 'k-collapse-heading-block'"
```
portikM, line 5: Instead of creating sub-classes for different variants, why don't you create a class `.has-trailing-trigger` (or something along those lines) and use combination of default styles (that would apply to `.k-collapse-heading`) and extend it when needed (through `.k-collapse-heading.has-trailing-trigger`)?

It would look something like:
```
:class="{ 'has-trailing-trigger': trailingTrigger }"
```

Then in your css:
```
.k-collapse-heading {
    display: flex !important;
    
    .k-collapse-title {
        margin-bottom: ... // here you can also nest default styles that apply to `.k-collapse-title` instead of creating `k-collapse-title-margin-right` since the condition is the same (`!trailingTrigger`)
    }
    ...
    &.has-trailing-trigger {
        display: block !important;
        
        .k-collapse-title {
            margin-right: ... // here you can also nest styles that apply to `.k-collapse-title` instead of creating `k-collapse-title-margin-right` since the condition is the same (`trailingTrigger === true`)
        }
    }
}
```

portikM (commented):
Great job @aanchalm01! Just a few semantic CSS suggestions and nits

docs/components/collapse.md
```diff
@@ -113,7 +113,7 @@ KCollapse can be controlled with `v-model`.
 </KCard>
 
 ```html
-<div class="mb-3">
+<div class="collapse-margin-bottom">
```
adamdehaven, line 116: issue: we don't want to name classes things that are similar to utilities.

You could have a class here like `collapse-label-wrapper` etc. or whatever makes sense. Same comment for whole PR

adamdehaven: I still don't think we want to build in classes like that; instead, we should have reusable SCSS mixins that can be applied _within_ a class.

adamdehaven: So as an example for your horizontal spacing:

```scss
@mixin horizontal-spacing {
  display: flex;
  column-gap: 4px;
}

// usage
.my-component-class {
  // other rules...
  @include horizontal-spacing;
}
```

adamdehaven: This way, we don't end up with utilities exposing themselves into our class names

portikM: Hmm I like that. @aanchalm01 let me know if you want to pair to tackle this :)

src/components/KCollapse/KCollapse.vue
```diff
@@ -37,7 +35,7 @@
                 data-testid="k-collapse-trigger-label"
               >
                 <KIcon
-                  class="k-collapse-trigger-chevron mr-1"
+                  class="k-collapse-trigger-chevron"
                   :icon="collapseIcon"
                   size="14"
```
portikM, line 40:
```suggestion
                  :size="KUI_ICON_SIZE_20"
```
issue: icon size needs to be replaced with a token (you'll need to import the const though)

aanchalm01: Should I also add in a comment about the size? Since, there is no icon size 14 and hence will be using size 12 here.

portikM: Not necessary. Just check visually to make sure the new icon size doesn't break anything or look very off. If it's all good it's expected change

docs/components/collapse.md
aanchalm01: I have pushed the recent changes.

docs/components/collapse.md
```diff
@@ -97,7 +97,7 @@ If no label is provided, only a caret will be displayed.
 
 KCollapse can be controlled with `v-model`.
 
-<div class="collapse-margin-bottom">
+<div class="collapse-parent-wrapper">
```
portikM, line 100: Can we not use the word `parent` in class name? `collapse-wrapper` is perfect

docs/components/collapse.md
```diff
@@ -269,17 +270,29 @@ export default defineComponent({
...
   margin-bottom: var(--kui-space-50, $kui-space-50) !important;
 }
 
-.icon-margin-right {
+@mixin icon-spacing {
   margin-right: var(--kui-space-40, $kui-space-40) !important;
 }
 
-.trigger-content-flex {
+@mixin display-type {
```
portikM, line 281: I think these mixins shoul live in `src/styles/_mixins.scss`. That way you can add an import of `@import '@/styles/mixins';` at the start of this style block and those mixins will be available in any other file where we might want to use them. Is that what you were thinking too @adamdehaven? If yes, then names of these mixins will need to be tweaked a little, but first let's make sure we meant the same thing.

docs/components/collapse.md
```diff
@@ -160,8 +160,8 @@ To set the default state (collapsed or expanded) without binding to v-model you
   <template #body>
     <KCollapse title="Look Mah!">
       <template #trigger-content>
-        <div class="d-flex">
-          <KIcon icon="help" size="16" color="var(--blue-500)" class="mr-2" />
+        <div class="trigger-wrapper">
+          <KIcon icon="help" :size="KUI_ICON_SIZE_30" color="var(--blue-500, var(--kui-color-text-primary, $kui-color-text-primary))" class="icon-style" />
```
adamdehaven, line 164: issue: you cannot use SCSS variables in the template.

docs/components/collapse.md
```diff
@@ -251,13 +251,14 @@ An example of theming the collapse:
 
 <style lang="scss">
 .k-collapse-wrapper {
-  --KCollapseTriggerColor: var(--red-500);
+  --KCollapseTriggerColor: var(--red-500, var(--kui-color-border-danger-weak, $kui-color-border-danger-weak));
```
adamdehaven, line 254: issue: you have to use interpolation here

docs/components/collapse.md
adamdehaven: These mixins don't need to be mixins as it's not really helpful to map one-off style rules to a mixin.

adamdehaven (changes requested):
I think there's a lot of confusion in this PR around what's expected, and better documentation is needed for this task in general. 

Please hold off on pushing anymore changes for now.

docs/components/collapse.md
aanchalm01: fixed

docs/components/collapse.md
aanchalm01: fixed

docs/components/collapse.md
aanchalm01: Talked to Adam, we don't need mixins for 1 line style properties. Moved the styles under the classes.

src/components/KCollapse/KCollapse.vue
```diff
@@ -156,18 +154,41 @@ watch(modelComputed, (newVal, oldVal) => {
...
 
+    .k-collapse-trigger-margin-left {
+      margin-left: var(--kui-space-auto, $kui-space-auto);
+    }
+
     .k-collapse-trigger-content {
-      color: var(--KCollapseTriggerColor, var(--blue-500));
-      font-size: var(--type-sm);
-      font-weight: 600;
+      color: var(--KCollapseTriggerColor, var(--blue-500, var(--kui-color-background-primary, $kui-color-background-primary)));
```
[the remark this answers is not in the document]
portikM: Please use `color-text` token for `color` property here

src/components/KCollapse/KCollapse.vue
```diff
@@ -180,16 +201,18 @@ watch(modelComputed, (newVal, oldVal) => {
...
-      .k-collapse-trigger-chevron.kong-icon {
-        &.kong-icon-chevronDown svg path,
-        &.kong-icon-chevronRight svg path {
-          stroke: var(--KCollapseTriggerColor, var(--blue-500));
+      .k-collapse-trigger-chevron {
+        margin-right: var(--kui-space-20, $kui-space-20) !important;
+        &.kong-icon {
+          &.kong-icon-chevronDown svg path,
+          &.kong-icon-chevronRight svg path {
+            stroke: var(--KCollapseTriggerColor, var(--blue-500, var(--kui-color-background-primary, $kui-color-background-primary)));
```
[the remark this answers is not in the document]
portikM: Same here, please use `color-text` token for `stroke` property here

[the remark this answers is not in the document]
aanchalm01: fixed

src/components/KCollapse/KCollapse.vue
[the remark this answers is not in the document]
aanchalm01: fixed

portikM (approved):
LGTM 🚀

adamdehaven (approved):

PR #1544, opened 2023-07-18
feat(kinput): introduce kui tokens [KHCP-7709]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-7709

Changes:
- introduces `kui-` design tokens in `KInput` component
- removes any usage of Kongponents utility classes in `KInput` component template (none found)

## PR Checklist

### DIFF
--- a/src/components/KInput/KInput.vue
+++ b/src/components/KInput/KInput.vue
@@ -324,13 +324,16 @@ export default {
   box-shadow: none !important;
 
   &.has-icon {
+    $kInputLineHeight: var(--kui-line-height-40, $kui-line-height-40);
+    // TODO: this block is repetitive and can be refactored into a mixin
     // input size medium
-    $kInputMediumSizingX: 10px;
-    $kInputMediumSizingY: var(--spacing-md, spacing(md));
-    $kInputMediumIconSize: 24px;
+    $kInputMediumSizingY: var(--kui-space-40, $kui-space-40);
+    $kInputMediumSizingX: var(--spacing-md, var(--kui-space-60, $kui-space-60));
+    $kInputMediumIconSize: var(--kui-icon-size-50, $kui-icon-size-50);
 
     ~ .input-icon {
-      top: $kInputMediumSizingX;
+      // (height of entire element (paddings + line height) - icon size) / 2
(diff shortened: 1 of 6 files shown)


src/components/KInput/KInput.vue
```diff
@@ -324,13 +324,16 @@ export default {
...
-    $kInputMediumSizingY: var(--spacing-md, spacing(md));
-    $kInputMediumIconSize: 24px;
+    $kInputMediumSizingY: var(--kui-space-40, $kui-space-40);
+    $kInputMediumSizingX: var(--spacing-md, var(--kui-space-60, $kui-space-60));
+    $kInputMediumIconSize: var(--kui-icon-size-50, $kui-icon-size-50);
 
     ~ .input-icon {
-      top: $kInputMediumSizingX;
+      // (height of entire element (paddings + line height) - icon size) / 2
+      top: calc((($kInputMediumSizingY + $kInputMediumSizingY + $kInputLineHeight) - $kInputMediumIconSize) / 2);
```
adamdehaven, line 336: question: this logic is very different than what was here before...?

Same comment applies to all the other blocks

src/components/KInput/KInput.vue
```diff
@@ -339,27 +342,28 @@ export default {
     }
 
     &.icon-start {
-      padding-left: calc($kInputMediumSizingY + var(--spacing-xs, spacing(xs)) + $kInputMediumIconSize) !important; // account for icon offset and width
+      padding-left: calc($kInputMediumSizingX + var(--spacing-xs, var(--kui-space-40, $kui-space-40)) + $kInputMediumIconSize) !important; // account for icon offset and width
       ~ .input-icon {
-        left: $kInputMediumSizingY;
+        left: $kInputMediumSizingX;
```
adamdehaven, line 347: question: variables here changed from using Y to X -- was that intentional?

Same comment applies to all of the blocks below

portikM: Yes, that is intentional - those variables got renamed because they were wrong before (notice how we were using Y variable for `left` property - which is X axis).

adamdehaven (changes requested):
Something broke here -- if I load the preview docs site, I can only type a single character into the first input on the page

PR #1572, opened 2023-07-24
feat(kskeleton): introduce kui tokens [KHCP-7723]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-7723

Changes:

- introduces `kui-` design tokens in `KSkeleton` component
- removes any usage of Kongponents utility classes in `KSkeleton` component template

## PR Checklist

### DIFF
--- a/src/components/KSkeleton/KSkeleton.vue
+++ b/src/components/KSkeleton/KSkeleton.vue
@@ -1,8 +1,8 @@
 <template>
   <div
     v-if="isVisible"
-    class="d-flex flex-wrap k-skeleton-container"
-    :class="{ 'w-100': type !== 'spinner', 'opacity-0': !isVisible }"
+    class="k-skeleton-container"
+    :class="{ 'is-spinner': type !== 'spinner' }"
   >
     <CardSkeleton
       v-if="type === 'card'"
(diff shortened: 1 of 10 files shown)


src/components/KSkeleton/KSkeletonBox.vue
```diff
@@ -39,8 +39,9 @@ defineProps({
     )
     repeat;
   background-size: 400% 100%;
-  border-radius: 3px;
+  border-radius: var(--kui-border-radius-10, $kui-border-radius-10);
   display: inline-flex;
+  margin-right: var(--kui-space-20, $kui-space-20) !important;
   transition: 1s;
```
portikM, line 45: For review: decided to leave some of the values untouched here (animation timing, gradient) as they are very specific to loader component, rather than our UIs so I don't see value in tokenizing those. Open to feedback.

src/components/KSkeleton/KSkeleton.vue
```diff
@@ -1,8 +1,8 @@
 <template>
   <div
     v-if="isVisible"
-    class="d-flex flex-wrap k-skeleton-container"
-    :class="{ 'w-100': type !== 'spinner', 'opacity-0': !isVisible }"
```
jillztom, line 5: We don't need opacity css here anymore?

portikM: We have a `v-if="isVisible"` on that element. So when `isVisible` (which is a prop) is false - the element won't render - which means that class will never apply. Now that I look at it again, I will add a commend to remove `isVisible` prop in the breaking change.

jillztom (approved):

jillztom (approved):

PR #1574, opened 2023-07-24
feat(kmodalfullscreen): introduce kui tokens [KHCP-7715]

aanchalm01:
# Summary
Addresses:
[https://konghq.atlassian.net/browse/KHCP-7715](https://konghq.atlassian.net/browse/KHCP-7715)

## PR Checklist

### DIFF
--- a/docs/components/modal-fullscreen.md
+++ b/docs/components/modal-fullscreen.md
@@ -477,14 +477,14 @@ export default {
 }
 
 .button-spacing {
-  margin-right: var(--kui-space-40, $kui-space-40) !important;
+  margin-right: $kui-space-40;
 }
 
 .k-switch {
-  border-top: var(--kui-border-width-10, $kui-border-width-10) solid #eaecef;
-  padding-top: var(--kui-space-80, $kui-space-80);
-  border-bottom: var(--kui-border-width-10, $kui-border-width-10) solid #eaecef;
-  padding-bottom: var(--kui-space-80, $kui-space-80);
+  border-top: $kui-border-width-10 solid #eaecef;
+  padding-top: $kui-space-80;
+  border-bottom: $kui-border-width-10 solid #eaecef;
+  padding-bottom: $kui-space-80;
 }
 
 .display {
@@ -504,20 +504,20 @@ export default {
   input {
     width: 100%;
     height: 80%;
-    border-radius: var(--kui-border-radius-20, $kui-border-radius-20);
-    border: var(--kui-border-width-10, $kui-border-width-10) solid var(--grey-300, var(--kui-color-border-neutral-weak, $kui-color-border-neutral-weak));
-    background-color: var(--white, var(--kui-color-background, $kui-color-background));
-    text-indent: var(--kui-space-40, $kui-space-40);
+    border-radius: $kui-border-radius-20;
+    border: $kui-border-width-10 solid var(--grey-300, $kui-color-border-neutral-weak);
+    background-color: var(--white, $kui-color-background);
+    text-indent: $kui-space-40;
   }
 
   label {
     position: absolute;
     top: -10px;
     left: 15px;
-    padding: var(--kui-space-20, $kui-space-20);
-    background-color: var(--white, var(--kui-color-background, $kui-color-background));
-    color: var(--grey-500, var(--kui-color-text-neutral, $kui-color-text-neutral));
-    font-size: var(--kui-font-size-10, $kui-font-size-10);
+    padding: $kui-space-20;
+    background-color: var(--white, $kui-color-background);
+    color: var(--grey-500, $kui-color-text-neutral);
+    font-size: $kui-font-size-10;
   }
 }
 </style>
--- a/docs/components/modal-fullscreen.md
+++ b/docs/components/modal-fullscreen.md
@@ -169,13 +169,13 @@ There are 6 designated slots you can use to display content in the fullscreen mo
     Install Plugin
   </template>
    <template #body-header>
-    <div class="ml-25">Configure a key auth plugin</div>
+    <div class="horizontal-spacing">Configure a key auth plugin</div>
   </template>
   <template #body-header-description>
-    <div class="ml-25">Lorem ipsum factum. <a>View documentation</a></div>
+    <div class="horizontal-spacing">Lorem ipsum factum. <a>View documentation</a></div>
   </template>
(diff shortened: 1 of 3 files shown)


docs/components/modal-fullscreen.md
```diff
@@ -472,15 +472,19 @@ export default {
...
+
 .k-switch {
-  border-top: 1px solid #eaecef;
-  padding-top: 26px;
-  border-bottom: 1px solid #eaecef;
-  padding-bottom: 26px;
+  border-top: var(--kui-border-width-10, $kui-border-width-10) solid #eaecef;
+  padding-top: var(--kui-space-80, $kui-space-80);
+  border-bottom: var(--kui-border-width-10, $kui-border-width-10) solid #eaecef;
+  padding-bottom: var(--kui-space-80, $kui-space-80);
```
adamdehaven, line 487: If this is in the markdown page docs styles section, you don't need the CSS variables and can just use the Sass tokens. 

They are only to be put in place where we need to allow for customization like in the component itself.

src/components/KModalFullscreen/KModalFullscreen.vue
```diff
@@ -260,40 +261,40 @@ $fullscreen-modal-padding: 64px;
...
-  border-bottom: 1px solid var(--grey-300);
+  background-color: var(--white, var(--kui-color-background, $kui-color-background));
+  border-bottom: var(--kui-border-width-10, $kui-border-width-10) solid var(--grey-300, var(--kui-color-border-neutral-weak, $kui-color-border-neutral-weak));
   display: flex;
   flex-direction: column;
-  padding: var(--spacing-lg) 0;
+  padding: var(--spacing-lg, var(--kui-space-80, $kui-space-80)) var(--kui-space-0, $kui-space-0);
   position: fixed;
-  top: 0;
+  top: var(--kui-space-0, $kui-space-0);
```
adamdehaven, line 270: issue: we don't use space tokens for positioning

This can be reverted. Same comment for all instances of positioning

src/components/KModalFullscreen/KModalFullscreen.vue
```diff
@@ -302,42 +303,43 @@ $fullscreen-modal-padding: 64px;
...
   & button,
   & :deep(button) {
-    font-size: 13px;
-    font-weight: 600;
-    height: 40px;
-    line-height: 13px;
-    margin-left: var(--spacing-md, spacing(md));
+    font-size: var(--kui-font-size-20, $kui-font-size-20);
+    font-weight: var(--kui-font-weight-semibold, $kui-font-weight-semibold);
+    height: var(--kui-space-100, $kui-space-100);
```
adamdehaven, line 326: We don't want to use space tokens for height and width, etc.

src/components/KModalFullscreen/KModalFullscreen.vue
```diff
@@ -302,42 +303,43 @@ $fullscreen-modal-padding: 64px;
...
+    line-height: var(--kui-line-height-10, $kui-line-height-10);
```
adamdehaven, line 327: The line height token should typically match the font size token
```suggestion
    line-height: var(--kui-line-height-20, $kui-line-height-20);
```

src/components/KModalFullscreen/KModalFullscreen.vue
```diff
@@ -302,42 +303,43 @@ $fullscreen-modal-padding: 64px;
...
-  padding-left: var(--spacing-lg);
-  padding-right: var(--spacing-lg);
+  color: var(--KModalFullscreenColor, var(--black-500, var(--kui-color-text, $kui-color-text)));
+  padding-left: var(--spacing-lg, var(--kui-space-80, $kui-space-80));
+  padding-right: var(--spacing-lg, var(--kui-space-80, $kui-space-80));
 
-  @media (min-width: $viewport-md) {
+  @media (min-width: $kui-breakpoint-phablet) {
     padding-left: 120px;
     padding-right: 120px;
```
portikM, line 344:
```suggestion
    padding-left: $tmp-spacing-120;
    padding-right: $tmp-spacing-120;
```

portikM: Same for other `$tmp` variables. But if we decide it doesn't become a token, we can replace it's usages easily later.

aanchalm01: This file also have:
```
 padding-left: 230px;
 padding-right: 230px;
```

portikM: I would create and use `$tmp` variables for those as well. Like I said above, it's easy to drop them later and switch back to pixel values.

adamdehaven: I don't think it makes sense to map these one-off sizes to temporary variables. Are they used anywhere else in the repo? If no, they shouldn't need a variable.

aanchalm01: I don't see 230px anywhere else, 120px is being used in 2 components..

portikM: Maybe that's true for `230px`, but 120px is used somewhere else (I don't remember exactly where, but I created and used it). The only way to know that it's a "one off" value is if the $tmp variable created for that value is only used once. When I or Aanchal encounters `230px` for the next time, we are not gonna remember that we had this value somewhere els=se before therefore now it deserves it's own `$tmp` variable.

portikM: I'm not sure I understand the question... they both use value `120px` but they are not related as they live in different components
> But are they used in two places because the values are directly related, or because they just happen to use the same value?

adamdehaven: If the temporary token that points to 120px changes to 220px, should both components change?

portikM: Just looked up, `$tmp-spacing-120` is used for positioning in one other component. Safe to say it's not needed there (I will fix that one occurrence in my next PR). Since there are no other usages of `120px`, I'm okay with leaving this a pixel value (same for `230px`).

aanchalm01:
@adamdehaven , @portikM - all the feedback has been addressed.

adamdehaven (approved):

PR #1579, opened 2023-07-25
feat(kmultiselect): introduce kui tokens [KHCP-7716]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-7716

Changes:

- introduces `kui-` design tokens in `KMultiselect` component
- removes any usage of Kongponents utility classes in `KMultiselect` component template

## PR Checklist

### DIFF
--- a/src/components/KMultiselect/KMultiselect.vue
+++ b/src/components/KMultiselect/KMultiselect.vue
@@ -1189,7 +1189,7 @@ onBeforeUnmount(() => {
       // Firefox workaround
       // since :has() selector isn't supported in Firefox be default
       .k-multiselect-list ~ .k-multiselect-dropdown-footer-sticky {
-        bottom: var(--kui-space-0, $kui-space-0);
+        bottom: 0;
         position: sticky;
       }
     }
--- a/src/components/KMultiselect/KMultiselect.vue
+++ b/src/components/KMultiselect/KMultiselect.vue
@@ -1066,8 +1066,8 @@ onBeforeUnmount(() => {
   }
 
   .k-multiselect-list {
-    margin: var(--kui-space-0) !important;
-    padding: var(--kui-space-0) !important;
+    margin: var(--kui-space-0, $kui-space-0) !important;
+    padding: var(--kui-space-0, kui-space-0) !important;
   }
 
   .k-multiselect-new-item {
--- a/src/components/KMultiselect/KMultiselect.vue
+++ b/src/components/KMultiselect/KMultiselect.vue
@@ -1000,22 +1000,22 @@ onBeforeUnmount(() => {
 
   .k-multiselect-icon {
     position: absolute;
-    right: var(--kui-space-10, $kui-space-10);
-    top: var(--kui-space-10, $kui-space-10);
+    right: 1px;
+    top: 1px;
     z-index: 1;
 
     .k-multiselect-chevron-icon {
       position: relative;
-      right: var(--kui-space-40, $kui-space-40);
-      top: var(--kui-space-40, $kui-space-40);
+      right: 10px;
+      top: 11px;
     }
 
     .k-multiselect-clear-icon {
       @include non-visual-button;
       padding: var(--kui-space-0, $kui-space-0) !important;
       position: absolute;
-      right: var(--kui-space-40, $kui-space-40);
-      top: var(--kui-space-40, $kui-space-40);
+      right: 10px;
+      top: 8px;
     }
   }
 
@@ -1140,7 +1140,6 @@ onBeforeUnmount(() => {
   .k-multiselect-popover {
     box-sizing: border-box;
     margin-top: var(--kui-space-0, $kui-space-0) !important;
-    margin-top: var(--kui-space-10, $kui-space-10) !important;
     width: 100%;
 
     &[x-placement^="top"] {
--- a/src/components/KMultiselect/KMultiselect.vue
+++ b/src/components/KMultiselect/KMultiselect.vue
@@ -39,7 +39,7 @@
         >
           <div
             ref="multiselectRef"
-            class="k-multiselect-trigger w-100"
+            class="k-multiselect-trigger"
             :class="{ focused: isFocused, hovered: isHovered, disabled: isDisabled, readonly: isReadonly }"
             data-testid="k-multiselect-trigger"
             role="listbox"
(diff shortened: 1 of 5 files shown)


src/components/KMultiselect/KMultiselect.vue
```diff
@@ -954,56 +954,76 @@ onBeforeUnmount(() => {
...
+      }
     }
   }
 
   .k-multiselect-icon {
     position: absolute;
-    right: 1px;
-    top: 1px;
+    right: var(--kui-space-10, $kui-space-10);
+    top: var(--kui-space-10, $kui-space-10);
```
adamdehaven, line 1004: shouldn't use space tokens for positioning; all PR instances

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -1108,32 +1139,33 @@ onBeforeUnmount(() => {
 
   .k-multiselect-popover {
     box-sizing: border-box;
-    margin-top: 2px !important;
+    margin-top: var(--kui-space-0, $kui-space-0) !important;
+    margin-top: var(--kui-space-10, $kui-space-10) !important;
```
adamdehaven, line 1143: issue: these two rules conflict

src/components/KMultiselect/KMultiselectItem.vue
```diff
@@ -74,27 +76,32 @@ const handleClick = (): void => {
...
     display: flex;
-    font-family: var(--font-family-sans);
-    font-size: var(--type-xs);
-    font-weight: 400;
-    line-height: 26px;
-    padding-left: var(--spacing-xxs);
+    font-family: var(--font-family-sans, var(--kui-font-family-text, $kui-font-family-text));
+    font-size: var(--type-xs, var(--kui-font-size-20, $kui-font-size-20));
+    font-weight: var(--kui-font-weight-regular, $kui-font-weight-regular);
+    line-height: var(--kui-line-height-40, $kui-line-height-40);
```
adamdehaven, line 103: question: shouldn't the line-height match the font size token?

adamdehaven (changes requested):

src/components/KMultiselect/KMultiselectItem.vue
portikM: No necessarily - sometime we might wanna give it some extra room maybe?

src/components/KMultiselect/KMultiselect.vue
portikM: Fixed

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -1034,35 +1054,46 @@ onBeforeUnmount(() => {
...
       }
+
+      &.is-toggled {
+        margin-top: var(--kui-space-20, $kui-space-20) !important;
+      }
     }
   }
 
+  .k-multiselect-list {
+    margin: var(--kui-space-0) !important;
```
kaiarrowood, line 1069: question: why is this `var()` with no fallback?

question: It seems like there is a significant increase in the usage `!important` for styles post tokenization. Is this desired? It seems like we should be trying to use more selective selectors to take precedence instead.

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -1108,32 +1139,33 @@ onBeforeUnmount(() => {
...
+      margin-top: var(--kui-space-0, $kui-space-0) !important;
     }
 
     &.k-multiselect-pop {
-      --KPopPaddingY: var(--spacing-xs);
-      --KPopPaddingX: var(--spacing-xs);
-      border: 1px solid var(--black-10);
+      --KPopPaddingY: var(--spacing-xs, var(--kui-space-40, #{$kui-space-40}));
+      --KPopPaddingX: var(--spacing-xs, var(--kui-space-40, #{$kui-space-40}));
+      border: var(--kui-border-width-10, $kui-border-width-10) solid $tmp-color-black-10;
```
kaiarrowood, line 1153: question: can you explain the need for `$tmp-color-black-10`? What happens to these when we do the big clean up effort to remove refs to the previous color vars? Or is this like a TODO saying we need this color to be a token?

src/components/KMultiselect/KMultiselectItem.vue
```diff
@@ -29,16 +29,17 @@
           >
             <KIcon
               class="selected-item-icon"
-              :color="item.selected ? 'var(--blue-200)' : 'var(--grey-400)'"
+              :class="{ 'is-item-selected': item.selected }"
+              :color="item.selected ? 'currentColor' : `var(--grey-400, var(--kui-color-text-neutral-weak, ${KUI_COLOR_TEXT_NEUTRAL_WEAK}))`"
```
kaiarrowood, line 33: question: why the switch to `currentColor` here? Which color does it get? I feel like readability suffers a bit with the usage

src/components/KMultiselect/KMultiselectItem.vue
```diff
@@ -106,62 +113,67 @@ const handleClick = (): void => {
...
       }
     }
 
     .kong-icon:not(.selected-item-icon) {
-      margin-right: var(--spacing-sm);
+      margin-right: var(--spacing-sm, var(--kui-space-50, $kui-space-50));
     }
 
     .k-multiselect-selected-icon-container {
       height: var(--spacing-lg);
```
kaiarrowood, line 161: What about this guy and the other references to `var(--spacing-lg)` for `width`?

src/components/KMultiselect/KMultiselectItem.vue
portikM: Switched to `currentColor` as a workaround to use one of the `$tmp-` variables for color. See `is-item-selected` class definition. We don't have a token for color I needed to use here so I needed to create a `$tmp-` variable - but because we can't use SCSS variables in the template (they won't compile) - I needed a way to assign color in style tag.

src/components/KMultiselect/KMultiselectItem.vue
adamdehaven: That doesn't match what Jason told us. Can you use the corresponding tokens?

portikM: Done

src/components/KMultiselect/KMultiselectItem.vue
kaiarrowood: But won't the old spacing vars eventually go away? I'd think then, we want to hardcode the value to `24px` (or whatever it is) instead? Or at least provide a fallback?

adamdehaven: Just hard-code the value here as the fallback.

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -1078,7 +1109,7 @@ onBeforeUnmount(() => {
       }
 
       &.input-placeholder-dark::placeholder {
-        color: var(--KInputColor, var(--black-70, rgba(0, 0, 0, 0.7))) !important;
+        color: var(--KInputColor, var(--black-70, var(--kui-color-text, $kui-space-40))) !important;
```
adamdehaven, line 1112: issue: this space token shouldn't be here

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -1158,7 +1189,7 @@ onBeforeUnmount(() => {
       // Firefox workaround
       // since :has() selector isn't supported in Firefox be default
       .k-multiselect-list ~ .k-multiselect-dropdown-footer-sticky {
-        bottom: 0;
+        bottom: var(--kui-space-0, $kui-space-0);
```
adamdehaven, line 1192: issue: don't use a space token for positioning

portikM: Fixed

src/components/KMultiselect/KMultiselect.vue
portikM: Oops, good catch, fixed

src/components/KMultiselect/KMultiselectItem.vue
portikM: Done

adamdehaven (approved):

PR #1582, opened 2023-07-25
feat(ktruncate): introduce kui tokens [KHCP-7732]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-7732

Changes:

- introduces `kui-` design tokens in `KTruncate` component
- removes any usage of Kongponents utility classes in `KTruncate` component template

## PR Checklist

### DIFF
--- a/src/components/KTruncate/KTruncate.vue
+++ b/src/components/KTruncate/KTruncate.vue
@@ -8,7 +8,7 @@
     <!-- Order switched for ease when using keyboard navigation -->
     <div
       v-if="!isTextContent && showToggle"
-      class="d-flex h-100 align-items-end"
+      class="k-truncate-expand-controls"
     >
       <div
         v-if="!expanded"
@@ -49,9 +49,9 @@
             @click="handleToggleClick"
           >
             <KIcon
-              color="var(--KTruncateCollapseIconColor, var(--blue-500, color(blue-500)))"
+              :color="`var(--KTruncateCollapseIconColor, var(--blue-500, var(--kui-color-text-primary, ${KUI_COLOR_TEXT_PRIMARY})))`"
               icon="chevronUp"
-              size="10"
+              :size="KUI_ICON_SIZE_10"
               title="Show less"
             />
           </KButton>
@@ -61,7 +61,7 @@
     <div
       v-if="isTextContent && (showToggle || expanded)"
       ref="textToggleControls"
-      class="place-self-end mt-2"
+      class="k-truncate-collapse-controls"
     >
       <div
         v-if="!expanded"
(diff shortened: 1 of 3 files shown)


src/components/KTruncate/KTruncate.vue
```diff
@@ -134,14 +136,15 @@ const kTruncateContainer = ref<HTMLDivElement>()
 const kTruncateWrapper = ref<HTMLDivElement>()
 const textToggleControls = ref<HTMLDivElement>()
 const wrapperHeight = ref<string>('0px')
+const gap = KUI_SPACE_40
```
adamdehaven, line 139: question: why do we need to set this to a `gap` variable instead of just referencing the `KUI_SPACE_40` variable directly?

src/components/KTruncate/KTruncate.vue
```diff
@@ -150,12 +153,13 @@ const setWrapperHeight = async (): Promise<void> => {
 
   if (kTruncateContainer.value && kTruncateContainer.value.children?.length) {
     const children = kTruncateContainer.value.children as unknown as HTMLElement[]
+    const gapNumber = Number(String(gap).replace(/px/gi, ''))
```
adamdehaven, line 156: Just in case 😬 
```suggestion
    const gapNumber = Number(String(gap).replace(/px$/gi, ''))
```

src/components/KTruncate/KTruncate.vue
portikM: Because we reference it twice - in case if that value changes at some point, so we don't forget about the other occurrence

adamdehaven: I'm not sure I understand. Why can't you just reference the space const?

portikM: Added a comment there. Basically this gap is used in one of the functions to perform some calculations. But it's also used in styles. So it's important, if we ever change one we use in styles, to not forget to update it in script too. If it's set to a local const - it's only defined in one place.

portikM: Does that make sense?

adamdehaven: you're talking about here? Why can't you reference in both?
```
gap: v-bind('KUI_SPACE_40');
```

portikM: Oh sorry - yes I can (don't even have to use JS const - I can use SCSS one instead) but what I was trying to do is to not allow this value be changed in styles but not in script. Basically if tomorrow we decide we want a bigger gap there a developer might come it and change it to:
```
gap: v-bind('KUI_SPACE_50');
```

However the change in script will remain untouched and that will break the component. If it's defined in one place, they will change the definition:
```
const gap = KUI_SPACE_50
```

But both usages in style and in script will reference the same value.

adamdehaven: Ah ok that makes more sense. Can you leave a comment above the gap variable explaining?

portikM: Done

adamdehaven (dismissed):

adamdehaven (approved):

PR #1631, opened 2023-08-16
feat(ktable): adds props.paginationPageNumber

kleinfreund:
# Summary

Adds support for `props.paginationPageNumber` which allows one to set KTable’s current page from an external source. This behaves exactly as interactions with the built-in pagination do.

## Notes

- I think this ability should be built into KTable. Currently, the only alternative would be updating the table’s `key` prop whenever a page changes externally. This would trigger `KTable`’s `onMounted` hook and let it read the now-changed page number from `props.initialFetcherParams`. While this works, I consider it a hack. It shouldn’t require a re-mount of KTable to change the page number in situations that KTable can’t control (e.g. browser history navigation).

## PR Checklist

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -485,6 +485,10 @@ If not provided the fetcher response should return a top-level property `total` 
 
 Pass in a number of pagination neighbors to be used by the pagination component. See more detail in the [Pagination](/components/pagination.html#neighbors) docs.
 
+### paginationPageNumber
+
+Can be used to update the current page from a source external to KTable (e.g. in response to a browser history navigation). Behaves exactly as interactions with the built-in pagination do.
+
 ### paginationPageSizes
 
 Pass in an array of page sizes for the page size dropdown. If not provided will default to the following:
--- a/src/components/KTable/KTable.vue
+++ b/src/components/KTable/KTable.vue
@@ -473,6 +473,13 @@ const props = defineProps({
     default: 1,
   },
   /**
+   * Can be used to update the current page from a source external to KTable (e.g. in response to a browser history navigation). Behaves exactly as interactions with the built-in pagination do.
+   */
+  paginationPageNumber: {
+    type: Number,
+    default: null,
+  },
+  /**
    * A prop to pass in an array of page sizes used by the pagination component
    */
   paginationPageSizes: {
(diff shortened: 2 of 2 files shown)


adamdehaven:
@kleinfreund I have the same concerns here as we chatted about yesterday as I don't understand the use-case for this prop.

Also, what happens if two things change the page simultaneously, e.g. the table itself and the host app prop are out of sync?

kleinfreund:
> I have the same concerns here as we chatted about yesterday as I don't understand the use-case for this prop.

I think the use case is pretty well explained. My assumption is that it’s expected by users that paginated documents (e.g. once containing tables) can be navigated using regular browser semantics including history navigation. For that to work, pagination state needs to be persisted on the client side (ideally in the URL). I think it’s reasonable that _that_ duty does _not_ fall to KTable as it can’t know _how_ a host application will do that.

With that assumption, I think KTable should support a way to control the current page. The next best alternative (which I think is an undesirable compromise) would be disabling KTable’s pagination (there’s a prop for that) and using KPagination in the host application. Then this can be wholly controlled on the host application side, but considering how easily KTable can be put in the position to handle this scenario, I think it’s far preferable to do that than hand-write the pagination interactions.

> Also, what happens if two things change the page simultaneously, e.g. the table itself and the host app prop are out of sync?

I think it can reasonably be assumed that, if someone decides to control the current page of a KTable instance from the outside, they would be the responsibly party for keeping the URLs up-to-date with what KTable provides (e.g. as a side effect of calling `props.fetcher` which will always hold that current page).

As far as I can tell, what I propose to add here can’t make the internal state of KTable inconsistent. If two interactions happen in quick succession, it’s either the new prop triggering last and superseding a KTable-internal interaction or the other way around. In both cases, two calls to `props.fetcher` would be triggered (unless they happen during the same event loop cycle I guess? Even then it’s consistently going to be the last one) and I think that’s very much expected.

adamdehaven:
> it’s expected by users that paginated documents (e.g. once containing tables) can be navigated using regular browser semantics including history navigation

I personally don’t agree with this assumption at all. The table is a component on a page; not the page itself. 

Coupling URL navigation to KTable IMO is outside the scope of the component and not the intended usage; however, if you’re wanting to control the KTable page based on the URL, the suggested incrementing of a key prop on the KTable to remount the component with an updated initial page seems acceptable IMO and is supported with the current implementation without introducing additional watchers and new props that are essentially doing the same thing, albeit without a remount.

adamdehaven:
This brings up a good point that the current implementation isn’t ideal for a lot of implementations and should likely allow the host app to perform the data fetching…

PR #1641, opened 2023-08-22
fix(tsconfig): extend vue/tsconfig

adamdehaven:
# Summary

Extend `@vue/tsconfig` and fix type imports

via https://github.com/Kong/public-ui-components/pull/729

## PR Checklist


DariaYeremina (approved):

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -280,13 +280,15 @@
 </template>
 
 <script setup lang="ts">
-import { computed, nextTick, onBeforeUnmount, onMounted, ref, useSlots, watch, PropType } from 'vue'
+import type { PropType } from 'vue'
+import { computed, nextTick, onBeforeUnmount, onMounted, ref, useSlots, watch } from 'vue'
 
-import KButton from '@/components/KButton/KButton.vue'
+import type KButton from '@/components/KButton/KButton.vue'
```
kleinfreund, line 286: **Issue**: This rule doesn't seem to work correctly. This import is not a type-only import. The component uses `KButton` in the template. This breaks now in projects that don’t register Kongponents globally.

kleinfreund: I know why this happens.

In this component, there are two lines containing `typeof KButton`. This seems to trip up the rule. Replacing those with `any` (for example) removes the false positive and the `KButton` import is allowed.

Not sure how to deal with this.

PR #1662, opened 2023-08-29
fix: making lefthook working, removing commitLint workflow [KHCP-8829]

ValeryG:
# Summary

- renaming `lefthook.yml to `lefthook.yaml` so that `yarn lefthook install` picks it up
- removing commitLint workflow


.github/CODEOWNERS
```diff
@@ -5,7 +5,7 @@
 
 # Workflows & CI
 /.github/ @adamdehaven @jillztom
-/lefthook.yml @adamdehaven @jillztom
+/lefthook.yaml @adamdehaven @jillztom
```
mihai-peteu, line 8: 🙃

mihai-peteu (approved):

.github/workflows/commitlint.yml
```diff
@@ -1,14 +0,0 @@
-name: Lint Commit Messages
```
adamdehaven, line 1: Why is this being removed? This enforces our commit hooks that are required for semantic release

ValeryG: This is all in JIRA and link to the slack thread. commitLints are part of the pre-commit hocks and enabled in left hook (wasn't working due to `.yaml`). This is the way we are doing things in other repos (public-ui-components), unification is good. listing commit messages after commit happens is too late in the game...

adamdehaven: So what happens if someone skips the pre-commit hooks when they push their commits? I don't see the harm in leaving the check in place

kaiarrowood (approved):

.github/workflows/commitlint.yml
ValeryG: same thing what happens in public-ui-components and other repos. If we need  keep  this workflow in this repo but not in any other public repos - we need a better reason and / or comment in this workflow as of why it so unique and why it needs to be there. Another reason to drop it - those failures we see for 2 weeks. https://github.com/Kong/kongponents/actions/workflows/commitlint.yml?page=2 - those are fixable but there is no reason to fix it. remove workflow is cleaner.

jillztom (approved):

adamdehaven:
@jillztom there is now nothing in this repository enforcing the commit format, as git hooks can easily be skipped. 

I’m very opposed as to why this PR was merged.

jillztom:
> @jillztom there is now nothing in this repository enforcing the commit format, as git hooks can easily be skipped.
> 
> I’m very opposed as to why this PR was merged.

Let's discuss this when you get back.

PR #1677, opened 2023-09-06
fix(multi-select): optimize overflow detection for selected items container

nekolab:
Issues observed with `<MultiSelect />`:
1. Excessive calls to `stageSelections()`. The current implementation schedules `items.length` macro tasks while `items` changed, leading to forced DOM reflows due to frequent access to `clientHeight`, and multiple changes to `stagingKey`.

2. The current strategy to prevent item container flickering involves the use of an offscreen div to gauge the maximum number of pills it can accommodate. This involves populating the pills and checking for remaining space sequentially, leading to a fixed `len(selected)` times DOM updates. Large selected items adversely affect performance.

Optimizations:
- Moved `stageSelections()` outside the loop to schedule only a single task post processing of all items.
- Modified the pill measurement strategy to maximize population at once by checking the `offsetTop`, reducing unnecessary DOM updates.

# Summary

## PR Checklist

### DIFF
--- a/src/components/KMultiselect/KMultiselect.cy.ts
+++ b/src/components/KMultiselect/KMultiselect.cy.ts
@@ -576,4 +576,23 @@ describe('KMultiselect', () => {
     cy.get('.k-multiselect-item').eq(3).should('contain.text', items[2].label)
     cy.get('.k-multiselect-item').eq(4).should('contain.text', items[4].label)
   })
+
+  it('should able to handle tons of items with no obvious lag', () => {
+    const items = Array.from(new Array(500)).map((_, i) => ({
+      label: `Item ${i}`,
+      value: `${i}`,
+      selected: i < 400,
+    }))
+
+    const startTime = Date.now()
+
+    mount(KMultiselect, {
+      props: {
+        testMode: true,
+        items,
+      },
+    }).then(() => {
+      expect(Date.now() - startTime).to.be.lessThan(3000)
+    })
+  })
 })
--- a/src/components/KMultiselect/KMultiselect.vue
+++ b/src/components/KMultiselect/KMultiselect.vue
@@ -604,10 +604,19 @@ const stageSelections = () => {
     if (elem) {
       const height = elem.clientHeight
       if (height > selectionsMaxHeight.value) {
-        const item = visibleSelectedItemsStaging.value.pop()
-        if (item && !invisibleSelectedItemsStagingSet.has(item.value)) {
-          invisibleSelectedItemsStagingSet.add(item.value)
-          invisibleSelectedItemsStaging.value.push(item)
+        // populate as much items as possible by checking the offsetTop
+        const overflowedElements = Array.from(elem.querySelectorAll('.k-multiselect-selection-badge'))
+          .filter(badge => (badge as HTMLElement).offsetTop >= selectionsMaxHeight.value)
+
+        // if there are overflowed items, move them to the invisibleSelectedItemsStaging array
+        const cutPoint = visibleSelectedItemsStaging.value.length - overflowedElements.length
+        const overflowedItems = visibleSelectedItemsStaging.value.splice(cutPoint, overflowedElements.length)
+
+        for (const item of overflowedItems) {
+          if (!invisibleSelectedItemsStagingSet.has(item.value)) {
+            invisibleSelectedItemsStagingSet.add(item.value)
+            invisibleSelectedItemsStaging.value.push(item)
+          }
         }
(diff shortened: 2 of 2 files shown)


nekolab:
Comparison for the case: 100 items with 90 selected

Before:
image

After:
image

src/components/KMultiselect/KMultiselect.cy.ts
```diff
@@ -576,4 +576,23 @@ describe('KMultiselect', () => {
...
+
+    const startTime = Date.now()
+
+    mount(KMultiselect, {
+      props: {
+        testMode: true,
+        items,
+      },
+    }).then(() => {
+      expect(Date.now() - startTime).to.be.lessThan(3000)
```
adamdehaven, line 595: question: why did you choose 3 seconds? Isn’t that still a considerable amount of lag?

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -604,11 +604,20 @@ const stageSelections = () => {
     if (elem) {
       const height = elem.clientHeight
       if (height > selectionsMaxHeight.value) {
-        const item = visibleSelectedItemsStaging.value.pop()
-        if (item && !invisibleSelectedItemsStagingSet.has(item.value)) {
-          invisibleSelectedItemsStagingSet.add(item.value)
-          invisibleSelectedItemsStaging.value.push(item)
-        }
+        // populate as much items as possible by testing the offsetTop
+        const $overflowedItems = Array.from(elem.querySelectorAll('.k-multiselect-selection-badge'))
```
adamdehaven, line 608: nit: we do not prefix variable names with `$` or any other prefix. Since there are two variants of the same variable name, it would be preferable to just name them differently. 

Can you please update all occurrences?

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -604,11 +604,20 @@ const stageSelections = () => {
...
+          .filter($badge => ($badge as HTMLElement).offsetTop >= selectionsMaxHeight.value)
+
+        // if there are overflowed items, move them to the invisibleSelectedItemsStaging array
+        const cutPoint = visibleSelectedItemsStaging.value.length - $overflowedItems.length
+        const overflowedItems = visibleSelectedItemsStaging.value.splice(cutPoint, $overflowedItems.length)
+
+        overflowedItems.forEach(item => {
```
adamdehaven, line 615: A regular `for` loop would be more efficient here

src/components/KMultiselect/KMultiselect.cy.ts
nekolab: Rendering 500 items does generate some pressure on the browser. On my Macbook, it took me ~380ms to render the component. Considering the CI machine is not that powerful, multiple 7-8 looks safe to me, so I chose 3 seconds.
Current implementation on CI takes ~1.2s, in comparison, the original implementation will take 40+ seconds or even cause the page to crash. I have to say the 3 seconds is a bit arbitrary, but I think it's okay to be the bottom line

adamdehaven: As long as we don't consistently see failures in the CI, that makes sense. Thanks for the extra info

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -604,10 +604,19 @@ const stageSelections = () => {
     if (elem) {
       const height = elem.clientHeight
       if (height > selectionsMaxHeight.value) {
-        const item = visibleSelectedItemsStaging.value.pop()
-        if (item && !invisibleSelectedItemsStagingSet.has(item.value)) {
-          invisibleSelectedItemsStagingSet.add(item.value)
-          invisibleSelectedItemsStaging.value.push(item)
+        // populate as much items as possible by checking the offsetTop
+        const overflowedElements = Array.from(elem.querySelectorAll('.k-multiselect-selection-badge'))
+          .filter($badge => ($badge as HTMLElement).offsetTop >= selectionsMaxHeight.value)
```
Leopoldthecoder, line 609: Do we want to remove the `$` prefix here as well?
```suggestion
          .filter(badge => (badge as HTMLElement).offsetTop >= selectionsMaxHeight.value)
```

nekolab: Ooooops

Leopoldthecoder (approved):
LGTM! Thanks for digging into this

adamdehaven (approved):
lgtm

PR #1720, opened 2023-09-12
fix(kbadge): non-interactive badge being focusable

kleinfreund:
# Summary

Fixes KBadge being able to receive focus when it’s not interactive. It now will only have `tabindex="0"` if a `click` handler is present.

Removes `tabindex="0"` from KBadge’s dismiss button if `props.hidden` isn’t true. Setting `tabindex="0"` is unnecessary as KButton is already focusable by virtue of being a button.

## PR Checklist

### DIFF
--- a/src/components/KBadge/KBadge.vue
+++ b/src/components/KBadge/KBadge.vue
@@ -8,7 +8,7 @@
       'clickable': isClickable
     } ]"
     :style="badgeCustomStyles"
-    :tabindex="hidden ? -1 : 0"
+    :tabindex="hidden ? -1 : isClickable ? 0 : undefined"
   >
     <component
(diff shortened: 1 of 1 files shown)


src/components/KBadge/KBadge.vue
```diff
@@ -8,7 +8,7 @@
       'clickable': isClickable
     } ]"
     :style="badgeCustomStyles"
-    :tabindex="hidden ? -1 : 0"
+    :tabindex="hidden ? -1 : isClickable ? 0 : undefined"
```
kleinfreund, line 11: **Comment**: In the case of `isClickable` being true, the `.k-badge` element should probably just be a button.

mihai-peteu (approved):

adamdehaven (approved):
lgtm

PR #1727, opened 2023-09-15
fix(ktable): reset offset when new query is issued

hfukada:
- fixes offset reset on new searchQuery input.


adamdehaven (dismissed):

adamdehaven (approved):
lgtm

PR #1728, opened 2023-09-15
feat(datetimepicker): update v-calendar dependency [MA-2030]

mihai-peteu:
# Summary

https://konghq.atlassian.net/browse/MA-2050

### Deploy Preview
https://deploy-preview-1728--kongponents.netlify.app/components/datetime-picker.html#date-and-time-range

## PR Checklist

### DIFF
--- a/docs/vite.config.ts
+++ b/docs/vite.config.ts
@@ -15,7 +15,7 @@ export default defineConfig({
       // This just ensures the style import isn't impacted by the separate v-calendar alias below
       'v-calendar/dist/style.css': path.resolve(__dirname, '../node_modules/v-calendar/dist/style.css'),
       // We must alias `v-calendar` here for the docs build to specifically utilize the esm build
-      'v-calendar': path.resolve(__dirname, '../node_modules/v-calendar/dist/v-calendar.es.js'),
+      'v-calendar': path.resolve(__dirname, '../node_modules/v-calendar/dist/es/index.js'),
       // We must alias `@kong/design-tokens` imports to specifically utilize the esm build
       '@kong/design-tokens/tokens/scss/variables': path.resolve(__dirname, '../node_modules/@kong/design-tokens/dist/tokens/scss/variables.scss'),
       '@kong/design-tokens': path.resolve(__dirname, '../node_modules/@kong/design-tokens/dist/tokens/js/'),
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
   },
   "dependencies": {
     "@kong/icons": "^1.5.0",
+    "@popperjs/core": "^2.11.8",
     "axios": "^0.27.2",
     "date-fns": "^2.30.0",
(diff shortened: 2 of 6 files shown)


adamdehaven (commented):
This one definitely needs to wait for the alpha branch

docs/vite.config.ts
```diff
@@ -15,7 +15,7 @@ export default defineConfig({
       // This just ensures the style import isn't impacted by the separate v-calendar alias below
       'v-calendar/dist/style.css': path.resolve(__dirname, '../node_modules/v-calendar/dist/style.css'),
       // We must alias `v-calendar` here for the docs build to specifically utilize the esm build
-      'v-calendar': path.resolve(__dirname, '../node_modules/v-calendar/dist/v-calendar.es.js'),
+      'v-calendar': path.resolve(__dirname, '../node_modules/v-calendar/dist/es/index.js'),
```
adamdehaven, line 18: This PR needs to wait for alpha

adamdehaven (changes requested):

docs/vite.config.ts
mihai-peteu: Understood.

package.json
```diff
@@ -50,6 +50,7 @@
     "semantic-release": "semantic-release"
   },
   "dependencies": {
+    "@popperjs/core": "^2.11.8",
```
mihai-peteu, line 53: @adamdehaven The reason why this was added:
https://github.com/nathanreyes/v-calendar/issues/853

Is there an alternative way to add the direct path to popper.js in the root `vite.config.ts`'s resolved aliases, so we can avoid adding this dependency?

jillztom: @adamdehaven is there any concerns to add this?

adamdehaven: Yes, possibly. 

We already have a dependency on the old popper.js package, so I'm unsure what effect(s) importing (and exporting/packaging) both will have. 

I can take a look when I'm back; since no repos should be picking up the alpha version yet I don't think there should be a rush to get this merged.

mihai-peteu: Sounds good, i'll let it hang then.

Without it, `v-calendar` breaks on `yarn docs:build` with 

```
Rollup failed to resolve import "@popperjs/core
```

Screenshot 2023-09-20 at 2 21 41 PM

adamdehaven: It only breaks in the docs build, or if you pull in the preview Kongponents package and try to use it?

mihai-peteu: It breaks on a `yarn build`

adamdehaven: I don't see any issues in the namespace with both dependencies (other than a larger package size)

@jillztom we could likely refactor our popover usages in this major version to get off of the old `"popper.js": "^1.16.1"` dependency as part of the `beta` work, before major release

adamdehaven (approved):
lgtm

adamdehaven (approved):

PR #1730, opened 2023-09-19
feat(kbutton): component reskinning [KHCP-8970]

portikM:
# Summary

`KButton` component reskinning

Ticket: https://konghq.atlassian.net/browse/KHCP-8970
Figma: https://www.figma.com/file/Yze0SWXl5nKjR0rFdilljK/Components?type=design&node-id=0%3A1&mode=dev

Links:

- Kongponents sandbox: https://deploy-preview-1730--kongponents-sandbox.netlify.app/#/button
- Kongponents docs preview: https://deploy-preview-1730--kongponents.netlify.app/
- khcp-ui adoption PR: https://github.com/Kong/khcp-ui/pull/6891
- khcp-ui adoption PR preview: https://pr-6891--khcp-ui.cloud-preview.konghq.tech/

## PR Checklist


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -99,13 +99,11 @@
...
       <div class="k-search-actions">
         <KButton
-          :appearance="isRegExpMode ? 'action-active' : 'outline'"
+          appearance="secondary"
           :aria-pressed="isRegExpMode"
           class="k-regexp-mode-button"
           data-testid="k-code-block-regexp-mode-button"
-          :is-rounded="false"
           size="small"
-          :title="`Use regular expression (${ALT_SHORTCUT_LABEL}+R)`"
```
kleinfreund, line 108: **Question**: Why are the titles being removed here?

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -166,9 +160,7 @@
           class="k-next-match-button"
           data-testid="k-code-block-next-match-button"
           :disabled="matchingLineNumbers.length === 0 || isFilterMode"
-          :is-rounded="false"
           size="small"
-          title="Next match (F3)"
           type="button"
           @click="jumpToNextMatch"
         >
```
kleinfreund, line 166: **Issue**: The previous and next match buttons in particular don’t look right.

image

In their enabled state, they have a white text color on a light grey background that’s barely visible.

Perhaps this would be a good time to revisit the custom button styles that were added to this component over time (though I’m not sure if there is perhaps the intention to do so when working on KCodeBlock specifically). For example, removing the background color override on `.k-code-block.theme-light .k-button:not(.increase-specificity)` produces this look:

image

I think that looks better.

two24studios:
This is looking great! There is one inconsistency that I'm noticing between what's in the browser and Figma. Also, this is nothing that you did wrong, it's just a difference in how Figma handles borders versus the browser.

In Figma, it doesn't add the border-width to the box model, so when we add borders to things it doesn't increase the width/height of a element.

Due to this, all the buttons in Figma are 4px smaller in height than what I'm seeing in the browser. The heights we chose for these buttons also line up with a lot of other components so it's important that we maintain them:

- Large should be 40px height
- Medium should be 32px height
- Small should be 24px height

Looks like we could just reduce the padding on top/bottom to adjust this, or set a fixed height since button labels generally never wrap anyways... Thoughts? Open to chat over Zoom if needed.

adamdehaven:
> Looks like we could just reduce the padding on top/bottom to adjust this, or set a fixed height

I'd prefer to adjust the padding so we don't have to deal with hard-coded heights (especially since they don't link to a token)

two24studios:
> I'd prefer to adjust the padding so we don't have to deal with hard-coded heights (especially since they don't link to a token)

I'm open to whatever y'all think is best.

src/components/KCodeBlock/KCodeBlock.vue
portikM: Hmm sorry for answering your question with a question, but can you tell me what was it used for here?

src/components/KCodeBlock/KCodeBlock.vue
portikM: Yes, for now I will fix this to make it look okay, until we get to KCodeBlock when we will reskin the component to match the look and feel provided by design.

portikM:
> This is looking great! There is one inconsistency that I'm noticing between what's in the browser and Figma. Also, this is nothing that you did wrong, it's just a difference in how Figma handles borders versus the browser.
> 
> In Figma, it doesn't add the border-width to the box model, so when we add borders to things it doesn't increase the width/height of a element.
> 
> Due to this, all the buttons in Figma are 4px smaller in height than what I'm seeing in the browser. The heights we chose for these buttons also line up with a lot of other components so it's important that we maintain them:
> 
> * Large should be 40px height
> * Medium should be 32px height
> * Small should be 24px height
> 
> Looks like we could just reduce the padding on top/bottom to adjust this, or set a fixed height since button labels generally never wrap anyways... Thoughts? Open to chat over Zoom if needed.

Just an update on this: Jason found a way in Figma to exclude borders outside of the box to match browser behaviour so he will do another iteration to make sure spacing/border width values are correct and then we revisit the code again to reflect those changes. After that look and feel in browser and in Figma should match sizing-wise (and, hopefully, everything else-wise).

two24studios:
@portikM Took another look at this... it's looking good, I did notice that the small buttons have the wrong border-radius sizing though. The small buttons should use `border-radius-20` or 4px.

two24studios:
Another thing I noticed... and maybe @adamdehaven can weigh in on this:

I saw you were using the `all` property in your transitions for the buttons. I was under the impression this was really bad for performance because the browser is looking for every property to transition. I was always taught to be very specific about the properties that are transitioning (e.g. `background-color, border-color`) and to never use `all`. Should we be doing that here?

adamdehaven:
> Another thing I noticed... and maybe @adamdehaven can weigh in on this:
> 
> I saw you were using the `all` property in your transitions for the buttons. I was under the impression this was really bad for performance because the browser is looking for every property to transition. I was always taught to be very specific about the properties that are transitioning (e.g. `background-color, border-color`) and to never use `all`. Should we be doing that here?

Targeting `all` is typically the norm (although as you mentioned, not ideal) - @portikM if it's easy enough, can you update to just transition the properties we are changing?

src/components/KCodeBlock/KCodeBlock.vue
adamdehaven: Hover state help text (e.g. hover over this button tells the user how to use this toggled feature) -- this should remain

src/components/KCodeBlock/KCodeBlock.vue
adamdehaven: The title attr should be left in place here as well

src/components/KCodeBlock/KCodeBlock.vue
portikM: In my understanding we shouldn't be using title attribute as a tooltip, but since it's out of the scope for this I'm adding all titles in this component back for now.

> Use of the title attribute is highly problematic for:
>
> - People using touch-only devices
> - People navigating with keyboards
> - People navigating with assistive technology such as screen readers or magnifiers
> - People experiencing fine motor control impairment
> -  People with cognitive concerns
>
> This is due to inconsistent browser support, compounded by the additional assistive technology parsing of the browser-rendered page. If a tooltip effect is desired, it is better to use a more accessible technique that can be accessed with the above browsing methods.

[source](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title)

adamdehaven: TBH something like `aria-label` might be better, or actual tooltips on the button; you can add this info to the `KCodeBlock` ticket in the reskin and we can revisit then 👍🏼

portikM:
Addressed border-radius for small size and transition properties @adamdehaven @two24studios.

src/components/KAlert/KAlert.vue
```diff
@@ -130,7 +130,7 @@ import type { PropType } from 'vue'
 import { computed, useSlots } from 'vue'
 import KButton from '@/components/KButton/KButton.vue'
 import KIcon from '@/components/KIcon/KIcon.vue'
-import type { AlertAppearance, AlertDismissType, AlertAppearanceRecord, AlertSize, AlertType } from '@/types'
+import { type AlertAppearance, type AlertDismissType, type AlertSize, type AlertType, AlertAppearances } from '@/types'
```
adamdehaven, line 133: This type import format is incorrect - why is the eslint rule not catching this?

adamdehaven (changes requested):

src/components/KAlert/KAlert.vue
adamdehaven:
```suggestion
import type { AlertAppearance, AlertDismissType, AlertSize, AlertType } from '@/types'
import { AlertAppearances } from '@/types'
```

src/components/KButton/KButton.vue
```diff
@@ -1,108 +1,58 @@
...
-      view-box="2 2 15 15"
-    />
   </component>
 </template>
 
 <script setup lang="ts">
 import type { PropType } from 'vue'
-import { computed, useSlots, useAttrs } from 'vue'
+import { computed, useAttrs, useSlots } from 'vue'
+import { type ButtonAppearance, ButtonAppearances, type ButtonSize, ButtonSizes } from '@/types'
```
adamdehaven, line 26:
```suggestion
import { ButtonAppearances, ButtonSizes } from '@/types'
import type { ButtonAppearance, ButtonSize } from '@/types'
```

src/types/toaster.ts
```diff
@@ -1,5 +1,4 @@
-import type { AlertAppearance, AlertAppearanceRecord } from '@/types/alert'
-import { appearances } from '@/components/KAlert/KAlert.vue'
+import { type AlertAppearance, type AlertAppearanceRecord, AlertAppearances } from '@/types/alert'
```
adamdehaven, line 1:
```suggestion
import type { AlertAppearance, AlertAppearanceRecord } from '@/types/alert'
import { AlertAppearances } from '@/types/alert'
```

adamdehaven (approved):
lgtm

PR #1759, opened 2023-10-05
feat(kinput): component reskinning [KHCP-8985]

portikM:
# Summary

[Jira](https://konghq.atlassian.net/browse/KHCP-8985)
[Figma](https://www.figma.com/file/Yze0SWXl5nKjR0rFdilljK/Components?type=design&node-id=218%3A646&mode=dev)

KInput component reskinning.

Links:

- [Kongponents sandbox](https://deploy-preview-1759--kongponents-sandbox.netlify.app/#/input)
- [Kongponents docs preview](https://deploy-preview-1759--kongponents.netlify.app/components/input.html)
- TODO: khcp-ui adoption PR
- TODO: khcp-ui adoption PR preview

## PR Checklist

### DIFF
--- a/docs/guide/migrating-to-version-9.md
+++ b/docs/guide/migrating-to-version-9.md
@@ -130,6 +130,7 @@ Kongponents styles are no longer designed to be utilized standalone, separately 
 
 * `SizeArray` and `IconPositionArray` constants have been removed
 * `Size` and `IconPosition` types have been removed
+* `help` property was removed from `LabelAttributes` interface (TODO: after KLabel is reskinned)
 
 #### Props
 
@@ -137,6 +138,7 @@ Kongponents styles are no longer designed to be utilized standalone, separately 
 * `size` prop has been removed (KInput only comes in 1 size now)
 * `iconPosition` prop has been removed
 * `testMode` prop has been removed
+* `help` property was removed from `labelAttributes` prop (TODO: after KLabel is reskinned)
 
 #### Slots
 
--- a/docs/guide/migrating-to-version-9.md
+++ b/docs/guide/migrating-to-version-9.md
@@ -68,10 +68,6 @@ Kongponents styles are no longer designed to be utilized standalone, separately 
 
 * `icon` slot is removed (TODO: [beta])
 
-#### Events
-
-No changes.
-
 ### KCard
 
 
@@ -144,10 +140,6 @@ No changes.
 #### Slots
 
 * `icon` slot has been removed. Instead, you can use new `before` and `after` slots
-
-#### Events
-
-No changes.
 
 ### KInputSwitch
 
--- a/docs/guide/migrating-to-version-9.md
+++ b/docs/guide/migrating-to-version-9.md
@@ -26,6 +26,8 @@ Kongponents styles are no longer designed to be utilized standalone, separately 
 
 #### Slots
 
+#### Events
+
 ### KBadge
 
 #### Structure
@@ -81,7 +83,17 @@ No changes.
 
 ### KCodeBlock
 
-- Added new `maxHeight` prop to control the `max-height` of the code block. Defaults to `none`.
+#### Structure
+
+#### Constants, Types & Interfaces
+
+#### Props
+
+* New `maxHeight` prop to control the `max-height` of the code block. Defaults to `none`.
+
+#### Slots
+
+#### Events
 
 
 ### KCollapse
@@ -110,6 +122,15 @@ No changes.
 
 ### KInput
 
+#### Structure
+
+#### Constants, Types & Interfaces
+
+#### Props
+
+#### Slots
+
+#### Events
 
 ### KInputSwitch
 
--- a/docs/guide/migrating-to-version-9.md
+++ b/docs/guide/migrating-to-version-9.md
@@ -140,7 +140,11 @@ Kongponents styles are no longer designed to be utilized standalone, separately 
 
 #### Slots
 
-* `icon` slot has been removed. Instead, you can use new `before` and `after` slots
+* `icon` slot has been removed. Instead, you can use new `before` and/or `after` slots
+
+#### Events
+
+* `icon:click` event has been removed. Instead, when using `before` and/or `after` slots, you can provide your custom click handlers bound to clicks on slotted content. See KInput's [slots](/components/input#after) for more details
 
 ### KInputSwitch
 
--- a/docs/guide/migrating-to-version-9.md
+++ b/docs/guide/migrating-to-version-9.md
@@ -124,7 +124,10 @@ No changes.
 
 #### Structure
 
-* `form-control` class is gone
+* `form-control` class is no longer present
+* `over-char-limit` class is no longer present
+* `has-error` class is no longer present
+* `help` class has been changed to `help-text`
 
 #### Constants, Types & Interfaces
 
--- a/docs/guide/migrating-to-version-9.md
+++ b/docs/guide/migrating-to-version-9.md
@@ -142,7 +142,7 @@ Kongponents styles are no longer designed to be utilized standalone, separately 
 
 #### Slots
 
-* `icon` slot has been removed. Instead, you can use new `before` and/or `after` slots
+* `icon` slot has been removed. Instead, you can use new `before` and/or `after` slots to provide icons on the left/right, inside the input
 
 #### Events
 
--- a/docs/guide/migrating-to-version-9.md
+++ b/docs/guide/migrating-to-version-9.md
@@ -89,7 +89,7 @@ No changes.
 
 #### Props
 
-* New `maxHeight` prop to control the `max-height` of the code block. Defaults to `none`.
+* New `maxHeight` prop to control the `max-height` of the code block. Defaults to `none`
 
 #### Slots
 
@@ -124,13 +124,26 @@ No changes.
 
 #### Structure
 
+* `form-control` class is gone
+
 #### Constants, Types & Interfaces
+
+* `SizeArray` and `IconPositionArray` constants are no longer exported
+* `Size` and `IconPosition` types are no longer exported
 
 #### Props
 
+* `overlayLabel` prop is no longer supported
+* `size` prop is no longer supported
+* `iconPosition` prop is no longer supported
+
 #### Slots
 
+* `icon` slot is no longer supported. Instead, you can use new `before` and `after` slots
+
 #### Events
+
+No changes.
 
 ### KInputSwitch
 
@@ -146,7 +159,17 @@ No changes.
 
 ### KModal
 
-- Added new `maxWidth` prop to control the `max-width` of the modal. Defaults to `500px`.
+#### Structure
+
+#### Constants, Types & Interfaces
+
+#### Props
+
+* New `maxWidth` prop to control the `max-width` of the modal. Defaults to `500px`
+
+#### Slots
+
+#### Events
 
 
 ### KModalFullscreen
@@ -163,7 +186,16 @@ No changes.
 
 ### KPrompt
 
-- Added new `maxHeight` prop to control the `max-height` of the prompt. Defaults to `400px`.
+#### Constants, Types & Interfaces
+
+#### Props
+
+* New `maxHeight` prop to control the `max-height` of the prompt. Defaults to `400px`
+
+#### Slots
+
+#### Events
+
 
 
 ### Komponent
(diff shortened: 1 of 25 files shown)


src/components/KInput/KInput.vue
```diff
@@ -319,196 +238,148 @@ export default {
...
-        ~ .input-icon {
-          left: $kInputSmallSizingX;
-        }
-      }
-      &.icon-end {
-        padding-right: calc($kInputSmallSizingX + var(--kui-space-40, $kui-space-40) + $kInputSmallIconSize) !important; // account for icon offset and width
-        ~ .input-icon {
-          right: $kInputSmallSizingX;
+        &:not(.kui-icon) {
+          display: none;
```
adamdehaven, line 309: issue: I don't think we should be opinionated on what developers slot in. It's up to them to not break it; the component should be "stupid"

Also, this directly ties the logic to a class in a different package that could change over time

src/components/KInput/KInput.vue
```diff
@@ -319,196 +238,148 @@ export default {
...
-    @include overlay-label-input;
+  &:disabled {
+    @include kInputDisabledRadOnly;
 
-    label:not(.disabled):not(.readonly).hovered,
-    label:not(.disabled):not(.readonly):hover {
-      color: var(--kui-color-text-primary, $kui-color-text-primary);
-    }
+    // need this to ensure user select is disabled
+    pointer-events: none;
```
adamdehaven: question: I don't think we want to disallow pointer events when disabled; what's the reason for this?

src/components/KInput/KInput.vue
```diff
@@ -319,196 +238,148 @@ export default {
...
+      top: 50%;
+      transform: translateY(-50%);
+
+      :deep(>) {
+        &:is(.kui-icon) {
+          color: var(--kui-color-text-neutral, $kui-color-text-neutral) !important;
+          /* stylelint-disable-next-line @kong/design-tokens/use-proper-token */
+          height: $kInputIconSize;
+          /* stylelint-disable-next-line @kong/design-tokens/use-proper-token */
+          width: $kInputIconSize;
```
adamdehaven, line 305: question: why is the eslint disable comment needed here? Since these are SCSS variables, the stylelint plugin shouldn't be throwing any errors.

docs/guide/migrating-to-version-9.md
```diff
@@ -110,6 +122,31 @@ No changes.
 
 ### KInput
 
+#### Structure
+
+* `form-control` class is no longer present
+* `over-char-limit` class is no longer present
+* `has-error` class is no longer present
```
adamdehaven, line 129: question: did this one get replaced? If not, please change as shown here:
```suggestion
* `has-error` class has been removed
```

docs/guide/migrating-to-version-9.md
```diff
@@ -110,6 +122,31 @@ No changes.
...
+* `help` class has been changed to `help-text`
+
+#### Constants, Types & Interfaces
+
+* `SizeArray` and `IconPositionArray` constants are no longer exported
+* `Size` and `IconPosition` types are no longer exported
```
adamdehaven, line 135: question: are they no longer exported, or have been removed?

docs/guide/migrating-to-version-9.md
```diff
@@ -110,6 +122,31 @@ No changes.
...
+
+#### Props
+
+* `overlayLabel` prop is no longer supported
+* `size` prop is no longer supported
+* `iconPosition` prop is no longer supported
```
adamdehaven, line 141: "has been removed" instead of supported - I'd also state next to the `size` prop that all inputs are a single size

docs/guide/migrating-to-version-9.md
```diff
@@ -110,6 +122,31 @@ No changes.
...
+
+#### Slots
+
+* `icon` slot is no longer supported. Instead, you can use new `before` and `after` slots
```
adamdehaven, line 145:
```suggestion
* `icon` slot has been removed. Instead, you can use new `before` and `after` slots
```

docs/guide/migrating-to-version-9.md
```diff
@@ -110,6 +122,31 @@ No changes.
...
+
+#### Events
+
+No changes.
```
adamdehaven, line 149: nit: if no changes, this section doesn't need to exist

src/components/KInput/KInput.vue
```diff
@@ -1,127 +1,66 @@
...
   </div>
 </template>
 
-<script lang="ts" setup>
+<script setup lang="ts">
 import type { PropType } from 'vue'
-import { computed, ref, watch, onMounted, useSlots, useAttrs } from 'vue'
-import type { IconPosition, Size, LabelAttributes, LimitExceededData } from '@/types'
-import { SizeArray, IconPositionArray } from '@/types'
+import { computed, ref, watch, useSlots, useAttrs } from 'vue'
```
adamdehaven, line 62: nit: can you move the type imports below the vue method imports?

src/components/KInput/KInput.vue
```diff
@@ -319,196 +238,138 @@ export default {
...
-      ~ .input-icon {
-        // (height of entire element (paddings + line height) - icon size) / 2
-        top: calc((($kInputSmallSizingY + $kInputSmallSizingY + $kInputLineHeight) - $kInputSmallIconSize) / 2);
+  .help-text {
+    color: var(--kui-color-text-neutral, $kui-color-text-neutral);
+    font-family: var(--kui-font-family-text, $kui-font-family-text);
+    font-size: var(--kui-font-size-20, $kui-font-size-20);
+    font-weight: var(--kui-font-weight-regular, $kui-font-weight-regular);
+    line-height: var(--kui-line-height-20, $kui-line-height-20);
+    margin: 0;
```
adamdehaven, line 286: nit: can't we use a space token here?

src/components/KInput/KInput.vue
```diff
@@ -319,196 +238,138 @@ export default {
...
+        padding-right: calc($kInputPaddingX * 2 + $kInputIconSize);
       }
     }
   }
 }
 
-.help {
+.k-input {
+  background-color: var(--kui-color-background, $kui-color-background);
+  border: 0;
```
adamdehaven, line 340: issue: the border should be mapped to a token

src/styles/_vars.scss
```diff
@@ -0,0 +1,7 @@
+/*
+* Shared SCSS variables that are used in Kongponents styles but are not exported.
+* Each must start with `$kongponents` prefix to avoid conflicts with other libraries.
+*/
+
+// transition duration and transition timing function
+$kongponentsTransitionDurationTimingFunction: var(--kui-animation-duration-20, $kui-animation-duration-20) ease-in-out;
```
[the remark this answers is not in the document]
adamdehaven: If we keep this, can the variable name be shorter? 😅

docs/guide/migrating-to-version-9.md
portikM: Changed this to `have been removed`

src/components/KInput/KInput.vue
```diff
@@ -253,7 +176,21 @@ const charLimitExceededError = computed((): string => {
...
+const helpText = computed((): string => {
+  // if character limit exceeded, return that error message
+  if (charLimitExceeded.value) {
+    return charLimitExceededErrorMessage.value
+  }
+
+  // if hasError prop is true and there is an error message, return that
+  if (props.hasError && props.errorMessage) {
+    return props.errorMessage
+  }
```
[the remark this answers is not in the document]
portikM: Hmm Y would think the opposite - error message may be addressing some more specific issues in the input (maybe even displays the error returned by the api), while exceeding character limit won't even let them submit the form. So first things first - fix the thing that gets in the way first - which is character limit, then go deeper. Does that make sense?

src/components/KInput/KInput.vue
portikM: No because we really mean to remove border here. If we use `$kui-border-width-0` and then someone defines their custom value for that token, we will end up having both `border` and `box-shadow` on that element.

src/styles/_vars.scss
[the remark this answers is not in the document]
portikM: Yes.. I made it shorter.

src/components/KInput/KInput.vue
portikM: I guess you're right, removing this.

portikM: Though we won't escape using a class from a different package because we do want to style the icon size from within here using `.kui-icon` class

docs/guide/migrating-to-version-9.md
adamdehaven: The same order is fine, but if a section isn't needed, there's no reason to clutter up the docs with empty header sections.

src/components/KInput/KInput.vue
adamdehaven: >  we do want to style the icon size from within here
You could generally target an SVG, but not the specific class

src/components/KInput/KInput.vue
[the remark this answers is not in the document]
adamdehaven: I still disagree, but I'm fine leaving as-is for now and we can revisit if needed later 👍🏼

src/components/KInput/KInput.vue
portikM: That won't work because icons exported by our library are wrapped by the span element. We could target the psan element but I feel like that's wrong.

Screen Shot 2023-10-11 at 10 28 57 AM

sandbox/pages/SandboxInput.vue
```diff
@@ -0,0 +1,216 @@
...
+        <template #before>
+          <SearchIcon @click.stop="onSlotContentClick('before')" />
+        </template>
+        <template #after>
+          <CloseIcon @click.stop="onSlotContentClick('after')" />
+        </template>
+      </KInput>
+    </SandboxSectionComponent>
+    <SandboxSectionComponent
+      description="before and after slots only render icons exported by @kong/icons library. If you provide other content, display: none; CSS rule will be aplied to it."
```
adamdehaven: This comment should be removed -- we aren't going to hide non-icon elements

src/components/KInput/KInput.vue
adamdehaven: For the color, as long as you set the button `color` properly, the icon would just inherit the color since it defaults to `currentColor`

For the sizing, I'm wondering if you could just provide the size of the slot in general instead of targeting an element.

e.g. make a note that the slots are specifically in place to slot icons inside of the input, and that any slot content has a width/height rule in place of `Npx`

portikM: Well yeah that works. But what if they decide to use KButton to make that icon clickable? I think we need to discuss this during our sync later today

adamdehaven: I'm not sure I understand what you mean; can demo in our meeting later 👍🏼

src/styles/styles.scss
```diff
@@ -2,6 +2,7 @@
 // Do not remove the comment above - allows other apps to identify the style block
 
 @import "mixins";
```
portikM, line 4: Now that I've looked into this file I wonder why we want to import mixins here. Any thoughts @adamdehaven?

adamdehaven: We can remove this one; however, instead of importing the mixins into every component, e.g.

```scss
@import '@/styles/mixins';
``` 

We could just add the import to the root vite config here like this:

```ts
additionalData: `
  @import "@kong/design-tokens/tokens/scss/variables";
  @import "@/styles/mixins"; // you need to verify this path properly resolves in the components
`,
```

portikM: Got it. Done.

adamdehaven (changes requested):

src/components/KInput/KInput.vue
```diff
@@ -191,20 +113,17 @@ const emit = defineEmits<{
...
 
 const { stripRequiredLabel } = useUtilities()
 const slots = useSlots()
 const attrs = useAttrs()
 
-const isDisabled = computed((): boolean => attrs?.disabled !== undefined && String(attrs?.disabled) !== 'false')
-const isReadonly = computed((): boolean => attrs?.readonly !== undefined && String(attrs?.readonly) !== 'false')
 const isRequired = computed((): boolean => attrs?.required !== undefined && String(attrs?.required) !== 'false')
-const inputId = computed((): string => attrs.id ? String(attrs.id) : props.testMode ? 'test-input-id-1234' : uuidv4())
+const inputId = computed((): string => attrs.id ? String(attrs.id) : uuidv4())
```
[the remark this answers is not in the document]
portikM: Because we remove the `testMode` prop

adamdehaven (approved):
lgtm

PR #1794, opened 2023-10-26
feat(kcopy): new component [khcp-9464]

aanchalm01:
# Summary

## PR Checklist


src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+        :style="widthStyle"
+      >
+        {{ contentFormat }}
+      </div>
+    </KTooltip>
+
+    <KTooltip
+      class="content-icon-wrapper"
+      :label="`${props.badgeText}` ? iconTitle : tooltipText"
+      max-width="500px"
```
adamdehaven, line 30: question: why this hard-coded `500px` value? What's special about `500px`?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
+<template>
+  <div
+    v-if="!isBadge"
+    class="k-copy"
+    :class="truncate ? null : 'copy-truncated'"
+  >
+    <KTooltip
+      v-if="format !== 'hidden'"
+      :class="truncate ? null : 'copy-content'"
```
adamdehaven, line 9:
```suggestion
      :class="{ 'copy-content': truncate }"
```

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+      :label="contentTooltip"
+      placement="bottomStart"
+      position-fixed
+    >
+      <div
+        :class="[
+          'content-container',
+          truncate ? 'truncated-content' : 'non-truncated-content',
+          monospace ? 'monospace' : null
+        ]"
```
adamdehaven, line 20: suggestion: can we move these dynamic classes to a computed variable?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+        :style="widthStyle"
```
adamdehaven, line 21: suggestion: instead of inlining the styles, can you add another dynamic class and control the width in the `<style>` block?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+      />
+    </KBadge>
+  </div>
+</template>
+
+<script setup lang="ts">
+import type { PropType } from 'vue'
+import { computed, ref } from 'vue'
+import KBadge from '@/components/KBadge/KBadge.vue'
+import KIcon from '@/components/KIcon/KIcon.vue'
```
adamdehaven, line 92: This can be removed if you use an icon from `@kong/icons`

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+          tabindex="0"
+          @click.stop="copyIdToClipboard(copyToClipboard)"
+        >
+          <KIcon
+            class="content-icon"
+            :hide-title="!!copyTooltip || undefined"
+            icon="copy"
+            size="16"
+            :title="iconTitle"
+          />
```
adamdehaven, line 47: Can we use an icon from `@kong/icons` here? There's no reason to add another `KIcon`

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+      max-width="500px"
+      placement="bottomStart"
+      position-fixed
+    >
+      <KClipboardProvider v-slot="{ copyToClipboard }">
+        <span
+          data-testid="copy-to-clipboard"
+          role="button"
+          tabindex="0"
+          @click.stop="copyIdToClipboard(copyToClipboard)"
```
adamdehaven, line 39: These attributes should just be able to be applied to the icon - do we need the `span` wrapper?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+            :title="iconTitle"
+          />
+        </span>
+      </KClipboardProvider>
+    </KTooltip>
+  </div>
+
+  <div
+    v-else
+    class="k-copy-badge"
```
adamdehaven, line 55: issue: the wrapper container should be the only element with a `k-*` class.

This component has two top-level containers; `k-copy` and `k-copy-badge` -- it should just be `k-copy` and then modifier classes added as needed for the badge format instead of an entirely new wrapper

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+    >
+      {{ badgeText }}
+    </span>
+    <KBadge
+      v-if="isBadge"
+      appearance="custom"
+      background-color="#f9fafb"
+      border-color="#d7d8fe"
+      class="id-badge"
+      color="#473cfb"
```
adamdehaven, line 69: question: should we be using design tokens here for colors?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+        class="id-badge-content"
+        :content="content"
+        :content-tooltip="contentTooltip"
+        :copy-tooltip="copyTooltip"
+        :monospace="monospace"
+        :success-tooltip="successTooltip"
+        :truncate="truncate"
+        :truncation-limit="truncationLimit"
+      />
+    </KBadge>
```
adamdehaven, line 84: question: maybe I'm thinking about this wrong, but if the `KCopy` element has a badge appearance, why are we recursively nesting another `KCopy` element within a `KBadge` to provide the copy functionality?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+<script setup lang="ts">
+import type { PropType } from 'vue'
+import { computed, ref } from 'vue'
+import KBadge from '@/components/KBadge/KBadge.vue'
+import KIcon from '@/components/KIcon/KIcon.vue'
+import KClipboardProvider from '@/components/KClipboardProvider'
+import KTooltip from '@/components/KTooltip/KTooltip.vue'
+
+const props = defineProps({
+  // text displayed before the copyable content when `isBadge` is true
```
adamdehaven, line 97: suggestion: can you use `/** comment */` block style comments above the props so they will show up when using the component?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+  },
+  // formatting for copyable content (default, hidden, redacted, deleted)
+  format: {
+    type: String as PropType<'default' | 'hidden' | 'redacted' | 'deleted'>,
+    required: false,
+    default: 'default',
+    validator: (val: string) => ['default', 'hidden', 'redacted', 'deleted'].includes(val),
+  },
+  // whether or not to display as a badge
+  isBadge: {
```
adamdehaven, line 125: issue: just change to `badge` so we follow our "no prefixed props" standard
```suggestion
  badge: {
```

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+import type { PropType } from 'vue'
+import { computed, ref } from 'vue'
+import KBadge from '@/components/KBadge/KBadge.vue'
+import KIcon from '@/components/KIcon/KIcon.vue'
+import KClipboardProvider from '@/components/KClipboardProvider'
+import KTooltip from '@/components/KTooltip/KTooltip.vue'
+
+const props = defineProps({
+  // text displayed before the copyable content when `isBadge` is true
+  badgeText: {
```
adamdehaven, line 98: question: should this be "label" or something instead? Maybe I'm just confused.

Also, why does the component need to provide the label? Couldn't the host app just choose to display text before the component?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+    type: String,
+    default: '',
+  },
+  // the copyable text
+  content: {
```
adamdehaven, line 103: suggestion: `content` sounds like HTML, etc. Could this just be `text`?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+    type: String,
+    required: true,
+  },
+  // tooltip text displayed on hover over the `content`
+  contentTooltip: {
+    type: String,
+    default: '',
+  },
```
adamdehaven, line 111: if we change `content` to `text`, can this be `textTooltip`?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+  tooltipText.value = msg
+  setTimeout(() => {
+    tooltipText.value = props.copyTooltip
+  }, 1800)
+}
+
+const copyIdToClipboard = (executeCopy: (prop: string) => boolean) => {
+  if (!executeCopy(props.content)) {
+    if (hasSuccessTooltip.value) {
+      updateTooltipText('Failed to copy id to clipboard')
```
adamdehaven, line 190: issue: sometimes the content may not be an `id` - should this just read "Failed to copy"?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+  .copy-content {
+    align-items: center;
+    display: inline-flex;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
+  .truncated-content {
+    margin-right: 1ch;
```
adamdehaven, line 234: question: this width can vary depending on the font being used. Is that ok?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+
+  .content-icon {
+    display: flex;
+  }
+}
+.k-copy-badge {
+  :deep(.k-badge) {
+    &.id-badge {
+      .k-badge-text {
+        max-width: 300px;
```
adamdehaven, line 260: question: what is the significance of `300px`?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+      }
+    }
+  }
+
+  .copy-badge-text {
+    color: $kui-color-text-neutral;
+    margin-right: $kui-space-20;
+  }
+
+  .id-badge-content .non-truncated-content{
```
adamdehaven, line 270: nit: spacing
```suggestion
  .id-badge-content .non-truncated-content {
```

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
+# Copy
+
+**KCopy** is a component that displays text content and copies it to clipboard.
```
adamdehaven, line 3: suggestion: wording
```suggestion
**KCopy** is a component that displays provided text content and allows a user to copy it to their clipboard.
```

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+  badge-text="Id"
+  :content="content"
+/>
+```
+
+### content
+
+The copyable text.
+
+### contentTooltip
```
adamdehaven, line 35: suggestion: can we group all of the `tooltip` props together?

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+  format="deleted"
+  :content="content"
+/>
+
+```html
+<KCopy
+  format="deleted"
+  :content="content"
+/>
+```
```
adamdehaven, line 138: docs: Can we add sub headers in this section for each type, along with an explanation?

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+
+### isBadge
+
+Whether or not to display as a badge.
+> Note: By default `isBadge` is set to false.
```
adamdehaven, line 143:
```suggestion
Whether or not to display as a badge. Defaults to `false`.
```

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+<KCopy
+  is-badge
+  truncate
+  :content="content"
+/>
+```
+
+## Events
+
+- `@copied` -  Emitted when `copy button` is clicked; the payload is the copied text.
```
adamdehaven, line 160: nit: can this event be `@copy` ?

Similar to native `@click` - not past-tense

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+    class="k-copy-badge"
+  >
+    <span
+      v-if="isBadge"
+      class="copy-badge-text"
+    >
+      {{ badgeText }}
+    </span>
+    <KBadge
+      v-if="isBadge"
```
adamdehaven, line 64: issue: you only need to check once for `isBadge`
```suggestion
    v-else
    class="k-copy-badge"
  >
    <span
      class="copy-badge-text"
    >
      {{ badgeText }}
    </span>
    <KBadge
```

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+  // false if `isBadge` is true
+  monospace: {
+    type: Boolean,
+    default: true,
+  },
+  // whether or not the content should be truncated
+  truncate: {
+    type: Boolean,
+    default: false,
+  },
```
adamdehaven, line 138: issue: this shouldn't be a prop, it should auto-truncate text always if the container doesn't have enough space to display the whole string based on the available width

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+  truncate
+  :content="content"
+  content-tooltip="Hello!"
+/>
+```
+
+### truncate
+
+Whether or not the content should be truncated.
+> Note: By default truncate is set to false.
```
adamdehaven, line 56:
```suggestion
Whether or not the content should be truncated. Defaults to `false`.
```

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+<KCopy
+  truncate
+  :content="content"
+/>
+```
+
+### truncationLimit
+
+Number of characters to truncate at.
+> Note: By default it is set to `8`.
```
adamdehaven, line 73:
```suggestion
Number of characters to truncate at. Defaults to `8`.
```

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+  truncate
+  :content="content"
+  truncation-limit=10
+/>
+```
+
+### copyTooltip
+
+Tooltip text displayed on hover copy button. 
+> Note: If `badgeText` exists then text is `Copy <badge-text>` else `Copy`
```
adamdehaven, line 92:
```suggestion
Tooltip text displayed on hover copy button. 

If the `badgeText` prop has a value, then the copy tooltip text is `Copy {badgeText}`; otherwise, `Copy`.
```

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+
+
+### successTooltip
+
+Tooltip text displayed on successful copy.
+> Note: By default `successTooltip` is set to `Copied!`.
```
adamdehaven, line 98:
```suggestion
Tooltip text displayed on successful copy. Defaults to `Copied!`.
```

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+
+<KCopy
+  is-badge
+  :content="content"
+/>
+
+```html
+<KCopy
+  is-badge
+  truncate
```
adamdehaven, line 153: issue: you have `truncate` here but not in the actual usage on the line above

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+  :content="content"
+  truncation-limit=10
+/>
+
+```html
+<KCopy
+  truncate
+  :content="content"
+  truncation-limit=10
+/>
```
adamdehaven, line 86: issue: This example sets the value to `10` but it's only showing `8` characters

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+
+```html
+<KCopy
+  truncate
+  :content="content"
+  truncation-limit=10
```
adamdehaven, line 85: issue: since this is a number, you have to pass it with v-bind (also, you were missing quotes `"`)
```suggestion
  :truncation-limit="10"
```

docs/components/copy.md
```diff
@@ -0,0 +1,164 @@
...
+
+### truncationLimit
+
+Number of characters to truncate at.
+> Note: By default it is set to `8`.
+
+<KCopy
+  truncate
+  :content="content"
+  truncation-limit=10
```
adamdehaven, line 78: issue: You're missing quotes around the attribute value in a few places throughout the docs
```suggestion
  :truncation-limit="10"
```

adamdehaven (changes requested):

src/components/KCopy/KCopy.vue
portikM: Plus to that, I see size being hardcoded here - can we use a token on new component instead?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,274 @@
...
+const iconTitle = computed(() => {
+  if (props.badgeText) {
+    return `Copy ${props.badgeText}`
+  }
+  return 'Copy'
+})
+
+const widthStyle = computed(() => {
+  return {
+    maxWidth: props.truncate ? props.truncationLimit + 'ch' : undefined,
```
portikM, line 167: Can we use `px` instead of `ch` unit here? We don't want to be using relative units

src/components/KCopy/KCopy.vue
adamdehaven: I'm good with not calling it `label` then; however, I'm still not sure I see the need for this within the component.

portikM: Ah missed your point there but now I agree - seems redundant since host app can take care of that. Ultimately if we want to make sure styling is always the same in text used in front of the badge variant - I would suggest making this a default slot and adding styling on a wrapping container.

src/components/KCopy/KCopy.vue
aanchalm01: It was in the original code, didnt change it in order to not break the existing width requirement/design.

src/components/KCopy/KCopy.vue
aanchalm01: Would a v-if/v-else make more sense here instead of the recursive nesting? I was trying not to have the same code repeated under both the blocks of v-if/v-else.

src/components/KCopy/KCopy.vue
aanchalm01: Going to add a `slot` with name `badgeText` to fulfil this requirement: `copyTooltip - tooltip text displayed on hover over copy button (default: if `badgeText` provided “Copy <badge-text>” else “Copy”)`

aanchalm01: @kaiarrowood I am not sure how to display `Copy <badge-text>` if there is no prop for badgeText. Can you clarify if we need that prop or not ?

src/components/KCopy/KCopy.vue
aanchalm01: widthStyle is calculated based on props value; what do you mean by control it in <style> block ?

src/components/KCopy/KCopy.vue
aanchalm01: These shades of purple are not available in tokens yet.

src/components/KCopy/KCopy.vue
kaiarrowood: This absolutely needs to be a prop and not a slot. We use the `badge-text` prop for both the label and the default copy tooltip text. I don't care what we call the prop but but there is no point adding a slot when in most use cases we intend to use this text in the copy tooltip also.

adamdehaven: Maybe I'm confused, but are we talking about the text that comes before the copyable content?

```md
My label: 123456789 []
```

In this example ^ "My label" is the text I'm referring to. Do we really need a prop/slot for this? I'm not sure I understand why the host app can't just put text before the element if desired? Sorry in advance if I'm missing something

src/components/KCopy/KCopy.vue
aanchalm01: For KCopy requirement, we need a `truncate` prop to be able to display the text either in truncated mode or non-truncated mode by default. Use cases are the different ways we display the urls and other content in `AppAboutSection`

src/components/KCopy/KCopy.vue
aanchalm01: The tooltip text on hover needs to display that text 'Copy My Label' instead of 'Copy'; if there is no prop how can that be extracted ?

kaiarrowood: Correct, we have logic to scrape the `:` off whatever is passed for `badge-text` (the label) and also use it in the tooltip displayed on hover of the copy button `Copy <badge-text>`

docs/components/copy.md
aanchalm01: fixed.

adamdehaven:
The sandbox build is failing; it looks like you may need to rebase from `alpha`

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+    <div
+      :class="[
+        'copy-container',
+        truncateStyles,
+        badgeStyles
+      ]"
+    >
+      <KTooltip
+        v-if="format !== 'hidden'"
+        class="copy-text"
```
adamdehaven, line 18: When I hover over this element, I always have a pointer cursor even if there is no tooltip

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+
+const props = defineProps({
+  /**
+    * Text displayed before the copyable text when
+    * `badge` is true
+    */
+  badgeText: {
+    type: String,
+    default: '',
+  },
```
adamdehaven, line 69: This prop name is a bit confusing; it's called `badgeText` but it doesn't appear within the badge?

image

docs/components/copy.md
```diff
@@ -0,0 +1,183 @@
...
+  truncate
+  :text="text"
+  :truncation-limit="15"
+/>
+```
+
+### copyTooltip
+
+Tooltip text displayed on hover copy button. 
+If the `badgeText` prop has a value, then the copy tooltip text is `Copy {badgeText}`; otherwise, `Copy`.
```
adamdehaven, line 90: this doesn't mention that you're stripping certain characters from the `badgeText` for the tooltip?

docs/components/copy.md
```diff
@@ -0,0 +1,183 @@
...
+  format="deleted"
+  :text="text"
+/>
+
+```html
+<KCopy
+  format="deleted"
+  :text="text"
+/>
+```
```
adamdehaven, line 175: suggestion: Can you add a header over each example so you can tell which is which?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+      <KTooltip
+        v-if="format !== 'hidden'"
+        class="copy-text"
+        :class="[truncateContent, useMono]"
+        data-testid="copy-text"
+        :label="textTooltip"
+        placement="bottomStart"
+        position-fixed
+      >
+        <span>
```
adamdehaven, line 25: question: what's the purpose of the `span` tag here?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+          {{ textFormat }}
+        </span>
+      </KTooltip>
+
+      <KTooltip
+        class="text-icon-wrapper"
+        :label="`${props.badgeText}` ? iconTitle : tooltipText"
+        max-width="500px"
```
adamdehaven, line 33: question: should the max-width be configurable?

docs/components/copy.md
```diff
@@ -0,0 +1,183 @@
...
+
+```html
+<KCopy
+  truncate
+  :text="text"
+  :truncation-limit="15"
+/>
+```
+
+### copyTooltip
```
adamdehaven, line 87: image
question: When I click to copy, the tooltip doesn't change when this prop is provided. Is that desired?

docs/components/copy.md
```diff
@@ -0,0 +1,183 @@
...
+<KCopy
+  truncate
+  :text="text"
+  text-tooltip="Hello!"
+/>
+```
+
+### successTooltip
+
+Tooltip text displayed on successful copy. Defaults to `Copied!`.
```
adamdehaven, line 126: The success tooltip appears to be `Copied` by default (no exclamation mark)

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+
+  // strip trailing colon from label if one exists
+  return props.badgeText.replace(/:$/, '')
+})
+
+const iconTitle = computed(() => {
+  if (props.badgeText) {
+    return `Copy ${badgeTextLabel.value}`
+  }
+  return 'Copy'
```
adamdehaven, line 157: question: this hard-coded text will be an issue when translation is a thing. Guessing that doesn't matter for now?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+const truncateContent = computed((): string | null => props.truncate ? 'truncate-content' : null)
+const useMono = computed((): string | null => props.badge ? null : 'monospace')
+
+const textFormat = computed(() => {
+  if (props.format === 'redacted') {
+    return '*****'
+  } else if (props.format === 'deleted') {
+    return `*${props.text.substring(0, 5)}`
+  }
+  return (props.truncate && props.truncationLimit && truncateLimitText.value) ? truncateLimitText.value.replace(/^"(.*)"$/, '$1') : props.text
```
adamdehaven, line 174: suggestion: can you leave a comment here explaining the return logic?

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+    }
+  }
+
+  .badge-styles {
+    background-color: #f0f4fa;
+    border-color: #d7d8fe;
+    border-radius: var(--kui-border-radius-round, $kui-border-radius-round);
+    border-style: solid;
+    border-width: var(--kui-border-width-10, $kui-border-width-10);
+    color: #473cfb;
```
adamdehaven, line 228: @portikM once the decorative tokens are published, you can probably use them here

src/components/KCopy/KCopy.vue
aanchalm01: added

src/components/KCopy/KCopy.vue
aanchalm01: KSelect/KPrompt returns hard-coded text too, and there is no usage of i18n in kongponents, so we should be good.

docs/components/copy.md
aanchalm01: Added `!`

src/components/KCopy/KCopy.vue
aanchalm01: Current implementation of CopyUuid had it set to 500px, I will add in a prop to make it configurable.

src/components/KCopy/KCopy.vue
aanchalm01: Added it just to provide a way to be able to access that text in some unforeseeable case it needs some styling change through host app. 
image 1644

docs/components/copy.md
aanchalm01: Does this look ok? 
image 1646

docs/components/copy.md
aanchalm01: Added.
image 1647

docs/components/copy.md
adamdehaven: nit: I'd just use h4/h5 tags, e.g.

`#### deleted` or h5 depending on where you are in the doc

src/components/KCopy/KCopy.vue
aanchalm01: This was named based on the naming that was finalized in the Component brief.

src/components/KCopy/KCopy.vue
adamdehaven: I don't think it's needed; the host app shouldn't be overriding styles

src/components/KCopy/KCopy.vue
adamdehaven: @jillztom was the brief finalized? IMO this is confusing naming but 🤷🏼

src/components/KCopy/KCopy.vue
aanchalm01: span removed.

docs/components/copy.md
aanchalm01: updated
image 1648

src/components/KCopy/KCopy.vue
jillztom: Let's use `badge-label` and go forward with it now.

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+}
+</script>
+
+<style lang="scss" scoped>
+.k-copy {
+  align-items: center;
+  display: flex;
+
+  &-badge-text {
+    margin-right: $kui-space-40;
```
portikM, line 206:
```suggestion
    margin-right: var(--kui-space-40, $kui-space-40);
```

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+    .truncate-content {
+      display: inline-block;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
+  }
+
+  .badge-styles {
+    background-color: #f0f4fa;
```
portikM, line 223:
```suggestion
    background-color: var(--kui-color-background-decorative-purple-weakest, $kui-color-background-decorative-purple-weakest);
```
You can use newly added decorative tokens here, just make sure to install latest version of `@kong/design-tokens`

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+    border-color: #d7d8fe;
```
portikM, line 224:
```suggestion
    border-color: var(--kui-color-border-decorative-purple, $kui-color-border-decorative-purple);
```

See the comment above

src/components/KCopy/KCopy.vue
portikM, line 228:
```suggestion
    color: var(--kui-color-text-decorative-purple, $kui-color-text-decorative-purple);
```
Same as previous, please pull the latest `@kong/design-tokens` to use this token

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+    display: inline-flex;
+  }
+
+  .copy-container {
+    align-items: center;
+    cursor: pointer;
+    display: flex;
+    font-size: $kui-font-size-20;
+    line-height: $kui-line-height-20;
+    padding: $kui-space-10 $kui-space-30;
```
portikM, line 238:
```suggestion
    font-size: var(--kui-font-size-20, $kui-font-size-20);
    line-height: var(--kui-line-height-20, $kui-line-height-20;
    padding: var(--kui-space-10, $kui-space-10) var(--kui-space-30, $kui-space-30);
```

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+
+  .monospace {
+    font-family: var(--kui-font-family-code, $kui-font-family-code);
+  }
+
+  .text-icon-wrapper {
+    align-items: center;
+    cursor: pointer;
+    display: flex;
+    margin-left: $kui-space-30;
```
portikM, line 250:
```suggestion
    margin-left: var(--kui-space-30, $kui-space-30);
```

src/components/KCopy/KCopy.vue
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  .text-icon {
+    display: flex;
+  }
+
+  .copy-badge-text {
+    color: $kui-color-text-neutral;
+    font-size: $kui-font-size-20;
+    margin-right: $kui-space-20;
```
portikM, line 260:
```suggestion
    color: var(--kui-color-text-neutral, $kui-color-text-neutral);
    font-size: var(--kui-font-size-20, $kui-font-size-20);
    margin-right: var(--kui-space-20, $kui-space-20);
```

Reminder: please use CSS vars when using tokens in Kongponents

src/components/KCopy/KCopy.vue
aanchalm01: Need a div here, otherwise the KTooltip won't display on hover.

docs/components/copy.md
aanchalm01: Fixed.

src/components/KCopy/KCopy.vue
aanchalm01: Updated.

src/components/KCopy/KCopy.vue
aanchalm01: Fixed.

aanchalm01:
@adamdehaven @portikM all the feedback has been addressed.

adamdehaven (approved):
lgtm

PR #1797, opened 2023-10-27
fix(kcodeblock): not searching long content with short queries

kleinfreund:
# Summary

Removes the limitation to only trigger a search when (1) a piece of code is shorter than 1000 characters or (2) the query is greater than 2 characters.

## PR Checklist

* [X] **Docs:** includes a technically accurate README

### DIFF
--- a/src/components/KCodeBlock/KCodeBlock.vue
+++ b/src/components/KCodeBlock/KCodeBlock.vue
@@ -613,12 +613,9 @@ function updateMatchingLineNumbers(): void {
   isProcessingInternally.value = true
   regExpError.value = null
 
-  // Avoids searching when one can expect a very large number of results. The numbers here are determined purely by gut feel and not by any scientific reasoning. Feel free to tweak them.
-  const isSmallEnoughForCostlySearch = query.value.length >= 3 || props.code.length < 1000
-  const shouldPerformSearch = query.value.length > 0 && (isRegExpMode.value || (!isRegExpMode.value && isSmallEnoughForCostlySearch))
   let totalMatchingLineNumbers: number[] = []
 
-  if (shouldPerformSearch) {
+  if (query.value.length > 0) {
     try {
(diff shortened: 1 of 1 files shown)


kleinfreund:
**Comment**: The Percy failure is in the date picker component so unrelated to this change.

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -613,12 +613,9 @@ function updateMatchingLineNumbers(): void {
   isProcessingInternally.value = true
   regExpError.value = null
 
-  // Avoids searching when one can expect a very large number of results. The numbers here are determined purely by gut feel and not by any scientific reasoning. Feel free to tweak them.
-  const isSmallEnoughForCostlySearch = query.value.length >= 3 || props.code.length < 1000
-  const shouldPerformSearch = query.value.length > 0 && (isRegExpMode.value || (!isRegExpMode.value && isSmallEnoughForCostlySearch))
   let totalMatchingLineNumbers: number[] = []
 
-  if (shouldPerformSearch) {
+  if (query.value.length > 0) {
```
adamdehaven, line 618: question: why would it ever make sense to query based on 1 character?

Also curious why you're removing the "expensive" restriction; I remember you wanting it in place originally

kleinfreund: It comes up occasionally when we do our test Friday sessions that people wonder why they can’t search for "certain numbers".

Originally I had this in place because we had some (what turned out to be unrelated) performance issues while searching. Those were caused by syntax highlighting being very slow for some obscure reasons. We don’t have those issues anymore and it should be fine to search longer content.

adamdehaven: The search is still debounced, correct?

kleinfreund: Yes, at 150 ms.

adamdehaven (approved):
lgtm

PR #1832, opened 2023-11-07
fix(kselect): kselect reskin [KHCP-8998]

portikM:
# Summary

[Jira](https://konghq.atlassian.net/browse/KHCP-8998)
[Figma](https://www.figma.com/file/Yze0SWXl5nKjR0rFdilljK/Components?type=design&node-id=283%3A2898&mode=dev)

`KSelect` component reskin.

Links:

- [Kongponents sandbox](https://deploy-preview-1832--kongponents-sandbox.netlify.app/#/select)
- [Kongponents docs preview](https://deploy-preview-1832--kongponents.netlify.app/components/select.html)

## PR Checklist


adamdehaven (changes requested):
@Leopoldthecoder or @sumimakito it looks like the `autosuggest` prop is being removed in this PR.

Wasn't it added for Kong Manager specifically?

portikM:
> @Leopoldthecoder or @sumimakito it looks like the `autosuggest` prop is being removed in this PR.
> 
> Wasn't it added for Kong Manager specifically?

Small correction: the async `items` fetching functionality - which in my understanding is what we needed the `autosuggest` prop for is still supported with `enableFiltering` and `@query-change` event: https://deploy-preview-1832--kongponents.netlify.app/components/select.html#loading

sumimakito:
I tested the preview package with the public-ui-components repo. It worked normally with Kong Manager's Admin API filtering mechanism. IMO, it should be okay to remove this prop as long as we update our usages accordingly in the public-ui-components repo.

As @portikM pointed out, the filtering is implemented on the host side (public-ui-components in this case), and it appears that `autosuggest` was only introduced as a hint that actually controls the underlying filter things (e.g., `filterIsEnabled` previously).

One more minor thing: I noticed the behavior of `filteredItems` will be inverted after these changes: Previously, when `autosuggest` was present, we would not call the filter function, while now we will call a default filter function if we `enable-filtering` but not providing a `filter-function`. 

```ts
const filteredItems = computed(() => {
  return props.enableFiltering ? props.filterFunction({ query: filterQuery.value, items: selectItems.value }) : selectItems.value
})
```

https://github.com/Kong/kongponents/pull/1832/files#diff-68c30a7348bc67fac2678056261caf1e95de49eac77fe252f3efa4c7fe2eecf9R428-R430

Would it be better if we bypass the `props.filterFunction` call when it is `null` or non-function?

portikM:
> I tested the preview package with the public-ui-components repo. It worked normally with Kong Manager's Admin API filtering mechanism. IMO, it should be okay to remove this prop as long as we update our usages accordingly in the public-ui-components repo.
> 
> As @portikM pointed out, the filtering is implemented on the host side (public-ui-components in this case), and it appears that `autosuggest` was only introduced as a hint that actually controls the underlying filter things (e.g., `filterIsEnabled` previously).
> 
> One more minor thing: I noticed the behavior of `filteredItems` will be inverted after these changes: Previously, when `autosuggest` was present, we would not call the filter function, while now we will call a default filter function if we `enable-filtering` but not providing a `filter-function`.
> 
> ```ts
> const filteredItems = computed(() => {
>   return props.enableFiltering ? props.filterFunction({ query: filterQuery.value, items: selectItems.value }) : selectItems.value
> })
> ```
> 
> https://github.com/Kong/kongponents/pull/1832/files#diff-68c30a7348bc67fac2678056261caf1e95de49eac77fe252f3efa4c7fe2eecf9R428-R430
> 
> Would it be better if we bypass the `props.filterFunction` call when it is `null` or non-function?

Good call, thanks for testing this Makito! I added an option to bypass default filter function by passing a function that returns `true` through `filterFunction` prop: https://github.com/Kong/kongponents/pull/1832/commits/fe76192430bb235be9afaef9153ee52f64890d0e.

adamdehaven:
image

A couple issues showing up on the sandbox:

1. The KSelect (and the sandbox container) extends beyond the right side of the page
2. The KSelect dropdown is wider than the input element

adamdehaven:
image

on the enableItemCreation example, there is a bunch of extra space in the dropdown below the added item

adamdehaven:
image
In the sandbox, there's raw code - something isn't right

docs/components/select.md
```diff
@@ -1,1013 +1,741 @@
...
 ```html
-<KSelect appearance="select" :items="items" clearable />
+<KSelect help="Helpful text." :items="selectItems" />
 ```
 
-### buttonText
+### error
 
-You can configure the button text when an item is selected, if `appearance` is type `button`.
+When set to `true`, error styling will apply.
```
adamdehaven, line 167: The input element has an `errorMessage` prop, but this component reuses the `help` prop? Seems confusing - do we have a standard?

docs/components/select.md
```diff
@@ -1,1013 +1,741 @@
...
-When `autosuggest` is enabled, you can use the `loading` prop to show a loading indicator while fetching data from API.
-By default, the loading indicator is a spinner icon, and you can implement your own indicator using the `loading` slot.
-See [autosuggest](#autosuggest) for an example.
+Use fixed positioning of the popover to avoid content being clipped by parental boundaries. See [KPopover](/components/popover#positionfixed) docs for more information.
 
-### reuseItemTemplate
+### enableItemCreation
 
-Use this prop to customize selected item element appearance by reusing content passed through [item-template](#item-template) slot. **Note:** this prop only applies when `appearance` prop is `select` (use [selected-item-template](#selected-item-template) slot in other cases).
+When used in conjunction with `enableFiltering` set to `true`, KSelect will suggest adding a new value when filtering produces no results.
```
adamdehaven, line 336: issue: we need to sanitize HTML and input to prevent things like this (same comment on the multiselect PR)
image

docs/components/select.md
```diff
@@ -1,1013 +1,741 @@
...
 
 ```html
-<KSelect width="250" :items="items" dropdown-max-height="150" />
+<KSelect dropdown-max-height="300" :items="selectItems" />
 ```
 
 ### dropdownFooterText
 
-Adds informational text to the bottom of the dropdown options which remains visible even if the content is scrolled. Can also be [slotted](#slots).
+Text to be displayed at the bottom of dropdown container. Can also be [slotted](#dropdown-footer-text).
```
adamdehaven, line 203:
```suggestion
Text to be displayed at the bottom of the dropdown container. Can also be [slotted](#dropdown-footer-text).
```

docs/components/select.md
```diff
@@ -1,1013 +1,741 @@
...
 ```
 
 ### dropdownFooterTextPosition
 
-By default, the dropdown footer text will be stuck to the bottom of the dropdown and will always be visible even if the dropdown content is scrolled.
-
-If you want to override the behaviour and have the footer text at the end of the dropdown list, use the value `static`. This ensures the footer text is visible only when the user scrolls to view the bottom of the list.
-
-Accepted values: `sticky` (default) and `static`.
+Defaults to `sticky`, but also accepts `static` value should you want text passed through `dropdownFooterText` prop to bi displayed at the bottom of dropdown container after all items.
```
adamdehaven, line 215:
```suggestion
Defaults to `sticky`, but also accepts `static` value should you want text passed through `dropdownFooterText` prop to be displayed at the bottom of dropdown container after all items.
```

docs/components/select.md
```diff
@@ -1,1013 +1,741 @@
...
-### positionFixed
-
-Use fixed positioning of the popover to avoid content being clipped by parental boundaries - defaults to `true`. See [`KPop` docs](popover.html#positionfixed) for more information.
-
 ### enableFiltering
 
-Use this prop to control whether or not the `KSelect` component with an `appearance` prop set to a value of `select` or `dropdown` allows filtering. By default, filtering is enabled for `dropdown` appearance and is disabled for `select` appearance.
-
-`button` style `appearance` does not have filter support because it is a button.
+By default, items passed to KSelect are not searchable. When `enableFiltering` prop is true, clicking on select element will turn it into input field, typing into which will filter matching items.
```
adamdehaven, line 227:
```suggestion
By default, items passed to KSelect are not searchable. When `enableFiltering` prop is true, clicking on the select element will turn it into input field. You may filter items by typing in the input which will hide non-matching items.
```

docs/components/select.md
```diff
@@ -1,1013 +1,741 @@
...
-::: tip NOTE
-`filterFunc` does not work with `autosuggest` enabled.
-For `autosuggest`, you are in charge of filtering the options, so `KSelect` won't filter them internally.
-See [autosuggest](#autosuggest) for more details.
-:::
-
-### v-model
-
-`KSelect` works as regular inputs do using v-model for data binding:
+Custom function to filter items by.
```
adamdehaven, line 239:
```suggestion
A custom function to perform item filtering.
```

docs/components/select.md
```diff
@@ -1,1013 +1,741 @@
...
-    <template v-slot:loading>
-      <div>Loading...</div>
-    </template>
-    <template v-slot:empty>
-      <div>No results found</div>
-    </template>
-  </KSelect>
-</ClientOnly>
+:::tip TIP
+Should you need to disable the default filter function while still allowing filtering (for example when you want to handle filtering asynchronously by fetching from the API), you can pass a function that always returns boolean `true` to this prop, like so:
```
adamdehaven, line 291:
```suggestion
Should you need to disable the default filter function while still allowing filtering (for example when you want to handle filtering asynchronously by querying an API), you can provide a function that always returns boolean `true` to this prop, like so: 
```

docs/components/select.md
```diff
@@ -1,1013 +1,741 @@
...
-
-::: tip NOTE
-The `query-change` event triggers immediately when the user types in the input.
-If you need to send API requests in the `query-change` event handler, you may want to implement a debounce function.
-The following is an example:
 :::
 
+### reuseItemTemplate
+
+Whether content passed through [`item-template` slot](#itemtemplate) should be applied for selected item.
```
adamdehaven, line 304:
```suggestion
Reuse the same display format provided via the [`item-template` slot](#itemtemplate) for the selected item.
```

docs/components/select.md
```diff
@@ -1,1013 +1,741 @@
...
-</script>
 ```
 
-### loading
+### positionFixed
 
-When `autosuggest` is enabled, you can use the `loading` prop to show a loading indicator while fetching data from API.
-By default, the loading indicator is a spinner icon, and you can implement your own indicator using the `loading` slot.
-See [autosuggest](#autosuggest) for an example.
+Use fixed positioning of the popover to avoid content being clipped by parental boundaries. See [KPopover](/components/popover#positionfixed) docs for more information.
```
adamdehaven, line 332:
```suggestion
Use fixed positioning for the select dropdown to avoid content being clipped by the boundaries of its parent container. See [KPopover](/components/popover#positionfixed) docs for more information.
```

docs/components/select.md
```diff
@@ -1,1013 +1,741 @@
...
 
-### enableItemCreation
-
-`KSelect` offers users the ability to add custom item to the list by typing the item they want to and then clicking the `... (Add new value)` item at the bottom of the list, which will also automatically select it.
-
-Newly created item will have a `label` consisting of the user input and a randomly generated id for the `value` to ensure uniqueness. It will also have an attribute `custom` set to `true`. This action triggers an `item:added` event containing the added item data.
+### loading
 
-Deselecting the item will completely remove it from the list and underlying data, and trigger a `item:removed` event containing the removed item's data.
+Pass `true` to display loader instead of items in the dropdown. KSelect's `item` prop is reactive to changes by design, so that should you need to perform async item fetching/filtering you can execute that logic on host app's side and pass items back to KSelect. The example below utilizes the [`@query-change` event](#query-change) to simulate async items fetching behind the scenes.
```
adamdehaven, line 348:
```suggestion
Pass `true` to display loader instead of items in the dropdown. KSelect's `item` prop is reactive to changes by design, so that should you need to perform async item fetching/filtering you can execute that logic within the host app and pass items back to KSelect. 

The example below utilizes the [`@query-change` event](#query-change) to simulate async item fetching behind the scenes.
```

docs/components/select.md
```diff
@@ -1,1013 +1,741 @@
...
+  :items="selectItems"
+/>
 ```
 
-:::tip Note:
-When utilizing the `label-tooltip` slot, the `info` `KIcon` will be shown by default. To utilize the the `help` icon instead, set the `label-attributes` `help` property to any non-empty string value.
-:::
+## Slots
+
+### label-tooltip
```
adamdehaven, line 467: suggestion: Can we move `label-tooltip` below the other, more important slots?

docs/guide/migrating-to-version-9.md
```diff
@@ -316,6 +316,7 @@ This component has been removed. Please refer to KBadge component which has been
 
 ### KPagination
 
+KPagination now uses [KDropdown](/components/dropdown) instead of [KSelect](/components/select) as underlying element for page size selector.
```
adamdehaven, line 319:
```suggestion
KPagination now uses [KDropdown](/components/dropdown) instead of [KSelect](/components/select) as the underlying component for page size selection.
```

docs/guide/migrating-to-version-9.md
```diff
@@ -365,6 +366,37 @@ Removed as of `v9`. Use `KBreadcumbs` instead.
...
+* `buttonText` prop has been removed
+* `testMode` prop has been removed
+* `overlayLabel` prop has been removed
+* `filterFunc` prop has been renamed to `filterFunction`
+* `autosuggest` prop has been removed and replaced with `enableFiltering` prop. Should you want to simply enable default component-handled filtering or perform async fetching behind the scenes, you can use this prop in combination with`@query-change` event to achieve that
+
+#### Events
+
+* `item:added` event has been changed to `item-added`
+* `item:removed` event has been changed to `item-removed`
```
adamdehaven, line 381: same question here as multiselect - why are we renaming the events?

src/components/KInput/KInput.vue
```diff
@@ -260,6 +263,21 @@ watch(() => props.error, (newVal, oldVal) => {
     helpTextKey.value += 1
   }
 })
+
+const beforeSlotElement = ref<HTMLElement>()
+const afterSlotElement = ref<HTMLElement>()
```
adamdehaven, line 268: nit: should be initialized as `null`
```suggestion
const beforeSlotElement = ref<HTMLElement | null>(null)
const afterSlotElement = ref<HTMLElement | null>(null)
```

src/components/KPagination/KPagination.vue
```diff
@@ -114,17 +114,21 @@
...
           :items="pageSizeOptions"
           :kpop-attributes="kpopAttrs"
-          :placeholder="`${ currentPageSize } items per page`"
-          position-fixed
-          :test-mode="!!testMode || undefined"
-          width="190"
-          @selected="updatePageSize"
-        />
+          selection-menu
+          show-caret
```
adamdehaven, line 121: Screenshot 2024-01-03 at 9 46 47 PM
question: should we add a max-height attribute to this pagination dropdown so that it's not super-tall and just scrolls internally?

src/components/KSelect/KSelect.vue
```diff
@@ -2,278 +2,195 @@
...
+          toggle()
+          return isToggled.value
+        }"
+        :position-fixed="positionFixed"
+        target=".select-wrapper"
+        @closed="() => {
+          if (selectedItem) {
+            filterQuery = selectedItem.label
+          }
+          if (isToggled.value) {
```
adamdehaven, line 20: suggestion: would it be cleaner to refactor these in-template function blocks `@closed` `on-popover-click` etc. to functions down in the script block? You should be able to pass around the diff callbacks/vars as needed; just wondering if it would be cleaner.

src/components/KSelect/KSelect.vue
```diff
@@ -2,278 +2,195 @@
...
-              :style="widthStyle"
-              @keyup="(evt: any) => triggerFocus(evt, isToggled)"
+            <template #after>
+              <CloseIcon
+                v-if="isClearVisible"
+                data-testid="clear-selection-icon"
+                role="button"
+                tabindex="0"
+                @click="clearSelection"
+              />
```
adamdehaven, line 69: issue: this clear icon needs a `focus-visible` rule for keyboard nav (box-shadow, etc.)

src/components/KSelect/KSelect.vue
```diff
@@ -2,278 +2,195 @@
...
               </slot>
-            </KButton>
+            </div>
+          </Transition>
+        </div>
+        <template #content>
+          <div
+            v-if="enableFiltering && loading"
+            class="select-loading"
+            data-propagate-clicks="false"
```
adamdehaven, line 104: question: what is `data-propagate-clicks`?

src/components/KSelect/KSelect.vue
```diff
@@ -366,25 +256,18 @@ const props = defineProps({
...
   },
   /**
-   * A flag for autosuggest mode
+   * Override default filter functionality of case-insensitive search on label
    */
-  autosuggest: {
-    type: Boolean,
-    default: false,
+  filterFunction: {
+    type: Function,
```
adamdehaven, line 269: question: is it possible to type this function any more specifically?

src/components/KSelect/KSelect.vue
```diff
@@ -555,63 +418,43 @@ const popoverContentMaxHeight = computed((): string => getSizeFromString(props.d
...
-
-const selectButtonText = computed((): string => {
-  if (props.buttonText && selectedItem.value) {
-    return props.buttonText
-  } else if (selectedItem.value) {
-    return selectedItem.value.label
-  }
-  return placeholderText.value
-})
+const placeholderText = computed((): string => props.placeholder || attrs.placeholder as string || 'Filter...')
```
adamdehaven, line 421: Should the default placeholder be `Select` instead? If filtering isn't enabled it doesn't make much sense

src/components/KSelect/KSelect.vue
```diff
@@ -555,63 +418,43 @@ const popoverContentMaxHeight = computed((): string => getSizeFromString(props.d
...
-const isClearVisible = computed((): boolean => props.appearance === 'select' && props.clearable && !!selectedItem.value)
+const isClearVisible = computed((): boolean => !isDisabled.value && (props.clearable && !!selectedItem.value))
 
-const hasCustomSelectedItem = computed((): boolean => !!(selectedItem.value && props.appearance === 'select' &&
+const hasCustomSelectedItem = computed((): boolean => !!(selectedItem.value &&
   (slots['selected-item-template'] || (props.reuseItemTemplate && slots['item-template']))))
 
+const filteredItems = computed(() => {
+  return props.enableFiltering && props.filterFunction({ query: filterQuery.value, items: selectItems.value }) !== true ? props.filterFunction({ query: filterQuery.value, items: selectItems.value }) : selectItems.value
+})
```
adamdehaven, line 430: This block is pretty busy - can you add comments or break it up at all?

src/components/KSelect/KSelect.vue
```diff
@@ -671,34 +512,38 @@ const triggerFocus = (evt: any, isToggled: Ref<boolean>):void => {
...
+
+  emit('query-change', filterQuery.value)
 }
 
 const onInputBlur = (): void => {
   inputFocused.value = false
 }
 
+const onSelectWrapperClick = (event: Event): void => {
+  if (isDisabled.value || (event?.target as HTMLElement)?.dataset.propagateClicks === 'false') {
```
adamdehaven, line 538: nit: whole PR 

There are a ton of changes; can you add comments in the code please?

src/components/KSelect/KSelect.vue
```diff
@@ -741,385 +586,162 @@ watch(() => props.items, (newValue, oldValue) => {
...
-
-<style lang="scss" scoped>
-@import '@/styles/tmp-variables';
-@import '@/styles/mixins';
-
-.k-select {
-  width: fit-content; // necessary for correct placement of popup
+watch(filterQuery, () => {
+  emit('query-change', filterQuery.value)
+})
```
adamdehaven, line 600: nit: use the watch value instead of the ref
```suggestion
watch(filterQuery, (q: string) => {
  emit('query-change', q)
})
```

src/components/KSelect/KSelect.vue
```diff
@@ -741,385 +586,162 @@ watch(() => props.items, (newValue, oldValue) => {
...
-    color: var(--kui-color-text-primary, $kui-color-text-primary);
-    display: flex;
-    font-weight: var(--kui-font-weight-regular, $kui-font-weight-regular);
-    margin-bottom: var(--kui-space-30, $kui-space-30);
-    padding: var(--kui-space-20, $kui-space-20) var(--kui-space-50, $kui-space-50) !important;
+onMounted(() => {
+  if (selectWrapperElement.value) {
+    resizeObserver.value = new ResizeObserver(entries => {
+      // TODO: use resizeObserverHelper(entries, () => { actualElementWidth.value = `${selectWrapperElement.value?.offsetWidth}px` })
+      // Wrapper 'window.requestAnimationFrame' is needed for disabling "ResizeObserver loop limit exceeded" error in DD
```
adamdehaven, line 606: question: is this TODO still valid?

src/components/KSelect/KSelect.vue
```diff
@@ -741,385 +586,162 @@ watch(() => props.items, (newValue, oldValue) => {
...
-    border-left: var(--kui-border-width-30, $kui-border-width-30) solid var(--kui-color-border-transparent, $kui-color-border-transparent);
-    border-right: var(--kui-border-width-30, $kui-border-width-30) solid var(--kui-color-border-transparent, $kui-color-border-transparent);
-    border-top: var(--kui-border-width-30, $kui-border-width-30) solid;
-    content: "";
-    display: inline-block;
-    height: 0;
-    margin-left: var(--kui-space-40, $kui-space-40);
-    vertical-align: middle;
-    width: 0;
+onBeforeUnmount(() => {
```
adamdehaven, line 621: question: `onUnmounted` instead?

kaiarrowood (changes requested):
Most of the functionality is good, but I noticed some things that don't appear right or are confusing.

Item creation appears to be broken, nothing happens when I hit enter - you will need to hook up to the emitted events and add the item into the list in order to verify the actual use case works as expected (and show devs how they are expected to use it).

image

Too much bottom padding on empty state/item create?

image

image


Disabled state should have disabled cursor, not pointer:

image

Sandbox issue - same KSelectItem instance being used in 2 different KSelect's, clicking one triggers a click on the other

image

Two items selected? Possibly caused by reusing the KSelectItem?

image

image

To actually test the functionality, this KSelect needs filtering enabled (input event should fire when you type): 

image

Horizontal scrollbars in docs site example for `dropdownMaxHeight`:

image

I'm not sure this is the best example, I sort of expect results for this query. Maybe make the other searchable field be a string instead of an array, or adjust your function to check if each item of the array includes the substring rather than just if the array includes it. I also recommend updating the item template to show the other field being searched so it's less visually confusing what's happening.

image

I understand we want to keep things simple but these examples feel broken since you can't see/use the non-loading/non-empty state. I'd kind of like to see something even as simple as a toggle or button to switch between normal state and whatever slot you're trying to show to ensure default functionality is working as expected with the slot.

image

It would be really helpful if we can link to or display what we expect `item` to look like here

image

src/components/KSelect/KSelect.vue
portikM: Sometimes clicks on containers are propagated up which results in unintentional clicks on other elements. I added this data property to control that when we can't use `pointer-events: none;`.

docs/components/select.md
adamdehaven: If a user enters HTML or script tags when creating a new item, it's adding that HTML directly into the DOM (XSS).

You need to sanitize the user input to prevent this.

src/components/KSelect/KSelect.vue
adamdehaven: can you document this in the comments somewhere?

portikM: Added a comment in `onSelectWrapperClick` method.

docs/components/select.md
portikM: Added a quick and easy way to sanitize user input: https://github.com/Kong/kongponents/pull/1832/commits/42cd593c12afc33cf797146d3074c346daf35c59. Though I would argue that this would need to be a separate ticket.

docs/components/select.md
portikM: Added a quick and easy way to sanitize user input: https://github.com/Kong/kongponents/pull/1832/commits/42cd593c12afc33cf797146d3074c346daf35c59. Though I would argue that this would need to be a separate ticket.

adamdehaven: This looks good; you'll need this on KMultiselect too

portikM:
> Most of the functionality is good, but I noticed some things that don't appear right or are confusing.
> 
> Item creation appears to be broken, nothing happens when I hit enter - you will need to hook up to the emitted events and add the item into the list in order to verify the actual use case works as expected (and show devs how they are expected to use it).
> 
> image
> 
> Too much bottom padding on empty state/item create?
> 
> image
> 
> image
> 
> Disabled state should have disabled cursor, not pointer:
> 
> image
> 
> Sandbox issue - same KSelectItem instance being used in 2 different KSelect's, clicking one triggers a click on the other
> 
> image
> 
> Two items selected? Possibly caused by reusing the KSelectItem?
> 
> image
> 
> image
> 
> To actually test the functionality, this KSelect needs filtering enabled (input event should fire when you type):
> 
> image
> 
> I'm not sure this is the best example, I sort of expect results for this query. Maybe make the other searchable field be a string instead of an array, or adjust your function to check if each item of the array includes the substring rather than just if the array includes it. I also recommend updating the item template to show the other field being searched so it's less visually confusing what's happening.
> 
> image
> 
> I understand we want to keep things simple but these examples feel broken since you can't see/use the non-loading/non-empty state. I'd kind of like to see something even as simple as a toggle or button to switch between normal state and whatever slot you're trying to show to ensure default functionality is working as expected with the slot.
> 
> image
> 
> It would be really helpful if we can link to or display what we expect `item` to look like here
> 
> image

Thanks for catching bugs here! Fixed a few bugs outlined here and added some improvements in documentation like specifying interfaces and better examples.

src/components/KSelect/KSelect.vue
```diff
@@ -671,34 +506,68 @@ const triggerFocus = (evt: any, isToggled: Ref<boolean>):void => {
...
+
+  emit('query-change', filterQuery.value)
 }
 
 const onInputBlur = (): void => {
   inputFocused.value = false
 }
 
+const onSelectWrapperClick = (event: Event): void => {
+  console.log('here')
```
portikM, line 532:
```suggestion
```

adamdehaven:
issue: I can tab to the "clearable" button, but I can't press any key to actually clear the select
image

kaiarrowood:
- bug: Item creation has a bug when hitting enter - my typed text is cleared. We should do nothing instead.

- bug: still seeing this
> Two items selected? Possibly caused by reusing the KSelectItem?
> 
> image
> 
> image

- request: Please can we do this?
> I'm not sure this is the best example, I sort of expect results for this query. Maybe make the other searchable field be a string instead of an array, or adjust your function to check if each item of the array includes the substring rather than just if the array includes it. **I also recommend updating the item template to show the other field being searched so it's less visually confusing what's happening.**
> 
> image

- suggestion: I still think this is a problem. We have basically provided non-usable examples (ex. a user is never going to have a scenario where you want loading state to always be visible). We could easily improve these examples by having a toggle for loading and whether or not items is an empty array.
> I understand we want to keep things simple but these examples feel broken since you can't see/use the non-loading/non-empty state. I'd kind of like to see something even as simple as a toggle or button to switch between normal state and whatever slot you're trying to show to ensure default functionality is working as expected with the slot.
> 
> image

- request: Please can we do this? You can just link to [here](https://deploy-preview-1832--kongponents.netlify.app/components/select.html#items) since the first thing under that prop is the interface for `SelectItem`.
> It would be really helpful if we can link to or display what we expect `item` to look like here
> 
> image

portikM:
> image In the sandbox, there's raw code - something isn't right

It's an example

kaiarrowood (commented):
Functionally verified sandbox and docs working as expected 🚀

adamdehaven (approved):

PR #1851, opened 2023-11-16
fix(kradio): update radio card styles [KHCP-9491]

portikM:
# Summary

[Jira](https://konghq.atlassian.net/browse/KHCP-9570)
[Figma](https://www.figma.com/file/Yze0SWXl5nKjR0rFdilljK/Components?type=design&node-id=763%3A12969&mode=dev)

Updates KRadio card styles

Screen Shot 2023-11-16 at 5 19 48 PM

Links:
- [Kongponents sandbox](https://deploy-preview-1851--kongponents-sandbox.netlify.app/#/radio)
- [Kongponents docs preview](https://deploy-preview-1851--kongponents.netlify.app/components/radio.html)

## PR Checklist

### DIFF
--- a/src/components/KRadio/KRadio.vue
+++ b/src/components/KRadio/KRadio.vue
@@ -392,6 +392,10 @@ $kRadioDotSize: 6px;
           box-shadow: var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-20, $kui-border-width-20) var(--kui-color-border-primary-strong, $kui-color-border-primary-strong) inset;
         }
 
+        &:focus-visible {
+          box-shadow: var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-20, $kui-border-width-20) var(--kui-color-border-primary-strong, $kui-color-border-primary-strong) inset, var(--kui-shadow-focus, $kui-shadow-focus);
+        }
+
         .radio-label {
           color: var(--kui-color-text-primary, $kui-color-text-primary);
         }
@@ -421,6 +425,10 @@ $kRadioDotSize: 6px;
 
           &:hover {
             box-shadow: var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-20, $kui-border-width-20) var(--kui-color-border-danger-strong, $kui-color-border-danger-strong) inset;
+          }
+
+          &:focus-visible {
+            box-shadow: var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-0, $kui-border-width-0) var(--kui-border-width-20, $kui-border-width-20) var(--kui-color-border-danger-strong, $kui-color-border-danger-strong) inset, var(--kui-shadow-focus, $kui-shadow-focus);
           }
 
           .radio-label {
--- a/src/components/KRadio/KRadio.vue
+++ b/src/components/KRadio/KRadio.vue
@@ -360,6 +360,10 @@ $kRadioDotSize: 6px;
         box-shadow: var(--kui-shadow-border-primary-weak, $kui-shadow-border-primary-weak);
       }
 
+      &:active {
+        box-shadow: var(--kui-shadow-border-primary-strongest, $kui-shadow-border-primary-strongest);
+      }
+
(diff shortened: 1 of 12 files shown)


src/components/KRadio/KRadio.vue
```diff
@@ -351,23 +344,24 @@ $kRadioDotSize: 6px;
     .radio-card-wrapper {
       align-items: center;
       background-color: var(--kui-color-background, $kui-color-background);
+      border: var(--kui-border-width-10, $kui-border-width-10) solid var(--kui-color-border, $kui-color-border);
```
adamdehaven, line 347: I think we need to use a box-shadow instead of border for the selected cards. If no cards are selected, and then you select one, the cards "jump" on the page in their container when the border is added. Alternatively, you could give all cards the same border-width border, but `transparent` by default, and then add the color when the item is selected.

src/components/KRadio/KRadio.vue
```diff
@@ -376,51 +370,83 @@ $kRadioDotSize: 6px;
 
       &.has-label, &.has-description {
         .card-content-wrapper {
+          height: auto;
```
adamdehaven, line 373: nit/question: why is this needed? Isn't it the default?

src/components/KRadio/KRadio.vue
```diff
@@ -376,51 +370,83 @@ $kRadioDotSize: 6px;
...
         }
       }
 
       .radio-label {
         @include labelDefaults;
       }
+
+      :deep(#{$kongponentsKongIconSelector}) {
+        color: var(--kui-color-text-neutral, $kui-color-text-neutral) !important;
+      }
```
adamdehaven, line 384: issue: Shouldn't the host app have control over the icon it slots? There would be no way to override this.

src/components/KRadio/KRadio.vue
```diff
@@ -376,51 +370,83 @@ $kRadioDotSize: 6px;
...
+          border: var(--kui-border-width-20, $kui-border-width-20) solid var(--kui-color-border-primary-strong, $kui-color-border-primary-strong);
+        }
+
+        .radio-label {
+          color: var(--kui-color-text-primary, $kui-color-text-primary);
+        }
+
+        :deep(#{$kongponentsKongIconSelector}) {
+          color: var(--kui-color-text-primary, $kui-color-text-primary) !important;
+        }
```
adamdehaven, line 401: issue: Shouldn't the host app have control over the icon it slots? There would be no way to override this.

src/components/KRadio/KRadio.vue
```diff
@@ -376,51 +370,83 @@ $kRadioDotSize: 6px;
...
+            border: var(--kui-border-width-20, $kui-border-width-20) solid var(--kui-color-border-danger-strong, $kui-color-border-danger-strong);
+          }
+
+          .radio-label {
+            color: var(--kui-color-text, $kui-color-text);
+          }
+
+          :deep(#{$kongponentsKongIconSelector}) {
+            color: var(--kui-color-text, $kui-color-text) !important;
+          }
```
adamdehaven, line 432: issue: Shouldn't the host app have control over the icon it slots? There would be no way to override this.

src/components/KRadio/KRadio.vue
portikM: Fixed

src/components/KRadio/KRadio.vue
portikM: This is needed, it overrides another rule just 3 lines above it

src/components/KRadio/KRadio.vue
portikM: Figma designs define icon color in different states so this should be controlled by the component.

src/components/KRadio/KRadio.vue
portikM: Same as above, Figma designs define icon color in different states so this should be controlled by the component.

src/components/KRadio/KRadio.vue
portikM: Same as above.

src/components/KRadio/KRadio.vue
adamdehaven: This means that we can NEVER use another color icon within the component? Is that desirable? What if we're using the "kong" icon with gradient, or one of the flag icons down the road?

portikM: It can't be overwritten using `id` selector.
```
...
<KongIcon id="specific-color-icon" />
...
#specific-color-icon {
  color: $kui-color-text-success !important;
}
```

Screen Shot 2023-11-20 at 10 30 28 AM

adamdehaven: Doesn't seem ideal IMO; that's the whole point of the `color` prop on the icons?

portikM: Fixed now

PR #1886, opened 2023-11-30
fix(utils): add resize observer helper class

portikM:
# Summary

Adds a `ResizeObserverHelper` util class to help reuse resize observer logic that's duplicated across multiple components 

## PR Checklist

### DIFF
--- a/src/utilities/resizeObserverHelper.ts
+++ b/src/utilities/resizeObserverHelper.ts
@@ -2,6 +2,26 @@
  * ResizeObserverHelper
  * @description Helper class for ResizeObserver
  * @param callback - Callback function on resize
+ *
+ * @example component usage
+ *
+ * import { ResizeObserverHelper } from '@/utilities/resizeObserverHelper'
+ *
+ * const someElementRef = ref<HTMLDivElement>()
+ * const resizeObserver = ref<ResizeObserverHelper>()
+ *
+ * const onResizeCallback = () => {
+ *  // Callback on resize
+ * }
+ *
+ * onMounted(() => {
+ *  resizeObserver.value = ResizeObserverHelper.create(onResizeCallback)
+ *  resizeObserver.value.observe(someElementRef.value as HTMLDivElement)
+ * }}
+ *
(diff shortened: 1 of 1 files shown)


src/components/KTruncate/KTruncate.vue
```diff
@@ -228,24 +229,14 @@ const widthStyle = computed((): Record<string, string> => {
...
+
   resizeObserver.value.observe(kTruncateContainer.value as HTMLDivElement)
   updateToggleVisibility()
 })
 
 onBeforeUnmount(() => {
-  if (resizeObserver.value) {
-    resizeObserver.value.unobserve(kTruncateContainer.value)
-  }
+  resizeObserver.value?.unobserve(kTruncateContainer.value as HTMLDivElement)
```
adamdehaven, line 239: issue: this should be done in `onUnmounted` instead:

> Use this hook to clean up manually created side effects such as timers, DOM event listeners or server connections.

[Source](https://vuejs.org/api/composition-api-lifecycle.html#onunmounted)

src/utilities/resizeObserverHelper.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * ResizeObserverHelper
+ * @description Helper class for ResizeObserver
+ * @param callback - Callback function on resize
+ */
+
+export class ResizeObserverHelper {
```
adamdehaven, line 7: question: why are we using a class instead of just a composable?

portikM: So it looks close to native ResizeObserver API implementation. With native one you would do:
```
resizeObserver.value = new ResizeObserver()
resizeObserver.value.observe(...)
...
resizeObserver.value.unobserve(...)
```

With this helper wrapper it's only slightly different:
```
resizeObserver.value = ResizeObserverHelper.create()
resizeObserver.value.observe(...)
...
resizeObserver.value.unobserve(...)
```

portikM: Done

PR #1904, opened 2023-12-11
fix(ktruncate): stop click propagation [KHCP 10009]

adamdehaven:
# Summary

Stop click propagation on the KTruncate expand/collapse buttons to prevent it bubbling up to a table row, etc.


kaiarrowood (approved):
LGTM! Did not functional test because no example in docs.

PR #1914, opened 2023-12-21
feat(ktabs): reskin component [KHCP-9003]

portikM:
# Summary

[Jira](https://konghq.atlassian.net/browse/KHCP-9003)
[Figma](https://www.figma.com/file/Yze0SWXl5nKjR0rFdilljK/Components?type=design&node-id=297%3A3525&mode=dev)

`KTabs` component reskin.

Links:

- [Kongponents sandbox](https://deploy-preview-1914--kongponents-sandbox.netlify.app/#/tabs)
- [Kongponents docs preview](https://deploy-preview-1914--kongponents.netlify.app/components/tabs.html)

## PR Checklist

### DIFF
--- a/docs/components/tabs.md
+++ b/docs/components/tabs.md
@@ -80,7 +80,7 @@ interface Tab {
 
 ### v-model
 
-By default the KTabs will set the first tab in the array as active. You can override this by passing in the hash of any other tab to be used with `v-model`.
+KTabs will set the first tab in the `tabs` array as active. You can override this by passing in the hash of any other tab to be used with `v-model`.
 
 <KTabs v-model="modelTab" :tabs="tabs">
   <template #tab1>
@@ -126,7 +126,7 @@ const tabs: Tab[] = [
 </script>
 ```
 
-If you want to keep your `v-model` in sync so that you can programmatically change the active tab after initialization, you also must respond to the `@changed` emitted event.
+If you want to keep your `v-model` in sync so that you can programmatically change the active tab after initialization, you must subscribe to the emitted `@changed` event.
 
 <div>
   <KTabs v-model="modelTabProgrammatic" :tabs="tabs" @changed="hash => modelTabProgrammatic = hash">
@@ -229,7 +229,7 @@ Here's an example (code only) of utilizing a dynamic `router-view` component wit
 
 The tab control defaults to the `tab.title` string. You may use the `#{tab.hash}-anchor` slot to customize the content of the tab control.
 
-In order provide the tab panel content (when the `hasPanels` prop is set to `true`) you must slot the content in the named slot, defined by the `tab.hash` string, without the `#`.
+In order provide the tab panel content (when the `hasPanels` prop is set to `true`) you must slot the content into the corresponding named slot, defined by the `tab.hash` string, without the `#`.
 
 <KTabs :tabs="slottedTabs">
   <template #gateway-anchor>
--- a/docs/components/tabs.md
+++ b/docs/components/tabs.md
@@ -1,135 +1,157 @@
 # Tabs
 
-**KTabs** - A mindblowing tabs component
+KTabs are horizontal controls that allow users to switch between multiple views within one page.
 
 <KTabs :tabs="tabs">
-  <template v-slot:tab1>
+  <template #tab1>
     <p>Tab 1 content</p>
   </template>
-  <template v-slot:tab2>
+  <template #tab2>
     <p>Tab 2 content</p>
   </template>
 </KTabs>
 
 ```html
 <KTabs :tabs="tabs">
-  <template v-slot:tab1>
+  <template #tab1>
     <p>Tab 1 content</p>
   </template>
-  <template v-slot:tab2>
+  <template #tab2>
     <p>Tab 2 content</p>
   </template>
+  <template #tab3>
+    <p>Tab 3 content</p>
+  </template>
 </KTabs>
-
-<script lang="ts">
-import { defineComponent } from 'vue'
-
-export default defineComponent({
-  setup () {
-    const tabs = [
-      {
-        hash: '#tab1',
-        title: 'Tab 1'
-      },
-      {
-        hash: '#tab2',
-        title: 'Tab 2'
-      }
-    ]
-
-    return {
-      tabs,
-    }
-  }
-})
-</script>
 ```
 
 ## Props
 
 ### tabs
 
-`KTabs` has one **required** prop, `tabs`, which is an array of tab objects with the following interface:
+Required prop, which is an array of tab objects with the following interface:
 
 ```ts
-export interface Tab {
+interface Tab {
   hash: string
   title: string
+  disabled?: boolean
 }
 ```
 
+<KTabs :tabs="tabs">
+  <template #tab1>
+    <p>Tab 1 content</p>
+  </template>
+  <template #tab2>
+    <p>Tab 2 content</p>
+  </template>
+</KTabs>
+
 ```html
-<template>
-  <KTabs :tabs="tabs" />
-</template>
-
-<script lang="ts">
-import { defineComponent } from 'vue'
-
-export default defineComponent({
-  setup () {
-    const tabs = [
-      { hash: '#pictures', title: 'Pictures' },
-      { hash: '#movies', title: 'Movies' },
-      { hash: '#books', title: 'Books' },
-    ]
-
-    return {
-      tabs,
+<KTabs
+  :tabs="[
+    {
+      hash: '#tab1',
+      title: 'Tab 1'
+    },
+    {
+      hash: '#tab2',
+      title: 'Tab 2'
+    },
+    {
+      hash: '#tab3',
+      title: 'Tab 3',
+      disabled: true
     }
-  }
-})
-</script>
+  ]"
+>
+  <template #tab1>
+    <p>Tab 1 content</p>
+  </template>
+  <template #tab2>
+    <p>Tab 2 content</p>
+  </template>
+  <template #tab3>
+    <p>Tab 3 content</p>
+  </template>
+</KTabs>
 ```
 
 ### v-model
 
-By default the `KTabs` will set the first tab in the array as active. You can override this by passing in the hash of any other tab to be used with `v-model`.
+By default the KTabs will set the first tab in the array as active. You can override this by passing in the hash of any other tab to be used with `v-model`.
 
-<KTabs v-model="defaultTab" :tabs="tabs">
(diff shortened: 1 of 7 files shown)


src/components/KTabs/KTabs.vue
```diff
@@ -8,18 +8,18 @@
...
+        :class="{ active: activeTab === tab.hash, disabled: tab.disabled }"
       >
         <div
-          :class="['tab-link', { 'has-panels': hasPanels }]"
+          :aria-controls="hasPanels ? `panel-${i}` : undefined"
+          :aria-selected="hasPanels ? (activeTab === tab.hash ? 'true' : 'false') : undefined"
+          class="tab-link"
+          :role="hasPanels ? 'tab' : undefined"
+          :tabindex="tab.disabled ? '-1' : '0'"
+          @click="handleTabChange(tab.hash)"
```
adamdehaven, line 20: suggestion: is there any reason not to use an anchor tag for the clickable element instead of a `div`?

portikM: Anchor tag requires `href` attribute to be provided at all times. Considering that usage of this component doesn't always imply using hash in the route, I think it's not ideal. Also makes things trickier with disabled item.

adamdehaven: I'm not sure I understand -- a clickable div isn't semantically correct.

> Anchor tag requires href attribute to be provided at all times.

`<a href="#" @click.prevent="doSomething" />`

This has nothing to do with a hash in the route?

> Also makes things trickier with disabled item

```scss
a.disabled {
  cursor: not-allowed;
}
```

In the `doSomething` method, check if disabled. If true, exit.

portikM: Ah you're right, `.prevent` will prevent it from adding the `#` in the route. Changed it to anchor tag.

adamdehaven:
issue: can you add `overflow-x: auto` to the `ul role="tablist"` element so that on screens where the tabs extend beyond the container, they will scroll horizontally? You may need to adjust the padding, Y overflow, etc.

docs/components/tabs.md
```diff
@@ -192,166 +216,124 @@ Here's an example (code only) of utlizing a dynamic `router-view` component with
...
 ```
 
 ## Slots
 
-### Anchor Content
+### anchor & panel
 
 The tab control defaults to the `tab.title` string. You may use the `#{tab.hash}-anchor` slot to customize the content of the tab control.
 
+In order provide the tab panel content (when the `hasPanels` prop is set to `true`) you must slot the content in the named slot, defined by the `tab.hash` string, without the `#`.
```
adamdehaven, line 232:
```suggestion
In order provide the tab panel content (when the `hasPanels` prop is set to `true`) you must slot the content into the corresponding named slot, defined by the `tab.hash` string, without the `#`.
```

adamdehaven: Can you add another sentence here as an example, e.g. "If the `tab.hash` is `#name-tab` the slot name would be `name-tab`"

... or whatever you think a good way to explain it would be

docs/components/tabs.md
```diff
@@ -192,166 +216,124 @@ Here's an example (code only) of utlizing a dynamic `router-view` component with
...
+  </template>
+  <template #docs><b>Documentation</b> tab content</template>
+</KTabs>
 ```
 
-### Panel Content
+## Events
 
-In order provide the tab panel content (when the `hasPanels` prop is set to `true`) you must slot the content in the named slot, defined by the `tab.hash` string, without the `#`.
+### changed
```
adamdehaven, line 276: Should the event be updated to `@change` instead so that it is present-tense, similar to `@click` etc.?

src/components/KTabs/KTabs.vue
```diff
@@ -98,57 +99,69 @@ watch(() => props.modelValue, (newTabHash) => {
...
-      &:hover {
-        .tab-link.has-panels,
-        .tab-link:not(.has-panels) :deep(> *) {
-          border-bottom: var(--kui-border-width-30, $kui-border-width-30) solid $tmp-color-aqua-50; // teal-300 is now aqua-50
-          color: var(--kui-color-text, $kui-color-text);
+      &.disabled {
+        cursor: not-allowed;
+
+        .tab-link {
+          color: var(--kui-color-text-disabled, $kui-color-text-disabled);
```
adamdehaven, line 163: image
The disabled color and the inactive color are almost indistinguishable.

src/components/KTabs/KTabs.vue
```diff
@@ -8,23 +8,24 @@
...
+        :class="{ active: activeTab === tab.hash, disabled: tab.disabled }"
       >
-        <div
-          :class="['tab-link', { 'has-panels': hasPanels }]"
+        <a
+          :aria-controls="hasPanels ? `panel-${i}` : undefined"
+          :aria-selected="hasPanels ? (activeTab === tab.hash ? 'true' : 'false') : undefined"
+          class="tab-link"
+          href="#"
+          :role="hasPanels ? 'tab' : undefined"
```
adamdehaven, line 19: question: wouldn't the role still be a `tab` even if there isn't a corresponding panel?

src/components/KTabs/KTabs.vue
```diff
@@ -8,23 +8,24 @@
...
-        tabindex="0"
-        @click="handleTabChange(tab.hash)"
-        @keydown.enter.prevent="handleTabChange(tab.hash)"
-        @keydown.space.prevent="handleTabChange(tab.hash)"
+        :class="{ active: activeTab === tab.hash, disabled: tab.disabled }"
       >
-        <div
-          :class="['tab-link', { 'has-panels': hasPanels }]"
+        <a
+          :aria-controls="hasPanels ? `panel-${i}` : undefined"
```
adamdehaven, line 15: question: I'm not sure we should be using the index as the identifier here; if there are multiple instances of tabs on the page, wouldn't this be invalid?

docs/components/tabs.md
```diff
@@ -142,35 +164,37 @@ For example, you could set the `hasPanels` prop to `false` and then your host ap
 
 Here's an example where we display the active tab hash:
 
-<KTabs :tabs="slottedTabs" :has-panels="false" @changed="hasPanelsChanged" />
-<p>Active hash: {{ hasPanelsActiveHash }} </p>
+<div>
```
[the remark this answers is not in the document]
portikM: Changed to `showPanels` since just `panels` seems too generic

src/components/KTabs/KTabs.vue
portikM: Yes. No big deal I think since we won't often have disabled tabs but something we should address with design for sure.

kaiarrowood (commented):
Sort of a nit? I'm not sure, maybe also a bug. This example has 2 links on the first tab's anchor. Clicking one gives you the expected behavior, the other switches the selected tab, but does not update the tab content correctly

Right content (notice the anchor in the bottom right) - this is before clicking, you can tell because `Tab 2` is still selected

image


Wrong content - this is after clicking, you can tell because `Tab 1` is selected, but we are still seeing `Tab 2` content

image

Bug: clicking a tab in the middle of the docs page automatically scrolls me back to the top of the page

suggestion: This example/explanation is confusing. It would probably be helpful to add some thick brightly colored borders around the KTab and the external content to illustrate the difference.

image

Where is the actual example that displays the event's content when it is emitted? We have a code snippet but nothing to play with.

image

kaiarrowood (commented):
Functionally verified with a couple non-blocking nits. 🚀 

nit: clicks on second example changes content on first example as well (there is no Tab 3?).

image

suggestion: I'm sorry this is late in the game, would it be possible to change `showPanels` to `hidePanels` so we can follow the best practice of defaulting boolean variables to `false`?

docs/components/tabs.md
```diff
@@ -1,180 +1,204 @@
 # Tabs
 
-**KTabs** - A mindblowing tabs component
+KTabs are horizontal controls that allow users to switch between multiple views within one page.
 
 <KTabs :tabs="tabs">
-  <template v-slot:tab1>
+  <template #tab1>
```
adamdehaven, line 6: issue: the third tab in the example here is disabled, but is not shown as disabled in the example. I'm not sure we need `disabled` here in the first example since we don't mention it regardless.

adamdehaven: ...and there's actually a disabled tab in **most** of the examples -- this is unexpected. Can it just be a normal tab please and use a `disabled` one under the corresponding docs section?

portikM: Adjusted the disabled item title to mention that it's disabled. I think it's useful to have it a few examples.

docs/components/tabs.md
```diff
@@ -1,180 +1,204 @@
...
+  </KTabs>
 
-<KButton @click="defaultProgrammaticTab = '#tab1'" class="horizontal-spacing">Activate Tab 1</KButton>
-<KButton @click="defaultProgrammaticTab = '#tab2'">Activate Tab 2</KButton>
+  <div class="horizontal-spacing spacing-top">
+    <KButton @click="modelTabProgrammatic = '#tab1'">Activate Tab 1</KButton>
+    <KButton @click="modelTabProgrammatic = '#tab2'">Activate Tab 2</KButton>
+    <KButton @click="modelTabProgrammatic = '#tab3'">Activate Tab 3</KButton>
+  </div>
+</div>
```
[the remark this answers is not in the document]
portikM: I left it there to show how the tab behaves when it's disabled and active - sort of like an item is selected and disabled in KSelect or KMultiselect?

portikM:
> * nit: clicks on second example changes content on first example as well (there is no Tab 3?).

1. That's okay for sandbox.
2. Done.
3. It is disabled but the example shows what happens if the tab is changed programmatically (I think it's useful to show how the tab behaves when it's disabled and active - sort of like an item is selected and disabled in KSelect or KMultiselect?).

docs/components/tabs.md
adamdehaven: ...but it's in ALL the examples?

docs/components/tabs.md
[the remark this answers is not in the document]
adamdehaven: It's confusing IMO; similar to what @kaiarrowood has commented on some of the docs examples: I think they're getting too complex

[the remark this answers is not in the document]
portikM: Okay maybe you're right, removed some examples.

kaiarrowood (commented):
Functionally verified sandbox and docs looking good 🚀

adamdehaven (approved):

PR #1925, opened 2024-01-03
feat(kcopy): reskin component [KHCP-9663]

portikM:
# Summary

[Jira](https://konghq.atlassian.net/browse/KHCP-9663)

`KCopy` component reskin

Links:

- [Kongponents sandbox](https://deploy-preview-1925--kongponents-sandbox.netlify.app/#/copy)
- [Kongponents docs preview](https://deploy-preview-1925--kongponents.netlify.app/components/copy.html)

## PR Checklist

### DIFF
--- a/docs/components/copy.md
+++ b/docs/components/copy.md
@@ -14,6 +14,73 @@ KCopy is a component that displays provided text and allows a user to copy it to
 
 The copyable text.
 
+### format
+
+Determines the display format of the copyable text. The component can take the following `format` values:
+
+- `default`
+- `hidden`
+- `redacted`
+- `deleted`
+
+#### default
+
+Displays regular copyable text.
+
+<KCopy :text="text" />
+
+```html
+<KCopy :text="text" />
+```
+
+#### hidden
+
+Displays just a copy button without text.
+
+<KCopy
+  format="hidden"
+  :text="text"
+/>
+
+```html
+<KCopy
+  format="hidden"
+  :text="text"
+/>
+```
+
+#### redacted
+
+Displays `*****`.
+
+<KCopy
+  format="redacted"
+  :text="text"
+/>
+
+```html
+<KCopy
+  format="redacted"
+  :text="text"
+/>
+```
+
+#### deleted
+
+Displays `*<first-5-chars-of-copyable-text>`.
+
+<KCopy
+  format="deleted"
+  :text="text"
+/>
+
+```html
+<KCopy
+  format="deleted"
+  :text="text"
+/>
+```
+
 ### badge
 
 Whether or not to display as a badge. Defaults to `false`.
@@ -38,14 +105,14 @@ Text displayed before the copyable text when `badge` is true.
 
 <KCopy
   badge
-  badge-label="Service ID"
+  badge-label="Service ID:"
   :text="text"
 />
 
 ```html
 <KCopy
   badge
-  badge-label="Service ID"
+  badge-label="Service ID:"
   :text="text"
 />
 ```
@@ -92,7 +159,7 @@ If the `badgeLabel` prop has a value, then the copy tooltip text is `Copy {badge
 <KCopy
   :text="text"
   badge 
-  badge-label="Service ID"
+  badge-label="Service ID:"
   copy-tooltip="Copy to clipboard"
 />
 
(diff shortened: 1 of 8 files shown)


adamdehaven:
Screenshot 2024-01-03 at 8 46 03 PM
Why do we show the tooltip when the whole string is visible?

adamdehaven:
image
Should the icon be 1 size smaller? It looks a bit large inside the badge

adamdehaven:
image
Issue: when the text is truncated, shouldn't we be displaying the entire string in the tooltip on hover?

adamdehaven:
image
What's the point of the hover tooltip on a redacted value?

portikM:
> image Should the icon be 1 size smaller? It looks a bit large inside the badge

On your screenshot you have the old badge styles for some reason. The new one matches the badge icon size.

Screenshot 2024-01-04 at 9 12 59 AM

kaiarrowood (commented):
question: should the text have a pointer cursor or default? Clicking the text doesn't do anything. I think only the copy button should have a pointer.

image

This is easy to overlook since the label is so short. Can we display `ID` instead so it's a little more noticeable or change the content to something with a longer label?

image

docs/components/copy.md
```diff
@@ -127,7 +127,7 @@ Tooltip text displayed on successful copy. Defaults to `Copied!`.
 
 ### monospace
 
-An indicator of whether the copyable text has `JetBrains Mono` font or not. Defaults to `true`, but `false` if `badge` is `true`.
+An indicator of whether the copyable text has `JetBrains Mono` font or not. Defaults to `undefined`, but `true` if `badge` is `false`.
```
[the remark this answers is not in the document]
portikM: We want to use monospace font when `badge` prop is `false`. If `monospace` prop is `false` be default - no matter what they pass through this prop - it's always gonna be a monospace font in a non-badge.

I don't like this pattern though - to set a `boolean` prop to default to `undefined` so I change default back to `false` -- this will have to be addressed if we want to not use the monospace in a non-badge in future.

src/styles/_vars.scss
```diff
@@ -9,3 +9,6 @@ $kongponentsKongIconSelector: ".kui-icon";
 
 // transition duration and transition timing function
 $kongponentsTransitionDurTimingFunc: var(--kui-animation-duration-20, $kui-animation-duration-20) ease-in-out;
+
+// KCopy tooltip selector
```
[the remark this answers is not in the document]
portikM: > I'm wondering if this should be a JavaScript constant instead of a SCSS variable

I think I can't use JS constant as a selector in style block, can I?

portikM:
> question: should the text have a pointer cursor or default? Clicking the text doesn't do anything. I think only the copy button should have a pointer.
> 
> image
> 
> This is easy to overlook since the label is so short. Can we display `ID` instead so it's a little more noticeable or change the content to something with a longer label?
> 
> image
> 
> This example doesn't actually show how to use the prop it's supposed to be for:
> 
> image
> 
> Can we show an example for users who might not know what `monospace` fonts look like off the top of their heads?
> 
> image

Fixed all of these now :)

docs/components/copy.md
```diff
@@ -127,7 +129,13 @@ Tooltip text displayed on successful copy. Defaults to `Copied!`.
 
 ### monospace
 
-An indicator of whether the copyable text has `JetBrains Mono` font or not. Defaults to `true`, but `false` if `badge` is `true`.
+When `badge` is `true`, use this prop to control whether the copyable text has `JetBrains Mono` font or not. Defaults to `false`.
+
+<KCopy badge monospace :text="text" />
```
adamdehaven, line 134: When the badge is monospace, it looks like a font weight of `600` is being applied; I don't think it should be

docs/components/copy.md
```diff
@@ -127,7 +129,13 @@ Tooltip text displayed on successful copy. Defaults to `Copied!`.
...
+
+```html
+<KCopy badge monospace :text="text" />
+```
```
adamdehaven, line 138: Should format prop be moved towards the top of the docs? Seems important

docs/components/copy.md
adamdehaven: image
But then on the sandbox it looks regular and the other badges look bold. Something is going on.

portikM: Badge are supposed to be bold (to match the KBadge styles) but I agree that monospaced ones don't. Fixed now.

adamdehaven (approved):

kaiarrowood (approved):
LGTM! 🚀

PR #1961, opened 2024-01-18
feat(kicon): add network icon

adamdehaven:
# Summary
Add `network` icon


DariaYeremina (approved):

portikM (changes requested):
Whoa, why is every single file changing? Is that expected?

portikM (approved):

PR #1969, opened 2024-01-23
fix(kcopy): expose copy method, tweak hover in kbutton

portikM:
# Summary

KCopy:
* expose `triggerCopy` method to allow executing copy action from outside the component
* tweak styling to take hover color set by KButton when KCopy is a child of KButton

## PR Checklist

### DIFF
--- a/src/components/KCopy/KCopy.vue
+++ b/src/components/KCopy/KCopy.vue
@@ -135,7 +135,7 @@ const props = defineProps({
   },
 })
 
-const copyButtonElementId = computed((): string => uuid4())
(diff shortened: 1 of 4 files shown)


src/components/KCopy/KCopy.vue
```diff
@@ -131,6 +135,8 @@ const props = defineProps({
   },
 })
 
+const copyButtonElementId = computed((): string => uuid4())
```
adamdehaven, line 138: nit: this doesn't need to be computed, it can just be a const

adamdehaven (approved):

PR #1972, opened 2024-01-24
feat(kmodal, kprompt): reskin components

portikM:
# Summary

Jira:
https://konghq.atlassian.net/browse/KHCP-8990
https://konghq.atlassian.net/browse/KHCP-8995

`KModal` and `KPrompt` component reskin

Links:

- [Kongponents sandbox](https://deploy-preview-1972--kongponents-sandbox.netlify.app/#/modals)
- [KModal Kongponents docs preview](https://deploy-preview-1972--kongponents.netlify.app/components/modal.html)
- [KPrompt Kongponents docs preview](https://deploy-preview-1972--kongponents.netlify.app/components/prompt.html)

## PR Checklist


docs/components/modal.md
```diff
@@ -1,348 +1,752 @@
...
 
-::: info
-Are you looking for a modal with a close icon, left-aligned text, right-aligned action buttons, and divider lines at the top and bottom of the modal? Try [KPrompt](/components/prompt.html) instead.
+KModal is a modal window component with overlay.
+
+:::tip NOTE
+Consider using [KPrompt component](/components/prompt) instead if your use case is one of the below:
+* you need a simple pop-up element to ask user to confirm their action
+* you need user's input
+* you don't need much customization options for modal window layout
```
adamdehaven, line 9:
```suggestion
If your use-case matches one of the scenarios below, consider using [KPrompt](/components/prompt) instead:
* you need a simple pop-up dialog to ask user to confirm their action
* the user performed an action and you first require them to provide input text
* you don't need many customization options for a simple modal with action buttons
```

src/components/KModal/KModal.vue
```diff
@@ -293,160 +235,110 @@ watch(() => props.isVisible, async (isVisible) => {
...
-      margin-right: calc(#{$k-modal-padding} * -1);
-      /* stylelint-disable-next-line custom-property-no-missing-var-function */
-      margin-top: calc(#{$k-modal-padding} * -1);
+  .modal-container {
+    background-color: var(--kui-color-background, $kui-color-background);
+    border: var(--kui-border-width-10, $kui-border-width-10) solid var(--kui-color-border, $kui-color-border);
+    border-radius: var(--kui-border-radius-40, $kui-border-radius-40);
+    box-shadow: var(--kui-shadow, $kui-shadow);
+    margin: var(--kui-space-0, $kui-space-0) var(--kui-space-auto, $kui-space-auto);
+    width: v-bind('widthValue');
```
adamdehaven, line 262: Unsure if this is the right line, but on all viewports (especially mobile) it would be nice to have a little padding to the left and right of the modal white background instead of it going edge to edge
image

src/components/KModal/KModal.vue
```diff
@@ -293,160 +235,110 @@ watch(() => props.isVisible, async (isVisible) => {
...
-      margin-bottom: var(--kui-space-80, $kui-space-80) !important;
-
-      &.header-centered {
-        margin-left: var(--kui-space-auto, $kui-space-auto);
-        margin-right: var(--kui-space-auto, $kui-space-auto);
-        text-align: center;
+      justify-content: space-between;
+      padding: var(--kui-space-70, $kui-space-70) var(--kui-space-80, $kui-space-80);
+
+      .modal-title {
```
adamdehaven, line 280: can we add `user-select: none` here so they can't "highlight" the modal title?

docs/components/modal.md
```diff
@@ -1,348 +1,752 @@
...
+    </div>
+    <KInputSwitch
+      v-model="data.actionEnabled"
+      label="Action button enabled"
+    />
+  </div>
 
-### actionButtonText
+  <KModal
+    :action-button-disabled="!data.actionEnabled"
```
adamdehaven, line 154: since we're talking about disabling the button, I think it should be disabled by default in the example. Applies to **all** disabled examples in the other props too

docs/components/modal.md
```diff
@@ -1,348 +1,752 @@
...
+  @cancel="handleModalClose"
+  @proceed="handleModalProceed"
+>
+  Modal header section is omitted when no header content is present.
+</KModal>
+```
+
+### proceedOnEnter
+
+Whether pressing on Enter should result in KModal emitting the [`proceed` event](#proceed). Defaults to `true`.
```
adamdehaven, line 305: This example does not function properly on the docs site; modal stays open and nothing happens.

docs/components/modal.md
```diff
@@ -1,348 +1,752 @@
...
+<KModal
+  width="90%"
+  title="Modal"
+  :visible="modalVisible"
+  @cancel="handleModalClose"
+  @proceed="handleModalProceed"
+/>
+```
+
+### maxHeight
```
adamdehaven, line 408: suggestion: move width and max-height above the action button props since these are likely more widely used

docs/components/modal.md
```diff
@@ -1,348 +1,752 @@
...
 
-There are 4 designated slots you can use to display content in the modal.
+### default
 
-- `header-image` - content displayed above the header, ignores padding
-- `header-content`
-- `body-content`
-- `footer-content` - Contains cancel & action buttons by default.
-- `action-buttons` - Contains action buttons which are right-aligned. This slot will not exist if using `footer-content` slot.
+Slot for modal content.
```
adamdehaven, line 445: Can you mention something here about the differences between `default` and `content` slots?

docs/components/modal.md
```diff
@@ -1,348 +1,752 @@
...
+
+:::tip NOTE
+KModal takes care of placement and spacing between the buttons when using `footer-actions` slot. However, should you want to mimic that container behaviour using, for example, [`footer`](#footer) slot, we recomend you use `$kui-space-40` [design token](https://github.com/Kong/design-tokens) for spacing between the buttons and place the container containing the buttons on the right using `margin-left: $kui-space-auto;` rule.
+:::
+
+### content
+
+By default KModal provides you with the standard layout: modal header (optional), content section and footer. However, should you need to make a custom modal window, you can use `content` slot for your content. It will provide you with just a padded container with white background.
+
+:::warning NOTE
```
adamdehaven, line 587:
```suggestion
:::warning IMPORTANT
```

docs/components/modal.md
```diff
@@ -1,348 +1,752 @@
...
+Focus-trap requires at least one tabbable element to be present in modal at all times. You can learn more about what elements are considered tabbable [here](https://github.com/focus-trap/tabbable).
```
adamdehaven, line 588: Can you provide guidance here on what this means? e.g. ensure your modal has at least one element ... blah

docs/components/modal.md
```diff
@@ -1,348 +1,752 @@
...
+:::
+
 ## Events
 
-| Event      | Description                                 |
-| :--------- | :------------------------------------------ |
-| `canceled` | Emitted when cancel/close button is clicked |
+### proceed
+
+Emitted when action button is clicked. Doesn't pass any payload.
```
adamdehaven, line 662:
```suggestion
Emitted when action button is clicked.
```

docs/components/prompt.md
```diff
@@ -1,266 +1,468 @@
...
-Are you looking for a modal with no close icon, centered text, and the ability to display an image in the header of the modal? Try [KModal](/components/modal.html) instead.
-:::
-
-The **KPrompt** component is used to display a dialog that prompts a user to take an action.
+KPrompt is a pop-up window that temporarily interrupts the user's interaction with the main content, usually to require user confirmation before proceeding with an action. This component uses [KModal](/components/modal) under the hood.
 
-<KButton appearance="primary" @click="defaultIsOpen = true">Prompt</KButton>
+:::tip NOTE
+Consider using [KModal](/components/modal) instead if your use case is one of the below:
+* you need to display some information to the user but their input is not required
```
adamdehaven, line 7:
```suggestion
* you need to display some information to the user but their input/interaction is not required
```

adamdehaven (changes requested):

src/components/KPrompt/KPrompt.vue
```diff
@@ -1,332 +1,189 @@
...
+      <div
+        v-if="confirmationText"
+        class="prompt-confirmation-container"
+      >
+        <p class="prompt-confirmation-text">
+          {{ confirmationPromptText[0] }} <span class="confirmation-text">"{{ confirmationText }}"</span>
+          {{ confirmationPromptText[1] ? confirmationPromptText[1] : '' }}
+        </p>
+        <KInput
+          v-model.trim="confirmationInput"
```
adamdehaven, line 37: issue: do not trim here; they **must** type the exact text
```suggestion
          v-model="confirmationInput"
```

sandbox/pages/SandboxModals.vue
```diff
@@ -0,0 +1,600 @@
...
+                <KButton
+                  @click="data.modalVisible = !data.modalVisible"
+                >
+                  KModal
+                </KButton>
+              </div>
+              <KInputSwitch
+                v-model="data.modalActionEnabled"
+                label="Action button enabled"
+              />
```
adamdehaven, line 137: These input switches in the sandbox examples; can't you just combine the two examples in the section to 1 instance of the KInputSwitch instead of having 1 for each? Applies to each section
image

docs/components/modal.md
adamdehaven: I think it's because the element isn't focused when opened - this is different behavior than existed previously, so it's broken

src/components/KModal/KModal.vue
```diff
@@ -100,191 +86,147 @@
...
 }
 
-watchEffect(() => {
-  if (typeof document !== 'undefined') {
-    if (props.isVisible) {
-      // Hide body overflow
-      document?.body?.classList.add('k-modal-overflow-hidden')
-    } else {
-      // Reset body overflow
-      document?.body?.classList.remove('k-modal-overflow-hidden')
```
adamdehaven, line 269: This functionality needs to be restored

docs/guide/migrating-to-version-9.md
```diff
@@ -319,16 +319,41 @@ This component has been removed. Please refer to KBadge component which has been
...
+* `hideTitle` prop has been removed
+* `hideDismissIcon` prop has been removed and replaced with `hideCloseIcon`
+* `dismissButtonTheme` prop has been removed
+* `textAlign` prop has been removed
+* `isVisible` prop has been removed and replaced with `visible` prop
+* `content` prop has been removed. You can use the `default` slot instead
 
 #### Slots
 
+* `header-image` slot has been removed. You can use `content` to pass custom content to KModal
```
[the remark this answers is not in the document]
portikM: What do you mean?

sandbox/pages/SandboxModals.vue
portikM: Actually I would prefer to keep those toggles separate for each component. Having one for each helps making sure that attributes don't leak since KPrompt is using KModal under the hood.

src/components/KModal/KModal.vue
```diff
@@ -1,97 +1,83 @@
 <template>
   <div
-    v-if="isVisible"
+    v-if="visible"
     :aria-label="title"
     aria-modal="true"
     class="k-modal"
+    :class="[attrs.class]"
+    v-bind="sanitizedAttrs"
```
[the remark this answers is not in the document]
portikM: I need it because of `inheritAttrs: false` (which I needed to add because some attributes were leaking from KPrompt)

src/components/KModal/KModal.vue
```diff
@@ -100,191 +86,147 @@
...
+
+const handleKeydown = (event: any): void => {
+  // close on escape key press
+  if (props.visible && event.key === 'Escape') {
     close(true)
   }
+
+  if (props.visible && event.key === 'Enter' && props.proceedOnEnter) {
+    proceed()
+  }
```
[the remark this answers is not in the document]
portikM: Fixed

docs/guide/migrating-to-version-9.md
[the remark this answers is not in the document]
adamdehaven: don't we need to document new things in this migration document as well?

src/components/KModal/KModal.vue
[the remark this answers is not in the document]
adamdehaven: Hmm... don't resolve this comment yet

docs/components/modal.md
adamdehaven: I'm not sure this makes sense -- hitting "Enter" on the modal/prompt doesn't do anything regardless. Did this break prior to all of this alpha work?

@kaiarrowood I think you originally implemented this?

portikM: It's working - you gotta blur from the KButton: https://www.loom.com/share/d8bc979195c342fb9ce50011b0a751a4?sid=4bd23e94-0cd9-4319-9c25-f2e8ebbba403

docs/guide/migrating-to-version-9.md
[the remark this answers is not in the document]
portikM: We don't typically document the new props/slots here - just the breaking changes (removed and renamed ones). I do mention `content ` slot as a recommendation here, however to learn more about it them can go to component docs page.

docs/components/modal.md
adamdehaven: This didn't used to be the case. IIRC this prop was added because at one point when the dialog opened you could immediately press enter; otherwise, the prop doesn't make sense..?

portikM: I think it works as expected. The focus is set on the button, not on KModal. When I don't set focus on anything - it closes on Enter: https://www.loom.com/share/8c340b2c79524eccaf5c28cb40165fa0?sid=8ce509ca-61dc-4b98-b0e0-e60e7b017feb

adamdehaven: What I'm saying is that the modal/prompt should be automatically focused, or an element within it, when opened. This used to work this way, so there has been a regression at some point.

adamdehaven:
image
When the modal title is too long, it extends beyond the viewport's width. This is from the first example in the sandbox.

src/components/KModal/KModal.vue
```diff
@@ -100,343 +87,270 @@
...
+  maxHeight: {
+    type: String,
+    required: false,
+    default: 'calc(100vh - 200px)',
+  },
+  closeOnBackdropClick: {
+    type: Boolean,
+    default: false,
+  },
+  proceedOnEnter: {
```
adamdehaven, line 159: There is some sort of bug here. In the sandbox, if I open the modal, don't focus it, and press `Enter` on my keyboard, the `proceed` event is emitted but the modal doesn't actually close.
image

adamdehaven: (I added a console log in the `proceed()` method to illustrate)

src/components/KModal/KModal.vue
```diff
@@ -100,343 +87,270 @@
...
+  return attributes
+})
+
+const showHeader = computed((): boolean => {
+  return !!props.title || !!slots.title || !props.hideCloseIcon
 })
 
-const handleKeydown = (e: any): void => {
-  if (props.isVisible && e.keyCode === 27) {
+const handleKeydown = (event: any): void => {
```
adamdehaven, line 195: This event is too "greedy" and listens for `Enter` no matter what element in the DOM is focused.

The `Enter` keypress should only be triggered if the modal or one of its children are focused.

src/components/KPrompt/KPrompt.vue
```diff
@@ -1,332 +1,189 @@
...
-  }
+  // delete attributes that are handled through modalAttributes prop
+  delete attributes['tabbable-options']
+  delete attributes.width
+  delete attributes['max-height']
+  delete attributes['close-on-backdrop-click']
+  delete attributes['proceed-on-enter']
+  // delete KModal props that are not available in KPrompt
+  delete attributes['hide-close-icon']
+  delete attributes['hide-cancel-button']
```
adamdehaven, line 146: I question whether some of this is necessary. Props are not automatically bound to children the same way classes and styles are.

adamdehaven (changes requested):

adamdehaven:
With the changes in this PR, the default KModal and default KPrompt are nearly identical. What's the point in having two components at this point instead of just a single "KModal" that can include the prompt input and confirmation text?

### Modal

image

### Prompt

image

src/components/KPrompt/KPrompt.vue
portikM: Yes, it is necessary - I tested it and the prop is prop is propagated to KModal unless deleted like so (because KModal is not aa child of KPrompt - it's the first tag in template)

src/components/KModal/KModal.vue
portikM: Changed to only have event listener when modal is open

portikM:
> image When the modal title is too long, it extends beyond the viewport's width. This is from the first example in the sandbox.

Fixed (changed `width` prop to `maxWidth`)

docs/components/modal.md
portikM: Fixed now (both in sandbox and in docs)

kaiarrowood: Chiming in on expected behavior. As far as focus trap goes, I don't think we ever had it working that `Enter` would immediately do something (I think that's an accessibility no-no, because we would manually be grabbing the user's focus and dropping it wherever we chose which could be confusing for non-sighted) - the only place I did see this behavior was v7 KModalFullscreen (didn't check v8). 

KPrompt/KModal have never supported immediate close with `Enter` and honestly not sure they should. We SHOULD allow immediate close with `Esc` though, that doesn't require any focus change - it's just treated as a native keyboard action that always tries to exit out of whatever is focused.

adamdehaven: So then what is the purpose of the `preventProceedOnEnter` prop? How did it ever function?

kaiarrowood: `preventProceedOnEnter` is breaking the tie between pressing Enter key and sending the proceed event when focused on the dialog. I believe the thinking was something like, if the user completely controls the buttons by using the footer slot, we don't want to send a proceed event since we don't know the state of the buttons (ie. you wouldn't send `@proceed` if the button is disabled).

adamdehaven: ...then the prop can also be removed

portikM: I'm fine with this approach. If needed, host app can take care of performing an action on enter:
```
// inside modal body
<KInput @enter.prevent="...
```

docs/guide/migrating-to-version-9.md
```diff
@@ -387,14 +412,31 @@ KPagination now uses [KDropdown](/components/dropdown) instead of [KSelect](/com
...
+* `body-content` slot has been changed to `default`
+* `action-buttons` slot has been removed. If you need to slot in custom action buttons, check out KModal component
+
 #### Events
 
+* `canceled` event has been renamed to `cancel`
+
+#### Structure
+
+* `k-prompt-confirm-text` class has been been changed to `prompt-confirmation-text`
```
kleinfreund, line 433: **Suggestion**: This is missing a removal note for `.k-prompt-proceed`.

kleinfreund, line 433: **Suggestion**: Typo + missing removal note:

```suggestion
* `k-prompt-confirm-text` class has been changed to `prompt-confirmation-text`
* `k-prompt-proceed` class has been removed
```

docs/components/modal.md
portikM: Removed the prop

adamdehaven (approved):
lgtm

PR #2009, opened 2024-02-09
feat(ktruncate): reskin component [KHCP-9009]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-9009

`KTruncate` component reskin

## PR Checklist

### DIFF
--- a/docs/components/truncate.md
+++ b/docs/components/truncate.md
@@ -44,14 +44,14 @@ Number of rows to truncate content. Default value is `1`.
 </template>
 ```
 
-### textContent
+### truncateText
 
 By default the component treats anything passed through the `default` slot as collection of HTML elements. Use this prop if you want to truncate elements that only contain text, such as one or more paragraph `<p>` tags.
 
 When this prop is set to `true`, the component applies different logic; truncation is achieved via the `-webkit-line-clamp` CSS property, rather than assessing the height of child elements to determine the row height.
 
 <KCard>
-  <KTruncate text-content :rows="3">
+  <KTruncate truncate-text :rows="3">
     <p>A good design decision to apply <a href="https://kongponents.konghq.com/components/truncate.html#textcontent">text truncation</a> would be when displaying a <b>large amount of textual content</b>, such as a list of articles or product descriptions, in a limited space, such as a mobile screen or a small widget. By truncating the text to a short summary, it is possible to present the information in a more organized and readable manner, allowing the user to quickly scan and understand the main points of each item.</p>
     <p>The truncated text can also serve as a teaser, encouraging the user to click or tap to view the full content.</p>
   </KTruncate>
@@ -59,7 +59,7 @@ When this prop is set to `true`, the component applies different logic; truncati
 
 ```html
 <template>
-  <KTruncate text-content :rows="3">
+  <KTruncate truncate-text :rows="3">
     <p>A good design decision to apply text truncation would be
       when displaying a <b>large amount of textual content</b>, such as a list
       of articles or product descriptions, in a limited space, such as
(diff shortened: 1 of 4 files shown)


docs/components/truncate.md
```diff
@@ -44,22 +44,22 @@ Number of rows to truncate content. Default value is `1`.
 </template>
 ```
 
-### isTextContent
+### textContent
```
adamdehaven, line 47: kind of wonder if we should "better name" this prop if we're changing it anyway. "textContent" kind of seems like I'm supposed to pass it a string?

docs/components/truncate.md
```diff
@@ -44,22 +44,22 @@ Number of rows to truncate content. Default value is `1`.
...
 
 <KCard>
-  <KTruncate is-text-content :rows="3">
-    <p>A good design decision to apply <a href="https://kongponents.konghq.com/components/truncate.html#istextcontent">text truncation</a> would be when displaying a <b>large amount of textual content</b>, such as a list of articles or product descriptions, in a limited space, such as a mobile screen or a small widget. By truncating the text to a short summary, it is possible to present the information in a more organized and readable manner, allowing the user to quickly scan and understand the main points of each item.</p>
+  <KTruncate text-content :rows="3">
+    <p>A good design decision to apply <a href="https://kongponents.konghq.com/components/truncate.html#textcontent">text truncation</a> would be when displaying a <b>large amount of textual content</b>, such as a list of articles or product descriptions, in a limited space, such as a mobile screen or a small widget. By truncating the text to a short summary, it is possible to present the information in a more organized and readable manner, allowing the user to quickly scan and understand the main points of each item.</p>
```
adamdehaven, line 55: This example seems odd because you're still passing it HTML?

src/components/KTruncate/KTruncate.vue
```diff
@@ -43,18 +44,20 @@
         >
           <ChevronUpIcon
             class="collapse-trigger"
-            :color="KUI_COLOR_TEXT_PRIMARY"
             role="button"
             tabindex="0"
             @click.stop="handleToggleClick"
+            @enter="handleToggleClick"
+            @keydown.space.prevent
+            @keyup.space="handleToggleClick"
```
adamdehaven, line 52: Couple questions:

- Shouldn't this be keydown.enter? 
- why are we binging space to keyup and then preventing default on keydown?

src/components/KTruncate/KTruncate.vue
```diff
@@ -282,30 +287,35 @@ onUnmounted(() => {
...
       box-shadow: var(--kui-shadow-focus, $kui-shadow-focus);
     }
 
     &:hover {
       background-color: var(--kui-color-background-primary-weaker, $kui-color-background-primary-weaker);
+      color: var(--kui-color-text-primary-strong, $kui-color-text-primary-strong) !important;
     }
 
-    &:focus {
+    &:focus, &:focus-within {
```
adamdehaven, line 304: nit: drop to new line
```suggestion
    &:focus, 
    &:focus-within {
```

docs/components/truncate.md
portikM: How about just `text`?

docs/components/truncate.md
portikM: I'm passing some HTML there just to show that it doesn't have to be plain text - some HTML is ok

src/components/KTruncate/KTruncate.vue
adamdehaven: - If I click enter or space on the expand button, it works.
- Only the space key works on the collapse button to close.
- The focus-visible state of the collapse button looks... odd
 
image

portikM: - Yes, my mistake - corrected
- Preventing default on space keydown so that it doesn't scroll to the element when pressing enter. Binding to keyup so that pressing space triggers the event

adamdehaven: For the expand button, it's not 100% obvious the count e.g. `+6` is a clickable button. Should it have a background color and border radius similar to the chevron icon?

src/components/KTruncate/KTruncate.vue
```diff
@@ -43,18 +44,20 @@
         >
           <ChevronUpIcon
             class="collapse-trigger"
-            :color="KUI_COLOR_TEXT_PRIMARY"
             role="button"
             tabindex="0"
             @click.stop="handleToggleClick"
+            @enter="handleToggleClick"
```
portikM, line 50:
```suggestion
            @keyup.enter="handleToggleClick"
```

docs/components/truncate.md
```diff
@@ -121,10 +121,10 @@ The content to truncate.
 **Note:** when using KTruncate for truncating collection of HTML elements (not text), to ensure the best experience with the component, please pass elements of equal height in the default slot. The component will base its guess for the height of the element on the height of each individual sibling, so passing elements of equal height will result in a more accurate and consistent output.
 
 :::tip TIP
-The component is reactive to its dimensions changes. To see this in action, try resizing your browser window and notice the behavior of the example below.
+The component is reactive to its dimensions changes. To see this in action, try resizing the container below.
 :::
 
-<KCard>
+<KCard class="resizable-card">
```
adamdehaven, line 127: nit: can you set a min-width on the resizable container?
image

docs/guide/migrating-to-version-9.md
```diff
@@ -570,6 +570,18 @@ Removed as of `v9`. Use `KBreadcumbs` instead.
 
 ### KTruncate
 
+#### Props
+
+* `isTextContent` prop has been changed to `textContent`
```
adamdehaven, line 575: the naming of this prop is a bit confusing; I would think I would pass in a string to `textContent` - is there a better name we could use?

docs/components/truncate.md
adamdehaven: `truncateText: boolean`? false by default 🤷🏼

portikM: Renamed

src/components/KTruncate/KTruncate.vue
portikM: Design approved it the way it is -- we can change it later when design comes up with Figmas for this component (right now we don't have any)

portikM: > * If I click enter or space on the expand button, it works.
> * Only the space key works on the collapse button to close.
> * The focus-visible state of the collapse button looks... odd
> 
> image

Updated th component so all buttons work on click, enter and space. As for `focus-visible` styling - again, maybe let's correct that if we need when/if design wants to change it?

adamdehaven: I don't think changing the default behavior of buttons in the browser is desired (for this, and the other PRs)

PR #2015, opened 2024-02-14
feat(ktoaster): component reskin [KHCP-9005]

portikM:
# Summary

[Jira](https://konghq.atlassian.net/browse/KHCP-9005)
[Figma](https://www.figma.com/file/Yze0SWXl5nKjR0rFdilljK/Components?type=design&node-id=497%3A9761&mode=dev&t=lcG4fMxcpZoeYwsC-1)

`KToaster` component reskin

Links:

- [Kongponents sandbox](https://deploy-preview-2015--kongponents-sandbox.netlify.app/#/toaster)
- [Kongponents docs preview](https://deploy-preview-2015--kongponents.netlify.app/components/toaster.html)

## PR Checklist

### DIFF
--- a/docs/components/toaster.md
+++ b/docs/components/toaster.md
@@ -1,8 +1,8 @@
 # Toaster
 
-**KToaster** - a popup notification typically used to show the result of an action. The toaster can close on its own but can also be manually dismissed.
+KToaster a popup notification component. The toaster will close automatically after a set timeout or can be dismissed by clicking on the close icon.
 
-KToaster is used via the `ToastManager` instance. All rendering is controlled from ToastManager via an intuitive, imperative api. It is recommended that you initialize `ToastManager` in your app via [`app.config.globalProperties`](https://vuejs.org/api/application.html#app-config-globalproperties) to allow you to access it on any component instance inside your application.
+KToaster is used via the `ToastManager` instance. All rendering is controlled from ToastManager via an intuitive, imperative API. It is recommended that you initialize `ToastManager` in your app via [`app.config.globalProperties`](https://vuejs.org/api/application.html#app-config-globalproperties) to allow you to access it on any component instance inside your application.
 
 ```ts
 // main.ts
@@ -30,30 +30,23 @@ declare module 'vue' {
 
 Once `ToastManager` is added as a global property, you can access it's methods via `this.$toaster`
 
-<KButton @click="$toaster.open('Basic Notification')">Open Toaster</KButton>
+<KButton @click="$toaster.open({ title: 'Basic Notification', message: 'Detailed message' })">Open Toaster</KButton>
 
 ```html
-<KButton @click="$toaster.open('Basic Toaster')">Open Toaster</KButton>
+<KButton @click="$toaster.open({ title: 'Basic Notification', message: 'Detailed message' })">Open Toaster</KButton>
 ```
 
 or within the `setup()` function in your component
 
-```html
-<script lang="ts">
-import { defineComponent, getCurrentInstance } from 'vue'
+```ts
+<script setup lang="ts">
+import { getCurrentInstance } from 'vue'
 
-export default defineComponent({
-  setup() {
-    const $toaster = getCurrentInstance()?.appContext.config.globalProperties.$toaster
+const $toaster = getCurrentInstance()?.appContext.config.globalProperties.$toaster
 
-    const showToast = (name: string) => {
-      $toaster.open(`Wow, ${name} is looking toasty!`)
-    }
-
-    return { showToast }
-  }
-})
-</script>
+const showToast = (name: string) => {
+  $toaster.open(`Wow, ${name} is looking toasty!`)
+}
 ```
 
 ::: warning NOTE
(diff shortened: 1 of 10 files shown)


src/components/KToaster/KToaster.vue
```diff
@@ -41,76 +61,128 @@ defineProps({
...
+        background-color: var(--kui-color-background-primary-weak, $kui-color-background-primary-weak); // info appearance as default in case of invalid appearance
+        border-radius: var(--kui-border-radius-circle, $kui-border-radius-circle);
+        display: flex;
+        height: 32px;
+        justify-content: center;
+        width: 32px;
+      }
+
+      .toast-title {
+        @include truncate;
```
adamdehaven, line 121: issue: i'm not sure it makes sense to truncate the toast title, instead of wrapping the title text. If we truncate, there is literally no way for the user to see the toast title content.

docs/components/toaster.md
```diff
@@ -64,192 +57,202 @@ If a global property conflicts with a component’s own property, the component'
...
-<KButton @click="$toaster.open('Default message here')">Open Toaster</KButton>
+<KButton @click="$toaster.open({ title: 'Long Title Gets Truncated With Ellipsis' })">Open Toaster</KButton>
+<KButton @click="$toaster.open('String Will Become A Title')" appearance="secondary">Open Toaster</KButton>
 ```
 
-### appearance
+### message
 
-The Toaster uses the same appearance values as [KAlert](/components/alert) and are applied the same way.
+Because a long title gets truncated, it is only good for short notifications. When a longer notification needs to be shown to the user, `message` property becomes helpful.
```
adamdehaven, line 88: See my comment on truncation

docs/components/toaster.md
```diff
@@ -64,192 +57,202 @@ If a global property conflicts with a component’s own property, the component'
...
+  <KButton
+    @click="openNotification('Basic Notification')"
+    appearance="secondary"
+  >
+    Open Toaster
+  </KButton>
+</div>
+
+<pre>
+  {{ JSON.stringify(toasters || [], null, 2) }}
```
adamdehaven, line 200: I like the idea of showing this data, but should the container have a max-height so the page doesn't grow/shrink?

docs/components/toaster.md
```diff
@@ -64,192 +57,202 @@ If a global property conflicts with a component’s own property, the component'
...
-- `timeoutMilliseconds`
+```ts
+$toaster.open({ title: 'Warning', appearance: 'warning' })
+```
 
-<KButton :disabled="timeLeft <= 3" @click="openNotificationElapse({timeoutMilliseconds: 3000, 'appearance': 'success', 'message': `This toaster has a 3 second timeout`})">{{timeLeft > 3 ? 'Open Toaster' : `Closing in ${timeLeft} seconds` }}</KButton>
+### timeoutMilliseconds
 
-```html
+The default timeout, in milliseconds, is `5000` (5 seconds) however you can change this to by passing the `timeoutMilliseconds` property.
```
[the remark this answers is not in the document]
portikM: I didn't, it was 5 seconds before - the docs were wrong

docs/components/toaster.md
portikM: Made it fixed height and scrollable

src/components/KToaster/KToaster.vue
```diff
@@ -41,76 +61,128 @@ defineProps({
...
+      }
+
+      .toast-close-icon {
+        border-radius: var(--kui-border-radius-20, $kui-border-radius-20);
+        cursor: pointer;
+        margin-left: var(--kui-space-auto, $kui-space-auto);
+        outline: none;
+
+        &:hover, &:focus {
+          color: var(--kui-color-text-neutral-weaker, #e0e4ea) !important; // TODO: add token
```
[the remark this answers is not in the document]
portikM: Yes, but you can help me get it done by stamping this: https://github.com/Kong/design-tokens/pull/231

src/components/KToaster/KToaster.vue
portikM: We discussed it with @two24studios in the Slack thread - I suggest we truncate the title when it’s too long so it doesn’t take too much space - this way we limit title to just 3-4 words. If that’s not enough to say whatever the notification needs to say - the longer message should be passed though `message` property.

adamdehaven: I understand what you're saying, but I don't think it makes sense to truncate does it? we're hiding something the user has no other way of seeing. Wouldn't it be the responsibility of the host app to pass in a reasonable string?

src/components/KToaster/KToaster.vue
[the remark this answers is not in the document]
adamdehaven: ✅

src/components/KToaster/KToaster.vue
portikM: Well yeah I guess that's what I meant, that it would be host app's responsibility to make sure to provide reasonable length string and compliment it with a message when needed. We can also add `title` attribute as a way for users to view the full title on hover.

adamdehaven: you can't really expect a user to hover on a tooltip that goes away

adamdehaven: I'll resolve the thread; just still not sure this is a great pattern

docs/components/toaster.md
```diff
@@ -30,30 +30,23 @@ declare module 'vue' {
 
 Once `ToastManager` is added as a global property, you can access it's methods via `this.$toaster`
```
adamdehaven, line 31: > Once `ToastManager` is added as a global property, you can access it's methods via `this.$toaster` if using the Vue Options API.

docs/components/toaster.md
```diff
@@ -64,192 +57,207 @@ If a global property conflicts with a component’s own property, the component'
...
+  message?: string
+  appearance?: ToasterAppearance
+  timeoutMilliseconds?: number
+}
+```
 
-<KButton @click="$toaster.open('Default message here')">Open Toaster</KButton>
+### title
+
+Notification title. When passing an object to `$toaster.open()` method, the `title` property is required. When passing a string, it will be rendered as title.
```
adamdehaven, line 74:
```suggestion
Notification title. When passing an object to `$toaster.open()` method, the `title` property is required. When passing only a string, the string will be rendered as the title.
```

adamdehaven: Can you also add a note here that titles are truncated?

docs/components/toaster.md
```diff
@@ -64,192 +57,207 @@ If a global property conflicts with a component’s own property, the component'
...
-<KButton @click="$toaster.open('Default message here')">Open Toaster</KButton>
+<KButton @click="$toaster.open({ title: 'Long Title Gets Truncated With Ellipsis' })">Open Toaster</KButton>
+<KButton @click="$toaster.open('String Will Become A Title')" appearance="secondary">Open Toaster</KButton>
 ```
 
-### appearance
+### message
 
-The Toaster uses the same appearance values as [KAlert](/components/alert) and are applied the same way.
+Because a long title gets truncated, it is only good for short notifications. When a longer notification needs to be shown to the user, `message` property becomes helpful.
```
adamdehaven, line 88:
```suggestion
The message prop allows for displaying longer strings of text to the user. This prop is good for more detailed messages, or displaying IDs, etc.

Because a long title gets truncated, it is only good for short notifications. When a longer notification needs to be shown to the user, the `message` property is ideal.
```

docs/components/toaster.md
```diff
@@ -64,192 +57,207 @@ If a global property conflicts with a component’s own property, the component'
...
+  Open Toaster
+</KButton>
 ```
 
-### timeout
+### appearance
 
-The default timeout, in milliseconds, is `5000` (5 seconds) however you can change this to by passing an override argument.
+Depending on the nature of notification, you might want to use different appearances. KToaster supports these 5 appearances:
+* `info` (default)
```
adamdehaven, line 109: nit: add empty line
```suggestion

* `info` (default)
```

docs/components/toaster.md
```diff
@@ -64,192 +57,207 @@ If a global property conflicts with a component’s own property, the component'
...
+
+const toasterOptions: Toast = {
+  appearance: 'success',
+  timeoutMilliseconds: 3000,
+  title: 'Notification',
+  message: 'This toaster has a 3 second timeout'
+}
+
+const openNotification = (options: Toast | string) => {
+  $toaster.open(options)
```
adamdehaven, line 172: I don't think this example works when using the Composition API? Same comment for all instances.

adamdehaven: I think it would need to be something like this, showing how you can utilize via composable:

```typescript
import { ToastManager } from '@kong/kongponents'
import type { ToasterNotification } from '@kong/kongponents'

// Initialize the ToastManager as a singleton so a new instance isn't created each time
const toaster = new ToastManager()

export default function useToaster() {
  const notify = async (notification: Partial<ToasterNotification>): Promise<void> => {
    toaster.open({
      ...notification,
    })
  }

  return {
    notify,
  }
}
```

docs/guide/migrating-to-version-9.md
```diff
@@ -570,6 +570,21 @@ Removed as of `v9`. Use `KBreadcumbs` instead.
 
 ### KToaster
 
+#### Attributes
+
+#### Structure
+
+* `toaster-container-outer` class has been replaced with `k-toaster`
+* `toaster-item` class has been replaced with `toast`
```
adamdehaven, line 578: seems like we're switching between "toaster" and "toast" - can we pick one? 

I would think "toaster" since that's the name of the component?

adamdehaven: Selectors, the constants and types, etc.

src/components/KToaster/KToaster.vue
```diff
@@ -41,76 +61,128 @@ defineProps({
...
-  max-width: 300px;
+  display: flex;
+  flex-direction: column;
+  gap: var(--kui-space-50, $kui-space-50);
   position: fixed;
   right: 16px;
-  transition: $transition;
-  width: auto;
+  width: 400px;
   z-index: 10000;
```
adamdehaven, line 92: suggestion: can we expose the `z-index` as a prop? Oftentimes you may need to override depending on the structure of the rest of your app. You can use this value as the default

docs/components/toaster.md
portikM: It works and it is how we have it in script section of this file. This example refers to `timeoutMilliseconds` argument.

docs/guide/migrating-to-version-9.md
portikM: Updated - only left the `Toast` interface name unchanged because it was called that before and I think it makes sense (interface of a **single Toast**)

src/components/KToaster/KToaster.vue
portikM: Same as in the other thread - we're not really exposing `z-index` as prop anywhere else so it's weird? Maybe leave that to CSS?

docs/components/toaster.md
adamdehaven: In this file yes because it's using the Options API right?

src/components/KToaster/KToaster.vue
adamdehaven: But why not? We expose max width, isn't it the same concept?

adamdehaven: My thinking: let's say you have the intercom widget that is zindex of 20000 - you'd have to globally override somewhere in your app instead of just passing it as a prop to override

portikM: How about this: I'll handle adding zIndex in all components that need it (KModal, KTooltip, KToaster, KSlideout) in a separate PR?

docs/components/toaster.md
portikM: No, it's using script setup with `getCurrentInstance` (screenshot). Ultimately, I think you are right and this component docs need to be updated to feature composable approach - I just don't really want to do it in this PR - how about we do it in `beta` when we refactor it not to create a second app instance?

Screenshot 2024-02-16 at 11 07 07 AM

src/components/KToaster/KToaster.vue
adamdehaven: works for me; just throw a ticket in the backlog; we can address it later if/when it becomes an issue (doesn't need to be part of the reskin effort)

docs/components/toaster.md
adamdehaven: works for me 👍🏼

adamdehaven (approved):

PR #2012, opened 2024-02-12
feat(kslideout): component reskin [KHCP-9000]

portikM:
# Summary

[Jira](https://konghq.atlassian.net/browse/KHCP-9000)
[Figma](https://www.figma.com/file/Yze0SWXl5nKjR0rFdilljK/Components?type=design&node-id=2020%3A9543&mode=dev&t=USFhEkebi5cmYxaK-1)

`KSlideout` component reskin

Links:

- [Kongponents sandbox](https://deploy-preview-2012--kongponents-sandbox.netlify.app/#/slideout)
- [Kongponents docs preview](https://deploy-preview-2012--kongponents.netlify.app/components/slideout.html)

## PR Checklist

* [ ] **Conventional Commits** all commits follow the conventional commit standards [outlined in the main README](https://github.com/Kong/kongponents#committing-changes).
* [ ] **Tests coverage:** test coverage was added for new features and bug fixes
* [ ] **Docs:** includes a technically accurate README

PR #2020, opened 2024-02-16
feat(ktruncate, kslideout, ktoaster): components reskin

portikM:
# Summary

`KTruncate`, `KSlideout` and `KToaster` components reskin

- https://github.com/Kong/kongponents/pull/2009
- https://github.com/Kong/kongponents/pull/2012
- https://github.com/Kong/kongponents/pull/2015

## PR Checklist

### DIFF
--- a/src/components/KSlideout/KSlideout.vue
+++ b/src/components/KSlideout/KSlideout.vue
@@ -124,16 +124,6 @@ const handleClose = (e: any, forceClose = false): void => {
   }
 }
 
-const toggleBodyScroll = (isScrollable: boolean): void => {
-  if (typeof document !== 'undefined') {
-    if (isScrollable) {
-      document.body.classList.remove('k-slideout-overflow-hidden')
-    } else {
-      document.body.classList.add('k-slideout-overflow-hidden')
-    }
-  }
-}
-
 const toggleEventListeners = (isActive: boolean): void => {
   if (typeof document !== 'undefined') {
     if (isActive) {
@@ -146,10 +136,8 @@ const toggleEventListeners = (isActive: boolean): void => {
 
 watch(() => props.visible, async (visible: boolean): Promise<void> => {
   if (visible) {
-    toggleBodyScroll(false)
     toggleEventListeners(true)
   } else {
-    toggleBodyScroll(true)
     toggleEventListeners(false)
   }
 }, { immediate: true })
@@ -167,6 +155,7 @@ onUnmounted(() => {
     box-shadow: var(--kui-shadow, $kui-shadow);
     display: flex;
     flex-direction: column;
+    flex-grow: 1;
     gap: var(--kui-space-50, $kui-space-50);
     height: calc(100vh - v-bind('offsetTopValue'));
     max-width: v-bind('props.maxWidth');
@@ -227,6 +216,10 @@ onUnmounted(() => {
       line-height: var(--kui-line-height-30, $kui-line-height-30);
       overflow-y: auto;
       padding-right: var(--kui-space-70, $kui-space-70);
+
+      :deep(> *:last-child) {
+        padding-bottom: var(--kui-space-70, $kui-space-70); // add padding to the last child to add some spacing before bottom of the page
+      }
     }
   }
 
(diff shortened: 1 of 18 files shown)


src/components/KSlideout/KSlideout.vue
```diff
@@ -117,178 +97,159 @@ const emit = defineEmits<{
...
+    }
 
-      .content-card {
-        border: none;
-      }
+    .slideout-content {
+      box-sizing: border-box;
+      color: var(--kui-color-text, $kui-color-text);
+      display: flex;
+      flex-direction: column;
```
kleinfreund, line 223: **Issue**: In our application, I find that the slideout content can now trigger a vertical scrollbar despite there being plenty of space.

| Before | After |
|:------:|:-----:|
| image | image |

There is probably a more elegant solution for this, but in the meantime, I found `flex-grow: 1` to work:


```suggestion
      flex-direction: column;
      flex-grow: 1;
```

johncowen:
Hey @portikM 👋 

little heads up https://github.com/kumahq/kuma-gui/pull/2184#issuecomment-1953976332

Just thinking about it myself but quick FYI seeing as this is still in draft

src/components/KSlideout/KSlideout.vue
```diff
@@ -1,79 +1,59 @@
...
-          </div>
-
-          <!-- title -->
-          <div class="k-slideout-main-title">
-            <p
-              class="k-slideout-title"
-              data-testid="k-slideout-title"
-              :title="title"
-            >
+            <slot name="title">
```
portikM, line 23: @johncowen @kleinfreund based on conversation in https://github.com/kumahq/kuma-gui/pull/2184, if I understand it correctly, these's two outstanding issues with the reskinned KSlideout:
1. Title prop and slot content styles are enforced which makes it harder to make the title custom (e.g. link in your example)
2. The `<body>` element becomes non-scrollable one KSlideout is open, which makes scrolling content in the background impossible

Did I miss anything? Just want to lay questions out before addressing them

portikM: Removing this feature from KSlideout. At first I and Adam thought it's a good idea to make body non-scrollable but we forgot to account for the fact that KSlideout is often used where we still want the content underneath it to be viewable.

portikM: As for my first point, there's nothing we want to do from KSlideout side - it will ship with pre-defined styles for title for consistency and if you need to use your custom title - you might have to overwrite those styles here and there.

johncowen: Gotcha, sounds good to me 👍 let us know if we can help/test anything new in regards to the scroll behaviour

src/components/KSlideout/KSlideout.vue
portikM: Addressed this issue here: https://github.com/Kong/kongponents/pull/2025

src/components/KSlideout/KSlideout.vue
johncowen: Approved that one, watching here and will try it out when its on here.

src/components/KSlideout/KSlideout.vue
portikM: Updated now, please pick up latest version of preview package

src/components/KSlideout/KSlideout.vue
portikM: Thank you! Merged now

src/components/KSlideout/KSlideout.vue
kleinfreund: All good now, thanks 👍🏻

src/components/KSlideout/KSlideout.vue
kleinfreund: Updated https://github.com/kumahq/kuma-gui/pull/2184 and this is working now.

johncowen: Looks much better to me. I do still get the double scrollbar if I manage to make the KSlideout content short enough to make it scroll. But I have to try really hard to do that, especially considering the content we put in these at the moment is relatively small. Plus, really you want to be able to scroll both things separately so its kinda hard to do anything better in this respect I think.

So all in all, I definitely think we've optimised for the most best and common case here 🙌 

@kleinfreund I tried this on my little scratch branch I had, does this look good to you also?

kaiarrowood (approved):
LGTM! 🚀

PR #2028, opened 2024-02-21
fix(ktoaster): icon, title and message styling

portikM:
# Summary

Changes:
* title becomes optional
* when only argument provided to `ToastManager.open()` is string - treat it like message
* styling tweaks displaying icon, title and message

Screenshot 2024-02-20 at 6 01 36 PM
Screenshot 2024-02-20 at 6 02 06 PM

## PR Checklist

### DIFF
--- a/docs/components/toaster.md
+++ b/docs/components/toaster.md
@@ -83,7 +83,7 @@ Notification title.
 
 The message string that allows for displaying longer strings of text to the user. This prop is good for more detailed messages.
 
-Alternatively, if you provide a string as the only argument, it will be treated as message.
+Alternatively, if you provide a string as the only argument to the `open()` method, it will be treated as message.
 
 <div class="horizontal-container">
   <KButton @click="$toaster.open({ 
--- a/docs/components/toaster.md
+++ b/docs/components/toaster.md
@@ -62,7 +62,7 @@ KToaster is the underlying component rendered by the `ToastManager` instance, so
 ```ts
 interface Toast {
   key?: any
-  title: string
+  title?: string
   message?: string
   appearance?: ToasterAppearance
   timeoutMilliseconds?: number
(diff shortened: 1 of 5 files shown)


docs/components/toaster.md
```diff
@@ -71,38 +71,38 @@ interface Toast {
...
+<KButton @click="$toaster.open({ title: 'Notification Title' })">Open Toaster</KButton>
 ```
 
 ### message
 
-The message prop allows for displaying longer strings of text to the user. This prop is good for more detailed messages, or displaying IDs, etc.
+The message string that allows for displaying longer strings of text to the user. This prop is good for more detailed messages.
 
-Because a long title gets truncated, it is only good for short notifications. When a longer notification needs to be shown to the user, the `message` property is ideal.
+Alternatively, if you provide a string as the only argument, it will be treated as message.
```
adamdehaven, line 86: If you pass a string as the only argument to what?

src/components/KToaster/KToaster.vue
```diff
@@ -11,33 +11,36 @@
...
-        {{ toaster.message }}
-      </p>
+      <CloseIcon
+        class="toaster-close-icon"
+        :color="KUI_COLOR_TEXT_NEUTRAL_WEAK"
+        data-testid="toaster-close-icon"
+        role="button"
+        :size="KUI_ICON_SIZE_50"
+        tabindex="0"
+        @click="() => $emit('close', toaster.key)"
```
adamdehaven, line 42: question: why `@click="() => $emit('close', toaster.key)"` and not `@click="$emit('close', toaster.key)"` ?

docs/components/toaster.md
portikM: It's clearer in context - we are talking about arguments passed to `ToastManager.open()` method in this section.

Screenshot 2024-02-21 at 9 02 37 AM

src/components/KToaster/KToaster.vue
portikM: Updated

docs/components/toaster.md
adamdehaven: Since it's further down the page, how about something like this?

```md
Alternatively, if you provide a string as the only argument to the `open()` method, it will be treated as the message.
```

portikM: Updated

adamdehaven (approved):

PR #2031, opened 2024-02-22
feat(kemptystate): component reskin [KHCP-8980]

portikM:
# Summary

[Jira](https://konghq.atlassian.net/browse/KHCP-8980)
[Figma](https://www.figma.com/file/Yze0SWXl5nKjR0rFdilljK/Components?type=design&node-id=2020%3A11252&mode=dev&t=oHjbEUosoGUvnzuN-1)

`KEmptyState` component reskin

Links:

- [Kongponents sandbox](https://deploy-preview-2031--kongponents-sandbox.netlify.app/#/emptystate)
- [Kongponents docs preview](https://deploy-preview-2031--kongponents.netlify.app/components/empty-state.html)

## PR Checklist


src/components/KEmptyState/KEmptyState.vue
```diff
@@ -1,143 +1,173 @@
 <template>
   <section
```
johncowen, line 2: Hey @portikM 👋 

Glad you are doing this component now, I'm hoping to make use of its refactor shortly over in kuma, kinda been waiting for it to happen 🎉 

---

I noticed this is still using `section`. I 'think' we removed sections from KCard, should we replace this one here with a div also?

> The <section> [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents a generic standalone section of a document, which doesn't have a more specific semantic element to represent it. Sections should always have a heading, with very few exceptions.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

I know of a few areas in Kuma where our empty states aren't strictly a 'document section' and/or they will already appear in a `section` we've added in our application (i.e. `<section><KEmptyState /></section>`) so not having a section in the component would avoid a doubled up nested `section` i.e. `<section><section class="k-empty-state">...</section></section>`

lmk, not a huge request, but considering I think we removed sections in KCard this might have just been an oversight not to replace this with a `div`

adamdehaven: Agree, let’s switch to a div

portikM: Done

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -1,143 +1,173 @@
...
-    default: '',
+  iconVariant: {
+    type: String as PropType<EmptyStateIconVariant>,
+    default: EmptyStateIconVariants.Default,
   },
 })
-</script>
 
-<style lang="scss" scoped>
+defineEmits(['action-click'])
```
adamdehaven, line 87: can this be `click-action` instead to be closer to a native click, e.g. `@click-action`

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -1,143 +1,173 @@
...
+    case EmptyStateIconVariants.Config:
+      return CogIcon
+    case EmptyStateIconVariants.File:
+      return FileEmptyIcon
+    case EmptyStateIconVariants.Search:
+      return SearchIcon
+    case EmptyStateIconVariants.Kong:
+      return KongIcon
+    default:
+      return AnalyticsIcon // default variant in case of invalid value
```
adamdehaven, line 104: Importing all of these icons is increasing the bundle size even when most of them won't be used in the component instance.

portikM: Removed a few, only left 3 - the rest can be slotted

docs/components/empty-state.md
```diff
@@ -1,245 +1,323 @@
...
+  title="Empty State Action Button"
+  message="Lorem ipsum dolor sit amet..."
+/>
 ```
 
-### handleClick
+### actionButtonShowIcon
 
-A function that is passed as the click handler for the call to action button
+Boolean to control whether to show or hide add icon in action button. Helpful when using action button for prompting user to create a resource. Defaults to `false`.
```
adamdehaven, line 77: issue: the name of this prop doesn't correlate to what it's actually doing -- you're only adding/hiding the "plus" icon with the prop, but it's named as if it always hides/shows the icon.

Separately I still think this isn't ideal; we don't offer props for any other icons. Seems like over-optimizing for internal use-cases.

adamdehaven (changes requested):

docs/components/empty-state.md
portikM: Removed the new prop

kaiarrowood (approved):
🚀

PR #2041, opened 2024-02-27
feat(kpagination): reskin component [KKHCP-8993]

portikM:
# Summary

## PR Checklist


docs/components/pagination.md
```diff
@@ -1,8 +1,8 @@
 # Pagination
 
-**KPagination** is a component used for pagination.
+KPagination is a component used for pagination.
```
adamdehaven, line 3: Can we do a better description? This is repetitive

docs/components/pagination.md
```diff
@@ -11,275 +11,298 @@
...
+<KPagination :total-count="50"/>
 ```
 
 ### pageSizes
 
-A list of page sizes that the user can choose from.
+An array of numbers of page sizes that the user can choose from. The default page sizes are: `[15, 30, 50, 75, 100]`.
 
-Prop is an Array of numbers. The default page sizes are: `[15, 30, 50, 75, 100]`.
+You can provide custom page sizes. The first one in the array will be the initial page size unless otherwise specified through [`initialPageSize` prop](#initialpagesize).
```
adamdehaven, line 21:
```suggestion
You can provide custom page sizes. The first number in the array will be the initial page size unless otherwise specified through the [`initialPageSize` prop](#initialpagesize).
```

docs/components/pagination.md
```diff
@@ -11,275 +11,298 @@
...
-<KPagination :totalCount="100" :pageSizes="[10, 20, 30, 40]"/>
+<KPagination :total-count="100" :page-sizes="[10, 20, 30, 40]" />
 ```
 
 ### initialPageSize
 
-An optional initial page size.
-
-Prop is a Number. If the value is not set, the first one of the available pageSizes is selected as a page size.
+An optional initial page size. If the value is not set, the first one of the available `pageSizes` is selected as a page size.
```
adamdehaven, line 31:
```suggestion
An optional initial page size. If the value is not set, the first number in the `pageSizes` array is selected as the default.
```

docs/components/pagination.md
```diff
@@ -11,275 +11,298 @@
...
+<script setup lang="ts">
+const letters = ref<string>(['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i'])
+const visibleLetters = ref<string>(['a', 'b', 'c'])
 </script>
 ```
 
 ### currentPage
 
-Manually control the current page instead of using native handling. If using this prop you MUST keep it up-to-date using the `@pageChanged` event in order to remain reactive to clicking the prev, next, and specific page buttons.
+Manually control the current page instead of using native handling. If using this prop you **must** keep it up-to-date using the `@page-change` event in order to remain reactive to clicking the prev, next, and specific page buttons.
```
adamdehaven, line 114: question: what is "native handling"? Description here should be a little more clear

docs/components/pagination.md
```diff
@@ -11,275 +11,298 @@
...
-    <span v-for="number in data.visibleLetters">{{ number }} </span>
+    <span v-for="letter in data.visibleLetters">{{ letter }}</span>
     <KPagination
       :items="data.letters"
-      :totalCount="data.letters.length"
-      :pageSizes="[3]"
-      :currentPage="data.currPage"
-      @pageChanged="({visibleItems, page}) => { data.visibleLetters = visibleItems; data.currPage = page }"/>
+      :total-count="data.letters.length"
+      :page-sizes="[3]"
```
adamdehaven, line 123: question: why would we only show 1 page size? If this was the case, you wouldn't show the UI element

docs/components/pagination.md
```diff
@@ -11,275 +11,298 @@
...
+  visibleLetters.value = offsetLetters.slice((currentOffset.value * 3), ((currentOffset.value * 3) + pageSize))
+}
 </script>
 ```
 
-### paginationType
+### offsetPreviousButtonDisabled
 
-Pass in the type of pagination to be used. Options are `default` (page/pageSize) or `offset` (offset/pageSize)
+Pass in a boolean value for whether or not the offset-based Previous button should be disabled.
```
adamdehaven, line 206:
```suggestion
Pass in a boolean value for whether or not the offset-based previous button should be disabled.
```

docs/components/pagination.md
```diff
@@ -11,275 +11,298 @@
...
 
-### offsetPrevButtonDisabled
+### offsetNextButtonDisabled
 
-Pass in a boolean value for whether or not the offset-based Previous button should be disabled
+Pass in a boolean value for whether or not the offset-based Next button should be disabled.
```
adamdehaven, line 210:
```suggestion
Pass in a boolean value for whether or not the offset-based next button should be disabled.
```

docs/components/pagination.md
```diff
@@ -11,275 +11,298 @@
...
 
-### offsetNextButtonDisabled
+## Events
 
-Pass in a boolean value for whether or not the offset-based Next button should be disabled
+### pageChange
 
-## Usage
+Emitted when the page has been changed. The event provides following information. All of the numbers are sent in a human-understandable base (starting at 1 instead of 0) to be used directly in UI.
```
adamdehaven, line 216:
```suggestion
Emitted when the page has been changed. The page number is returned starting at 1 rather than 0 to be easier to use directly in the UI.
```

docs/guide/migrating-to-version-9.md
```diff
@@ -414,6 +414,31 @@ This component has been removed. Please refer to KBadge component which has been
 
 KPagination now uses [KDropdown](/components/dropdown) instead of [KSelect](/components/select) as the underlying component for page size selection.
 
+#### Props
+
+* `testMode` prop has been removed
+* `paginationType` prop has been removed. You can use new `offset` prop instead
```
adamdehaven, line 420:
```suggestion
* `paginationType` prop has been removed. You can use the new `offset` prop instead
```

docs/guide/migrating-to-version-9.md
```diff
@@ -414,6 +414,31 @@ This component has been removed. Please refer to KBadge component which has been
...
+#### Events
+
+* `pageChanged` event has been renamed to `pageChange`
+* `pageSizeChanged` event has been renamed to `pageSizeChange`
+* `getPrevOffset` event has been renamed to `getPreviousOffset`
+
+#### Structure
+
+* `card-pagination-bar` has been removed and replaced with `k-pagination`
+* `k-pagination-container` has been replaced with `k-pagination`
```
adamdehaven, line 432: So both of these were the same element?

src/components/KPagination/KPagination.vue
```diff
@@ -1,150 +1,138 @@
...
+        <li :class="{ 'arrow-spacing': !disablePageJump }">
+          <KButton
+            appearance="tertiary"
+            aria-label="Go to the previous page"
+            class="pagination-button arrow"
+            data-testid="previous-button"
+            :disabled="backDisabled"
+            size="large"
+            type="button"
+            @click.prevent="pageBack"
```
adamdehaven, line 24: question: why do you need prevent here?

src/components/KPagination/KPagination.vue
```diff
@@ -1,150 +1,138 @@
...
-              href="#"
-              @click.prevent="changePage(1)"
-            >1</a>
-          </li>
-          <li
-            v-if="!disablePageJump && firstDetached"
-            class="pagination-button placeholder"
+            data-testid="page-1-button"
+            type="button"
+            @click.prevent="changePage(1)"
```
adamdehaven, line 37: question: why do you need prevent? Same question for all of the button elements

src/components/KPagination/KPagination.vue
```diff
@@ -1,150 +1,138 @@
...
+      <KDropdown
+        class="page-size-dropdown"
+        data-testid="page-size-dropdown"
+        :items="pageSizeOptions"
+        :kpop-attributes="kpopAttrs"
+        selection-menu
+        show-caret
+        @change="updatePageSize"
+      >
+        <KButton
```
adamdehaven, line 116: Does this button need a `type`?

src/components/KPagination/PaginationOffset.vue
```diff
@@ -1,118 +1,57 @@
...
-      data-testid="prev-btn"
-      :disabled="prevButtonDisabled"
-      @click.prevent="getPrevOffset"
+      data-testid="previous-button"
+      :disabled="previousButtonDisabled"
+      type="button"
+      @click.prevent="emit('getPreviousOffset')"
     >
       <template #icon>
-        <KIcon
```
adamdehaven, line 13: Aren't we deprecating the icon slot in favor of using KButton's default slot?

Same question for all instances

docs/components/pagination.md
portikM: Updated

docs/components/pagination.md
portikM: Are you suggesting we don't display page size selection dropdown if there's only 1 option to choose from? This makes sense to me but I feel like we still need to indicate users ho many items per page are displayed...

portikM: I suggest we display disabled dropdown and hide the caret icon, like so:

Screenshot 2024-02-29 at 11 30 49 AM

docs/guide/migrating-to-version-9.md
portikM: Not quite, `card-pagination-bar` was a wrapper div inside `k-pagination-container` but I found it to be unnecessary

src/components/KPagination/KPagination.vue
portikM: I guess we don't, removed

src/components/KPagination/KPagination.vue
portikM: Yes, in case it's rendered inside of a `form` element so it doesn't trigger submit action (learnings from the incident we had related to `KSegmentedControl`)

src/components/KPagination/PaginationOffset.vue
portikM: We are in beta. Right now to make the button square when only using icon you need to use `icon` slot. After we've deprecated the KIcon, the `icon` prop in KButton will handle that (it will become boolean vs. now it's a string that's passed down to underlying KIcon).

docs/components/pagination.md
adamdehaven: I think disabling is fine but curious if it makes sense as an example on the docs site

src/components/KPagination/KPagination.vue
adamdehaven: I think it was missing the type attribute which is why I commented?

docs/components/pagination.md
portikM: I think it does - those examples are there to serve as example on page navigation, page size has it's own example

src/components/KPagination/PaginationOffset.vue
portikM: We can't - it won't be square.

src/components/KPagination/KPagination.vue
portikM: Oh, right, fixed now, thanks

src/components/KPagination/PaginationOffset.vue
adamdehaven: ...we are everywhere else 😬

portikM: Slot is what we instruct to use when using only icon in KButton to make it square: https://alpha--kongponents.netlify.app/components/button.html#icon-1

Not using the slot results in KButton having more padding on left and right (compare the screenshots).

Screenshot 2024-02-29 at 12 51 45 PM
Screenshot 2024-02-29 at 12 51 10 PM

adamdehaven (approved):

PR #2030, opened 2024-02-22
feat(kviewswitcher): remove component [KHCP-9010]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-9010

Remove `KViewSwitcher` component

## PR Checklist

* [ ] **Conventional Commits** all commits follow the conventional commit standards [outlined in the main README](https://github.com/Kong/kongponents#committing-changes).
* [ ] **Tests coverage:** test coverage was added for new features and bug fixes
* [ ] **Docs:** includes a technically accurate README

PR #2043, opened 2024-02-28
feat(kalert, kemptystate, kpagination, kviewswitcher): component reskin

portikM:
# Summary

`KAlert`, `KEmptyState`, `KPagination` component reskin and `KViewSwitcher` removal

- https://github.com/Kong/kongponents/pull/2021
- https://github.com/Kong/kongponents/pull/2031
- https://github.com/Kong/kongponents/pull/2041
- https://github.com/Kong/kongponents/pull/2030

## PR Checklist


src/components/KEmptyState/KEmptyState.vue
```diff
@@ -1,143 +1,167 @@
...
+        <slot name="default">
+          <p>
+            {{ message }}
+          </p>
         </slot>
       </div>
     </div>
-  </section>
+    <div
+      v-if="actionButtonVisible && (actionButtonText || $slots.action)"
```
kleinfreund, line 34: **Issue**: This is a regression preventing the use of the action slot while `props.actionButtonVisible` is `false`. If `action` content is slotted in, no further condition should prevent the slot from rendering. If `action` content is not slotted in and both `props.actionButtonVisible` and `props.actionButtonText` are set, then the slot should render, too.

```suggestion
      v-if="$slots.action || (actionButtonVisible && actionButtonText)"
```

Seeing that `props.actionButtonVisible` defaults to true and that `props.actionButtonText` defaults to `''` (i.e. it’s required for users to provide the text if they want to rely on the action slot’s fallback content), the visibility prop is superfluous and can be removed, simplifying the check to:

```suggestion
      v-if="$slots.action || actionButtonText"
```

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -1,143 +1,167 @@
...
+            :size="KUI_ICON_SIZE_60"
+          />
+        </slot>
       </div>
+      <span
+        v-if="title"
+        class="empty-state-title"
+      >
+        {{ title }}
+      </span>
```
johncowen, line 21: Oh soz, I missed this in the previous PR

Q: How much do we expect people might what to add a `h*` in here for the title/header of the EmptyState? i.e. should the title be a slot?

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -1,143 +1,167 @@
...
-            appearance="primary"
-            size="small"
-            @click.prevent="() => handleClick && handleClick()"
-          >
-            {{ ctaText }}
-          </KButton>
+        <slot name="default">
+          <p>
+            {{ message }}
+          </p>
```
johncowen, line 29: Kinda similar to the above:

Q: Is anyone ever likely to have a 2 paragraph empty message? Or some other semantic detail that wrapping it with a `p` could be problematic?

edit: I think this might be problematic in the case of KAlert (which also enforces a single `p` here). Specifically our KAlert message in places is a `ul` which is not [Phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content) and is therefore not allowed within a p tag:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p#technical_summary

I think, as a very general point, these types of semantic elements like `p` and `h*`s shouldn't be used inside the majority of reusable components as it makes them less reusable. Just to stress this is a very general point, i.e. we should think twice when adding them. and I guess there will always be exceptions.

johncowen (commented):
👋 hey @portikM sorry I missed these q's in the previous PR, just two other questions incase. lmk either way 👍

src/components/KEmptyState/KEmptyState.vue
johncowen: Oh whoops! Ignore me 😅 I think I read this wrong, we can overwrite the p with the slot.

src/components/KEmptyState/KEmptyState.vue
kleinfreund: We can workaround this by dynamically setting `actionButtonVisible` to `true` if we’re using `action` slot content and, technically, since `actionButtonText` isn’t set initially, we could leave `actionButtonVisible` unset where we sometimes want an action button here, but I think the second suggestion above makes using this a little simpler because it allows removing one prop.

src/components/KEmptyState/KEmptyState.vue
portikM: PR adding the title slot back: https://github.com/Kong/kongponents/pull/2053

PR #2054, opened 2024-03-05
fix(ktable): kemptystate props

portikM:
# Summary

Changes:

`KTable.vue`:

- remove `empty-state-action-button-icon` prop because it was utilizing KIcon - we no longer want to be doing that
- remove `empty-state-icon` - same reason as above
- remove `empty-state-icon-color` - no longer supported, KEmptyState sets the default color for the icon
- remove `empty-state-icon-size` - not supported, icon comes in default size
- add `emptyStateIconVariant` prop
- add `empty-state-action-icon` slot
- all of the above documented in migration guide and reflected in KTable docs

`KEmptyState.vue`:

- add kong icon variant

## PR Checklist

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -843,7 +843,7 @@ Using a `KPop` inside of a clickable row requires some special handling. Non-cli
 
 <template>
   <KCard>
-    <template v-slot:body>
+    <template #default>
       <div v-if="eventType">
         {{eventType}} on: {{row}}
       </div>
@@ -1231,7 +1231,7 @@ the section below or completely slot in your own content.
 - `error-state` - Slot content to be displayed when in an error state
 
 <KCard>
-  <template v-slot:body>
+  <template #default>
     <KTable :fetcher="emptyFetcher" :headers="headers">
       <template v-slot:empty-state>
         <div style="text-align: center;">
@@ -1280,7 +1280,7 @@ If using a CTA button, a `@ktable-empty-state-cta-clicked` event is fired when c
 
 <br/>
 <KCard>
-  <template v-slot:body>
+  <template #default>
     <KTable :fetcher="() => { return { data: [] } }" />
   </template>
 </KCard>
@@ -1288,7 +1288,7 @@ If using a CTA button, a `@ktable-empty-state-cta-clicked` event is fired when c
 ```html
 <template>
   <KCard>
-    <template v-slot:body>
+    <template #default>
       <KTable :fetcher="fetcher" :headers="headers" />
     </template>
   </KCard>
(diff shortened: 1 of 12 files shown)


docs/components/table.md
```diff
@@ -1270,12 +1270,10 @@ Set the following properties to handle empty state:
 
 - `emptyStateTitle` - Title text for empty state
 - `emptyStateMessage` - Message for empty state
-- `emptyStateIcon` - Icon for empty state
-- `emptyStateIconColor` - Color for empty state icon
-- `emptyStateIconSize` - Size for empty state icon
 - `emptyStateActionRoute` - Route for empty state action
 - `emptyStateActionMessage` - Button text for empty state action
-- `emptyStateActionButtonIcon` - Icon for the empty state action button
+- `emptyStateActionButtonShowPlusIcon` - Boolean prop to show/hide plus icon in empty state action button
```
adamdehaven, line 1275: Is this really a prop? 34 character prop name...

adamdehaven: name it something more concise

kaiarrowood: You could try something to indicate the usage instead of what it is like: `emptyStateCreationActionButton` - this is used on create CTAs

portikM: How about `emptyStateActionCreation`?

adamdehaven: `actionButtonPlusIcon: boolean` ?

adamdehaven: What if we just simplified for KTable and didn't show an icon in the empty state? Wouldn't that make things way easier? Surely a user is smart enough to read the button and see that it says `Add {entity}`? 🤷🏼 

Again, this seems extremely use-case specific for Konnect.

portikM: I feel like it's not our decision to make.. maybe slot will be a better alternative? E.g. `action-button-text` that would take precedence over `actionButtonText` prop if provided.

adamdehaven: action button slot - then you have to slot the whole button? 
I don't think adding a prop for a component nested 3 levels deep is ideal

portikM: Okay, scratch that - removed icon-related props and slots - we're just gonna use `empty-state` slot to slot in the whole `KEmptyState` component with all needed buttons and icons

adamdehaven (dismissed):
image
Something isn't working - the empty states on the KTable page are completely missing

portikM:
> image Something isn't working - the empty states on the KTable page are completely missing

Fixed now

kaiarrowood (approved):
LGTM! 🚀

PR #2064, opened 2024-03-13
feat(ktreelist): reskin component [KHCP-9008]

portikM:
# Summary

[Jira](https://konghq.atlassian.net/browse/KHCP-9008)

`KTreeList` component reskin

Links:

- [Kongponents sandbox](https://deploy-preview-2064--kongponents-sandbox.netlify.app/#/treelist)
- [Kongponents docs preview](https://deploy-preview-2064--kongponents.netlify.app/components/tree-list.html)

## PR Checklist

### DIFF
--- a/src/components/KTreeList/KTreeItem.vue
+++ b/src/components/KTreeList/KTreeItem.vue
@@ -7,6 +7,7 @@
     }"
     :data-testid="`tree-item-${item.id}`"
     role="button"
+    tabindex="0"
     @click.prevent="handleClick"
   >
     <div
--- a/src/components/KTreeList/KTreeItem.vue
+++ b/src/components/KTreeList/KTreeItem.vue
@@ -1,31 +1,27 @@
 <template>
   <a
-    class="k-tree-item"
+    class="tree-item"
     :class="{
       'not-draggable': disabled,
       'selected': item.selected
     }"
-    :data-testid="`k-tree-item-${item.id}`"
+    :data-testid="`tree-item-${item.id}`"
     href="#"
     role="button"
     @click.prevent="handleClick"
   >
     <div
       v-if="hasIcon"
-      class="k-tree-item-icon"
-      data-testid="k-tree-item-icon"
+      class="tree-item-icon"
+      data-testid="tree-item-icon"
     >
       <slot name="item-icon">
-        <KIcon
-          :icon="itemIcon"
-          :secondary-color="iconSecondaryColor"
-          :size="KUI_ICON_SIZE_40"
-        />
+        <ServiceDocumentIcon />
       </slot>
     </div>
     <div
-      class="k-tree-item-label"
-      data-testid="k-tree-item-label"
+      class="tree-item-label"
+      data-testid="tree-item-label"
     >
       <slot name="item-label">
         {{ item.name }}
@@ -38,8 +34,7 @@
 import type { PropType } from 'vue'
 import { computed, useSlots } from 'vue'
 import type { TreeListItem } from '@/types'
-import { KUI_ICON_SIZE_40, KUI_COLOR_BORDER_DISABLED } from '@kong/design-tokens'
-import KIcon from '@/components/KIcon/KIcon.vue'
+import { ServiceDocumentIcon } from '@kong/icons'
 
 export const itemsHaveRequiredProps = (items: TreeListItem[]): boolean => {
   return items.every(i => i.name !== undefined && i.id !== undefined && (!i.children?.length || itemsHaveRequiredProps(i.children)))
@@ -57,6 +52,10 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+  hideIcons: {
+    type: Boolean,
+    default: false,
+  },
 })
 
 const emit = defineEmits<{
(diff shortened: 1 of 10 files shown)


src/components/KTreeList/KTreeItem.vue
```diff
@@ -1,45 +1,39 @@
...
+      class="tree-item-icon"
+      data-testid="tree-item-icon"
     >
       <slot name="item-icon">
-        <KIcon
-          :icon="itemIcon"
-          :secondary-color="iconSecondaryColor"
-          :size="KUI_ICON_SIZE_40"
-        />
+        <ServiceDocumentIcon />
```
adamdehaven, line 18: suggestion: can you use `DocumentIcon` or `DocumentListIcon` instead?

src/components/KTreeList/KTreeItem.vue
```diff
@@ -1,45 +1,39 @@
 <template>
-  <a
-    class="k-tree-item"
+  <div
```
adamdehaven, line 2: question: why is this a `div` rather than a `button` or an `a` tag?

---

If you use a `div` we'll need `tabindex="0"` as well

src/components/KTreeList/KTreeList.vue
```diff
@@ -203,24 +205,27 @@ onMounted(() => {
...
   }
 }
 
 // override cursor as grabbing when an item is being dragged
-.k-tree-list-grabbing *,
-.k-tree-item-grabbing * {
+.tree-list-grabbing *,
+.tree-item-grabbing * {
   cursor: move !important; /* fallback: no `url()` support or images disabled */
   cursor: grabbing !important; /* W3C standards syntax, should come least */
```
adamdehaven, line 230: "should come last" - misspelled

adamdehaven (commented):
image

Unsure where to put this, but the selected item should still have a border.

Also, IMO the click area for each items is a bit small now; could we increase the vertical padding a couple pixels?

src/components/KTreeList/KTreeItem.vue
portikM: Adding `tabindex`. Changed it from `a` to `div` because it's definitely not a link and on the docs site it causes issues (page jump).

src/components/KTreeList/KTreeItem.vue
portikM: Tried `DocumentIcon` and I think I like `ServiceDocumentIcon` better. `DocumentListIcon` is for list, not actual items.

Screenshot 2024-03-13 at 4 37 07 PM

src/components/KTreeList/KTreeList.vue
portikM: Ohh last but not least, right, thanks 😅

src/components/KTreeList/KTreeItem.vue
adamdehaven: IIRC we want to get rid of the service one

src/components/KTreeList/KTreeItem.vue
adamdehaven: any reason not to use a button?

portikM:
> image
> 
> Unsure where to put this, but the selected item should still have a border.
> 
> Also, IMO the click area for each items is a bit small now; could we increase the vertical padding a couple pixels?

I tried adding border to the active item and I don't like it...

Screenshot 2024-03-13 at 4 39 58 PM

src/components/KTreeList/KTreeItem.vue
portikM: No particular reason apart from "doesn't feel right" since this is draggable items that aren't really buttons.. do you think it should be a button?

src/components/KTreeList/KTreeItem.vue
portikM: How about `FileEmptyIcon`?

adamdehaven: IMO I like the docs ones better than the empty file - but let's just leave the service one for now 
nbd

src/components/KTreeList/KTreeItem.vue
portikM: @kaiarrowood since you were the one working on this originally, any reason it is not using list tags like `ul` and `li`?

adamdehaven: I think a list structure makes the most sense TBH unless there's a strong reason not to use it.

portikM: Kai went through the comments on the original PR adding this component and couldn't find any arguments against using `ul` here so I'll work on refactoring this tomorrow.

portikM: Hi team! So I pushed my changes implementing the ordered list structure, also recorded a Loom walking through the resulting tree structure and outlining couple small caveats that came with this change:
https://www.loom.com/share/013dcb4129b44626b898782a9d5c5134?sid=e66bd70d-db41-4c2a-bf37-9959e83373f8

portikM: Update: the bug with multiple drag zone blue lines is fixed now.

portikM: Update: we decided to leave the idea utilizing the list element structure because it's hard to keep the HTML structure semantic, especially after user reordering when a deeply nested item can end up being the first-level item and so on.

adamdehaven:
When dragging items on the docs site, I'm getting an error sometimes: 
image

portikM:
> When dragging items on the docs site, I'm getting an error sometimes: image

Actually the example where that error was occurring doesn't make much sense so I removed it.

adamdehaven (approved):
lgtm

PR #2069, opened 2024-03-15
fix(kcard): add title tag prop

portikM:
# Summary

Adds `titleTag` prop in `KCard`

## PR Checklist


docs/components/card.md
```diff
@@ -74,6 +74,16 @@ String to be used as a title. Can also be [slotted](#slots).
 <KCard title="Card Title" />
 ```
 
+### titleTag
+
+HTML element you want title to be rendered as. Defaults to `div`.
+
+<KCard title-tag="h4" title="Card Title H4 Heading" />
```
adamdehaven, line 81: question: Why wouldn't this just be accomplished via the slot?

portikM: Because a slot would require extra 3 lines instead of just 1 and because this gives more flexibility without a burden of being too opinionated

src/components/KCard/KCard.cy.ts
```diff
@@ -13,14 +13,16 @@ describe('KCard', () => {
...
 
     mount(KCard, {
       props: {
         title: titleProp,
+        titleTag,
       },
     })
 
-    cy.get('.k-card').find('.card-title').should('contain', titleProp)
+    cy.get('.k-card').find(`${titleTag}.card-title`).should('contain', titleProp)
```
adamdehaven, line 25: since a prop is determining the HTML tag, I'd also add a test case in here to ensure that the prop.titleTag = the HTML tag element

src/components/KCard/KCard.vue
```diff
@@ -42,6 +43,10 @@ const props = defineProps({
     type: String,
     default: '',
   },
+  titleTag: {
+    type: String,
```
adamdehaven, line 47: question: should you do any validation on the prop value? I'd think you could have a validator that only allows these:
```md
['div', 'span', 'p', 'a', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6',]
```

src/components/KCard/KCard.cy.ts
portikM: It is exactly what this test is doing: I'm setting `titleTag` prop and then checking whether correct element is rendered.. or did I miss something?

src/components/KCard/KCard.vue
portikM: Doen

src/components/KCard/KCard.cy.ts
adamdehaven: Ah I missed the `${}` selector. Yep.

docs/components/card.md
```diff
@@ -74,6 +74,16 @@ String to be used as a title. Can also be [slotted](#slots).
 <KCard title="Card Title" />
 ```
 
+### titleTag
+
+HTML element you want title to be rendered as. Defaults to `div`.
```
adamdehaven, line 79: suggestion: might be worth documenting the allowed tags

adamdehaven (dismissed):

docs/components/card.md
portikM: Done

adamdehaven (approved):
lgtm

PR #2077, opened 2024-03-20
fix(klabel, kselect, kmultiselect): a11y attributes [KHCP-111026]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-11026

Changes:
* add `aria-labelledby` on KLabel's tooltip
* add `aria-live` attribute in KSelect and KMultiselect
* change element to bind attributes to in KMultiselect

## PR Checklist

### DIFF
--- a/src/components/KLabel/KLabel.vue
+++ b/src/components/KLabel/KLabel.vue
@@ -1,6 +1,6 @@
 <template>
   <label
-    :id="labelId"
+    :aria-describedby="tooltipId"
     class="k-label"
     :class="{ 'required': required }"
   >
@@ -8,10 +8,10 @@
 
     <KTooltip
       v-if="hasTooltip"
-      :aria-labelledby="labelId"
       v-bind="tooltipAttributes"
       class="label-tooltip"
       position-fixed
+      :tooltip-id="tooltipId"
     >
       <InfoIcon
         class="tooltip-trigger-icon"
@@ -27,7 +27,7 @@
 
 <script setup lang="ts">
 import type { PropType } from 'vue'
-import { computed, useAttrs, useSlots } from 'vue'
+import { computed, useSlots } from 'vue'
 import KTooltip from '@/components/KTooltip/KTooltip.vue'
 import type { TooltipAttributes } from '@/types'
 import { InfoIcon } from '@kong/icons'
@@ -64,11 +64,10 @@ const props = defineProps({
 })
 
 const slots = useSlots()
-const attrs = useAttrs()
 
 const hasTooltip = computed((): boolean => !!(props.help || props.info || slots.tooltip))
 
-const labelId = attrs.id ? String(attrs.id) : uuidv4()
+const tooltipId = uuidv4()
 </script>
 
 <style lang="scss" scoped>
--- a/src/components/KLabel/KLabel.vue
+++ b/src/components/KLabel/KLabel.vue
@@ -1,11 +1,14 @@
 <template>
   <label
+    :id="labelId"
     class="k-label"
     :class="{ 'required': required }"
   >
     <slot />
+
     <KTooltip
       v-if="hasTooltip"
+      :aria-labelledby="labelId"
       v-bind="tooltipAttributes"
       class="label-tooltip"
       position-fixed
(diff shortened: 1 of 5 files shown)


src/components/KSelect/KSelect.vue
```diff
@@ -101,68 +101,70 @@
...
-          </div>
-          <div
-            v-else
-            class="select-items-container"
-            data-propagate-clicks="false"
-          >
-            <KSelectItems
-              :items="filteredItems"
-              @selected="handleItemSelect"
+          <div :aria-live="enableFiltering ? 'polite' : 'off'">
```
portikM, line 104: Note for PR review: the change here is adding a wrapper div with `aria-live` attribute

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -173,47 +173,49 @@
...
-                </template>
-              </KMultiselectItems>
-              <KMultiselectItem
-                v-if="enableItemCreation && uniqueFilterStr && !$slots.empty"
-                key="multiselect-add-item"
-                class="multiselect-add-item"
-                data-testid="multiselect-add-item"
-                :item="{ label: `${filterString} (Add new value)`, value: 'add_item' }"
-                @selected="handleAddItem"
+              <div aria-live="polite">
```
portikM, line 176: Note for PR review: the change here is adding a wrapper div with aria-live attribute

src/components/KLabel/KLabel.vue
```diff
@@ -1,11 +1,14 @@
...
   <label
+    :id="labelId"
     class="k-label"
     :class="{ 'required': required }"
   >
     <slot />
+
     <KTooltip
       v-if="hasTooltip"
+      :aria-labelledby="labelId"
```
adamdehaven, line 11: question: I'm not sure this makes sense in the context?

You have a label, which essentially provides info about another element, e.g. input.

This is then saying that the KTooltip is "labeled by the label"

portikM: Yes, it's providing accessible association between the tooltip and label. Otherwise it's impossible for screen readers to know the tooltip is somehow related to the label.

adamdehaven: What I mean is I think the relationship here is backwards - the tooltip provides extra info about the label, not the other way around

portikM: Yes, I get what you are saying but `aria-labelledby` needs to be set on an interactive elements - which in this case is the tooltip. Doing the other way around (setting it on label element) would be incorrect.

adamdehaven: But isn't saying "the tooltip is described by the label" incorrect? 

The tooltip is offering additional information within the label itself, about the label content

portikM: Labelled by, not described by - those are 2 different attributes. And that actually might be a good solution here - to set `aria-describedby` on label element to point at tooltip content.. only thing I'm not sure about is whether the two are interchangeable (if we did add `aria-describedby` on label would we still need to keep `aria-labelledby` on the tooltip). Let me do some reading and get back to you.

adamdehaven: You should be able to drop labelledby; aria-attributes are additive, meaning since the Tooltip is contained within the label, it's inferred

portikM: Okay, updated now

adamdehaven (approved):

src/components/KTooltip/KTooltip.vue
```diff
@@ -79,6 +83,10 @@ const props = defineProps({
     type: String,
     default: '',
   },
+  tooltipId: {
+    type: String,
+    default: uuidv4(),
```
adamdehaven, line 88: @portikM you can’t directly call this function inside the default prop

portikM: It's working fine but I guess I can make it a function ruturning uuid, fixing here: https://github.com/Kong/kongponents/pull/2081

portikM: Yup already saw that. Making it a computed property in this PR

PR #2081, opened 2024-03-21
fix(kmodal): toggle focus trap off on unmounted

portikM:
# Summary

## PR Checklist

* [ ] **Conventional Commits** all commits follow the conventional commit standards [outlined in the main README](https://github.com/Kong/kongponents#committing-changes).
* [ ] **Tests coverage:** test coverage was added for new features and bug fixes
* [ ] **Docs:** includes a technically accurate README

PR #2085, opened 2024-03-22
fix(ktooltip): content id

Leopoldthecoder:
# Summary

The way the default value of `tooltipId` is generated has caused unit tests in host app to fail.

image

## PR Checklist

### DIFF
--- a/src/components/KTooltip/KTooltip.vue
+++ b/src/components/KTooltip/KTooltip.vue
@@ -15,7 +15,7 @@
 
     <template #content>
       <div
-        :id="tooltipId"
+        :id="tooltipId || uuidv4()"
         role="tooltip"
       >
         <slot
@@ -85,7 +85,7 @@ const props = defineProps({
   },
   tooltipId: {
     type: String,
-    default: uuidv4(),
+    default: '',
   },
 })
 


nekolab (approved):
LGTM

src/components/KTooltip/KTooltip.vue
```diff
@@ -15,7 +15,7 @@
 
     <template #content>
       <div
-        :id="tooltipId"
+        :id="tooltipId || uuidv4()"
```
adamdehaven, line 18: suggestion: can you move this to a computed property?

portikM (changes requested):
Sorry for trouble, fixing the issue here: https://github.com/Kong/kongponents/pull/2081 (includes this fix and something else for full screen modal, previously approved by Kai)

src/components/KTooltip/KTooltip.vue
portikM: No need making it computed since it's static - a constant will do. Fixing here: https://github.com/Kong/kongponents/pull/2081

adamdehaven: I was saying this but yeah static how you have it is fine 
```
const var = computed((): string => props.tooltipId || randomTooltipId)
```

Leopoldthecoder:
Thank you both!

Issue #1868, opened 2023-11-23
KCodeBlock: filter/regexp mode buttons not showing pressed state in dark theme, disabled buttons not looking disabled


kleinfreund:
See https://alpha--kongponents.netlify.app/components/codeblock.html#theme

**Dark theme**: Neither mode button is pressed & the previous/next match buttons are disabled:

![image](https://github.com/Kong/kongponents/assets/5774638/8263fcba-4ecc-4f35-b93d-d668e2bd9b00)

**Dark theme**: Both mode buttons are pressed & the previous/next match buttons are disabled:

![image](https://github.com/Kong/kongponents/assets/5774638/2a56a087-63fc-49e1-ba77-624231919f3b)

For comparison:

**Light theme**: Neither mode button is pressed & the previous/next match buttons are disabled:

![image](https://github.com/Kong/kongponents/assets/5774638/7095f350-e479-451d-954e-67c4093b0f2c)

kleinfreund:
Closing this as it was addressed in https://github.com/Kong/kongponents/pull/2100

PR #2100, opened 2024-03-27
feat(kcodeblock): component reskin [KHCP-8975]

portikM:
# Summary

[Jira](https://konghq.atlassian.net/browse/KHCP-8975)
[Figma](https://www.figma.com/file/Yze0SWXl5nKjR0rFdilljK/Components?type=design&node-id=2020%3A11253&mode=dev&t=wfYkBJZ5Ry0OUS2m-1)

`KCodeBlock` component reskin

Links:

- [Kongponents sandbox](https://deploy-preview-2100--kongponents-sandbox.netlify.app/#/codeblock)
- [Kongponents docs preview](https://deploy-preview-2100--kongponents.netlify.app/components/codeblock.html)

## PR Checklist


src/components/KCodeBlock/CodeBlockIconButton.vue
```diff
@@ -0,0 +1,129 @@
...
+})
+</script>
+
+<style lang="scss" scoped>
+.code-block-action-tooltip {
+  // display: inline-block;
+}
+
+// expand upon original tertiary button styles changing some colors and states
+.code-block-action-button {
```
kleinfreund, line 73: **Issue**: As far as I can tell, the styles for this button now only apply to exactly this button. However, the code block has a `secondary-actions` slot which we use to (sometimes) slot in extra buttons. The styles here should also apply to those buttons. Otherwise, you end up needing to customize those styles in the consuming application to avoid this:

image

kleinfreund (commented):
I had a look at this to get a head start to creating the adoption PR on our end (https://github.com/kumahq/kuma-gui/pull/2374) and noticed two things:

- The default styles when the editor only displays a single line of code (e.g. via a query combined with filter mode) don’t allow the full content to be displayed vertically and trigger a scrollbar. That should be avoided, I think.

  Screenshot from 2024-04-03 10-06-07

  (On a somewhat unrelated note, I wonder if we really need `props.singleLine`. It would be nicer if the component’s default styles without `props.singleLine` displays properly in the related use cases)
- The search input is practically invisible (see screenshot in the previous point) in both themes. I think there are two issues here:
  - The search input being left-aligned visually removes it from the related actions that are still right-aligned. That just seems wrong, especially on larger screens. **If** that is really how we want this to look, the input should at least take up the full width of the space. Currently, it only takes up about 500 pixels at the maximum.
  - The search input has no border or distinct background color making it nearly impossible to know where you can actually interact with it (outside of guessing that, probably, the placeholder text is a part of the input). This, too, is worse on larger screens because there’s just that much more seemingly empty area to consider.

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -5,207 +5,150 @@
...
+              data-testid="code-block-processing-icon"
+              :size="KUI_ICON_SIZE_30"
+            />
+          </Transition>
 
-      <div class="k-search-actions">
-        <!-- temporary workaround to get state to "stick" -->
-        <KButton
-          :appearance="isRegExpMode ? 'secondary' : 'tertiary'"
-          :aria-pressed="isRegExpMode"
```
kleinfreund, line 104: **Issue**: This being removed is a regression. The buttons that currently have the `aria-pressed` state are toggle buttons and _need_ this to properly expose their pressed state to the accessibility tree.

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -5,207 +5,150 @@
...
         >
-          <span class="visually-hidden">RegExp mode enabled</span>
-
-          .*
-        </KButton>
-
-        <!-- temporary workaround to get state to "stick" -->
-        <KButton
-          :appearance="isFilterMode ? 'secondary' : 'tertiary'"
-          :aria-pressed="isFilterMode"
```
kleinfreund, line 120: **Issue**: This being removed is a regression. The buttons that currently have the `aria-pressed` state are toggle buttons and _need_ this to properly expose their pressed state to the accessibility tree.

portikM:
> I had a look at this to get a head start to creating the adoption PR on our end ([kumahq/kuma-gui#2374](https://github.com/kumahq/kuma-gui/pull/2374)) and noticed two things:
> 
> * The default styles when the editor only displays a single line of code (e.g. via a query combined with filter mode) don’t allow the full content to be displayed vertically and trigger a scrollbar. That should be avoided, I think.
>   Screenshot from 2024-04-03 10-06-07
>   (On a somewhat unrelated note, I wonder if we really need `props.singleLine`. It would be nicer if the component’s default styles without `props.singleLine` displays properly in the related use cases)
> * The search input is practically invisible (see screenshot in the previous point) in both themes. I think there are two issues here:
>   
>   * The search input being left-aligned visually removes it from the related actions that are still right-aligned. That just seems wrong, especially on larger screens. **If** that is really how we want this to look, the input should at least take up the full width of the space. Currently, it only takes up about 500 pixels at the maximum.
>   * The search input has no border or distinct background color making it nearly impossible to know where you can actually interact with it (outside of guessing that, probably, the placeholder text is a part of the input). This, too, is worse on larger screens because there’s just that much more seemingly empty area to consider.

Thanks for prompt review @kleinfreund! Let me address your concerns in order:

1. Scroll issue when only single line shows is fixed (screenshot). As to whether we need `singleLine` prop - I think we do as apart from displaying a single line of code properly style-wise it also takes care of not showing the line numbers and bunch of other things ([prop docs](https://deploy-preview-2100--kongponents.netlify.app/components/codeblock.html#singleline)).
2. I have removed the limit of `max-width: 500px;` on the input - that should make seemingly empty area more "interactive". As to the fact that the input field is nearly invisible - I'll pass on the feedback to design and let them address this bit.

Screenshot 2024-04-03 at 9 09 22 AM

src/components/KCodeBlock/CodeBlockIconButton.vue
kleinfreund: For reference, how we use the `secondary-actions` slot:

https://github.com/kumahq/kuma-gui/blob/master/src/app/common/code-block/ResourceCodeBlock.vue#L19-L40

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -5,207 +5,152 @@
...
-          data-testid="k-code-block-filter-mode-button"
-          icon="filter"
-          size="small"
+          <RegexIcon />
+        </CodeBlockIconButton>
+
+        <CodeBlockIconButton
+          :active="isFilterMode"
+          :aria-label="`Filter results (${ALT_SHORTCUT_LABEL}+F)`"
+          :aria-pressed="isRegExpMode"
```
kleinfreund, line 99: **Issue**: Small issue from bringing this back:

```suggestion
          :aria-pressed="isFilterMode"
```

portikM: Oops, good catch - fixed.

src/components/KCodeBlock/CodeBlockIconButton.vue
portikM: Added export and docs for `KCodeBlockIconButton` component: https://deploy-preview-2100--kongponents.netlify.app/components/codeblock.html#kcodeblockiconbutton

kleinfreund:
@portikM Also, you can make this PR resolve #1868, I think.

src/components/KCodeBlock/CodeBlockIconButton.vue
kleinfreund: Nice, thanks, I’ll have a look now.

kleinfreund: Yep, I think I can adapt our implementation to use this helper component:
image

portikM: Awesome! Notice that you don't need to add spacing to that component (margin) - the KCodeBlock slot styles take care of that for you.

src/components/KCodeBlock/KCodeBlockIconButton.vue
```diff
@@ -0,0 +1,129 @@
...
+      appearance="tertiary"
+      class="code-block-action-button"
+      :class="[`theme-${theme}`, { 'code-block-action-button-active': active }]"
+      v-bind="attrs"
+      @click="onClick"
+    >
+      <!-- TODO: [beta] update once icon slot is deprecated -->
+      <template #icon>
+        <slot />
+      </template>
```
kleinfreund, line 18: **Question**: Shouldn’t this just re-expose the button’s default slot? 

```suggestion
      <slot />
```

Seems simpler and I don’t think anyone would need it to be the icon slot specifically currently.

portikM: We want it to be the icon slot so that the button is square (padding top and bottom = padding left and right). Otherwise it will look like this.

Screenshot 2024-04-04 at 9 03 37 AM

adamdehaven: Isn't that going to make us have to do another followup PR to change it in a few weeks..?

portikM: Yes but I'm planning to make it backwards compatible

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -780,491 +710,217 @@ async function copyCode(event: Event): Promise<void> {
...
+      pre {
+        .line-number-rows {
+          .line {
+            .line-anchor {
+              color: var(--kui-color-text-neutral-weak, $kui-color-text-neutral-weak);
+            }
+
+            &.line-is-match {
+              &::before {
+                background-color: $tmp-color-white-opaque-12; // we don't have a token for this now and we might not ever need one so using a temp variable
```
portikM, line 896: @adamdehaven fyi I decided to use tmp variable for this value

adamdehaven: sgtm

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -5,207 +5,152 @@
...
+              aria-label="Clear query"
+              class="clear-query-button"
+              data-testid="clear-query-button"
+              role="button"
+              tabindex="0"
+              title="Clear query"
+              @click="clearQuery"
+              @keydown.enter="clearQuery"
+              @keydown.space="clearQuery"
+            />
```
kleinfreund, line 39: **Issue**: This must use an actual button. There are only disadvantages to "re-implementing" buttons because one has to add a bunch of functionality that are already standard features of `button`: things like appropriate keyboard interactions (here added via two `keydown` listeners), accessibility properties (some added here via `role` and `tabindex`), etc. all come with `button` by default.

```suggestion
            <button
              v-if="searchQuery"
              aria-label="Clear query"
              class="clear-query-button"
              data-testid="clear-query-button"
              type="button"
              @click="clearQuery"
            >
              <CloseIcon title="Clear query" />
            </button>
```

(This will need adjustments in the selectors attached to `.clear-query-button` I believe)

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -780,491 +710,217 @@ async function copyCode(event: Event): Promise<void> {
...
+            color: var(--kui-color-text-neutral-weaker, $kui-color-text-neutral-weaker);
+          }
+        }
+
+        .clear-query-button,
+        .code-block-search-icon {
+          color: var(--kui-color-text-neutral-weaker, $kui-color-text-neutral-weaker) !important;
+        }
+
+        .clear-query-button[role="button"] {
```
kleinfreund, line 874: (Needed for the other suggestion I made related to this element)

```suggestion
        .clear-query-button {
```

src/components/KCodeBlock/KCodeBlock.vue
portikM: Good idea, applied

sandbox/pages/SandboxCodeBlock.vue
```diff
@@ -0,0 +1,219 @@
...
+      >
+        <KCodeBlock
+          id="secondary-actions-slot"
+          :code="code"
+          language="json"
+        >
+          <template #secondary-actions>
+            <KButton>
+              Secondary Action
+            </KButton>
```
kleinfreund, line 156: **Suggestion**: Should this use a KCodeBlockIconButton? That would help verify that using it has the right styles just by looking at the sandbox.

PR #2105, opened 2024-03-28
feat(ktable): add column resize support [khcp-11161]

kaiarrowood:
# Summary

Add column resize support to `KTable` for [KHCP-11161](https://konghq.atlassian.net/browse/KHCP-11161).

image

## PR Checklist

[KHCP-11161]: https://konghq.atlassian.net/browse/KHCP-11161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KTable/KTable.vue
+++ b/src/components/KTable/KTable.vue
@@ -696,6 +696,7 @@ const headerHeight = computed((): string => {
 
   return 'auto'
 })
+
 const startResize = (evt: MouseEvent, colKey: string) => {
   let x = 0
   let width = 0
@@ -1062,8 +1063,6 @@ export const defaultSorter = (key: string, previousKey: string, sortOrder: strin
 </script>
 
 <style lang="scss" scoped>
-@import '@/styles/tmp-variables';
-@import '@/styles/mixins';
 
 .k-table-wrapper {
   overflow: auto;
@@ -1086,7 +1085,7 @@ export const defaultSorter = (key: string, previousKey: string, sortOrder: strin
 
   th,
   td {
-    border-right: var(--kui-border-width-20, $kui-border-width-20) solid $kui-color-border-transparent;
+    border-right: var(--kui-border-width-20, $kui-border-width-20) solid var(--kui-color-border-transparent, $kui-color-border-transparent);
     padding: var(--kui-space-50, $kui-space-50) var(--kui-space-60, $kui-space-60);
     vertical-align: middle;
     white-space: nowrap;
--- a/src/components/KTable/KTable.vue
+++ b/src/components/KTable/KTable.vue
@@ -84,18 +84,25 @@
         :data-tableid="tableId"
       >
         <thead :class="{ 'is-scrolled': isScrolled }">
-          <tr :class="{ 'is-scrolled': isScrolled }">
+          <tr
+            ref="headerRow"
+            :class="{ 'is-scrolled': isScrolled }"
+          >
             <th
               v-for="(column, index) in tableHeaders"
               :key="`k-table-${tableId}-headers-${index}`"
               :aria-sort="!disableSorting && column.key === sortColumnKey ? (sortColumnOrder === 'asc' ? 'ascending' : 'descending') : undefined"
               class="k-table-headers"
               :class="{
+                'resize-hover': currentFocusedColumn === column.key && resizeColumns,
+                'truncated-column resizable': resizeColumns,
                 'sortable': !disableSorting && !column.hideLabel && column.sortable,
                 'active-sort': !disableSorting && !column.hideLabel && column.sortable && column.key === sortColumnKey,
                 [sortColumnOrder]: !disableSorting && column.key === sortColumnKey && !column.hideLabel,
                 'is-scrolled': isScrolled
               }"
+              :data-testid="`k-table-header-${column.key}`"
+              :style="columnStyles[column.key]"
               @click="() => {
                 if (!disableSorting && column.sortable) {
                   $emit('sort', {
@@ -106,13 +113,20 @@
                   sortClickHandler(column)
                 }
               }"
+              @mouseleave="currentFocusedColumn = ''"
+              @mouseover="currentFocusedColumn = column.key"
             >
-              <span class="k-table-headers-container">
+              <div class="k-table-headers-container">
                 <slot
                   :column="getGeneric(column)"
                   :name="getColumnSlotName(column.key)"
                 >
-                  <span :class="{'sr-only': column.hideLabel}">
+                  <span
+                    :class="{
+                      'sr-only': column.hideLabel,
+                      'truncated-column': resizeColumns,
+                    }"
+                  >
                     {{ column.label ? column.label : column.key }}
                   </span>
                 </slot>
@@ -125,7 +139,13 @@
                   icon="chevronDown"
                   :size="KUI_ICON_SIZE_20"
                 />
-              </span>
+
+                <div
+                  v-if="resizeColumns"
+                  class="resize-handle"
+                  @mousedown="(evt) => startResize(evt, column.key)"
+                />
+              </div>
             </th>
           </tr>
         </thead>
@@ -143,6 +163,11 @@
               v-for="(value, index) in tableHeaders"
               v-bind="cellAttrs({ headerKey: value.key, row, rowIndex, colIndex: index })"
               :key="`k-table-${tableId}-cell-${index}`"
+              :class="{
+                'resize-hover': currentFocusedColumn === value.key && resizeColumns,
+                'truncated-column': resizeColumns
+              }"
+              :style="columnStyles[value.key]"
               v-on="tdlisteners(row[value.key], row)"
             >
               <slot
@@ -230,6 +255,13 @@ const props = defineProps({
     type: Object,
     default: () => null,
     required: false,
+  },
+  /**
+   * Allow columns to be resized
+   */
+  resizeColumns: {
+    type: Boolean,
+    default: false,
   },
   /**
    * Enable client side sort - only do this if using a fetcher
@@ -509,7 +541,9 @@ const defaultFetcherProps = {
   offset: null,
 }
 const data = ref<Record<string, any>[]>([])
+const headerRow = ref<HTMLDivElement>()
 const tableHeaders: Ref<TableHeader[]> = ref([])
+const currentFocusedColumn = ref('')
 const total = ref(0)
 const isScrolled = ref(false)
 const page = ref(1)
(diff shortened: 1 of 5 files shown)


portikM:
Noticed an issue: when I start dragging and move my cursor (while dragging) away from the header, the border becomes invisible - is there any way we can have it visible while dragging even when not hovering the header?

https://github.com/Kong/kongponents/assets/36751160/028b66ba-20c6-4bd0-8c7b-28ee54cc99f1

src/components/KTable/KTable.vue
```diff
@@ -977,11 +1086,24 @@ export const defaultSorter = (key: string, previousKey: string, sortOrder: strin
...
     vertical-align: middle;
     white-space: nowrap;
   }
 
+  th.resize-hover {
+    border-right: var(--kui-border-width-20, $kui-border-width-20) solid var(--kui-color-border-decorative-purple, $kui-color-border-decorative-purple);
+  }
+
+  td.resize-hover {
+    border-right: var(--kui-border-width-20, $kui-border-width-20) solid rgba(175, 183, 197, .2); /** $kui-color-border as rgb */
```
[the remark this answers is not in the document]
kaiarrowood: I tried the `neutral` shades but `weakest` was either too subtle or too strong. I'm happy to use whatever token you think looks best. I just used opacity because it was quickest to get the shade I was trying to have.

src/components/KTable/KTable.vue
```diff
@@ -632,6 +680,114 @@ const tdlisteners = computed((): any => {
...
+  let x = 0
+  let width = 0
+
+  resizingColumn.value = colKey
+
+  // get the current column's element
+  let col: HTMLElement | null = null
+  headerElems.value?.forEach((elem) => {
+    if (elem.getAttribute('data-testid') === `k-table-header-${colKey}`) {
+      col = document.querySelector(`[data-testid="k-table-header-${colKey}"]`)
```
johncowen, line 755: I only had a brief look at this PR so I might be missing something here, but what would happen here if you had two tables in the same page? Is `colKey` unique to each KTable instance?

src/components/KTable/KTable.vue
```diff
@@ -125,7 +148,16 @@
                   icon="chevronDown"
                   :size="KUI_ICON_SIZE_20"
                 />
-              </span>
+
+                <div
+                  v-if="resizeColumns && index !== tableHeaders.length - 1"
+                  class="resize-handle"
+                  @click.stop
+                  @mousedown="(evt) => startResize(evt, column.key)"
```
adamdehaven, line 156: (just providing info)

for these native events, passing in with an arrow function works, but you can also use the special `$event` variable if it makes things simpler:


```ts
@mousedown="startResize($event, column.key)"
```

[Docs](https://vuejs.org/guide/essentials/event-handling.html#accessing-event-argument-in-inline-handlers)

src/components/KTable/KTable.vue
```diff
@@ -632,6 +680,114 @@ const tdlisteners = computed((): any => {
   }
 })
 
+const columnWidths = ref<Record<string, number>>({})
```
adamdehaven, line 683: question: is there a reason you initialize with an empty object instead of just undefined `ref()` ?

src/components/KTable/KTable.vue
```diff
@@ -632,6 +680,114 @@ const tdlisteners = computed((): any => {
...
+      return resizerHoveredColumn.value
+    case 'th-hover':
+      return currentHoveredColumn.value
+    default:
+      return ''
+  }
+})
+
+// get the resizable header divs to be used for the resize observers
+const headerElems = computed(() => headerRow.value?.querySelectorAll('th.resizable'))
```
adamdehaven, line 734: suggestion: add a return type on the computed

src/components/KTable/KTable.vue
adamdehaven: Same comment ^

src/components/KTable/KTable.vue
```diff
@@ -632,6 +680,114 @@ const tdlisteners = computed((): any => {
...
+    // Update column width to match
+    col?.setAttribute('style', `width: ${width + dx}px`)
+    columnWidths.value[colKey] = width + dx
+  }
+
+  // done resizing
+  const mouseUpHandler = (): void => {
+    resizingColumn.value = ''
+    document.removeEventListener('mousemove', mouseMoveHandler)
+    document.removeEventListener('mouseup', mouseUpHandler)
```
adamdehaven, line 773: nit: optionally chain here just to make sure the document is available
```suggestion
    document?.removeEventListener('mousemove', mouseMoveHandler)
    document?.removeEventListener('mouseup', mouseUpHandler)
```

src/components/KTable/KTable.vue
```diff
@@ -632,6 +680,114 @@ const tdlisteners = computed((): any => {
...
+  x = evt.clientX
+
+  if (col) {
+    // set current column's width
+    const styles = window.getComputedStyle(col)
+    width = parseInt(styles.width, 10)
+
+    // event listeners for resizing
+    document.addEventListener('mousemove', mouseMoveHandler)
+    document.addEventListener('mouseup', mouseUpHandler)
```
adamdehaven, line 787:
```suggestion
    document?.addEventListener('mousemove', mouseMoveHandler)
    document?.addEventListener('mouseup', mouseUpHandler)
```

src/components/KTable/KTable.vue
```diff
@@ -632,6 +680,114 @@ const tdlisteners = computed((): any => {
...
+    emitTablePreferences()
+  }
+
+  // get mouse position
+  x = evt.clientX
+
+  if (col) {
+    // set current column's width
+    const styles = window.getComputedStyle(col)
+    width = parseInt(styles.width, 10)
```
adamdehaven, line 783: question: what if `styles.width` is undefined?

src/components/KTable/KTable.vue
```diff
@@ -632,6 +680,114 @@ const tdlisteners = computed((): any => {
...
+})
+
+// get the resizable header divs to be used for the resize observers
+const headerElems = computed(() => headerRow.value?.querySelectorAll('th.resizable'))
+const headerHeight = computed((): string => {
+  const elem = headerElems.value?.item(0)
+  if (elem) {
+    const styles = window.getComputedStyle(elem)
+
+    return `${parseInt(styles.height, 10)}px`
```
adamdehaven, line 740: question: what if `styles.height` is undefined?

src/components/KTable/KTable.vue
```diff
@@ -955,6 +1063,7 @@ export const defaultSorter = (key: string, previousKey: string, sortOrder: strin
 
 <style lang="scss" scoped>
 @import '@/styles/tmp-variables';
+@import '@/styles/mixins';
```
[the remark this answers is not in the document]
adamdehaven: I had the same thought - why are these imports needed?

src/components/KTable/KTable.vue
```diff
@@ -982,6 +1140,20 @@ export const defaultSorter = (key: string, previousKey: string, sortOrder: strin
     white-space: nowrap;
   }
 
+  th.resize-hover {
+    // creates a 2px "border" on the right - can't use the border because it will "jump"
+    box-shadow: calc(-1 * var(--kui-border-width-20, $kui-border-width-20)) 0 0 0 var(--kui-color-border-decorative-purple, $kui-color-border-decorative-purple) inset;
```
adamdehaven, line 1145: question: I'm guessing this was already discussed, but just to be sure, how does the user know the table columns are resizable without first hovering on the table? There is no icon/indicator on the table headers that hint that the columns are resizable.

src/components/KTable/KTable.vue
```diff
@@ -1028,6 +1200,25 @@ export const defaultSorter = (key: string, previousKey: string, sortOrder: strin
       padding: var(--kui-space-50, $kui-space-50) var(--kui-space-60, $kui-space-60);
       text-align: left;
 
+      &.resizable {
+        min-width: 20px;
```
adamdehaven, line 1204: suggestion: I think the minimum column width should be at least large enough to see the first character of the column name (maybe `40px`?); otherwise, you can see table cell data but not the header and it's a bit odd

image

src/components/KTable/KTable.vue
```diff
@@ -1028,6 +1200,25 @@ export const defaultSorter = (key: string, previousKey: string, sortOrder: strin
       padding: var(--kui-space-50, $kui-space-50) var(--kui-space-60, $kui-space-60);
       text-align: left;
 
+      &.resizable {
```
adamdehaven, line 1203: suggestion: Is it possible to limit the maximum width of the column headers as well? 

A user should never need to expand the column to a width greater than the column cell with the largest amount of content.

I believe there's logic in the KTruncate component that might be able to be repurposed here in order to determine when all of the cell content is visible, and at that point, restrict increasing the table width beyond that value. Otherwise, the user can infinitely expand the column width.

src/components/KTable/KTable.vue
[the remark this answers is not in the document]
kaiarrowood: I totally thought I accepted this suggestion already 🤔

src/components/KTable/KTable.vue
kaiarrowood: Good catch, I will roll the `tableId` into the query

src/components/KTable/KTable.vue
kaiarrowood: It makes it easier to not have to check if the object is defined or not down below when I start adding entries for the column widths.

src/components/KTable/KTable.vue
```diff
@@ -509,7 +553,11 @@ const defaultFetcherProps = {
   offset: null,
 }
 const data = ref<Record<string, any>[]>([])
+const headerRow = ref<HTMLDivElement>()
```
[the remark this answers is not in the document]
kaiarrowood: I would have to define it as `ref<HTMLDivElement | null>` then.

image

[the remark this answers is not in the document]
adamdehaven: Ok?
https://vuejs.org/guide/essentials/template-refs.html#accessing-the-refs

src/components/KTable/KTable.vue
kaiarrowood: As discussed in stand up, we will use current implementation with the understanding that design team will revisit during `KTable` reskin.

src/components/KTable/KTable.vue
kaiarrowood: I'm not sure I understand what you're saying here. If we don't allow columns to expand beyond the largest cell's width doesn't that mean we can't guarantee the table width? (ie. I want the table to span the entire page's width and I'd like the columns to have relatively even spacing. This means sometimes I will want columns with minimal content to take up more space than they need.

To a degree this also comes down to the user's preference. As long as the table is still functional (you can continue to use the resize/other controls) shouldn't we allow them to expand the table as wide as they want?

adamdehaven: I'm not sure I understand why a user would need to be able to expand a table cell extremely beyond the width of the cell's content?

kaiarrowood: Not extremely beyond, I'm just not sure how to limit it when you may have a table like this one where it would look strange with the content not evenly spaced because we would need to enforce a `max-width`.

image

None of these columns content extends to the current width of the columns, so this would make the table much smaller. Each of the columns would immediately shrink to fit the content.

adamdehaven: what about maxing out at `100vw` or something?

adamdehaven (approved):
lgtm based on the previous reviews; I only reviewed the scroll container changes

PR #2114, opened 2024-04-04
feat(ktable): toggle column visibility [khcp-11162]

kaiarrowood:
# Summary

Update `KTable` to add ability to toggle the visibility of columns.
For [KHCP-11162](https://konghq.atlassian.net/browse/KHCP-11162).

image

image

## PR Checklist

[KHCP-11162]: https://konghq.atlassian.net/browse/KHCP-11162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adamdehaven:
suggestion: I think we need a header or something at the top of the popover so the user knows what they're doing with these checkboxes
image

src/components/KTable/ColumnVisibilityMenu.vue
```diff
@@ -0,0 +1,91 @@
...
+})
+</script>
+
+<style lang="scss" scoped>
+.k-table-column-visibility-menu {
+  margin-left: auto;
+
+  .apply-button {
+    margin-left: auto;
+    margin-right: auto;
```
portikM, line 88:
```suggestion
  margin-left: var(--kui-space-auto, $kui-space-auto);

  .apply-button {
    margin-left: var(--kui-space-auto, $kui-space-auto);
    margin-right: var(--kui-space-auto, $kui-space-auto);
```

src/components/KTable/ColumnVisibilityMenu.vue
```diff
@@ -0,0 +1,91 @@
...
+    type: Object as PropType<Record<string, boolean>>,
+    default: () => ({}),
+  },
+})
+
+const visibilityMap = ref<Record<string, boolean>>({})
+
+const handleApply = () => {
+  // pass by ref problems
+  emit('update:visibility', JSON.parse(JSON.stringify(visibilityMap.value)))
```
portikM, line 69: question: why do we need `JSON.parse(JSON.stringify(` here? Can't we just emmit the original object, or does that produce mutations?

docs/components/table.md
```diff
@@ -474,6 +475,30 @@ Example headers array:
 </script>
 ```
 
+#### `allowHide`
```
portikM, line 478:
```suggestion
#### allowHide
```

nit: I'm trying to bring all headers to one standard which is not to apply any styling to header text of any level

src/components/KTable/ColumnVisibilityMenu.vue
```diff
@@ -0,0 +1,91 @@
...
+        </KDropdownItem>
+        <KDropdownItem>
+          <KButton
+            appearance="tertiary"
+            class="apply-button"
+            @click="handleApply"
+          >
+            Apply
+          </KButton>
+        </KDropdownItem>
```
portikM, line 35: Apply button doesn't need to be wrapped by KDropdownItem (makes it look weird)

Screenshot 2024-04-05 at 10 21 29 AM

src/components/KTable/ColumnVisibilityMenu.vue
```diff
@@ -0,0 +1,91 @@
+<template>
+  <div class="k-table-column-visibility-menu">
+    <KDropdown>
+      <KTooltip text="Show/Hide Columns">
```
portikM, line 4: suggestion: we probably want to make tooltip text a prop (possible localization considerations?). Also I like sentence case better (_Show/hide columns_) but that might be just a personal preference.

src/components/KTable/ColumnVisibilityMenu.vue
```diff
@@ -0,0 +1,91 @@
...
+        <KButton
+          appearance="secondary"
+          class="menu-button"
+        >
+          <template #icon>
+            <CogIcon />
```
portikM, line 10: question: did we decide it will the cog icon? I remember we were discussing using columns icon or something

src/components/KTable/ColumnVisibilityMenu.vue
```diff
@@ -0,0 +1,91 @@
+<template>
+  <div class="k-table-column-visibility-menu">
+    <KDropdown>
+      <KTooltip text="Show/Hide Columns">
+        <KButton
+          appearance="secondary"
```
portikM, line 6: suggestion: maybe we should use `tertiary` appearance here?

docs/components/table.md
```diff
@@ -438,6 +438,7 @@ Pass in an array of header objects for the table.
 | `key`              | string  | A unique key for the column                                                                                                                |
 | `label`            | string  | The label displayed on the table for the column                                                                                            |
 | `sortable`         | boolean | Enables or disables server-side sorting for the column (`false` by default)                                                                |
+| `allowHide`        | boolean | Enable show/hide for this column                                                                                                           |
```
adamdehaven, line 441: nit: can we rename this prop to align with the other properties? 
```suggestion
| `hidable`        | boolean | Enable show/hide for this column                                                                                                           |
```

docs/components/table.md
adamdehaven: See my comment on the prop naming 😄

src/components/KTable/ColumnVisibilityMenu.vue
adamdehaven: Yes I think a different icon would be more ideal. Material Symbols has some column icons I believe we could add to `@kong/icons`

src/components/KTable/ColumnVisibilityMenu.vue
```diff
@@ -0,0 +1,91 @@
...
+
+const handleApply = () => {
+  // pass by ref problems
+  emit('update:visibility', JSON.parse(JSON.stringify(visibilityMap.value)))
+}
+
+onBeforeMount(() => {
+  // initialize visibility state
+  visibilityMap.value = props.columns.reduce((acc, col: TableHeader) => {
+    acc[col.label] = props.visibilityPreferences[col.label] === undefined ? true : props.visibilityPreferences[col.label]
```
adamdehaven, line 75: question: do we need to point to a clone of the object instead of the object itself?

src/components/KTable/ColumnVisibilityMenu.vue
```diff
@@ -0,0 +1,91 @@
...
+
+    return acc
+  }, {} as Record<string, boolean>)
+})
+</script>
+
+<style lang="scss" scoped>
+.k-table-column-visibility-menu {
```
adamdehaven, line 83: @portikM should we be using `k-*` classes here since this component isn't externally exposed?

src/components/KTable/KTable.vue
```diff
@@ -93,7 +99,7 @@
               :class="{ 'is-scrolled': isScrolled }"
             >
               <th
-                v-for="(column, index) in tableHeaders"
+                v-for="(column, index) in visibleHeaders"
                 :key="`k-table-${tableId}-headers-${index}`"
```
adamdehaven, line 103: can we use the column instead of the index as the key?

src/components/KTable/KTable.vue
```diff
@@ -169,11 +175,11 @@
               v-on="hasSideBorder ? tdlisteners(row, row) : {}"
             >
               <td
-                v-for="(value, index) in tableHeaders"
+                v-for="(value, index) in visibleHeaders"
                 v-bind="cellAttrs({ headerKey: value.key, row, rowIndex, colIndex: index })"
                 :key="`k-table-${tableId}-cell-${index}`"
```
adamdehaven, line 180: we shouldn't be using index in the key

src/components/KTable/KTable.vue
```diff
@@ -1040,6 +1066,18 @@ const getTestIdString = (message: string): string => {
...
+  const newVisibility = newVals[0]
+  const newHeaders = newVals[1]
+  const newVisibleHeaders = newHeaders.filter((header: TableHeader) => newVisibility[header.label] !== false)
+
+  if (JSON.stringify(newVisibleHeaders) !== JSON.stringify(visibleHeaders.value)) {
+    emitTablePreferences()
+  }
+
+  visibleHeaders.value = newVisibleHeaders
+}, { deep: true, immediate: true })
```
adamdehaven, line 1079: question: why does this need to be `immediate`? I wouldn't think it could change on mount?

src/components/KTable/KTable.vue
```diff
@@ -1040,6 +1066,18 @@ const getTestIdString = (message: string): string => {
...
+watch([columnVisibility, tableHeaders], (newVals) => {
+  const newVisibility = newVals[0]
+  const newHeaders = newVals[1]
+  const newVisibleHeaders = newHeaders.filter((header: TableHeader) => newVisibility[header.label] !== false)
+
+  if (JSON.stringify(newVisibleHeaders) !== JSON.stringify(visibleHeaders.value)) {
+    emitTablePreferences()
+  }
+
+  visibleHeaders.value = newVisibleHeaders
```
adamdehaven, line 1078: question: should we set the value before emitting?

Also, if the objects are equal, do we still need to set the `visibleHeaders.value` again?

src/components/KTable/ColumnVisibilityMenu.vue
portikM: I thought it will be externally exposed since we want to have control over where we put it.. am I wrong @kaiarrowood?

docs/components/table.md
portikM: I don't think "hidable" is a word. Besides, this is not a prop but a property within an object that a prop accepts

src/components/KTable/ColumnVisibilityMenu.vue
portikM: In that case yes, don't use `k-*` prefixed classes in this component

docs/components/table.md
portikM: Oh okay, let's make it `hidable` then!

kaiarrowood:
> suggestion: I think we need a header or something at the top of the popover so the user knows what they're doing with these checkboxes. Also may make sense to have a Cancel button image

@adamdehaven I added a tooltip, I still think we should use a different icon like you suggested but that's easy enough to change whenever design picks one.

image

I've added support to "cancel" when the dropdown is closed without clicking `Apply`. It will reset at that point.

src/components/KTable/ColumnVisibilityMenu.vue
kaiarrowood: For which? Aren't props non-mutatable?

adamdehaven: was thinking where it evals to, but may be fine

```ts
acc[col.label] = props.visibilityPreferences[col.label]
```

src/components/KTable/KTable.vue
kaiarrowood: I will change this but it's old code. Realistically this could wait until the reskin to fix?

adamdehaven: You can skip for now

src/components/KTable/KTable.vue
adamdehaven: You can skip for now

src/components/KTable/ColumnVisibilityMenu.vue
kaiarrowood: This component won't be exposed. We would end up needing the prop added to `KTable` which I don't think we want. Thoughts @adamdehaven ?

src/components/KTable/ColumnVisibilityMenu.vue
kaiarrowood: Picking up the new icon can be addressed in a followup PR once we have the desired icon in `@kong/icon`

src/components/KTable/KTable.vue
kaiarrowood: Because `visibleHeaders` needs to be initialized on mount once `tableHeaders` value is set.

src/components/KTable/ColumnVisibilityMenu.vue
adamdehaven: I believe we have other hard-coded text already in Kongponents? If yes I'd say just leave it hard-coded so we don't have prop-drilling 2-3 levels deep

src/components/KTable/ColumnVisibilityMenu.vue
adamdehaven: TBH I don't think this is ideal; if the table starts getting used before then, it's going to be confusing for Konnect users to see a random "settings" icon and then have it change a few days later. 

You should be able to get an icon pretty easily from Jason in `#ask-kong-design-system`

src/components/KTable/ColumnVisibilityMenu.vue
portikM: Yes we do. Okay, since this component won't be exported it's okay to leave it as is for now (or would be an overkill to add a prop just for this).

src/components/KTable/ColumnVisibilityMenu.vue
portikM: ^ Second that, let's use the proper icon from beginning - I don't think design will have strong opinions against the columns icon

src/components/KTable/ColumnVisibilityMenu.vue
```diff
@@ -0,0 +1,141 @@
+<template>
+  <div class="table-column-visibility-menu">
+    <KDropdown
+      data-testid="table-column-visibility-menu"
+      @toggle-dropdown="handleDropdownToggle"
+    >
+      <KTooltip text="Show/Hide Columns">
+        <KButton
+          appearance="secondary"
+          class="menu-button"
```
portikM, line 10: issue: since this button will essentially be rendered in KTable toolbar - where often times we will have KInput and/or KSelect, this button needs to have `size="large"` to render equal size next to those components. However it will be too big when it's the only thing in the toolbar which isn't ideal - so maybe this is something we can control with a prop? Something like `columnVisibilityButtonSize` - seems oddly specific and I know @adamdehaven will hate it (so do I) but I'm sure displaying medium size button next to KInputs will be ab issue down the line

portikM (dismissed):

src/components/KTable/ColumnVisibilityMenu.vue
adamdehaven: If we have inputs in the toolbar regularly, then just make the button size large by default? It shouldn't "change" depending on the siblings; we want consistency

portikM: ^ Works for me

kaiarrowood: image

image

adamdehaven: (regardless of the button/icon size) IMO the button just kind of "floating" there by itself is a bit odd; e.g. there's nothing anchoring it to the table or showing that it's related

portikM (approved):
LGTM 🚀

PR #2113, opened 2024-04-04
fix(kcopy): not using actual button

kleinfreund:
# Summary

Fix KCopy not using an actual button for its click interaction. This is strictly better because you don’t have to worry about re-implementing button functionality and it’s default accessibility properties.

Signed-off-by: Philipp Rudloff <philipp.rudloff@konghq.com>

## Notes

- There is a pre-existing "Extraneous non-props attributes (class, data-testid) were passed to component but could not be automatically inherited because component renders fragment or text root nodes." warning in this component.

## PR Checklist

* [x] **Conventional Commits** all commits follow the conventional commit standards [outlined in the main README](https://github.com/Kong/kongponents#committing-changes).
* [x] **Tests coverage:** test coverage was added for new features and bug fixes
* [x] **Docs:** includes a technically accurate README

PR #2119, opened 2024-04-05
fix: using icons as buttons

kleinfreund:
# Summary

Fix using icons as buttons in KModal, KMultiSelect, KSelect, KSlideout, KToaster, and KTruncate.

It’s important to use actual buttons for click interactions to ensure all the expected behavior is made available to users. Native buttons can be focused, handle keyboard interactions correctly, expose their accessibility properties (e.g. `role="button"`), and more all without requiring developer attention beyond using the appropriate markup. It should be noted that other means to expose an accessible name (done here using `aria-label`) are available and can be considered.

See also https://github.com/Kong/kongponents/pull/2113 and https://github.com/Kong/kongponents/pull/2100#discussion_r1553172711.

## PR Checklist

### DIFF
--- a/src/components/KCopy/KCopy.vue
+++ b/src/components/KCopy/KCopy.vue
@@ -37,6 +37,7 @@
         <KClipboardProvider v-slot="{ copyToClipboard }">
           <button
             :id="copyButtonElementId"
+            class="copy-to-clipboard-button"
             data-testid="copy-to-clipboard"
             type="button"
             @click.stop="copyIdToClipboard(copyToClipboard)"
@@ -306,6 +307,10 @@ onUnmounted(() => {
     font-size: var(--kui-font-size-20, $kui-font-size-20);
     line-height: var(--kui-line-height-20, $kui-line-height-20);
     margin-right: var(--kui-space-20, $kui-space-20);
+  }
+
+  .copy-to-clipboard-button {
+    @include defaultButtonReset;
   }
 }
 </style>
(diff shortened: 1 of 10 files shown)


portikM:
Good idea but styling needs to be fixed

Screenshot 2024-04-05 at 9 48 57 AM
Screenshot 2024-04-05 at 9 46 48 AM

kleinfreund:
> but styling needs to be fixed

> I suggest you create file `src/styles/mixins/_buttons.scss` (and import it in `src/styles/_mixins.scss`) where you can create a mixin for resetting default button styles. Depending on whether all these buttons share same styles or not, you could also create one mixin for styling all of them or style each of them individually in components style blocks (look up their default and hover/active state styles in classes that apply to them).

I made similar amends here now. I stuck mostly with the additions you made in https://github.com/Kong/kongponents/pull/2100/commits/c49f3cce88b3517f93dd74373fc746154ae4bea4, but needed a few more additions:

- `cursor: pointer`
- `color: inherit`

Note also that some class names are a little weird now because they make mention of `-icon` while no longer being used on an icon. I intentionally kept those to avoid unnecessary breaking changes.

kleinfreund:
I added another [commit](https://github.com/Kong/kongponents/pull/2119/commits/db5ead132f0dda84d5b5f627e6838e1783556dee) fixing the button styles not being reset in KCopy (caused by https://github.com/Kong/kongponents/pull/2113)

portikM (approved):
LGTM 🚀

adamdehaven (approved):
lgtm

PR #2123, opened 2024-04-08
fix(kalert,kemptystate): icon slot test

kaiarrowood:
# Summary

## PR Checklist

### DIFF
--- a/src/components/KAlert/KAlert.cy.ts
+++ b/src/components/KAlert/KAlert.cy.ts
@@ -1,8 +1,9 @@
+import { h } from 'vue'
 import { mount } from 'cypress/vue'
 import KAlert from '@/components/KAlert/KAlert.vue'
 import { AlertAppearances } from '@/types'
 
-const rendersCorrectVariant = (variant) => {
+const rendersCorrectVariant = (variant: string) => {
   it(`renders ${variant} variant`, () => {
     mount(KAlert, {
       props: {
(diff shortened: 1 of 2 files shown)


portikM (approved):
LGTM 🚀

src/components/KAlert/KAlert.cy.ts
```diff
@@ -85,21 +86,25 @@ describe('KAlert', () => {
...
-        default: `<span data-testid="default-slot-content">${defaultSlotContent}</span>`,
+        default: h('span', {}, defaultSlotContent),
       },
     })
 
     cy.get('.alert-message').should('be.visible').and('have.text', defaultSlotContent)
   })
 
   it('displays icon passed through icon slot', () => {
+    const dragon = '🐉'
```
adamdehaven, line 97: can we just check for a string here instead of an emoji?

adamdehaven (changes requested):

src/components/KAlert/KAlert.cy.ts
kaiarrowood: I can change this, but for the record the test is completely functional as is and this shouldn't be considered blocking feedback ☹️

adamdehaven: We shouldn't be using emojis in tests 🤷🏼

adamdehaven (approved):

PR #2130, opened 2024-04-11
feat: add zIndex prop to overlaying components [KHCP-10747]

vaibhavrajsingh2001:
# Summary

Adds a `zIndex` prop to all overlaying components (that need to be shown above the default layer) so that they can be assigned custom `z-index` values.

Components updated:
- `KModal`
- `KPop`
- `KSlideout`
- `KToaster`
- `KTooltip`
- `KSkeleton`
  - `FullScreenKongSkeleton`
  - `FullScreenGenericSpinner`

[KHCP-10747](https://konghq.atlassian.net/browse/KHCP-10747)

## PR Checklist

* [X] **Conventional Commits** all commits follow the conventional commit standards [outlined in the main README](https://github.com/Kong/kongponents#committing-changes).

[KHCP-10747]: https://konghq.atlassian.net/browse/KHCP-10747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/components/modal.md
+++ b/docs/components/modal.md
@@ -352,6 +352,20 @@ Maximum height of the content area. When content overflows, content area becomes
 </KModal>
 ```
 
+### zIndex
+
+`z-index` value for the modal. Defaults to `1100`.
+
+```html
+<KModal
+  z-index="9999"
+  title="Modal"
+  :visible="modalVisible"
+  @cancel="handleModalClose"
+  @proceed="handleModalProceed"
+/>
+```
+
 ### fullScreen
 
 Use this prop to make modal window take up almost the whole screen. When set to `true`, `maxWidth` and `maxHeight` props will have no effect.
--- a/src/components/KModal/KModal.cy.ts
+++ b/src/components/KModal/KModal.cy.ts
@@ -223,6 +223,31 @@ describe('KModal', () => {
     cy.get('.k-modal .modal-backdrop.modal-full-screen').should('be.visible')
   })
 
+  it('renders modal with correct zIndex when prop is passed', () => {
+
+    mount(KModal, {
+      props: {
+        visible: true,
+        zIndex: 1200,
+      },
+    })
+
+    // cypress only allows assertion with strings in CSS assertions
+    // https://github.com/cypress-io/cypress/blob/0e1a49dc461a670a3a1dd9a6e139eeb2f00c7c46/cli/types/cypress.d.ts#L5632
+    cy.get('.k-modal .modal-backdrop').should('have.css', 'z-index', '1200')
+  })
+
+  it('renders modal with correct default zIndex when prop is not passed', () => {
+
+    mount(KModal, {
+      props: {
+        visible: true,
+      },
+    })
+
+    cy.get('.k-modal .modal-backdrop').should('have.css', 'z-index', '1100') // default z-index
+  })
+
   it('emits proceed event when action button is clicked', () => {
     mount(KModal, {
       props: {
--- a/src/components/KModal/KModal.vue
+++ b/src/components/KModal/KModal.vue
@@ -170,6 +170,10 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+  zIndex: {
+    type: Number,
+    default: 1100,
+  },
 })
 
 const emit = defineEmits<{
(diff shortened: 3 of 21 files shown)


src/components/KSlideout/KSlideout.vue
```diff
@@ -224,7 +228,7 @@ onUnmounted(() => {
     position: fixed;
     right: 0;
     top: v-bind('offsetTopValue');
-    z-index: 9999;
+    z-index: v-bind(zIndex); // same value as z-index of slideout-container
```
vaibhavrajsingh2001, line 231: Since `slideout-backdrop` and `slideout-container` had the same `z-index` value, bounded both of them to the prop.

If only `z-index` of `slideout-container` was bound to the prop, it would allow someone to pass a `z-index` to `slideout-container` that could be less than the `z-index` of `slideout-backdrop`.

Another way to achieve this would be adding a validator to the `z-index` prop that would prevent passing a value lower than `9999` to it.

adamdehaven: this looks fine 👍🏼

src/components/KModal/KModal.cy.ts
```diff
@@ -223,6 +223,31 @@ describe('KModal', () => {
...
+
+    mount(KModal, {
+      props: {
+        visible: true,
+        zIndex: 1200,
+      },
+    })
+
+    // cypress only allows assertion with strings in CSS assertions
+    // https://github.com/cypress-io/cypress/blob/0e1a49dc461a670a3a1dd9a6e139eeb2f00c7c46/cli/types/cypress.d.ts#L5632
```
vaibhavrajsingh2001, line 236: Should I document this somewhere else, since it applies to every CSS assertion?

docs/components/modal.md
```diff
@@ -352,6 +352,20 @@ Maximum height of the content area. When content overflows, content area becomes
...
+```html
+<KModal
+  z-index="9999"
+  title="Modal"
+  :visible="modalVisible"
+  @cancel="handleModalClose"
+  @proceed="handleModalProceed"
+/>
+```
+
```
vaibhavrajsingh2001, line 368: Since a custom z-index won't be visually discernable, there's not much benefit of adding a try-out component.

src/components/KModal/KModal.cy.ts
adamdehaven: Prob. not needed

src/components/KModal/KModal.vue
```diff
@@ -299,7 +303,7 @@ onBeforeUnmount(async () => {
     justify-content: center;
     padding: var(--kui-space-70, $kui-space-70) var(--kui-space-50, $kui-space-50) var(--kui-space-0, $kui-space-0) var(--kui-space-50, $kui-space-50);
     position: fixed;
-    z-index: 1100;
+    z-index: v-bind(zIndex);
```
portikM, line 306:
```suggestion
    z-index: v-bind('zIndex');
```

issue: `v-bind` parameter needs to be wrapped by single quotes

src/components/KPop/KPop.vue
```diff
@@ -459,7 +466,7 @@ export default defineComponent({
   padding: var(--kui-space-80, $kui-space-80) var(--kui-space-60, $kui-space-60);
   text-align: left;
   white-space: normal;
-  z-index: 1000;
+  z-index: v-bind(zIndex);
```
portikM, line 469:
```suggestion
  z-index: v-bind('zIndex');
```

Same here and other occurrences

src/styles/mixins/_loaders.scss
```diff
@@ -1,6 +1,6 @@
 /* Loaders styles */
 
-@mixin fullScreenLoadingContainer {
+@mixin fullScreenLoadingContainer($zIndex: 10500) {
```
portikM, line 3: Fallback value, nice ✨

src/components/KModal/KModal.vue
adamdehaven: Yep, let's standardize by wrapping: https://vuejs.org/api/sfc-css-features.html#v-bind-in-css

portikM: According to the docs seems like it isn't always required but I think we should always wrap since it's safer while still correct

adamdehaven: it's required when accessing properties of an object, etc.

In this particular case, referencing the prop should really be done via `v-bind('props.zIndex')` I believe? But yes quoted regardless

portikM (approved):
LGTM 🚀

PR #2157, opened 2024-04-24
fix(kdropdownitem): attribute handling

portikM:
# Summary

Bind attributes to underlying `.dropdown-item-trigger` instead of wrapper `li` element in `KDropdownItem`

## PR Checklist


src/components/KDropdown/KDropdownItem.vue
```diff
@@ -1,12 +1,12 @@
...
     class="k-dropdown-item"
-    :class="{
+    :class="[{
       'has-divider': hasDivider,
       'disabled': disabled,
       'danger': danger || isDangerous,
-      'dropdown-selected-option': selected
-    }"
+      'dropdown-selected-option': selected,
+    }, $attrs.class]"
```
adamdehaven, line 9: question: should we purposefully bind inline styles to the parent as well? I don't think they should fallthrough

portikM: I'm fine with whichever we pick

kaiarrowood, line 9: I'm confused, wasn't the point to bind the `data-testid` to the element that actually triggers the click (ie. the `button`, `li`, etc)? If the child `<component>` still has a `data-testid` (which it does - `dropdown-item-trigger`) won't we still have the original problem attempting to catch the clicks in DD?

portikM: No - if we provide `data-testid` it will override the original value

kaiarrowood: But how? I'm not understanding what this `availableComponent` thing is

portikM: It binds provided attributes after it sanitizes them a bit: https://github.com/Kong/kongponents/blob/75576a48ae122e9be72d891638d00b925289f3cf/src/components/KDropdown/KDropdownItem.vue#L156

portikM: Going on with your suggestion to bind inline styles to the parent

src/components/KDropdown/KDropdownItem.vue
```diff
@@ -1,12 +1,12 @@
...
     class="k-dropdown-item"
-    :class="{
+    :class="[{
       'has-divider': hasDivider,
       'disabled': disabled,
       'danger': danger || isDangerous,
-      'dropdown-selected-option': selected
-    }"
+      'dropdown-selected-option': selected,
+    }, $attrs.class, $attrs.style]"
```
adamdehaven, line 9: question: why are we moving all of the attributes? This is a larger breaking change

src/components/KDropdown/KDropdownItem.vue
```diff
@@ -24,11 +24,17 @@
   </li>
 </template>
 
-<script lang="ts" setup>
+<script lang="ts">
 import type { DropdownItem, DropdownItemRenderedRecord, DropdownItemRenderedType, DropdownItemType } from '@/types'
 import type { PropType } from 'vue'
 import { computed, useAttrs } from 'vue'
 
+export default {
```
adamdehaven, line 32: suggestion: if you need this you can use defineOptions

portikM: Done

adamdehaven (approved):
lgtm

PR #2178, opened 2024-05-09
fix(ktooltip): suppress vue attr passthrough warning

nekolab:
it's safe to set `inheritAttrs` to `false` because we have set `v-bind="$attrs"` on `<KPop />` manually

# Summary

## PR Checklist

### DIFF
--- a/src/components/KTooltip/KTooltip.vue
+++ b/src/components/KTooltip/KTooltip.vue
@@ -41,6 +41,10 @@ import type { PopPlacements } from '@/types'
 import { PopPlacementsArray } from '@/types'
 import { v4 as uuidv4 } from 'uuid'
 
+defineOptions({
+  inheritAttrs: false,
+})
+
 const props = defineProps({
   /**
   * Text to show in tooltip


src/components/KTooltip/KTooltip.vue
```diff
@@ -41,6 +41,10 @@ import type { PopPlacements } from '@/types'
 import { PopPlacementsArray } from '@/types'
 import { v4 as uuidv4 } from 'uuid'
 
+defineOptions({
+  inheritAttrs: false,
```
adamdehaven, line 45: Would an equivalent change be to just remove the `v-bind=“$attrs”` on the root element since they are essentially bound by default?

nekolab: Nope, just removing the `v-bind=“$attrs”` still triggers the warning because the `<slot />` is still wrapped by a `<template />`

image

Although we can also change the
```
  <template v-else>
    <slot />
  </template>
```
to 
```
  <div v-else>
    <slot />
  </div>
```
to suppress the warning but it will introduce an extra wrapper to the slot, and the `$attrs` will bind to the div wrapper even for the "no-text-to-show" scenario in the tooltip

Personally, I want to avoid this behavior. I'd prefer to treat the `<KToolip />` as a conditional wrapper, when the text is provided, it wraps the children and renders a popover when hovered. But if there's no text to show, it should render the children directly without any wrap. And this is the current behavior.

Please let me know if you think the "always wrap" behavior is better, cc @Leopoldthecoder since he introduced the conditional render logic here

Leopoldthecoder: I can provide some context: this fix is mainly about getting rid of the console warning. [`KCopy` binds `class` and `data-testid` to `KTooltip`](https://github.com/Kong/kongponents/blob/411927ff00a8523e1de65e6151a62410c71aaf9b/src/components/KCopy/KCopy.vue#L15-L16), which confuses Vue because KTooltip's root node is conditionally rendered.

adamdehaven (approved):

PR #2184, opened 2024-05-10
docs(*): add comments in interfaces

portikM:
# Summary

## PR Checklist

### DIFF
--- a/docs/components/select.md
+++ b/docs/components/select.md
@@ -44,7 +44,7 @@ Prop for providing select item options. Supports grouping items under one group 
 interface SelectItem {
   label: string
   value: string | number
-  key?: string
+  key?: string // optional parameter that will be appended with `-selected` when selected
   selected?: boolean
   disabled?: boolean
   group?: string
--- a/src/types/catalog.ts
+++ b/src/types/catalog.ts
@@ -13,7 +13,7 @@ export interface CatalogItem extends Record<string, any> {
   id?: string
   title: string
   description: string
-  key?: string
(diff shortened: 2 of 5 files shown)


docs/components/select.md
```diff
@@ -44,7 +44,7 @@ Prop for providing select item options. Supports grouping items under one group
 interface SelectItem {
   label: string
   value: string | number
-  key?: string
+  key?: string // optional parameter that will be appended with `-selected` when selected
```
adamdehaven, line 47: should you add the comment as a block comment on the actual interface as well?

portikM: Done

src/types/catalog.ts
```diff
@@ -13,7 +13,7 @@ export interface CatalogItem extends Record<string, any> {
   id?: string
   title: string
   description: string
-  key?: string
+  key?: string // optional parameter to be used as key in v-for, has to be unique for each item
```
adamdehaven, line 16: suggestion: If you add as an actual block comment above the property, it will show up in intellisense (same comment for the other instance)

```suggestion
  /** Optional parameter to be used as key in v-for, has to be unique for each item */
  key?: string
```

portikM: > it will show up in intellisense

You mean like so?
Screenshot 2024-05-10 at 2 25 17 PM

portikM: If that's what you mean than inline comments show up too

portikM: Or am I getting it wrong?

adamdehaven (approved):

PR #2191, opened 2024-05-16
fix(ktable): emit preferences only in success state [KHCP-11890]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-11890

## PR Checklist

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -207,6 +207,10 @@ Boolean to control whether the component should display the error state. Default
 
 Allow table column width to be resizable (defaults to `false`). Adjusting a column's width will trigger an [`update:table-preferences` event](#updatetable-preferences).
 
+:::warning WARNING
+`update:table-preferences` event only fires when KTable is in `success` state (see [`state` event](#state) for details).
+:::
+
 <KTable
   resize-columns
   :fetcher="basicFetcher"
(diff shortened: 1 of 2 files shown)


kaiarrowood (commented):
I think this approach makes sense if we can get it to work. Until the table data has finished loading there is no need to load the table preferences IMO. I would ask that we update the logic for the menu trigger button to handle the other table states though (if we aren't already).

```
const tableState = computed((): TableState => isTableLoading.value ? 'loading' : fetcherError.value ? 'error' : 'success')
```

If `tableState` is `loading` lets `disable` the menu button; if it's `error` lets hide the button.

portikM:
> I think this approach makes sense if we can get it to work. Until the table data has finished loading there is no need to load the table preferences IMO. I would ask that we update the logic for the menu trigger button to handle the other table states though (if we aren't already).
> 
> ```
> const tableState = computed((): TableState => isTableLoading.value ? 'loading' : fetcherError.value ? 'error' : 'success')
> ```
> 
> If `tableState` is `loading` lets `disable` the menu button; if it's `error` lets hide the button.

That's how it works already (column visibility and pagination controls are hidden when table is on loading or error states)

kaiarrowood (approved):
LGTM! 🚀

PR #2214, opened 2024-06-10
fix(*): replace component uuid usage [KHCP-11956]

portikM:
# Summary

* adds plugin options object that gets accepts `uniqueIdFn` parameter
* in components, instead of using `uuid` for generating random ids by default, use provided `uniqueIdFn` function (falls back to `uuid` if none was provided)

## PR Checklist


docs/guide/usage.md
```diff
@@ -81,6 +81,42 @@ this path instead: import '~@kong/kongponents/dist/style.css' */
 </style>
 ```
 
+## SSR
+
+If you're using Kongponents in SSR mode, we advice [registering Kongponents as a Vue Plugin](#vue-plugin).
+
+Kongponents use external function to generate random strings that are mostly used in accessibility-related attributes. In server-side rendering, that will produce hydration mismatch because the function returns different values on server and client side. However, you can provide your own SSR-compatible function for generating random strings or ids through plugin options.
```
adamdehaven, line 88:
```suggestion
Kongponents uses an external function to generate UUID strings that are mostly used in accessibility-related attributes. 

In server-side rendering, these unique ids will produce hydration mismatch errors because the function returns different values on the server and client. 

You can provide your own SSR-compatible function for generating random strings or ids through the Vue plugin options.
```

docs/guide/usage.md
```diff
@@ -81,6 +81,42 @@ this path instead: import '~@kong/kongponents/dist/style.css' */
...
+
+const app = createApp(App)
+
+// Install and register all Kongponents as a plugin
+app.use(Kongponents, {
+  kongponentsId: useId
+})
+
+app.mount('#app')
+```
```
adamdehaven, line 110: suggestion: let's update this to our actual use-case
```suggestion

Here's an example of how to initialize the Kongponents plugin in a [Nuxt plugin](https://nuxt.com/docs/guide/directory-structure/plugins#vue-plugins).

```ts
// plugins/kongponents.ts

// Import Kongponents plugin
import Kongponents from '@kong/kongponents'

// Import Kongponents styles
import '@kong/kongponents/dist/style.css'

export default defineNuxtPlugin({
  name: 'kongponents',
  setup(nuxtApp) {
    nuxtApp.vueApp.use(Kongponents, {
      kongponentsId: useId // Utilize the `useId()` function from Nuxt/Vue (v3.5+)
    })
  },
})
```
```

File: sandbox/index.ts [hunk 3]
```diff
@@ -7,6 +7,9 @@ import Kongponents from '../src'
 import { SandboxLayout } from '@kong-ui-public/sandbox-layout'
 import type { SandboxNavigationItem } from '@kong-ui-public/sandbox-layout'
 import '@kong-ui-public/sandbox-layout/dist/style.css'
+import { v4 as uuidv4 } from 'uuid'
+
+const kongponentsSandboxId = () => `kongponents-sandbox-${uuidv4()}`
```
adamdehaven, line 12: suggestion: we typically won't want a prefix before the UUID since we _already_ usually prefix them in the Kongponent
```suggestion
const kongponentsSandboxId = uuidv4
```

src/composables/useGetRandomId.ts
```diff
@@ -0,0 +1,11 @@
...
+import { kongponentsRandomIdFunction } from '@/types'
+import { v4 as uuidv4 } from 'uuid'
+
+export default function useGetRandomId() {
+  const kongponentsId: () => string = inject<() => string>(kongponentsRandomIdFunction, uuidv4)
+
+  return {
+    kongponentsId,
+  }
+}
```
adamdehaven, line 11: suggestion: since there's no reason to export anything else from this file, let's simplify it a bit:
```suggestion
export default inject<() => string>(kongponentsRandomIdFunction, uuidv4)
```

A couple other changes:
1. Let's rename the composable to `useUniqueId`
2. The usage would also change:

    ```ts
    import useUniqueId from '@/composables/useUniqueId' // or maybe just from `@/composables` ?

    const randomTooltipId = useUniqueId()
    ```

src/types/kongponents-plugin.ts
```diff
@@ -0,0 +1,12 @@
+import { v4 as uuidv4 } from 'uuid'
+
+export const kongponentsRandomIdFunction = 'kongponents-id' as const
```
adamdehaven, line 3: nit: can we rename this a bit:
```suggestion
export const kongponentsGetUniqueId = 'kongponents-get-unique-id' as const
```

src/types/kongponents-plugin.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface KongponentsPlugin {
+  kongponentsId: () => string
+}
+
+export const defaultKongponentsPlugin: KongponentsPlugin = {
+  kongponentsId: uuidv4,
```
adamdehaven, line 10: This will impact other files, but since this is an option within the `Kongponents` plugin, we don't need to prefix it
```suggestion
  uniqueIdFn: () => string
}

export const defaultKongponentsPlugin: KongponentsPlugin = {
  uniqueIdFn: uuidv4,
```

adamdehaven (changes requested):

sandbox/index.ts
adamdehaven: ah ok

src/composables/useGetRandomId.ts
portikM: This doesn't work (I kept getting `useUniqueId is not a function error`)

adamdehaven: is it getting initialized before the composable is ready?

portikM: I'm not sure - in the console there were two messages:
```
inject() can only be used inside setup() or functional components
```
and the second one `useUniqueId is not a function`

adamdehaven: ah wait that's the issue; you can't inject from a composable

portikM: I tried moving the import inside `<script setup...` but that didn't help

adamdehaven: You're going to have to inject in every component

portikM: Does that mean in every component I need to make sure the import statement is within `<script setup...`?

portikM: Okay let me give it a try

portikM: Nope, getting the same error. I suggest we live this the way it is - we can always improve this later if we want to

portikM: I named it `uniqueIdFn` - will that do?

adamdehaven: sgtm

docs/guide/usage.md
```diff
@@ -81,6 +81,45 @@ this path instead: import '~@kong/kongponents/dist/style.css' */
...
+import Kongponents from '@kong/kongponents'
+
+// Import Kongponents styles
+import '@kong/kongponents/dist/style.css'
+
+export default defineNuxtPlugin({
+  name: 'kongponents',
+  setup(nuxtApp) {
+    nuxtApp.vueApp.use(Kongponents, {
+      kongponentsId: useId // Utilize the `useId()` function from Nuxt/Vue (v3.5+)
```
adamdehaven, line 110: don't forget to update the usage and in other places to `uniqueIdFn`

adamdehaven, line 110:
```suggestion
      uniqueIdFn: () => useId(), // Utilize the `useId()` function from Nuxt/Vue (v3.5+)
```

portikM: Done

docs/guide/usage.md
```diff
@@ -81,6 +81,45 @@ this path instead: import '~@kong/kongponents/dist/style.css' */
...
+  },
+})
+
+Plugin options object accepts these parameters:
+
+```ts
+interface KongponentsPlugin {
+  uniqueIdFn: () => string
+}
+```
```
adamdehaven, line 121: docs: once we decide on a solution, we should also document that you can just call `app.provide` in the host app if you don't initialize all Kongponents as a plugin

PR #2231, opened 2024-06-14
feat: switch from yarn to pnpm

adamdehaven:
# Summary

Switch from `yarn` to `pnpm`


portikM (commented):
Looks good to me but will defer to Val for final review.

portikM (approved):

PR #2241, opened 2024-06-17
fix(kfileupload, kselect, kmultiselect): kinput id [KHCP-12336]

portikM:
# Summary

Binding id to KInput in KFileUpload, KSelect and KMultiselect through `v-bind-once` directive creates a problem where the id gets assigned to the wrong element because the directive ignores the `inherritAttrs: false` setting. That means:
* in cases when we don't pass a static id and automatically generate one in the component, the id attribute on input element is different from for attribute value on label element
* in cases when we pass static id, two elements (the input and the wrapping element) end up having the same id

Solution: do not rely on `v-bind-once` in said components and
* when we don't set static id, in `onMounted` hook assign label for attribute programatically based on value of input id attribute
* use static value when one is provided

## PR Checklist

### DIFF
--- a/sandbox/pages/SandboxFileUpload.vue
+++ b/sandbox/pages/SandboxFileUpload.vue
@@ -3,14 +3,7 @@
     :links="inject('app-links', [])"
     title="KFileUpload"
   >
-    <KFileUpload
-      :accept="['image/*']"
-      label="Label"
-    />
-    <div
-      v-if="false"
-      class="kfileupload-sandbox"
-    >
+    <div class="kfileupload-sandbox">
       <!-- Props -->
       <SandboxTitleComponent
         is-subtitle
--- a/sandbox/pages/SandboxFileUpload.vue
+++ b/sandbox/pages/SandboxFileUpload.vue
@@ -3,7 +3,14 @@
     :links="inject('app-links', [])"
     title="KFileUpload"
   >
-    <div class="kfileupload-sandbox">
+    <KFileUpload
+      :accept="['image/*']"
+      label="Label"
+    />
+    <div
+      v-if="false"
+      class="kfileupload-sandbox"
+    >
       <!-- Props -->
       <SandboxTitleComponent
         is-subtitle
--- a/src/components/KFileUpload/KFileUpload.vue
+++ b/src/components/KFileUpload/KFileUpload.vue
@@ -267,6 +267,11 @@ const setLabelAttributes = () => {
   }
 }
 
+watch(fileInputKey, async () => {
+  await nextTick()
+  setLabelAttributes()
+})
+
 watch(() => attrs.id, async () => {
   fileInputKey.value++
   await nextTick()
--- a/src/components/KFileUpload/KFileUpload.vue
+++ b/src/components/KFileUpload/KFileUpload.vue
@@ -1,9 +1,13 @@
 <template>
-  <div class="k-file-upload">
+  <div
+    class="k-file-upload"
+    v-bind="modifiedAttrs"
+  >
     <KLabel
       v-if="label"
-      v-bind-once="{ for: inputId }"
       v-bind="labelAttributes"
+      ref="labelElement"
+      :for="$attrs.id ? String($attrs.id) : undefined"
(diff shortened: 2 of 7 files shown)


src/components/KFileUpload/KFileUpload.vue
```diff
@@ -26,9 +30,9 @@
       </span>
 
       <KInput
+        v-bind="attrs"
```
Leopoldthecoder, line 33: The [sandbox preview deploy](https://deploy-preview-2241--kongponents-sandbox.netlify.app/#/fileupload) is broken because when we set `model-value` on a KFileUpload, it'll be passed to the `<input>` element. I suggest changing this line to:
```suggestion
        v-bind="attrs.id ? { id: String(attrs.id) } : {}"
```

portikM: Done

src/components/KFileUpload/KFileUpload.vue
```diff
@@ -235,6 +252,35 @@ const resetInput = (): void => {
...
+
+watch(fileInputKey, async () => {
+  await nextTick()
+  setLabelAttributes()
+})
+
+watch(() => attrs.id, async () => {
+  fileInputKey.value++
+  await nextTick()
+  setLabelAttributes()
```
adamdehaven, line 278: issue: you're calling `setLabelAttributes` in both functions, so you're essentially doing 2 nextTick awaits and 2 setLabelAttributes calls every single time?

src/components/KFileUpload/KFileUpload.vue
```diff
@@ -64,14 +68,19 @@
   </div>
 </template>
 
+<script lang="ts">
+export default {
+  inheritAttrs: false,
+}
+</script>
+
```
Leopoldthecoder, line 76: nit: we can add `defineOptions` in the `<script lang="ts" setup>` block:
```ts
defineOptions({
  inheritAttrs: false,
})
```

Then we don't need another `<script>` block.

adamdehaven: I’d actually recommend making this change anywhere we’re already using inheritors: false

src/components/KFileUpload/KFileUpload.vue
portikM: Good catch, fixed

src/components/KFileUpload/KFileUpload.vue
portikM: That produces typescript errors

Screenshot 2024-06-18 at 10 51 57 PM

portikM: Oh, sorry, you suggest using `defineOptions`. Pushed a fix

portikM: I can handle updating that in every component other than ones scoped in this PR in a separate PR

sandbox/pages/SandboxFileUpload.vue
```diff
@@ -3,7 +3,14 @@
...
   >
-    <div class="kfileupload-sandbox">
+    <KFileUpload
+      :accept="['image/*']"
+      label="Label"
+    />
+    <div
+      v-if="false"
+      class="kfileupload-sandbox"
+    >
```
Leopoldthecoder, line 13: LGTM. Please feel free to merge after these changes are reverted.

Leopoldthecoder (approved):

src/components/KFileUpload/KFileUpload.vue
```diff
@@ -235,6 +250,33 @@ const resetInput = (): void => {
...
+    const inputElementId = fileInputElement.value?.$el?.querySelector('input')?.id
+
+    if (inputElementId) {
+      labelElement.value?.$el?.setAttribute('for', inputElementId)
+    }
+  }
+}
+
+watch(fileInputKey, async () => {
+  await nextTick()
```
adamdehaven, line 269: Why does this code need to be in a separate watcher instead of in the one below that triggers the key to increment?

PR #2262, opened 2024-07-02
fix(ktable): `props.headers` should be reactive [MA-3028]

adorack:
# Summary

- Copy changes to internal ref on prop change.
- Update tests.

### DIFF
--- a/src/components/KTable/KTable.vue
+++ b/src/components/KTable/KTable.vue
@@ -432,7 +432,7 @@ const props = defineProps({
    * A prop to pass in an array of headers for the table
    */
   headers: {
-    type: Array,
+    type: Array as PropType<TableHeader[]>,
     default: () => [],
   },
   /**
@@ -861,7 +861,7 @@ const initData = () => {
 
   // get table headers
   if (props.headers && props.headers.length) {
-    tableHeaders.value = props.headers as TableHeader[]
+    tableHeaders.value = props.headers
   }
 
   // trigger setting of tableFetcherCacheKey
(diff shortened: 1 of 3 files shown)


src/components/KTable/KTable.vue
```diff
@@ -868,6 +868,13 @@ const initData = () => {
   hasInitialized.value = true
 }
 
+// Ensure `props.headers` are reactive.
+watch(() => props.headers as TableHeader[], (newVal: TableHeader[]) => {
+  if (newVal && newVal.length) {
+    tableHeaders.value = newVal
+  }
+})
```
adamdehaven, line 876: for this to be actually reactive, e.g. if you changed a property of an individual header, you would need a `deep` watcher.

As an example, if you changed `hidable: true` to `false`

src/components/KTable/KTable.vue
```diff
@@ -868,6 +868,13 @@ const initData = () => {
   hasInitialized.value = true
 }
 
+// Ensure `props.headers` are reactive.
+watch(() => props.headers as TableHeader[], (newVal: TableHeader[]) => {
```
adamdehaven, line 872: suggestion: rather than force-casting the props, could we add the proper interface within the `defineProps` macro?

```typescript
headers: {
  type: Array as PropType<TableHeader[]>,
  default: () => [],
},
```

adamdehaven: We'll need to ensure that existing usage doesn't choke on the type change (e.g. pull the package preview into `konnect-ui-apps` in a draft PR and run the `typecheck` command recursively)

adamdehaven (changes requested):

src/components/KTable/KTable.vue
adorack: Good catch -- done.

src/components/KTable/KTable.vue
adorack: https://github.com/Kong/konnect-ui-apps/pull/3608 -- I tweaked the table header type to match how it's actually used in the wild, and that seems to work.

adamdehaven (approved):

PR #2276, opened 2024-07-11
fix(kpagination): detect and fix overflow [KHCP-10930]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-10930

Changes:
* KPaginatiion to automatically detect horizontal overflow and adjust number of displayed pagination items until overflow is fixed
* Styling to stack on mobile (screenshot)

Screenshot 2024-07-12 at 11 52 09 AM

### DIFF
--- a/src/components/KPagination/KPagination.vue
+++ b/src/components/KPagination/KPagination.vue
@@ -2,6 +2,7 @@
   <nav
     ref="kPaginationElement"
     class="k-pagination"
+    :class="{ 'full': !disablePageJump && !offset }"
     data-testid="k-pagination"
   >
     <template v-if="!offset">
@@ -105,6 +106,15 @@
       @get-previous-offset="getPreviousOffset"
     />
     <div class="page-size-select">
+      <span
+        v-if="!disablePageJump && !offset"
+        class="pagination-text-mobile"
+        data-testid="visible-items"
+      >
+        <span class="pagination-text-pages">{{ pagesString }}</span>
+        {{ pageCountString }}
+      </span>
+
       <KDropdown
         class="page-size-dropdown"
         data-testid="page-size-dropdown"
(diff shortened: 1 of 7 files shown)


src/components/KPagination/KPagination.vue
```diff
@@ -131,12 +132,13 @@
...
 <script setup lang="ts">
 import type { Ref, PropType } from 'vue'
-import { ref, computed, watch } from 'vue'
+import { ref, computed, watch, onMounted, onUnmounted, nextTick } from 'vue'
 import KDropdown from '@/components/KDropdown/KDropdown.vue'
 import KButton from '@/components/KButton/KButton.vue'
 import PaginationOffset from './PaginationOffset.vue'
 import type { PageSizeChangeData, PageChangeData, DropdownItem } from '@/types'
 import { BackIcon, ForwardIcon, ChevronDownIcon } from '@kong/icons'
+import { ResizeObserverHelper } from '@/utilities/resizeObserverHelper'
```
adamdehaven, line 141: (I realize this is still a draft)
If we’re going to expand the usage of this resize observer helper, I think we first need to debug the errors it seems to be throwing in Datadog on a regular basis

portikM: Is it a blocker for this PR? I remember Kai mentioning something he wanted to try that could solve the ResizeObserver issue but bigger picture I think we need to create a ticket and assign a person responsible for fixing or silencing that error in DD (I feel like we circled around this topic a few times already but never got to action)

adamdehaven: Yes, I think we need to create a ticket and see if it can be picked up next sprint. Not a blocker here, but just make sure this addition properly unmounts, prevents loops, etc.

Logging in the callback(s) and playing around in the sandbox can likely help identify some initial issues

portikM: Sgtm, will bring it up in the standup parking lot for discussion 👍

PR #2282, opened 2024-07-16
fix(kselect): blur input on click

adamdehaven:
# Summary

Prevent the KSelect input triggering the keyboard on mobile when filtering is not enabled


portikM (approved):
LGTM 🚀

PR #2301, opened 2024-07-23
fix(kpagination): remove dynamic size props [KHCP-12680]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-12680

KPagination is throwing hydration errors in SSR mode because of conditional props dependant on screen size. Instead of relying on js for that, add media queries to override certain styles on large and small screens.

### DIFF
--- a/src/components/KPagination/PaginationOffset.vue
+++ b/src/components/KPagination/PaginationOffset.vue
@@ -7,7 +7,7 @@
       data-testid="previous-button"
       :disabled="previousButtonDisabled"
       icon
-      :size="isWideScreen ? 'medium' : 'small'"
+      size="small"
       type="button"
       @click.prevent="emit('getPreviousOffset')"
     >
@@ -21,7 +21,7 @@
       data-testid="next-button"
       :disabled="nextButtonDisabled"
       icon
-      :size="isWideScreen ? 'medium' : 'small'"
+      size="small"
       type="button"
       @click.prevent="emit('getNextOffset')"
     >
@@ -32,8 +32,6 @@
 
 <script setup lang="ts">
 import { BackIcon, ForwardIcon } from '@kong/icons'
-import { useMediaQuery } from '@vueuse/core'
-import { KUI_BREAKPOINT_MOBILE } from '@kong/design-tokens'
 
 defineProps({
   previousButtonDisabled: {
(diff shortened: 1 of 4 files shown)


src/components/KPagination/PaginationOffset.vue
```diff
@@ -7,7 +7,7 @@
       data-testid="previous-button"
       :disabled="previousButtonDisabled"
       icon
-      :size="isWideScreen ? 'medium' : 'small'"
+      size="small"
```
adamdehaven, line 10: I'm thinking we should do the inverse: default to medium and apply the small styles on the smaller viewport? 

The most common use-case for all of Konnect is always going to be Desktop, so having the classes match the rendered DOM seems more "normal" with `medium` as the default.

portikM: That would mean we'd take desktop-first approach. I always thought we want to stick to mobile-first approach, across the board?

adamdehaven: The styles can be mobile-first, but the class reflected in the DOM would be un-manipulated

portikM: I'd like to stick to this approach because reversing it would require writing a desktop-first query too (`@media (max-width: $kui-breakpoint-mobile)` instead of more traditional and preferable IMO, `min-width`).

adamdehaven: Couldn't the `medium` button appearance still be used, but your style rules would still apply?

portikM: No. It's `small` by default and on screen size above mobile I apply `medium` size styles. If we made it `medium` by default I would need to apply `small` size styling on screen < mobile.

adamdehaven: What I'm worried about is someone trying to target `medium` buttons for padding via customization. I guess since we have unique classes on these buttons it's fine; just don't love the `button.small/button.medium` still there but 🤷🏼

adamdehaven:
image

Should you also adjust the popover styles on smaller screens for pagination?

src/components/KPagination/PaginationOffset.vue
portikM: Yes, I think if someone is using `medium` or `small` classes for customization that's a bad approach in the first place but also chances of us doing those overrides anywhere in Konnect are very low.

portikM:
> image
> 
> Should you also adjust the popover styles on smaller screens for pagination?

We don't really do that anywhere. I think it's a good idea but if we were to do that I suggest that we do it in a separate ticket.

adamdehaven (approved):

PR #2302, opened 2024-07-24
fix(kcodeblock): correct the line height for single line code block

nekolab:
Adjusts the line height to center the text in single-line code blocks

Before & After:
image

JIRA: KM-347


adamdehaven (approved):
lgtm

PR #2303, opened 2024-07-24
fix(kpagination): ensure currentPage is highlighted

johncowen:
If your current page is not in the set of page buttons that KPagination defaults to (e.g. [1] [2] [3] [...] [20] with a current page of `7`). This fix ensures that the correct set of page buttons is displayed plus highlights the button that corresponds to the current page.

Honestly, I've kinda no idea what this is doing (I've no idea what `firstDetached` means for example), but I fiddled with things a little and noticed the hardcoded `false` in related code. Following this I copy/pasta'd some code in from somewhere else that seemed to be doing something important 🤷 and things began to work 😅 

I also added an example to the sandbox which I used to figure this out.

---

FYI: Our use case is that we always pass bookmarkable `?page=7` parameters to our pages, previous to this fix a '7 page button' would not be shown or highlighted, also see https://github.com/kumahq/kuma-gui/issues/2769

### DIFF
--- a/sandbox/pages/SandboxPagination.vue
+++ b/sandbox/pages/SandboxPagination.vue
@@ -17,6 +17,7 @@
       </SandboxSectionComponent>
       <SandboxSectionComponent title="pageSizes">
         <KPagination
+          :current-page="7"
           :page-sizes="[10, 20, 30, 40]"
           :total-count="100"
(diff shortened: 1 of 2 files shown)


adamdehaven (changes requested):
This appears to break the functionality on `main` branch; blocking merge until discussion is resolved

PR #2317, opened 2024-07-31
fix(kprompt): proceed on enter

portikM:
# Summary

Allow trigger proceed action on _Enter_ upon typing in confirmation text in KPrompt

### DIFF
--- a/docs/components/prompt.md
+++ b/docs/components/prompt.md
@@ -356,7 +356,7 @@ interface ModalAttributes {
 
 ### errorMessage
 
-Error message text that will be displayed once user attempts to proceed after having typed in the wrong confirmation prompt. Default value is "Confirmation text does not match.".
+Error message text that will be displayed once user attempts to proceed after having typed in the wrong confirmation prompt. Default value is "Please enter the text as indicated above.".
 
 <KButton @click="prompt14Visible = true">Prompt</KButton>
 <KPrompt
--- a/docs/components/prompt.md
+++ b/docs/components/prompt.md
@@ -354,16 +354,41 @@ interface ModalAttributes {
 />
 ```
 
+### errorMessage
+
+Error message text that will be displayed once user attempts to proceed after having typed in the wrong confirmation prompt. Default value is "Confirmation text does not match.".
+
+<KButton @click="prompt14Visible = true">Prompt</KButton>
+<KPrompt
+  confirmation-text="confirm"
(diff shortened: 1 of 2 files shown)


docs/components/prompt.md
```diff
@@ -86,7 +86,7 @@ A string to be displayed as the prompt dialog title. Can also be [slotted](#titl
 
 ### confirmationText
 
-A string the user must type before the action button becomes enabled.
+A string the user must type before the action button becomes enabled. Pressing _Enter_ after typing in the confirmation text will trigger proceed action too.
```
adamdehaven, line 89:
```suggestion
A string the user must type before the action button becomes enabled. Pressing `Enter` in the confirmation text input will trigger the `proceed` action.
```

src/components/KPrompt/KPrompt.vue
```diff
@@ -139,6 +146,7 @@ const sanitizedAttrs = computed(() => {
 })
 
 const confirmationInput = ref<string>('')
+const displayErrorState = ref<boolean>(false)
```
adamdehaven, line 149: image

issue: if the error state is triggered, and then I close the prompt and re-open it, the error state persists. I believe this ref should be reset when closing/canceling

src/components/KPrompt/KPrompt.vue
```diff
@@ -112,11 +115,15 @@ const props = defineProps({
     type: Object as PropType<ModalAttributes>,
     default: () => ({}),
   },
+  errorMessage: {
+    type: String,
+    default: 'Please enter the text as indicated above.',
```
adamdehaven, line 120:
```suggestion
    default: 'You must enter the text to proceed.',
```

portikM: It's not just enter the text, it also has to match.. so how about _You must enter the text as indicated above to proceed._?

src/components/KPrompt/KPrompt.vue
portikM: Done

src/components/KPrompt/KPrompt.vue
adamdehaven: sgtm, except what about `"to confirm" instead of "proceed"?

portikM: Done

adamdehaven (approved):
lgtm

PR #2326, opened 2024-08-06
fix(kpop): bind aria-controls to trigger [KHCP-12832]

portikM:
# Summary

Add `aria-label` to dropdown trigger button

### DIFF
--- a/src/components/KDropdown/KDropdown.vue
+++ b/src/components/KDropdown/KDropdown.vue
@@ -29,6 +29,7 @@
             <KButton
               v-if="triggerButtonText"
               :appearance="appearance"
+              :aria-label="isToggled.value ? 'Close dropdown' : 'Open dropdown'"
               class="dropdown-trigger-button"
               data-testid="dropdown-trigger-button"
               :disabled="disabled"
--- a/src/components/KDropdown/KDropdown.vue
+++ b/src/components/KDropdown/KDropdown.vue
@@ -30,6 +30,7 @@
               v-if="triggerButtonText"
               :appearance="appearance"
               class="dropdown-trigger-button"
+              data-dropdown-trigger="true"
               data-testid="dropdown-trigger-button"
               :disabled="disabled"
             >
--- a/src/components/KDropdown/KDropdown.vue
+++ b/src/components/KDropdown/KDropdown.vue
@@ -29,7 +29,6 @@
             <KButton
               v-if="triggerButtonText"
               :appearance="appearance"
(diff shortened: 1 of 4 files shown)


src/components/KDropdown/KDropdown.vue
```diff
@@ -29,6 +29,7 @@
             <KButton
               v-if="triggerButtonText"
               :appearance="appearance"
+              :aria-label="isToggled.value ? 'Close dropdown' : 'Open dropdown'"
```
adamdehaven, line 32: I'm not 100% sure this should be hard-coded as the Kongponent has no idea as to the context in which it's being used (e.g. there may be far better text here the app would be able to provide on its own).

e.g. if the dropdown is to "pick a region" and already has a label or other contextual text around the dropdown, adding _more_ labels would actually be harmful to the experience

portikM: I think you're right here. Decided to go with a different approach

src/components/KPop/KPop.vue
```diff
@@ -287,11 +287,15 @@ onMounted(() => {
     // we don't set any other click event listeners on purpose to avoid conflict of event listeners
     document?.addEventListener('click', clickHandler)
 
-    if (popoverTrigger.value && props.trigger === 'hover') {
-      popoverTrigger.value.addEventListener('mouseenter', showPopover)
-      popoverTrigger.value.addEventListener('focus', showPopover)
-      popoverTrigger.value.addEventListener('mouseleave', hidePopover)
-      popoverTrigger.value.addEventListener('blur', hidePopover)
+    if (popoverTrigger.value) {
+      popoverTrigger.value.setAttribute('aria-controls', popoverId)
```
adamdehaven, line 291: what if the trigger already has this attribute set?

portikM: Changed the logic a bit (to make sure to bind it to the button and not the wrapper div)

src/components/KDropdown/KDropdown.vue
```diff
@@ -30,6 +30,7 @@
               v-if="triggerButtonText"
               :appearance="appearance"
               class="dropdown-trigger-button"
+              data-dropdown-trigger="true"
```
adamdehaven, line 33: if this isn't documented we're likely not going to know what it's for 😬

portikM: Added a comment in KPop (not where the attribute is defined but rather where it's being used - I'm guessing someone who wants to find out what it's for will try searching for references in the project)

PR #2334, opened 2024-08-12
fix(ktreelist): fix multiple emits

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-12985

1. Fixes multiple `selected` event emits on item select (screenshots)
2. Fixes weird dragging behaviour where after starting dragging a second click is needed to "release" dragged element

Before:
```
@selected="onSelected"

const onSelected = (item: TreeListItem) => {
  console.log('onSelected', item)
}
```
Screenshot 2024-08-12 at 3 08 27 PM

After:
Only 1 event is emitted


src/components/KTreeList/KTreeList.vue
```diff
@@ -111,6 +111,14 @@ const emit = defineEmits<{
   (event: 'selected', item: TreeListItem): void
 }>()
 
+const { useDebounce } = useUtilities()
+
+// use debounce function to avoid emitting multiple same events at once
+// even a 1ms debounce will do the trick
+const { debouncedFn: debouncedEmit } = useDebounce((item: TreeListItem) => {
+  emit('selected', item)
+}, 1)
```
adamdehaven, line 120: tbh this seems a bit like a hack. Can you explain the actual issue here?

portikM: I'm unsure what the cause of the problem is, this solution worked for me while I was working on something. I do agree that this might not be ideal solution, I'll move this back to draft and investigate this a bit more when I have time.

adamdehaven: I think investigating why a 1ms debounce is impacting selection is the way to go.

Is the DOM not yet updated for something? Then you could `await nextTick()` for a DOM cycle instead, etc.

adamdehaven (approved):

PR #2338, opened 2024-08-13
feat(ktableview): add component [KHCP-9499]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-9499

Adds KTableView component

Screenshot 2024-08-15 at 12 35 45 PM


docs/components/table-view.md
```diff
@@ -0,0 +1,1011 @@
...
+
+Boolean to control whether table should display hover state upon hovering rows. Defaults to `true`.
+
+### tablePreferences
+
+Table preferences object for persisting some configs.
+
+```ts
+interface TablePreferences {
+  pageSize?: number // the number of items to display per page
```
adamdehaven, line 304: If pagination isn’t included, what value does the page size serve here?

adamdehaven (commented):
Couple other thoughts:

- might be nice to be able to provide a non-generated `data-testid` and `id` for each row (optional) 
- An option to truncate cell content if over the width (if this isn’t already included)
- Sticky table headers?

docs/components/table-view.md
```diff
@@ -0,0 +1,1011 @@
...
+Notice that in the example above the _Username_ column is `sortable` and the _Email_ column is `hidable`.
+:::
+
+#### Reserved Data Keys
+
+:::warning NOTE
+This feature is only available in KTableView and not available in [KTable](/components/table).
+:::
+
+- `to` - if you wish to make the whole row a link, you can pass `to` parameter in each `data` entry. If an object is passed, it will be rendered as `<router-link>` component or otherwise if a string is passed, it will be rendered as an anchor tag. Clicking anywhere on table row will trigger route change. Refer to the example below for usage.
```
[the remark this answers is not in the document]
portikM: > this would require mutating table data

Since data manipulations are handled by the host app rather than KTableView component, I think this is acceptable?

Right-click support is a deal breaker and a long awaited feature for table component. In my opinion, a reserved key is the easiest way to turn the entire row into a link. I'm not sure how we could implement that using a slot like you suggested. Could you provide more details as to how you imagine I could make it work?

docs/components/table-view.md
portikM: This interface is shared between KTable and KTableView and [useTablePreferences composable](https://github.com/Kong/public-ui-components/tree/main/packages/core/core/src/useTablePreferences) uses it.

docs/components/table-view.md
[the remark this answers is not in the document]
adamdehaven: > Since data manipulations are handled by the host app rather than KTableView component, I think this is acceptable?

A component that is meant to display data shouldn’t restrict the data/column names for that data. 

I’m not sure anything I mentioned involves right-click support? The table row itself can be a link, but it shouldn’t be tied to a property of the data object the table is displaying. 

It could be a config on the cell itself, slot content, or maybe even a factory function for each row to generate the link.

docs/components/table-view.md
adamdehaven: So then can’t we utilize a separate interface that extends the base one…?

src/types/table.ts
```diff
@@ -13,6 +15,16 @@ export interface TablePreferences {
   columnVisibility?: Record<string, boolean>
 }
 
+export interface TableDataEntry {
+  [key: string]: any
```
adamdehaven, line 19: Is `any` appropriate here? I wouldn’t think an array or other object structures would be valid?

src/types/table.ts
```diff
@@ -13,6 +15,16 @@ export interface TablePreferences {
...
+  /** RouteLocationRaw or url string for row link */
+  to?: RouteLocationRaw | string
+  /** Target for row link */
+  target?: string
```
adamdehaven, line 23: suggestion: there only only some strings that are valid here; should we tighten the interface to only allowed values? 

(Should still discuss these “reserved” properties)

docs/components/table-view.md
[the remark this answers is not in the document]
portikM: How about a new prop `rowLink: Function`. It would take a function as a parameter and the reason for that is because we need a way for component to pass back row value so that the function returns the correct dynamic route.

docs/components/table-view.md
portikM: I think that would be confusing? I'm fine with removing references to these parameters from this docs section but I don't think we should create a separate interface because then we would have to create a new or alter the existing `useTablePreferences` composable. Since those fields are optional anyways, I think we can still use the original interface for both components.

src/types/table.ts
portikM: An array or object would probably not be rendered appropriately by default be we can still use slots to alter the way data is rendered in the cells. So yes, it can be `any`thing 😅

src/types/table.ts
portikM: Good call. Let's resolve the reserved properties discussion and then come back to this.

docs/components/table-view.md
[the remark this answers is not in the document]
portikM: Actually I think this is something we can introduce and support in both KTable and KTableView (I think we should maintain parity between the two components, to an extend).

docs/components/table-view.md
adamdehaven: You can use the same composable since the property is optional, but for the KTableVoew you can `Omit` the extra property from a similarly-named interface since it doesn't do anything. 

I understand wanting to keep a close parity between components, but inevitably over time they will both need to differ in places since they are fundamentally different under the hood.

adamdehaven:
> That can be done using `rowAttrs` [prop (both KTable and KTableView)](https://deploy-preview-2338--kongponents.netlify.app/components/table-view.html#rowattrs). Or did you mean default values provided by component itself?

Ok then that is sufficient.

src/styles/mixins/_tables.scss
```diff
@@ -0,0 +1,220 @@
...
+                min-width: 80px !important; // needs important because resizing will set min-width inline
+              }
+
+              &.sortable.has-tooltip {
+                min-width: 100px !important; // needs important because resizing will set min-width inline
+              }
+
+              .resize-handle {
+                cursor: col-resize;
+                height: v-bind("headerHeight");
```
Leopoldthecoder, line 100: `headerHeight` is deleted from the source code, so now `.resize-handle` can't be interacted with, which essentially breaks `resizeColumns`:

image

portikM: Good catch, thanks!

portikM: Pushed a fix

kaiarrowood:
Initial pass looks good to me 🚀

docs/components/table-view.md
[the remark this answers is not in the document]
portikM: So I decided to use `rowAttrs` prop for this. The prop takes a function that returns an object which have optional parameters `to` and `target` for making the whole row a link (see [docs](https://deploy-preview-2338--kongponents.netlify.app/components/table-view.html#row-links)). I think this is much better than including those keys in `data` prop. Lmk what you think.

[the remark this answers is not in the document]
adamdehaven: Utilizing `rowAttrs` here is a bit confusing to be honest. Turning a row into a link isn't really related to HTML attributes being passed to the row (e.g. like `data-testid`, etc.)

It's better than the previous iteration, but is there a reason there can't be a separate prop specifically for a row link?

src/types/table.ts
```diff
@@ -30,6 +34,15 @@ export interface TableHeader {
   useSortHandlerFunction?: boolean
 }
 
+export enum TableViewHeaderKeys {
+  ACTIONS = 'actions',
+}
+
+export interface TableViewHeader extends Omit<TableHeader, 'useSortHandlerFunction'> {
+  /** Must be unique for each column. If 'actions' value is provided, the column is treated as a column that only displays action dropdown menus for each row and no label (hideLabel: true) */
+  key: string | `${TableViewHeaderKeys}`
```
adamdehaven, line 43: question: why is any string acceptable here?

src/types/table.ts
```diff
@@ -50,3 +63,11 @@ export interface TableStatePayload {
   state: TableState
   hasData: boolean
 }
+
+export interface TableRowAttributes {
+  [key: string]: string | RouteLocationRaw | undefined
```
adamdehaven, line 68: issues:

1. Since you're binding the `to` property below specifically for row links, you don't need to also include the `RouteLocationRaw` here for the catch-all property type
2. undefined syntax
```suggestion
  [key: string]?: string
```

docs/components/table-view.md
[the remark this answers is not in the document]
portikM: I think it can be - I was just trying to avoid adding an extra arbitrary prop but maybe it's better to just add one rather than complicate existing ones.. is that your thinking here too?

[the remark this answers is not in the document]
adamdehaven: Yes I think since having "row links" is a net-new feature, adding it as a separate row-level "config" is ideal. So it can work similar to `rowAttrs` but something like `rowLink` instead

src/types/table.ts
adamdehaven, line 43: issue: this type will not resolve property if more than `'actions'` is provided in the `TableViewHeaderKeys` - I think just type to a `string` here.

src/components/KTableView/KTableView.vue
```diff
@@ -0,0 +1,896 @@
...
+                  :class="{ 'resized': resizingColumn === column.key }"
+                >
+                  <slot
+                    :column="getGeneric(column)"
+                    :name="getColumnSlotName(column.key)"
+                  >
+                    <span
+                      class="table-header-label"
+                      :class="{
+                        'sr-only': column.hideLabel || (column.key === TableViewHeaderKeys.ACTIONS && column.hideLabel !== false),
```
adamdehaven, line 141: since "actions" is special, can we just check for it in the component (you can define the const in here) rather than importing `TableViewHeaderKeys`?

The other places it's used won't evaluate correctly for types as-is

src/components/KTableView/KTableView.vue
```diff
@@ -0,0 +1,896 @@
...
+                    >
+                      <MoreIcon
+                        class="more-icon"
+                        decorative
+                      />
+                    </KButton>
+
+                    <template #items>
+                      <slot
+                        :name="`${TableViewHeaderKeys.ACTIONS}-items`"
```
adamdehaven, line 246: issue: should be `action-items` not `actions-items`

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
+# Table View
+
+Component for displaying data in table format.
+
+:::tip NOTE
+KTableView is a display-only component, it does not provide any data management capabilities like fetching, pagination, sorting or searching. If you are looking for a component that includes those functions out of the box, check out [KTable](/components/table).
```
adamdehaven, line 6:
```suggestion
KTableView does not handle data management capabilities like data fetching, functional pagination, sorting or searching. If you are looking for a component that integrates with the data layer, check out [KTable](/components/table).
```

Doesn't KTableView handle sorting though?

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+## Props
+
+### headers
+
+Array of objects that represent table columns along with some configurations that should apply to each of the columns (whether a column is sortable, hidable, etc).
+
+```ts
+interface TableViewHeader {
+  key: string | 'actions' // must be unique for each column, see Reserved Header Keys section for more information about 'actions' key value
```
adamdehaven, line 29: issue: I'm not sure `actions` should be in the interface here; see my other comment.

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+  label: string // visible column header text
+  sortable?: boolean // in a nutshell, this property defines whether sort icon should be displayed next to the column header and whether the column header will emit sort event upon clicking on it
+  hidable?: boolean // allow toggling column visibility
+  tooltip?: string // when provided, an info icon will be rendered next to the column label, upon hovering on which the tooltip will be revealed
+  hideLabel?: boolean // whether column header text should be hidden (only visible to screen readers)
+}
+```
+
+:::tip NOTE
+If at least one column is `hidable` in the table, KTableView will render a dropdown on the right of the table toolbar, directly above the table, which will provide an interface for showing/hiding columns to the user.
```
adamdehaven, line 39:
```suggestion
If at least one column is `hidable` in the table, KTableView will render a dropdown on the right of the table toolbar directly above the table, which will provide an interface for showing/hiding columns to the user.
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+:::
+
+For an example of `headers` prop usage please refer to [`data` prop documentation](#data) below.
```
adamdehaven, line 42: Update sentence for after moving `data` above `headers`

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+#### Reserved Header Keys
+
+:::warning NOTE
+This feature is only available in KTableView and not available in [KTable](/components/table).
+:::
```
adamdehaven, line 48:
```suggestion
```
I don't think we need these notes; there is no assumption that props are available in both components or that the component interfaces are the same.

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+- `actions` - the column is treated as a column that only displays action dropdown menus for each row and displays no label (as if `hideLabel` was `true`, however you can still display the label by setting `hideLabel` parameter to `false`). KViewTable will automatically render the trigger button with an icon, however you will need to pass dropdown items through the [`actions-items` slot](#actions-items).
```
adamdehaven, line 50:
```suggestion
- `actions` - the column displays an "Actions" `KDropdownMenu` button for each row and displays no label (as if `hideLabel` was `true`; set `hideLabel` to `true` to show the column header label). KTableView will automatically render the actions dropdown button with an icon, and then simply provide the dropdown items via the [`action-items` slot](#actions-items).
```

- Changed `actions-items` to `action-items`
- Changed `KViewTable` to `KTableView`

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+### data
+
+Data to be rendered in the table. Accepted type is array of objects where each key should correspond to a certain `key` in [`headers` prop](#headers).
```
adamdehaven, line 54:
```suggestion
Data to be rendered in the table. Accepted interface is an array of objects where each property key should have a corresponding `key` in the [`headers` prop](#headers).
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+```ts
+type TableData = Array<Record<string, any>>
```
adamdehaven, line 57: nit: use our regular format
```suggestion
type TableData = Record<string, any>[]
```

src/types/table.ts
```diff
@@ -13,6 +15,8 @@ export interface TablePreferences {
   columnVisibility?: Record<string, boolean>
 }
 
+export type TableData = Array<Record<string, any>>
```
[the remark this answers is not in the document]
adamdehaven: issue: `TableData` interface name is a bit misleading here since we have multiple table components, but only applies to this new component. This should be renamed `TableViewData` and update the docs/usage accordingly.

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+</script>
+```
+
+:::tip NOTE
+Notice that in the example above the _Username_ column is `sortable` and the _Email_ column is `hidable`.
+:::
+
+### loading
+
+Boolean to control whether the component should display the loading state.
```
adamdehaven, line 180: Add "Defaults to `false`"

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+/>
+```
+
+### rowHover
+
+Boolean to control whether table should display hover state upon hovering rows. Defaults to `true`.
+
+### tablePreferences
+
+Table preferences object for persisting some configs.
```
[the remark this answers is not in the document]
adamdehaven: question: why do we say "persisting" here?

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+<KTableView
+  :cell-attrs="({ headerKey, row }) => { return { 'data-testid': `column-${headerKey}-row-${row.id}` } }"
+  :data="tableData"
+  :headers="headers"
+/>
+```
+
+### maxHeight
+
+Pass a number (px) you want to limit table height to. If the height table exceeds the specified number, it will be scrollable. Table header is a `position: sticky;` element and will always be visible.
```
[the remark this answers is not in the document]
adamdehaven: question: is the table header **always** sticky?

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+```html
+<KTableView
+  error
+  :data="tableData"
+  :headers="headers"
+/>
+```
+
+Set the following properties to customize error state:
```
adamdehaven, line 457:
```suggestion
Set the following properties to customize the error state:
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+- `errorStateTitle` - Title text for error state
+- `errorStateMessage` - Message for error state
+- `errorStateActionMessage` - Button text for error state action
+- `errorStateActionRoute` - Route for error state action
+
+A `error-action-click` event is fired when error state action button is clicked.
```
adamdehaven, line 464:
```suggestion
- `errorStateTitle` - Title text for the error state
- `errorStateMessage` - Message for the error state
- `errorStateActionMessage` - Button text for the error state action
- `errorStateActionRoute` - Route for the error state action

A `error-action-click` event is fired when error state action button is clicked.
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+  :headers="headers"
+/>
+```
+
+Set the following properties to customize error state:
+
+- `errorStateTitle` - Title text for error state
+- `errorStateMessage` - Message for error state
+- `errorStateActionMessage` - Button text for error state action
+- `errorStateActionRoute` - Route for error state action
```
adamdehaven, line 462: issue/question: is `errorStateActionRoute` a new prop for this component? What if it's an external link? the interface defined is just `Object | string` this seems wrong?

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+### Column Header
+
+You can slot in your custom content into each column header. For that, use column `key` value prefixed with `column-*` like in the example below.
+
+Slot props:
+* `column` - column header object
+
+:::tip NOTE
+Header slot container is a `display: flex;` element that takes care of spacing between slotted items so you can slot in your items without having to worry about adding margin between them.
+:::
```
adamdehaven, line 499: nit: I'm not sure this is needed; we don't typically call out CSS implementation in the docs; it should be obvious here.
```suggestion
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+>
+  <template #column-email="{ column }">
+    {{ column.label }} <KBadge>Beta</KBadge>
+  </template>
+</KTableView>
+```
+
+### Cell
+
+You can slot in each individual cell content. Each cell slot is name after the `key` it corresponds to.
```
adamdehaven, line 523:
```suggestion
You can provide each individual cell's content via slot. Each cell slot is named after the header `key` it corresponds to.
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+Slot props:
+* `row` - table row object
+* `rowKey` - table row index
+* `rowValue` - the cell value
+
+:::warning NOTE
+This slot is not supported for [`actions` column](#reserved-header-keys).
```
adamdehaven, line 531:
```suggestion
This slot is not supported for the [`actions` column](#reserved-header-keys).
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+>
+  <template #email="{ rowValue }">
+    <KCopy :text="rowValue" />
+  </template>
+</KTableView>
+```
+
+### Header Tooltip
+
+If you want to utilize HTML in the column header's tooltip, use the slot. Similar to column header slot, use column `key` value prefixed with `tooltip-*` like in the example below.
```
adamdehaven, line 556:
```suggestion
Utilize HTML in the column header's tooltip by utilizing this slot. Similar to the column header slot, it uses the column `key` value prefixed with `tooltip-*` as shown in the example below.
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+>
+  <template #tooltip-email>
+    HubSpot Id: <code>8576925e-d7e0-4ecd-8f14-15db1765e69a</code>
+  </template>
+</KTableView>
+```
+
+### toolbar
+
+Toolbar is rendered directly above the table. Useful for slotting table controls like search or filter fields.
```
adamdehaven, line 583:
```suggestion
The toolbar is rendered directly above the table and is useful for providing table controls like search or filter fields.
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+:::tip NOTE
+Toolbar slot container is a `display: flex;` element that takes care of spacing between slotted items so you can slot in your items without having to worry about adding margin between them.
+:::
```
adamdehaven, line 587: nit: I don't think we need to call this out
```suggestion
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+### empty-state-action-icon
+
+Slot for icon to be displayed in front of action button text in empty state. See [empty state](#empty) section for example of usage of this slot.
+
+### error-state
+
+Slot content to be displayed when in error state.
+
+### actions-items
```
adamdehaven, line 635: see other comments, rename to `action-items`

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+:::warning NOTE
+This slot is only available in KTableView and not available in [KTable](/components/table).
+:::
```
adamdehaven, line 639: Warning isn't needed, this is a different component
```suggestion
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+Slot for passing actions dropdown items. See [KDropdownItem component docs](/components/dropdown#kdropdownitem) for details.
```
adamdehaven, line 641:
```suggestion
Slot for passing action dropdown items. See [KDropdownItem component docs](/components/dropdown#kdropdownitem) for details.
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+Slot props:
+* `row` - table row object
+* `rowKey` - table row index
+
+:::tip NOTE
+This slot is only available when `actions` header hey is present in [`headers` prop](#reserved-header-keys).
```
adamdehaven, line 648:
```suggestion
This slot is only available when the `actions` header key is present in [`headers`](#reserved-header-keys).
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+    </KDropdownItem>
+  </template>
+</KTableView>
+```
+
+### after
+
+:::warning NOTE
+This slot is only available in KTableView and not available in [KTable](/components/table).
+:::
```
adamdehaven, line 691: Warning isn't needed, this is a different component
```suggestion
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+Slot for content that should be rendered below the table. For example, could be used for displaying pagination component.
```
adamdehaven, line 693: Is this slot still needed now that we're baking in pagination component?

docs/components/table-view.md
```diff
@@ -0,0 +1,1016 @@
...
+
+Emitted when empty state action button is clocked.
+
+### error-action-click
+
+Emitted when error state action button is clicked.
+
+### update:table-preferences
+
+Fired when the user performs sorting, resizes columns or toggles column visibility. Event payload is object of type `TablePreferences` interface (see [`tablePreferences` prop](#tablepreferences) for details).
```
adamdehaven, line 815:
```suggestion
Emitted when the user performs sorting, resizes columns or toggles column visibility. Event payload is object of type `TablePreferences` interface (see [`tablePreferences` prop](#tablepreferences) for details).
```

docs/components/table-view.md
adamdehaven, line 464: Replace all instances of `fired` with `emitted`

adamdehaven (changes requested):

src/types/table.ts
portikM: This is obsolete now

docs/components/table-view.md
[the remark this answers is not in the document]
portikM: I suppose that's driven by our own use case?

docs/components/table-view.md
portikM: Now that pagination is included, I suppose we don't need a new interface?

docs/components/table-view.md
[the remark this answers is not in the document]
portikM: Yes. Doesn't matter whether the height is limited or now - that makes the logic simpler

docs/components/table-view.md
portikM: It's not new - migrated over from KTable

docs/components/table-view.md
portikM: Agreed

docs/components/table.md
```diff
@@ -2,6 +2,10 @@
 
 Component that takes care of fetching and rendering data in a table format.
 
+:::tip NOTE
+If you are looking for a display-only component for rendering data in table format, check out [KTableView](/components/table-view).
+:::
+
```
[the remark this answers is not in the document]
portikM: Is it still the case? Don't see the error locally nor on preview - might have fixed in one of the commits I pushed today?

src/components/KTableView/KTableView.vue
```diff
@@ -0,0 +1,896 @@
...
+            <tr
+              ref="headerRow"
+              :class="{ 'is-scrolled': isScrolled }"
+            >
+              <th
+                v-for="(column, index) in visibleHeaders"
+                :key="`table-${tableId}-headers-${index}`"
+                :aria-sort="column.key === sortColumnKey ? (sortColumnOrder === 'asc' ? 'ascending' : 'descending') : undefined"
+                class="table-headers"
+                :class="getHeaderClasses(column, index)"
```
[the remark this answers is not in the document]
portikM: We only need index here to check if the it's the last column or not (in which case, we don't show the column resize handle, given that resizing is enabled)

src/components/KTableView/KTableView.vue
```diff
@@ -0,0 +1,896 @@
...
+          :message="emptyStateMessage"
+          :title="emptyStateTitle"
+        >
+          <template
+            v-if="!!emptyStateActionMessage"
+            #action
+          >
+            <KButton
+              :appearance="emptyStateButtonAppearance"
+              :data-testid="getTestIdString(emptyStateActionMessage)"
```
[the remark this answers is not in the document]
portikM: Done

docs/components/table-view.md
portikM: No, it only emits an event (can't handle sorting since `data` prop could only be accepting a paginated piece of otherwise larger dataset).

docs/components/table-view.md
adamdehaven: Does it utilize _all_ of the same properties? (are there any the new component does not utilize?)

docs/components/table-view.md
portikM: Yes - part of component brief.

docs/components/table-view.md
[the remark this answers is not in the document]
adamdehaven: I'm wondering if we may end up with some issues with the sticky header in Konnect, but I can't remember what area I was thinking this could be an issue

docs/components/table-view.md
portikM: Nope, all of them

docs/components/table-view.md
adamdehaven: It's a very loose interface 🤔

docs/components/table-view.md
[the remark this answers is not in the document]
portikM: It's been sticky in KTable since... at least reskin but maybe forever

docs/components/table-view.md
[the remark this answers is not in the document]
adamdehaven: Do the preferences actually persist with the built-in functionality? If not, I think we should reword even if we do something different in Konnect

docs/components/table.md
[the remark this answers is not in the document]
adamdehaven: Seems resolved now 🤷🏼

docs/components/table-view.md
adamdehaven: what would a user slot in below the pagination?

docs/components/table-view.md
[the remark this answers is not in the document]
adamdehaven: ah 😅

docs/components/table-view.md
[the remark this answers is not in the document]
portikM: Removed the word "persisting" from this sentence - still sense and isn't contradictory.

docs/components/table-view.md
portikM: Agreed, pushed an improvement

docs/components/table-view.md
portikM: Maybe pagination that's not KPagination? Can't think of our internal use case though. Remove the slot?

portikM: No we have a `hidePagination` prop

src/components/KTableView/KTableView.vue
```diff
@@ -0,0 +1,938 @@
...
+   */
+  emptyStateMessage: {
+    type: String,
+    default: 'There is no data to display.',
+  },
+  /**
+   * A prop to pass in a custom empty state action route
+   */
+  emptyStateActionRoute: {
+    type: [Object as PropType<RouteLocationRaw>, String],
```
adamdehaven, line 373: I don't believe this syntax is valid

adamdehaven: Maybe this?
```suggestion
    type: [Object, String] as PropType<RouteLocationRaw | string>,
```

src/components/KTableView/KTableView.vue
```diff
@@ -0,0 +1,938 @@
...
+   */
+  errorStateMessage: {
+    type: String,
+    default: 'Data cannot be displayed due to an error.',
+  },
+  /**
+   * A prop to pass in a custom error state action route
+   */
+  errorStateActionRoute: {
+    type: [Object as PropType<RouteLocationRaw>, String],
```
adamdehaven, line 416: Same comment as above on syntax

docs/components/table-view.md
adamdehaven: So if a user wanted to utilize their own KPagination (or something) instead of the built-in, do they need a slot, or can they just place it under the table? 

If they don't need the slot, let's just remove it.

adamdehaven:
I believe @DariaYeremina is going to review as well. Before merging here, let's pull the preview package into the new Portal project and see if this new component can easily be swapped in for the Credentials table and satisfies our basic use-case

src/components/KTableView/KTableView.vue
portikM: Fixed

DariaYeremina:
> I believe @DariaYeremina is going to review as well. Before merging here, let's pull the preview package into the new Portal project and see if this new component can easily be swapped in for the Credentials table and satisfies our basic use-case

I reviewed it and it can be easely adapted to the Credentials table

docs/components/table-view.md
```diff
@@ -0,0 +1,1039 @@
...
+Component for displaying data in table format.
+
+:::tip NOTE
+KTableView does not handle data management capabilities like data fetching, functional pagination, sorting or searching. If you are looking for a component that integrates with the data layer, check out [KTable](/components/table).
+:::
+
+<KTableView
+  :data="basicData"
+  :headers="basicHeaders()"
+/>
```
adamdehaven, line 12: issue: the pagination on this example is not functional. If that's expected, we should remove the pagination (need to check the rest of the examples as well)

Also, several of the pagination counts say `1 to 0 of 0`

docs/components/table-view.md
```diff
@@ -0,0 +1,1039 @@
...
+
+:::tip NOTE
+If at least one column is `hidable` in the table, KTableView will render a dropdown on the right of the table toolbar directly above the table, which will provide an interface for showing/hiding columns to the user.
+:::
+
+For an example of `headers` prop usage please refer to [`data` prop documentation](#data) below.
+
+#### Reserved Header Keys
+
+- `actions` - the column displays an actions [KDropdow](/components/dropdown) button for each row and displays no label (as if `hideLabel` was `true`; you can set `hideLabel` parameter to `false` to show the label). KTableView will automatically render the actions dropdown button with an icon and you simply need to provide dropdown items via the [`action-items` slot](#action-items).
```
adamdehaven, line 46: typo
```suggestion
- `actions` - the column displays an actions [KDropdown](/components/dropdown) button for each row and displays no label (as if `hideLabel` was `true`; you can set `hideLabel` parameter to `false` to show the label). KTableView will automatically render the actions dropdown button with an icon and you simply need to provide dropdown items via the [`action-items` slot](#action-items).
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1039 @@
...
+/>
+```
+
+### rowHover
+
+Boolean to control whether table should display hover state upon hovering rows. Defaults to `true`.
+
+### tablePreferences
+
+Table preferences object for some configuration options.
```
adamdehaven, line 234: docs: this sentence doesn't accurately describe what this prop is for, or how to use it

docs/components/table-view.md
```diff
@@ -0,0 +1,1039 @@
...
+
+```ts
+interface TablePreferences {
+  pageSize?: number // the number of items to display per page
+  sortColumnKey?: string // the column to sort by's `key` defined in the table headers
+  sortColumnOrder?: SortColumnOrder // the order by which to sort the column, one of `asc` or `desc`
+  columnWidths?: Record<string, number> // the customized column widths, if resizing is allowed
+  columnVisibility?: Record<string, boolean> // column visibility, if visibility is toggleable
```
adamdehaven, line 242: nit: can you move the comments above each property so you can read them without having to scroll horizontally?

docs/components/table-view.md
```diff
@@ -0,0 +1,1039 @@
...
+  offsetPreviousButtonDisabled?: boolean
+  offsetNextButtonDisabled?: boolean
+}
+```
+
+<KTableView
+  :pagination-attributes="{ totalCount: 100, currentPage: 3 }"
+  :data="basicData"
+  :headers="basicHeaders()"
+/>
```
adamdehaven, line 390: issue: This example isn't functional; the current page never updates and changing the page size doesn't do anything

docs/components/table-view.md
```diff
@@ -0,0 +1,1039 @@
...
+      :total-count="tableData.length"
+      @page-change="onPageChange"
+    />
+  </template>
+</KTableView>
+```
+
+## Events
+
+### Row Events
```
adamdehaven, line 736: issue: the row events and cell events examples are pretty confusing; there is so much going on in each cell I'm not sure what the example is trying to show me. Can we simplify?

adamdehaven (changes requested):

docs/components/table-view.md
portikM: Added hidePagination everywhere where pagination is not relevant in the example.

docs/components/table-view.md
portikM: I expanded this sentence a little, not sure if it became clearer.

docs/components/table-view.md
portikM: Done

docs/components/table-view.md
portikM: Fixed

docs/components/table-view.md
portikM: I mean they are but I think it's one of those things that you need to know what you're looking for? The button, the input field and an external link are there to just demonstrate that you can both click on row/cell and other interactive elements within the table and handle those clicks separately.

portikM: In other words, it doesn't look confusing once you know what too look for in that example but I also don't know how to make it less confusing for those who don't know what exactly they are looking for

docs/components/table-view.md
adamdehaven: But if it wasn't functional, should the table know not to show it or something?

portikM: Eh so it can't be fully "smart" like that but I've pushed a change so that it relies on paginationAttributes a bit more to determine if it needs to show pagination (hide if `props.paginationAttributes.totalCount <= props.data.length`)

adamdehaven: It now looks like the pagination is now not visible on any table examples on the docs page; I'm not sure that's intended/desirable either?

docs/components/table-view.md
adamdehaven: The pagination isn't visible now?

docs/components/table-view.md
adamdehaven: I don't think the "rowEvents" and "cellEvents" examples should look exactly the same - I can't tell what the difference is between the two examples

image

We should rework this

portikM: Oh, that's a good point, they don't have to be identical. Changed them up a bit and added a sentence in front of the example to better clarify what the example is demonstrating.

docs/components/table-view.md
portikM: Sorry, made it visible & functional

docs/components/table-view.md
portikM: It's hidden on all examples because it's irrelevant everywhere except for `paginationAttributes` prop example (there it is visible & functional)

docs/components/table-view.md
```diff
@@ -0,0 +1,1084 @@
...
+
+`@row:{event}` - returns the `Event`, the row item, and the type. `row-click` event is emitted whenever a row is clicked and the row click event handler is emitted, returns the row `data`.
+
+To avoid firing row clicks by accident, the row click handler ignores events coming from `a`, `button`, `label`, `input`, and `select` elements (unless they have the `disabled` attribute). As such click handlers attached to these element types do not require stopping propagation via `@click.stop`.
+
+The table in the example below contains buttons, inputs and links to demonstrate how KTableView handles clicks on different interactive elements within the table as well as clicks on its rows.
```
adamdehaven, line 753: Let's add a small playground suggestion:
```suggestion
The table in the example below contains buttons, inputs and links to demonstrate how KTableView handles clicks on different interactive elements within the table as well as clicks on its rows.

Try clicking below on the table row, and then within elements inside the table cells.
```

docs/components/table-view.md
```diff
@@ -0,0 +1,1084 @@
...
+```html
+<KTableView
+  :data="tableData"
+  :headers="headers"
+  @row:click="onRowClick"
+  hide-pagination
+/>
+```
+
+### Cell Events
```
adamdehaven, line 789: When cell clicks are enabled in the table, should the cursor be a `pointer` on hover for cells with click events? Unsure how complex that CSS would be, but nice if possible.

portikM: That's actually easy to add, done

adamdehaven (approved):
lgtm

PR #2339, opened 2024-08-13
fix(ktreelist): type interfaces

adamdehaven:
# Summary

Update the type interfaces for `KTreeList`


kaiarrowood (approved):
LGTM! Everything functioning as expected

PR #2342, opened 2024-08-15
feat(kradio): support setting to null

kaiarrowood:
# Summary

Support setting the `modelValue` to `null` to unset the selection.

### DIFF
--- a/src/components/KRadio/KRadio.vue
+++ b/src/components/KRadio/KRadio.vue
@@ -99,7 +99,7 @@ const props = defineProps({
    * Sets whether or not radio is selected
    */
   modelValue: {
-    type: [String, Number, Boolean, Object],
+    type: [String, Number, Boolean, Object, null] as PropType<string | number | boolean | object | null>,
     required: true,
   },
   /**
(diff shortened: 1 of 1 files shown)


src/components/KRadio/KRadio.vue
```diff
@@ -99,7 +99,7 @@ const props = defineProps({
    * Sets whether or not radio is selected
    */
   modelValue: {
-    type: [String, Number, Boolean, Object],
+    type: [String, Number, Boolean, Object, null] as PropType<string | number | boolean | object | null>,
```
adamdehaven, line 102: Why null instead of undefined (which I believe is the default)?

kaiarrowood: The admin api doesn't recognize `undefined`, it's got to be explicitly `null`. I'm following [this pattern](https://vuejs.org/guide/components/props.html#nullable-type). I know this is a bit opinionated but VFG can't use KRadio otherwise 🤔

adamdehaven (approved):
Lgtm

PR #2345, opened 2024-08-16
fix(*): fixable eslint-plugin-vuejs-accessibility issues [KHCP-11024]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-11024

### DIFF
--- a/src/components/KMultiselect/KMultiselectItem.vue
+++ b/src/components/KMultiselect/KMultiselectItem.vue
@@ -1,14 +1,12 @@
 <template>
   <div
     :key="item.key"
+    :aria-selected="item.selected"
     class="multiselect-item"
     :data-testid="`multiselect-item-${item.value}`"
+    role="option"
   >
-    <div
-      :aria-selected="item.selected"
-      class="multiselect-item-container"
-      role="option"
-    >
+    <div class="multiselect-item-container">
       <button
         :class="{ selected: item.selected }"
         :disabled="item.disabled === true ? true : undefined"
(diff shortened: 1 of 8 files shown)


src/components/KSelect/KSelectItem.vue
```diff
@@ -14,6 +14,7 @@
         :disabled="item.disabled === true ? true : undefined"
         type="button"
         :value="item.value"
+        @click="handleClick"
```
adamdehaven, line 17: question: how was this working previously if this is required?

portikM: It was previously bound to div on line 6. I moved it to button

src/components/KMultiselect/KMultiselectItem.vue
```diff
@@ -6,6 +6,7 @@
     role="listitem"
   >
     <div
+      :aria-selected="item.selected"
```
adamdehaven, line 9: question: should this go on the parent?

src/components/KSelect/KSelectItem.vue
```diff
@@ -3,9 +3,9 @@
     :key="item.key"
     class="select-item"
     :data-testid="`select-item-${item.value}`"
-    @click="handleClick"
   >
     <div
+      :aria-selected="item.selected"
```
adamdehaven, line 8: question: should this go on the parent?

src/components/KMultiselect/KMultiselectItem.vue
portikM: No, it belongs on the `role="option"` element. Same in KSelectItem

adamdehaven (approved):

PR #2352, opened 2024-08-21
fix(ktable): remove next page clicked logic [KHCP-13030]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-13030

Fixes bug when KTable sets page to 1 unexpectedly


src/components/KTable/KTable.vue
```diff
@@ -815,12 +814,6 @@ const fetchData = async () => {
...
     if (!res.pagination?.offset) {
       offset.value = null
-
-      // reset to first page if no pagiantion data is returned unless the "next page" button was clicked
-      // this will ensure buttons display the correct state for cases like search
-      if (!nextPageClicked.value) {
-        page.value = 1
-      }
     } else {
       offset.value = res.pagination.offset
```
2eha0, line 818: I don’t quite understand why we updated the offset here and then [fixed it in the fetcher](https://github.com/Kong/public-ui-components/pull/1577/commits/bb3261dabf5a43b691e1e3b171343f6fa3d94a68#diff-3566b8fd0bfd0a04b9e6714bbcd61a580e818d8d38e9b0c3e2d010c1a1f3084aR42). Wouldn’t it make more sense to save `res.pagination.offset` into a variable like `nextPageOffset` and update the offset only when moving to the next page?

2eha0: I don't have permission to push a branch, I changed it locally and it seems to work
image

PR #2355, opened 2024-08-23
fix(ktable): store offset of next page in nextOffset

Leopoldthecoder:
# Summary

This PR includes:
- changes in https://github.com/Kong/kongponents/pull/2352
- store the offset of the next page in a new ref `nextOffset`, so that the original `offset` ref will always equal the offset of the current page.
- send the user to the first page when the last item on the last page is deleted
- test cases


adamdehaven (commented):
The changes here look correct to me; we need to thoroughly test both with offset and cursor pagination to ensure all scenarios are functional

portikM (commented):
These changes look good to me. I've pulled these changes into konnect-ui-apps for sanity check: https://github.com/Kong/konnect-ui-apps/pull/4037. I'll tag a few people to check offset and cursor pagination functionality in different apps just to make sure we don't miss anything.

portikM (approved):
LGTM 🚀

PR #2357, opened 2024-08-26
fix(kselect): allow arrow keys navigation [khcp-12967]

aanchalm01:
# Summary
Addresses:
[https://konghq.atlassian.net/browse/KHCP-12967](https://konghq.atlassian.net/browse/KHCP-12967)

This PR introduces the ability to navigate through the select options using the arrow keys. Users can now:

- Use the up and down arrow keys to move through the list of options.
- Press Enter or Space to select the highlighted option.

### DIFF
--- a/src/components/KSelect/KSelectItem.vue
+++ b/src/components/KSelect/KSelectItem.vue
@@ -1,20 +1,19 @@
 <template>
   <div
     :key="item.key"
+    :aria-selected="item.selected"
     class="select-item"
     :data-testid="`select-item-${item.value}`"
+    role="option"
     tabindex="0"
-    @click="handleClick"
   >
-    <div
-      class="select-item-container"
-      role="option"
-    >
+    <div class="select-item-container">
       <button
         :class="{ selected: item.selected }"
         :disabled="item.disabled === true ? true : undefined"
         type="button"
         :value="item.value"
+        @click="handleClick"
       >
         <span class="select-item-label">
           <slot name="content">{{ item.label }}</slot>
--- a/src/components/KSelect/KSelectItem.vue
+++ b/src/components/KSelect/KSelectItem.vue
@@ -1,18 +1,20 @@
 <template>
   <div
     :key="item.key"
-    :aria-selected="item.selected"
     class="select-item"
     :data-testid="`select-item-${item.value}`"
-    role="option"
+    tabindex="0"
+    @click="handleClick"
   >
-    <div class="select-item-container">
+    <div
+      class="select-item-container"
+      role="option"
+    >
       <button
         :class="{ selected: item.selected }"
         :disabled="item.disabled === true ? true : undefined"
         type="button"
         :value="item.value"
-        @click="handleClick"
       >
         <span class="select-item-label">
           <slot name="content">{{ item.label }}</slot>
--- a/src/components/KSelect/KSelectItems.vue
+++ b/src/components/KSelect/KSelectItems.vue
@@ -2,7 +2,9 @@
   <KSelectItem
     v-for="item in nonGroupedItems"
     :key="item.key"
+    ref="focusableItems"
     :item="item"
+    @keydown="($event: KeyboardEvent) => handleKeyDown($event, item)"
     @selected="handleItemSelect"
   >
     <template #content>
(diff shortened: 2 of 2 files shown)


src/components/KSelect/KSelectItem.vue
```diff
@@ -1,18 +1,20 @@
 <template>
   <div
     :key="item.key"
-    :aria-selected="item.selected"
     class="select-item"
     :data-testid="`select-item-${item.value}`"
-    role="option"
```
adamdehaven, line 7: I believe this is the case of a bad merge? @portikM didn't you move this role attribute, etc. to this parent?

adamdehaven (changes requested):

src/components/KSelect/KSelectItems.vue
```diff
@@ -59,7 +63,87 @@ const emit = defineEmits<{
...
+    .filter((el: HTMLElement) => {
+    // Check if the element has a disabled button among its children
+      const hasDisabledButton = Array.from(el.querySelectorAll('button')).some(button => button.disabled)
+      // Exclude the element if it contains a disabled button
+      return !hasDisabledButton
+    })
+
+  const currentIndex = elements.indexOf(activeElement)
+
+  if (['ArrowUp', 'ArrowDown', 'Tab', ' ', 'Enter'].includes(event.key)) {
```
adamdehaven, line 89: question: shouldn't we be using `code` here instead of `key`?

adamdehaven (commented):
Tabbing backwards with `Shift + Tab` is now broken with the changes in the PR.

src/components/KSelect/KSelectItem.vue
aanchalm01: let me fix the bad rebase; for my changes I just need tabindex="0" **change** for this file.

src/components/KSelect/KSelectItems.vue
```diff
@@ -59,7 +63,87 @@ const emit = defineEmits<{
...
-const handleItemSelect = (item: SelectItem) => emit('selected', item)
+const handleItemSelect = (item: SelectItem) => {
+  return emit('selected', item)
+}
+
+// Ref to store focusable items
+const focusableItems = ref<HTMLElement[]>([])
+
+// Function to handle the keydown event
+const handleKeyDown = (event: KeyboardEvent, item: SelectItem) => {
```
adamdehaven, line 74: (not sure where to place this comment)
Pressing "Enter" to select an element does not close the dropdown

src/components/KSelect/KSelectItem.vue
aanchalm01: fixed.

src/components/KSelect/KSelectItem.vue
```diff
@@ -5,6 +5,7 @@
     class="select-item"
     :data-testid="`select-item-${item.value}`"
     role="option"
+    tabindex="0"
```
adamdehaven, line 8: @portikM this will technically steal the focus away from the internal button which I believe conflicts with your previous changes?

src/components/KSelect/KSelectItems.vue
aanchalm01: We can, wasnt sure that we have a preference.

adamdehaven: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

> The KeyboardEvent.code property represents a physical key on the keyboard (as opposed to the character generated by pressing the key). In other words, this property returns a value that isn't altered by keyboard layout or the state of the modifier keys.

adamdehaven: Let me check something

adamdehaven: Ignore this for now

src/components/KSelect/KSelectItem.vue
portikM: Yes, we don't need `tabindex` here as we have button element to play role of interactive element

src/components/KSelect/KSelectItems.vue
aanchalm01: Can you recheck? I have tried both Space/Enter key and it closes as expected

portikM (changes requested):
Logic-wise code looks correct but very lengthy for what is a minor feature.. let me try something to see if we can simplify this a bit, please pause on this PR til EOD 🙏

PR #2358, opened 2024-08-26
fix(kselect): arrow down navigation [KHCP-12967]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-12967

Adds arrow up and down navigation through items in KSelect

### DIFF
--- a/src/components/KSelect/KSelectItem.vue
+++ b/src/components/KSelect/KSelectItem.vue
@@ -13,7 +13,8 @@
         type="button"
         :value="item.value"
         @click="handleClick"
-        @keydown="onKeyPress"
+        @keydown.down.prevent="$emit('arrow-down')"
+        @keydown.up.prevent="$emit('arrow-up')"
       >
         <span class="select-item-label">
           <slot name="content">{{ item.label }}</slot>
@@ -50,16 +51,6 @@ const handleClick = (e: MouseEvent): void => {
   }
 
   emit('selected', props.item)
-}
-
-const onKeyPress = (event: any) => {
-  if (event.code === 'ArrowDown') {
-    event.preventDefault()
-    emit('arrow-down')
-  } else if ((event.code === 'ArrowUp')) {
-    event.preventDefault()
-    emit('arrow-up')
-  }
 }
 </script>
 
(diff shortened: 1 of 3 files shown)


src/components/KSelect/KSelectItems.vue
```diff
@@ -67,6 +73,39 @@ const groups = computed((): string[] =>
...
+
+const setFocus = () => {
+  if (kSelectItem.value) {
+    kSelectItem.value[0]?.$el?.querySelector('button').focus()
+  }
+}
+
+const shiftFocus = (key: SelectItem['key'], direction: 'down' | 'up') => {
+  const index = props.items.findIndex(item => item.key === key)
+  if (index === -1) return // Exit if the item is not found
```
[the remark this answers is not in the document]
aanchalm01: Done

src/components/KSelect/KSelectItem.vue
```diff
@@ -13,6 +13,7 @@
         type="button"
         :value="item.value"
         @click="handleClick"
+        @keydown="onKeyPress"
```
portikM, line 16: @aanchalm01 I think here we can have two events `@keydown.up.prevent="$emit('arrow-up')"` and `@keydown.down.prevent="$emit('arrow-down')"` - this way we don't even need the `onKeyPress` function

Docs: https://vuejs.org/guide/essentials/event-handling#key-modifiers

aanchalm01: updated.

sichvoge:
@aanchalm01 Does this PR also introduce selecting items via keyboard? So once I use the up and down arrow keys to navigate to an item, can I press Enter or Space to select it?

portikM:
> https://deploy-preview-2358--kongponents-sandbox.netlify.app

Yes, selecting an item via keyboard is now supported (that was introduced in [PR](https://github.com/Kong/kongponents/pull/2345/files#diff-5b3ceae168aa63b6d9ff276027bbcc9e1606b76f7262f3c0921b2bcacc52ff40) that was merged earlier last week).

portikM (commented):
These changes look good to me both code-wise and functionally but I can't approve since I'm a collaborator on the PR.

kaiarrowood (approved):
Everything looks good.

One non-blocking nit: it's a little confusing that hitting the `up` arrow when my focus is on the select input drops my focus into the drop down. I would expect this for `down` press, but not for `up` press. I don't think it's a blocking problem so approving.

PR #2375, opened 2024-09-06
fix(ktable, ktableview): `resizeColumns` should ignore right clicks

Justineo:
# Summary

The column resizer should ignore right clicks.

Before:

Sep-06-2024 17-46-49

After:

Sep-06-2024 17-48-23


adamdehaven (approved):
lgtm

PR #2376, opened 2024-09-06
feat(ktableview): add expandable rows feature [KHCP-13108]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-13108

Adds expandable rows feature in KTableView

Changes:
* add `expandableRows` prop (`boolean`) to enable expandable rows in the table
* add `hideHeader` prop (`boolean`) to hide table header in nested tables
* add `nested` prop (`boolean`) to disable some features in nested tables
* add `row-expanded` slot for passing expandable content

Custom content:
Screenshot 2024-09-11 at 11 15 01 AM

Nested table with it's own header:
Screenshot 2024-09-11 at 11 16 07 AM

Nested table with hidden header:
Screenshot 2024-09-11 at 11 16 44 AM


docs/components/table-view.md
```diff
@@ -922,6 +1093,47 @@ Slot props:
 </KTableView>
 ```
 
+### row-expanded
```
adamdehaven, line 1096: Is there an event that is emitted when a row is expanded, before the slot content is actually shown?

We likely would want to do some data fetching once a user clicks to expand a row, before the row is actually expanded

docs/components/table-view.md
```diff
@@ -510,6 +510,48 @@ interface TablePaginationAttributes {
 
 A boolean to hide pagination element (defaults to `false`).
 
+### expandableRows
```
adamdehaven, line 513: What if we only want to allow a subset of rows to be expandable?

docs/components/table-view.md
```diff
@@ -510,6 +510,48 @@ interface TablePaginationAttributes {
...
+  :headers="headers"
+  :pagination-attributes="{ totalCount: tableData.length }"
+>
+  <template #row-expanded>
+    Lorem ipsum odor amet...
+  </template>
+</KTableView>
+```
+
+### hideHeader
```
adamdehaven, line 547: suggestion: should this be plural? “hide headers” since it hides all of them?

docs/components/table-view.md
```diff
@@ -510,6 +510,48 @@ interface TablePaginationAttributes {
...
+
+A boolean to hide table header. Only recomended when used in nested table. Refer to [Expandable Rows](#expandable-rows) section documentation for more details. Defaults to `false`.
```
adamdehaven, line 549:
```suggestion
A boolean to hide table headers. Only recomended when used in nested table. Refer to [Expandable Rows](#expandable-rows) section documentation for more details. Defaults to `false`.
```

docs/components/table-view.md
```diff
@@ -510,6 +510,48 @@ interface TablePaginationAttributes {
...
+
+### nested
```
adamdehaven, line 551: question: why do we need a prop for this? We should be able to target a nested table with CSS and JavaScript without the need for a prop

docs/components/table-view.md
```diff
@@ -622,6 +664,135 @@ A `error-action-click` event is emitted when error state action button is clicke
 />
 ```
 
+## Expandable Rows
+
+Data presented in a table often requires futher clarification or specification. The [`expandableRows` prop](#expandablerows) allows each row to be toggled, revealing or hiding additional information without the need to navigate away from the current view. Any content can be passed using the [`row-expanded` slot](#row-expanded). However, when displaying a nested table in expanded rows, there are a few important considerations to be aware of.
```
adamdehaven, line 669: What if we don’t want to allow every row to be expanded?

docs/components/table-view.md
```diff
@@ -622,6 +664,135 @@ A `error-action-click` event is emitted when error state action button is clicke
...
+### Nested Table With Same Columns
+  
+A nested table that complements the parent table requires some special handling:
+
+* The `nested` prop must be set to `true`
+* Table headers can be hidden using the [`hideHeader` prop](#hideheader). Value still need to be passed through the `headers` prop, and this is where the `nestedHeaders` slot prop becomes useful. It returns an array of header objects, simplifying the synchronization of column visibility when it is enabled in the parent table
+* To better align the columns of the parent and nested tables, the `columnWidths` slot prop can be utilized. It returns an object that can be passed to the nested table through the [`tablePreferences` prop](#tablepreferences). Each time a column is resized in the parent table, the nested table will be updated accordingly
+
+:::warning NOTE
+If bulk actions is enabled in parent table, it will be disabled in the nested table.
```
adamdehaven, line 726: question: why this restriction? You may want to act on the parent and children

docs/components/table-view.md
```diff
@@ -922,6 +1093,47 @@ Slot props:
...
 ```
 
+### row-expanded
+
+Slot for passing custom content that will be revealed once user expands one of the table rows when [`expandableRows` prop](#expandablerows) is `true`.
+
+Slot props:
+
+* `row` - table row object
+* `rowKey` - table row index
```
adamdehaven, line 1103: question: what’s the purpose of exposing an index? Child data likely wouldn’t have any knowledge of the parent, but even so, what would the index of the parent be used for?

src/styles/mixins/_tables.scss
```diff
@@ -236,20 +238,135 @@
...
+              }
+            }
+          }
+
+          &.last-row {
+            border-bottom: none;
+          }
+
+          &.expandable-content-row {
+            background-color: var(--kui-color-background-neutral-weakest, $kui-color-background-neutral-weakest);
```
adamdehaven, line 294: question: do we always want to enforce a background color?

src/styles/mixins/_tables.scss
```diff
@@ -236,20 +238,135 @@
...
+                  max-height: none;
+                  padding-right: var(--kui-space-0, $kui-space-0);
+                }
+
+                /* stylelint-disable-next-line selector-pseudo-class-no-unknown */
+                :deep(.k-table-view) {
+                  background-color: var(--kui-color-background-neutral-weakest, $kui-color-background-neutral-weakest);
+
+                  .table-wrapper {
+                    max-height: $expandedRowMaxHeight;
```
adamdehaven, line 324: question: should the max height be configurable?

src/styles/mixins/_tables.scss
```diff
@@ -236,20 +238,135 @@
...
+                  }
+
+                  tr.is-scrolled {
+                    background-color: var(
+                      --kui-color-background-neutral-weakest,
```
adamdehaven, line 329: same question here on enforcing a background color

docs/components/table-view.md
portikM: Nested table doesn't have room for bulk actions dropdown since toolbar is hidden. Act on both parent and nested from the parent table dropdown is unrealistic in majority of use cases since parent and nested items are usually different entities (debatable but see my next point). For not this limitation is coming purely from the fact that nested table doesn't display toolbar. Maybe in future we revisit this when we have a use case.

docs/components/table-view.md
portikM: Good point, implemented.

docs/components/table-view.md
portikM: I'm not sure what application would it have but we are exposing the same property in cell slots - maybe there's usage of it somewhere.

src/styles/mixins/_tables.scss
portikM: As per Figma designs, expanded row content has this background.

src/styles/mixins/_tables.scss
portikM: Maybe but I thought it could be hardcoded for now until we (if ever) face a need to have this configurable?

src/styles/mixins/_tables.scss
portikM: Just following designs

docs/components/table-view.md
adamdehaven: Example, a list of control planes. Some are composite and have expandable rows, others are not composite.

docs/components/table-view.md
adamdehaven: I’m not sure I understand. Bulk actions column has a reserved header key

docs/components/table-view.md
adamdehaven: Added example

docs/components/table-view.md
adamdehaven: I’m not sure children would have a need for the parent’s index?

docs/components/table-view.md
portikM: Oh.. you're right. I guess this prop needs to be a function that returns a boolean

docs/components/table-view.md
portikM: Yes but we don't set any special classes on bulk action column elements.. I guess we could though. What about resizable columns though, do you think we should allow column resizing in nested tables?

docs/components/table-view.md
portikM: Okay, removing this slot prop.

docs/components/table-view.md
portikM: For some reason resizing columns doesn't work in nested table and I think instead of fixing that we should just disable it but open to discussion here.

src/styles/mixins/_tables.scss
```diff
@@ -236,20 +238,135 @@
...
+            border-bottom: none;
+          }
+
+          &.expandable-content-row {
+            background-color: var(--kui-color-background-neutral-weakest, $kui-color-background-neutral-weakest);
+            height: auto;
+
+            > td {
+              padding: var(--kui-space-50, $kui-space-50) var(--kui-space-0, $kui-space-0)
+                var(--kui-space-50, $kui-space-50) var(--kui-space-60, $kui-space-60);
```
Justineo, line 299: Should this apply to `.expandable-content-wrapper` so that it won't produce a gap above the scrollbar?

image

Justineo: It seems to be inconsistent with how we are dealing with scrollable containers with `maxHeight` in other components like `<CodeBlock>` or `<Modal>`?

portikM: Asked the team in the thread whether we have a preference.

portikM: We decided to leave this as is for now

kaiarrowood (approved):
This looks really sharp (🔪 happy friday the 13th 😈), nice work @portikM !

PR #2380, opened 2024-09-09
fix(ktable): disable column visibility menu while loading

portikM:
# Summary

Instead of hiding column visibility menu while KTable is in loading state, disable it

### DIFF
--- a/src/components/KTable/KTable.vue
+++ b/src/components/KTable/KTable.vue
@@ -12,6 +12,7 @@
       <ColumnVisibilityMenu
         v-if="hasColumnVisibilityMenu"
         :columns="visibilityColumns"
+        :disabled="isTableLoading || loading"
         :table-id="tableId"
         :visibility-preferences="visibilityPreferences"
         @update="(columnMap: Record<string, boolean>) => columnVisibility = columnMap"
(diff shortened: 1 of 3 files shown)


src/components/KTable/KTable.vue
```diff
@@ -532,7 +533,7 @@ const hasHidableColumns = computed((): boolean => tableHeaders.value.filter((hea
 const hasColumnVisibilityMenu = computed((): boolean => {
   // has hidable columns, no error/loading/empty state
   return !!(hasHidableColumns.value &&
-    !props.error && !isTableLoading.value && !props.loading && (data.value && data.value.length))
+    !props.error && (data.value && data.value.length))
```
Justineo, line 536: I think If the table is in the loading state, there’s a good chance that `data.value` isn’t available yet?

Justineo: Maybe we should evaluate whether the dropdown menu should always be visible?

portikM: No, `data.value` get's reassigned once response is received, which means it has the old value while new data is being loaded (you can test that in the sandbox)

Justineo: `data.value` can be `[]` in multiple circumstances, like the initial value or search results. Any changes to the fetcher parameters may result in an empty array being returned, causing an unexpected layout shift, especially when we have customized the toolbar content (which users may be interacting with).

portikM: I apologize, I thought we hide the toolbar in empty state altogether but it doesn't seem to be the case. I think you're right and I'll push a fix shortly but just in case gonna tag @kaiarrowood to ask whether there was a reason why we decided to hide the column visibility menu in loading and empty states in the first place.

portikM: @Justineo feedback addressed

PR #2385, opened 2024-09-10
fix(kcodeblock): keyboard shortcut for copy should work

Justineo:
# Summary

* make code block focusable upon click so that keyboard events can work
* fix typo: Options -> Option


portikM (approved):
LGTM 🚀

PR #2388, opened 2024-09-11
fix(kcopy): truncated text should have max-width on tooltip

Justineo:
# Summary

Added `max-width` for the tooltip for `<KCopy>`. I currently set it to `500px` , same as the copy button tooltip. Not sure if we have any design guideline here.

https://github.com/Kong/kongponents/blob/aa5192b3058c5cdd602689dfa3a2a1d54a2f68b5/src/components/KCopy/KCopy.vue#L29-L34

Before:

image

After:

image

### DIFF
--- a/sandbox/pages/SandboxCopy.vue
+++ b/sandbox/pages/SandboxCopy.vue
@@ -35,6 +35,15 @@
           :text="uuid1"
           truncate
         />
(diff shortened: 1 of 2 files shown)


portikM (approved):
LGTM 🚀

PR #2389, opened 2024-09-11
fix(ktreeitem): ignore clicks on some slotted elements [KHCP-13213]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-13213

Ignore clicks on interactive slotted elements in KTreeItem


src/components/KTreeList/KTreeItem.vue
```diff
@@ -70,7 +71,23 @@ const slots = useSlots()
...
 
-const handleClick = () => {
+const handleClick = (event: any) => {
+  if (event.target) {
+    const ignoredElements = ['a', 'button:not([data-tree-item-trigger])', 'label', 'input', 'select']
+
+    // check whether target is an ignored elem
+    if (ignoredElements.includes(event.target.tagName.toLowerCase())) {
+      return
+    }
```
Justineo, line 81: Can we make `ignoredElements` a selector like `a, button:not([data-tree-item-trigger]), label, input, select` use `event.target.closest(ignored) !== null` here?

As `closest` starts searching from the element itself, we don't have to match the element itself and parents separately.

Ps. we can add `textarea` to the list, and maybe add `:not(:disabled)` to form elements to make it more complete.

portikM: Removed the unneeded element check and added the `:not(:disabled)` selector.

As for `textarea` selector, we don't need it in this list as no one should try to slot in textarea into the tree list item label - the component is way too small for that.

src/components/KTreeList/KTreeItem.vue
```diff
@@ -70,7 +71,18 @@ const slots = useSlots()
 
 const hasIcon = computed((): boolean => !props.hideIcons || !!slots['item-icon'])
 
-const handleClick = () => {
+const handleClick = (event: any) => {
+  if (event.target) {
+    const ignoredElements = ['a:not(.disabled)', 'button:not([data-tree-item-trigger]):not(disabled)', 'label', 'input:not(disabled)', 'select:not(disabled)']
```
Justineo, line 76: `disabled` -> `:disabled`

And I think we can use a selector group so that we don’t have to use a for loop:

```js
const ignoredElements = 'a:not(.disabled), button:not([data-tree-item-trigger]):not(:disabled), label, input:not(:disabled), select:not(:disabled)'

if (event.target.closest(ignoredElements) !== null) {
  return
}
```

portikM: Sorry, didn't understand what you mean by making it a string in the previous comment but thanks for clarifying. Updated, much cleaner this way!

src/components/KTreeList/KTreeItem.vue
```diff
@@ -70,7 +71,16 @@ const slots = useSlots()
 
 const hasIcon = computed((): boolean => !props.hideIcons || !!slots['item-icon'])
 
-const handleClick = () => {
+const handleClick = (event: any) => {
+  if (event.target) {
+    const ignoredElements = 'a:not(.disabled), button:not([data-tree-item-trigger]):not(:disabled), label, input:not(:disabled), select:not(:disabled)'
```
Justineo, line 76: Sorry, there was a slight error in my previous comment. I forgot that disabled elements do not respond to click events and do not bubble up to the container and we are already inside a click handler here. Therefore, there's no need to add `:not(:disabled)` here (although `a:not(.disabled)` might still be necessary).

portikM: I tested it in the sandbox and it works as expected without `:not(:disabled)`

Justineo (approved):
LGTM 🚀

jillztom (approved):

PR #2392, opened 2024-09-13
fix(ktableview): disable column visibility dropdown [KHCP-13248]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-13248

KTable and KTableView toolbar default items include bulk actions dropdown (only supported in KTableView) and column visibility menu (screenshot).

Screenshot 2024-09-16 at 11 08 49 AM

Changes:
Instead of hiding toolbar default items in loading and empty states at all times, only hide them in those states when toolbar slot is NOT provided. Otherwise, display those toolbar default items in disabled state to avoid layout shifts when table switches states (e.g. loading -> empty -> loading -> success).

### DIFF
--- a/sandbox/pages/SandboxTableView/SandboxTableView.vue
+++ b/sandbox/pages/SandboxTableView/SandboxTableView.vue
@@ -4,6 +4,57 @@
     title="KTableView"
   >
     <div class="k-table-view-sandbox">
+      <SandboxSectionComponent title="Toolbar default items">
+        <KComponent
+          v-slot="{ data }"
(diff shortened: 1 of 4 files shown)


Justineo (commented):
Can we update the sandbox to help testing? The code looks good to me.

portikM:
> Can we update the sandbox to help testing? The code looks good to me.

I added a [sandbox example](https://deploy-preview-2392--kongponents-sandbox.netlify.app/#/table-view) for functional verification but I will remove it before merging this PR (I don't think we need to keep the example for every arbitrary use case in the sandbox).

Justineo (approved):

kaiarrowood (approved):
Changes look good, sorry for the confusion. Should we call out this behavior in the docs somewhere?

PR #2405, opened 2024-09-18
feat(ktabledata): create component [KHCP-13110]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-13110

Changes:
* deprecate KTable component
* create new KTableData component that replicates the KTable but uses KTableView, providing data fetching layer on top


src/types/table.ts
```diff
@@ -17,6 +17,9 @@ export interface TablePreferences {
 
 export type TableViewData = Record<string, any>[]
 
+/**
+ * @deprecated in favor of TableDataHeader
+ */
```
adamdehaven, line 22: issue: it doesn't make sense to deprecate this interface if the `TableDataHeader` is still extending it

src/types/table.ts
```diff
@@ -76,3 +81,14 @@ export interface TablePaginationAttributes {
...
+
+export interface SortHandlerFunctionParam {
+  /** The key of the column to be sorted */
+  key: string
+  /** The key of the column previously sorted */
+  prevKey: string
+  /** The order in which to sort the column (asc or desc) */
+  sortColumnOrder: 'asc' | 'desc'
+  /** The data returned from the fetcher function response */
+  data: Array<any>
```
adamdehaven, line 93: issue: doesn't the interface require an array of objects?

src/types/table.ts
```diff
@@ -76,3 +81,14 @@ export interface TablePaginationAttributes {
 }
 
 export type RowBulkAction = boolean | { enabled: boolean, disabledTooltip?: string }
+
+export interface SortHandlerFunctionParam {
+  /** The key of the column to be sorted */
+  key: string
+  /** The key of the column previously sorted */
+  prevKey: string
```
adamdehaven, line 89: question: is there a time when this is undefined? (e.g. on init?)

src/components/KTableView/KTableView.vue
```diff
@@ -1167,9 +1175,10 @@ const setActualColumnWidths = (): void => {
   actualColumnWidths.value = widths
 }
 
-watch([columnVisibility, tableHeaders], (newVals) => {
+watch([columnVisibility, tableHeaders, hasExpandableRows], (newVals) => {
   const newVisibility = newVals[0]
   const newHeaders = newVals[1]
+  const newExpandableRows = newVals[2]
```
adamdehaven, line 1181: referencing the new values by index, now that there are 3, seems less than ideal. Could you extract them as variables and then reference them below as needed?

```ts
const [newVisibility, newHeaders, newExpandableRows] = newVals
```

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
+# Table Data
+
+Component that takes care of fetching and rendering data in a table format.
```
adamdehaven, line 3:
```suggestion
A table component that wraps [KTableView](/components/table-view) with built-in data fetching functionality.
```

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+```html
+<KTableData
+  :fetcher="fetcher"
+  :headers="headers"
+/>
+```
+
+## Underlying KTableView Component
+
+KTableData uses KTableView component providing data fetching layer on top. KTableData supports all KTableView [props](/components/table-view#props) (with the exception of `data` prop, which is replaced with [`fetcher` prop](#fetcher) in KTableData), [slots](/components/table-view#slots) and [events](/components/table-view#events) (with the exception of [pagination events](/components/table-view#pagination-events) which are not emitted by KTableData). `headers` and `hidePaginationWhenOptional` props as well as `update:table-preferences` event are handled slightly differently in KTableData. See below sections for details.
```
adamdehaven, line 23:
```suggestion
## Base KTableView Component

KTableData wraps the KTableView component, adding data fetching functionality. 

KTableData supports all [KTableView props](/components/table-view#props) (with the exception of `data` prop, which is replaced with [`fetcher` prop](#fetcher) in KTableData), [slots](/components/table-view#slots) and [events](/components/table-view#events) (with the exception of [pagination events](/components/table-view#pagination-events) which are not emitted by KTableData). `headers` and `hidePaginationWhenOptional` props as well as `update:table-preferences` event are handled slightly differently in KTableData. See below sections for details.
```

Suggesting an update to this paragraph, but you need to break up the sentences still a bit more. That's a long sentence 😅

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+
+## Props
+
+### headers
+
+In addition to all header object properties supported in KTableView, KTableData also allows specifying whether a column should be sortable via a function provided via [`sortHandlerFunction` prop](#sorthandlerfunction).
```
adamdehaven, line 29:
```suggestion
In addition to all header object properties supported by KTableView, KTableData also allows specifying whether a column should be sortable  by providing a [`sortHandlerFunction` prop](#sorthandlerfunction).
```

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+  /** Whether KTableData should use function passed through sortHandlerFunction prop to apply sorting logic to this column */
+  useSortHandlerFunction?: boolean
+}
+```
+
+For an example of `headers` prop usage please refer to [`fetcher` prop documentation](#fetcher) below.
+
+### fetcher
+
+Function that handles data fetching, server-side search and pagination. It takes a single param as an argument which you can conveniently pass as a prop (see [`initialFetcherParams` prop](#initialfetcherparams) for details).
```
adamdehaven, line 54: suggestion: "server-side search" implies this component does something in SSR or communicates with the server. I think this should be reworded to something that suggests that it supports querying new data from an API.

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+
+Fetcher function is expected to return an object with the following properties:
```
adamdehaven, line 56:
```suggestion
The fetcher function is expected to return an object with the following properties:
```

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+    }
+
+    return data
+  })
+}
+</script>
+```
+
+:::tip NOTE
+Notice that in the example above the _Username_ column is `sortable`. In this example sorting is handled on the client side, however should you want to apply server-side logic to handle sorting, you can bind your logic to the [`sort` event](#sort) and perform re-fetching of data as needed.
```
adamdehaven, line 153:
```suggestion
Notice in the example above the _Username_ column is `sortable`. In this example sorting is handled client-side; however, should you want to apply server-side logic to handle sorting, you can bind your logic to the [`sort` event](#sort) and perform re-fetching of data as needed.
```

Again, "server-side" logic here doesn't make sense; needs to be reworded.

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+:::
+
+### initialFetcherParams
+
+Fetcher function takes a single param - an object with the following properties:
```
adamdehaven, line 158:
```suggestion
The fetcher function accepts a single parameter, an object with the following properties:
```
suggestion: Try not to abbreviate in docs, e.g. `param` -> `parameter`

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+
+```ts
+interface FetcherParams {
+  pageSize?: number // the number of items to display per page
+  page?: number // the currently visible page
```
adamdehaven, line 163:
```suggestion
  page?: number // the currently active page
```

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+  query?: string // a text string to filter table data on (defined in the searchInput prop)
+  sortColumnKey?: string // sortable column key (defined in the headers prop)
+  sortColumnOrder?: 'asc' | 'desc' // sorting order
+  offset?: string | null // the value of the offset for offset-based pagination. offset MUST be included in the fetcher params for offset-based pagination to work properly
```
adamdehaven, line 167:
```suggestion
  offset?: string | null // the value of the offset for offset-based pagination. Offset must be included in the fetcher params for offset-based pagination to work properly
```

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+  query: '',
+  sortColumnKey: '',
+  sortColumnOrder: '',
+  offset: null
+}
+```
+
+### clientSort
+
+Boolean to enable client side sorting if using a fetcher that returns unpaginated data. Defaults to `false`. Use this in conjunction with [`sortHandlerFunction` prop](#sorthandlerfunction).
```
adamdehaven, line 186:
```suggestion
Boolean to enable client-side sorting if using a fetcher that returns unpaginated data. Defaults to `false`. Use this in conjunction with [`sortHandlerFunction` prop](#sorthandlerfunction).
```

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+
+### sortHandlerFunction
+
+Pass a custom sort handler function to handle sorting table data for specific columns.
+
+:::warning WARNING
+If [`clientSort` prop](#clientsort) is `true` but no `sortHandlerFunction` provided, KTableData will apply some default logic for client side sorting. This may produce unexpected results. We recommend **always** providing a `sortHandlerFunction` when using client side sorting.
```
adamdehaven, line 193:
```suggestion
If [`clientSort` prop](#clientsort) is `true` but no `sortHandlerFunction` provided, KTableData will apply some default logic for client side sorting. This may produce unexpected results. We recommend **always** providing a `sortHandlerFunction` when using client-side sorting.
```

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+  /** The data returned from the fetcher function response */
+  data: Array<any>
+}
+```
+
+:::tip NOTE
+Please note that you have to set [`clientSort` prop](#clientsort) to `true` as well as to pass `sortable: true` and `useSortHandlerFunction: true` in the respective [headers object](#headers) to make a column sortable via `sortHandlerFunction`.
+:::
+
+Please refer to [`fetcher` prop documentation](#fetcher) for example.
```
adamdehaven, line 215:
```suggestion
Please refer to the [`fetcher` prop documentation](#fetcher) for an example.
```

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+  } catch (error) {
+    // handle error
+  }
+}
+</script>
+```
+
+### searchInput
+
+Pass in a string of search query for server-side table filtering.
```
adamdehaven, line 292: Shouldn't reference "server-side filtering" - needs to be reworded to explain that the searchInput is returned as a param to the fetcher that can then be used to modify the API request

docs/components/table-data.md
```diff
@@ -0,0 +1,479 @@
...
+
+Set this to `true` to hide pagination when the table record count is less than or equal to the page size. Defaults to `false`.
+
+### hideToolbar
+
+Prop for hiding toolbar. Useful when elements provided in the toolbar are not actionable (for example, when toolbar contains filter controls, however the fetcher returns no results as there are no records to filter by).
+
+## Events
+
+### state
```
adamdehaven, line 308: This event is named the same as it was in KTable, but an event name of "state" is very broad and doesn't actually return the "state" of a large portion of the table's configuration/data

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,706 @@
...
+    default: null,
+  },
+  /**
+   * Enable client side sort - only do this if using a fetcher that returns static data
+   */
+  clientSort: {
+    type: Boolean,
+    default: false,
+  },
+  sortHandlerFunction: {
```
adamdehaven, line 338: nit: can you add inline comments for all props

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,706 @@
...
+  },
+  hideToolbar: {
+    type: Boolean,
+    default: false,
+  },
+})
+
+const emit = defineEmits<{
+  (e: 'cell-click', value: { data: any }): void
+  (e: 'row-click', value: { data: any }): void
```
adamdehaven, line 361: question: wouldn't the cell & row data at least conform to `Record<string, any>` rather than just `any`?

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,706 @@
...
+
+  return Object.keys(slots).filter((slot) => tableHeaders.value.some((header) => header.key === slot))
+})
+
+const isInitialFetch = ref<boolean>(true)
+const fetchData = async () => {
+  const searchInput = props.searchInput
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
adamdehaven, line 437: question: why is the ts-ignore needed here? if absolutely needed, please add an inline comment next to the ignore explaining

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,706 @@
...
+    } else {
+      nextOffset.value = res.pagination.offset
+
+      if (!offsets.value[page.value]) {
+        offsets.value.push(res.pagination.offset)
+      }
+    }
+
+    hasNextPage.value = (res.pagination && 'hasNextPage' in res.pagination) ? res.pagination.hasNextPage : true
+  }
```
adamdehaven, line 461: nit: can you add comments on this block?

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,706 @@
...
+
+  if (props.paginationAttributes?.offset) {
+    offset.value = fetcherParams.offset
+    offsets.value.push(fetcherParams.offset)
+  }
+
+  // trigger setting of tableFetcherCacheKey
+  hasInitialized.value = true
+}
```
adamdehaven, line 500: nit: can you add comments on this block?

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,706 @@
...
+
+watch([stateData, tableState], (newData) => {
+  emit('state', {
+    state: newData?.[1], // newData[tableState]
+    hasData: newData?.[0]?.hasData, // newData[stateData].hasData
+  })
+})
+
+// handles debounce of search query
+watch(() => props.searchInput, (newValue: string) => {
```
adamdehaven, line 660: nit: can you rename `newValue` to something more descriptive?

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,706 @@
...
+    case swrvState.VALIDATING_HAS_DATA:
+      isTableLoading.value = isRevalidating.value
+      break
+    default:
+      isTableLoading.value = false
+      break
+  }
+}, { immediate: true })
+
+watch([stateData, tableState], (newData) => {
```
adamdehaven, line 652: nit: can you rename `newData` to something more descriptive?

Also, it would be better to extract the variables rather than referencing below by index:

```ts
const [newStateData, newTableState] = newData
```

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,706 @@
...
+
+  if (newValue === '') {
+    search(newValue)
+  } else {
+    debouncedSearch(newValue)
+  }
+}, { immediate: true })
+
+const isRevalidating = ref<boolean>(false)
+watch([query, page, pageSize], async (newData, oldData) => {
```
adamdehaven, line 673: nit: can you rename newData and oldData to something more descriptive?

Also, it would be better to extract the variables rather than referencing below by index:

```ts
const [newQuery, newPage] = newData
```

Same for "oldData"

adamdehaven:
I'm not sure of a way to reliably test everything here before merge; it might be nice to replace a few existing tables in `public-ui-components` and `konnect-ui-apps` in draft PRs and verify functionality before merging.

src/components/KTableData/KTableData.vue
portikM: Done

src/components/KTableData/KTableData.vue
portikM: Done

src/components/KTableData/KTableData.vue
portikM: Done (still using `newData, oldData` though cuz it's generic enough to accommodate this use case)

src/components/KTableView/KTableView.vue

src/types/table.ts
portikM: Once we are ready to remove it we can just swap it without changing any of the properties

src/types/table.ts
portikM: It's never undefined, if there is no column that was previously sorted, this value is empty string (`''`)

src/types/table.ts
portikM: Yes, fixed

portikM:
> I'm not sure of a way to reliably test everything here before merge; it might be nice to replace a few existing tables in `public-ui-components` and `konnect-ui-apps` in draft PRs and verify functionality before merging.

This is what I was thinking too

src/types/table.ts
adamdehaven: You can't extend a deprecated interface in an interface you are _not_ deprecating. 

If you're wanting to rename, you'll need to copy/paste the interface into the new one, and mark the old one as deprecated.

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,714 @@
...
+    default: () => false,
+  },
+  hideHeaders: {
+    type: Boolean,
+    default: false,
+  },
+  nested: {
+    type: Boolean,
+    default: false,
+  },
```
adamdehaven, line 289: suggestion: if you define the props with pure type annotations, you can import the shared table props in both the `KTableView` and `KTableData` components and merge the object with any that are needed specifically in this component:

```ts
// Import this interface
interface MyProps {
  msg?: string
  labels?: string[]
}

const { 
  msg = 'hello', 
  labels = ['one', 'two'],
  // This prop only lives in this component
  total = 0,
} = defineProps<MyProps & {
  // Another prop that only lives here
  total: number
}>()
```

This works for Vue 3.5 and above; for 3.4 you'll need to use `withDefaults` [as described here](https://vuejs.org/guide/typescript/composition-api.html#props-default-values). 

If we use this without consideration for 3.4, we'll need to bump the peerDependency range and ensure that `konnect-ui-apps`, `portal` and `konnect-portal`, `kong-admin`, and mesh are all on 3.5

adamdehaven (changes requested):

src/components/KTableData/KTableData.vue
portikM: Done

src/types/table.ts
portikM: Done

src/components/KTableData/KTableData.vue
adamdehaven: You'll need to check the other repositories above before this merges to ensure they're on 3.5+

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,714 @@
...
+      v-if="$slots['row-expanded']"
+      #row-expanded="{ columnWidths, nestedHeaders, row }"
+    >
+      <slot
+        :column-widths="columnWidths"
+        name="row-expanded"
+        :nested-headers="nestedHeaders"
+        :row="row"
+      />
+    </template>
```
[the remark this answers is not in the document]
adamdehaven: So you'd have to have all expandable row content defined in a single slot, and then utilize `v-if` logic inside the slot (or something) to make it dynamic per row?

If that's the case, can you update the docs example in this PR accordingly to prove the concept?

docs/components/table.md
```diff
@@ -1,1124 +0,0 @@
-# Table
```
adamdehaven, line 1: issue: If we're just deprecating the component, we shouldn't delete the docs file.

You can put a "danger" note at the top explaining that it's deprecated and to use KTableData instead

package.json
```diff
@@ -69,7 +69,7 @@
   },
   "peerDependencies": {
     "axios": "^1.7.7",
-    "vue": ">= 3.3.4 < 4",
+    "vue": ">= 3.5.4 < 4",
```
adamdehaven, line 72:
```suggestion
    "vue": ">= 3.5.0 < 4",
```

adamdehaven (changes requested):

src/components/KTableData/KTableData.vue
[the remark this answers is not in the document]
adamdehaven: Another question on the KTableView implementation (since we're here).

The "getGeneric" function that returns the row object: why can't it accept an actual typescript generic and return the same type as to not lose the interface of the row?

Something like this:

```ts
const getGeneric = <TRow>(obj: TRow): TRow => obj
```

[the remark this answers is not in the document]
portikM: > If that's the case, can you update the docs example in this PR accordingly to prove the concept?

Added an [example in docs](https://deploy-preview-2405--kongponents.netlify.app/components/table-view.html#row-expanded) for this.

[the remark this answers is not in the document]
portikM: > Another question on the KTableView implementation (since we're here).

Implemented your suggestion, however I'm not sure it help as the passed object still is of type `any`.

package.json
portikM: Done

docs/components/table.md
portikM: Restored the component docs page, only did not add it to the sidebar nav.

src/components/KTableData/KTableData.vue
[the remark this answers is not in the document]
adamdehaven: Is it getting set to any somewhere else? I know for the `row` it's always going to show `Record<string, any>` but for the column it should show that it follows the header interface?

docs/components/table.md
adamdehaven: It needs to be in the sidebar nav as well; just add the deprecation message to the top

adamdehaven: it's still used in our apps for a while so we don't want to hide documentation for existing instances. Deprecated doesn't mean no longer used

src/types/table.ts
```diff
@@ -76,3 +98,141 @@ export interface TablePaginationAttributes {
...
+}
+
+export interface CellAttrsParam {
+  headerKey: string
+  row: Record<string, any>
+  rowIndex: number
+  colIndex: number
+}
+
+export interface TableProps {
```
adamdehaven, line 120: nit: since we have two tables, could we rename this to `TablePropsShared`?

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,591 @@
...
+  fetcherCacheKey?: string
+  cacheIdentifier?: string
+  searchInput?: string
+  initialFetcherParams?: TableDataFetcherParams
+  clientSort?: boolean
+  sortHandlerFunction?: (param: SortHandlerFunctionParam) => Record<string, any>[]
+  sortable?: boolean
+  hidePaginationWhenOptional?: boolean
+  hideToolbar?: boolean
+}
```
adamdehaven, line 192: nit: can we move this to the types file?

src/components/KTableView/KTableView.vue
```diff
@@ -407,180 +402,47 @@ enum TableViewHeaderKeys {
...
-    default: null,
-  },
-  /**
-   * A prop to pass in a custom error state action message
-   */
-  errorStateActionMessage: {
-    type: String,
-    default: '',
-  },
+interface TableViewProps extends TableProps {
```
adamdehaven, line 405: nit: can we move this to the types file?

adamdehaven: (for both comments like this, my reasoning is that if they are defined here, they are not exported for import by the consuming app)

portikM: > (for both comments like this, my reasoning is that if they are defined here, they are not exported for import by the consuming app)

Oh maybe this is why I can't see the inline comments in the consuming app. But then the shared interface doesn't have to be exported?

adamdehaven: > But then the shared interface doesn't have to be exported?

If the actual table prop interfaces are exported, the "TablePropsShared" does not need to be exported on it's own

```ts
interface TablePropsShared { ... }

export interface TableViewProps extends TablePropsShared & { ... }
export interface TableDataProps extends TablePropsShared & { ... }
```

portikM: Done

src/components/KTableData/KTableData.vue
portikM: Done

src/types/table.ts
portikM: Done

src/components/KTableData/KTableData.vue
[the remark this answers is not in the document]
portikM: So I had to revert this change because it breaks usage in [public-ui-components](https://github.com/Kong/public-ui-components/actions/runs/11034925613/job/30649674841?pr=1654). We could try and fix it forward but I think that will make adoption of the new component too hard while adding very little benefit.

src/components/KTableData/KTableData.vue
```diff
@@ -0,0 +1,576 @@
...
+    :empty-state-title="emptyStateTitle"
+    :error="error"
+    :error-state-action-message="errorStateActionMessage"
+    :error-state-action-route="errorStateActionRoute!"
+    :error-state-message="errorStateMessage"
+    :error-state-title="errorStateTitle"
+    :headers="tableHeaders"
+    :hide-headers="hideHeaders"
+    :hide-pagination="hidePagination || !showPagination"
+    :hide-pagination-when-optional="hidePaginationWhenOptional"
```
Leopoldthecoder, line 20: It seems `hide-pagination-when-optional` is not implemented in `KTableView`.

portikM: Fixed now. It's handled slightly differently in KTableView and KTableData since KTableData has more context about data and API responses.

docs/components/table.md
```diff
@@ -1,11 +1,11 @@
...
 
-Component that takes care of fetching and rendering data in a table format.
-
-:::tip NOTE
-If you are looking for a simpler table component that does not integrate data fetching, check out [KTableView](/components/table-view).
+:::danger DANGER
+KTable has been deprecated. Please use [KTableData](/components/table-data) instead.
 :::
 
+Component that takes care of fetching and rendering data in a table format.
```
adamdehaven, line 7:
```suggestion
A table component that takes care of fetching and rendering data in a table format.
```

src/components/KTableView/KTableView.vue
```diff
@@ -963,11 +812,11 @@ const startResize = (evt: MouseEvent, colKey: string) => {
 }
 
 const showPagination = computed((): boolean => {
-  if (props.hidePagination) {
+  if (props.hidePagination || props.nested) {
```
adamdehaven, line 815: There appear to be changes unrelated to KTableData in this PR as well? So now we'll need to test _all_ of the added functionality?

src/types/table.ts
```diff
@@ -77,6 +99,176 @@ export interface TablePaginationAttributes {
...
+  key: string
+  /** The key of the column previously sorted */
+  prevKey: string
+  /** The order in which to sort the column (asc or desc) */
+  sortColumnOrder: 'asc' | 'desc'
+  /** The data returned from the fetcher function response */
+  data: Array<Record<string, any>>
+}
+
+export interface CellAttrsParam {
```
adamdehaven, line 113: nit: add a comment what this interface is for

src/utilities/tableHelpers.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param {TablePreferences} tablePreferences
+ * @param {TablePaginationAttributes} paginationAttributes
+ * @returns {number} The initial page size
+ */
+export const getInitialPageSize = (tablePreferences: TablePreferences, paginationAttributes: TablePaginationAttributes): number => {
+  if (tablePreferences.pageSize) {
+    return tablePreferences.pageSize
+  } else if (paginationAttributes.initialPageSize) {
+    return paginationAttributes.initialPageSize
+  } else if (paginationAttributes.pageSizes) {
```
adamdehaven, line 16: issue: I think you need to check that it's an array?
```suggestion
  } else if (typeof paginationAttributes.pageSizes === 'object' && paginationAttributes.pageSizes.length) {
```

docs/components/table-view.md
```diff
@@ -699,6 +699,10 @@ interface TablePaginationAttributes {
 
 A boolean to hide pagination element (defaults to `false`).
 
+### hidePaginationWhenOptional
+
+Set this to `true` to hide pagination when the table record count is less than or equal to the page size. Defaults to `true`.
```
Justineo, line 704: Shall we flip the logic and use something like `autoHidePagination` so that it defaults to `false`?

portikM: I suggest we don't in order to keep component API as close to original KTable API as possible.

docs/components/table-view.md
```diff
@@ -741,6 +745,10 @@ A boolean to hide table headers. Only recomended when used in nested table. Refe
 
 A boolean to disable some of the table features like column visibility, column resizing, bulk actions and hide toolbar element in nested tables. Refer to [Expandable Rows](#expandable-rows) section documentation for more details. Defaults to `false`.
 
+### hideToolbar
+
+Prop for hiding toolbar. Useful when elements provided in the toolbar are not actionable (for example, when toolbar contains filter controls, however the fetcher returns no results as there are no records to filter by).
```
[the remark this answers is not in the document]
portikM: This is just an example of how this prop can be used, nuances are left up to host apps to handle.

src/components/KTableView/KTableView.vue
portikM: Wdym? Only functionality added to KTableView in this PR is `hidePaginationWhenOptional` prop, which adds minor change and has been tested. All other changes are either refactor related to props handling (`withDefaults`) or minor fixes found during KTableData adoption (such as this line).

src/types/table.ts
portikM: Done

src/utilities/tableHelpers.ts
portikM: Checking type isn't really necessary since if something other than array of numbers was passed through `paginationAttributes` prop Vue would throw a type error, however checking array length is useful.

src/components/KTableView/KTableView.vue
adamdehaven: > minor fixes found during KTableData adoption

This explains what I was asking 👍🏼

src/utilities/tableHelpers.ts
adamdehaven: If you passed `paginationAttributes.pageSizes = 'my-name'` you would get a length of `7`

You'd get prop validation _warnings_ in dev mode, but not necessarily errors in a consuming app, depending on the setup

docs/components/table-view.md
adamdehaven: FWIW this is a new component, there is no requirement to follow the old component APIs 1:1 as long as the changes are documented correctly and picked up in Konnect accordingly.

portikM: Okay, even if so, `autoHidePagination` sounds confusing to me. `hidePaginationWhenOptional` is a prop name our engineers are familiar with from KTable and KCatalog, it's intuitive.

src/utilities/tableHelpers.ts
portikM: Added a check to make sure `paginationAttributes.pageSizes` is an array

docs/components/table-view.md
Justineo: What I mainly want to say is that we should strive to set the default value of `boolean` props to false, but the specific name in this case is of course open for discussion.

It seems that the `<KTable>` component's `hidePaginationWhenOptional` prop [defaults to `false`](https://github.com/Kong/kongponents/blob/be5df2d4467087add361cd0564110bafde52dd4d/src/components/KTable/KTable.vue#L494), are we changing its default value in the new tables?

portikM: Yes, it's `true` in KTableView but `false` in KTableData.

src/components/KTableView/KTableView.vue
```diff
@@ -382,206 +382,55 @@ import type {
...
+  errorStateTitle: 'An error occurred',
+  errorStateMessage: 'Data cannot be displayed due to an error.',
+  errorStateActionMessage :'',
+  maxHeight: 'none',
+  hidePagination: false,
+  paginationAttributes: () => ({}),
+  rowExpandable: () => false,
+  hideHeaders: false,
+  nested: false,
+  hidePaginationWhenOptional: true,
```
adamdehaven, line 427: I think it's a bit confusing that `hidePaginationWhenOptional` defaults to `true` in KTableView, but defaults to `false` in `KTableData`

adamdehaven, line 427: I'm also curious what the guidance is for when we set this to true/false when used in our host apps? 

There doesn't seem to be any standard pattern

docs/components/table-view.md
portikM: We point that out in docs here: https://deploy-preview-2405--kongponents.netlify.app/components/table-data.html#base-ktableview-component. Is that sufficient or would it be better if the default value for this prop was matching across the two components (in that case I would suggest it defaults to `false`)?

src/components/KTableView/KTableView.vue
portikM: Yes I realize this is rather misleading. Since KTableData implementation should ideally align with KTable (where it's `false` by default), I suggest that `hidePaginationWhenOptional` defaults to `false` in both KTableView and KTableData.

portikM: Pushed a commit to change `hidePaginationWhenOptional` default value to `false` in KTableView.

docs/components/table-view.md
portikM: Pushed a commit to change `hidePaginationWhenOptional` default value to `false` in KTableView.

src/components/KTableView/KTableView.vue
portikM: I think the prop description in docs makes it pretty clear how and when to use it?

adamdehaven (approved):

PR #2419, opened 2024-09-27
feat(kmultiselect): add search-placeholder prop [KHCP-13352]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-13352

Changes (KMultiselect):
* add `searchPlaceholder` prop that acts as search input placeholder
  * in collapsed-context mode - the trigger element is the search input so text passed through `searchPlaceholder` prop displays in the trigger element when no items are selected
  * otherwise, `searchPlaceholder` sets a placeholder in search input inside multiselect dropdown
  * in non-collapsed-context mode, existing `placeholder` prop (if has value) is used as trigger element text when no items are selected
* add `placeholder` and `searchPlaceholder` props documentation

### DIFF
--- a/src/components/KMultiselect/KMultiselect.vue
+++ b/src/components/KMultiselect/KMultiselect.vue
@@ -50,7 +50,7 @@
                 data-testid="multiselect-input"
                 :disabled="isDisabled"
                 :model-value="filterString"
-                :placeholder="placeholderText"
+                :placeholder="triggerElementText"
                 :readonly="isReadonly ? true : undefined"
                 type="text"
                 @blur="() => isFocused = false"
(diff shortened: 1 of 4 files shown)


src/components/KMultiselect/KMultiselect.vue
```diff
@@ -348,6 +349,10 @@ const props = defineProps({
     type: String,
     default: '',
   },
+  searchPlaceholder: {
+    type: String,
+    default: '',
+  },
```
Justineo, line 355: Shall we warn users when `collapsedContext` is `true` but `placeholder` is not empty?

portikM: I don't think that's necessary: the default displayed text for both modes is _0 items selected_ - I don't think we need to specifically point that out.

Justineo (approved):
LGTM

adamdehaven (approved):

PR #2424, opened 2024-09-30
feat(kcatalog): support offest pagination [KHCP-13325]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-13325

Support offset pagination in KCatalog

### DIFF
--- a/src/components/KCatalog/KCatalog.vue
+++ b/src/components/KCatalog/KCatalog.vue
@@ -132,15 +132,20 @@
     </div>
 
     <KPagination
-      v-if="!disablePagination && fetcher && !(hidePaginationWhenOptional && total <= paginationPageSizes[0]) && !error && data && data.length"
+      v-if="showPagination"
       class="card-pagination"
       :current-page="page"
       data-testid="catalog-pagination"
       :disable-page-jump="disablePaginationPageJump"
       :initial-page-size="pageSize"
       :neighbors="paginationNeighbors"
+      :offset="paginationOffset"
+      :offset-next-button-disabled="!nextOffset || !hasNextPage"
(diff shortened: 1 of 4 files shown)


src/components/KCatalog/KCatalog.vue
```diff
@@ -360,20 +373,37 @@ const emit = defineEmits<{
...
 
 const data = ref<CatalogItem[]>([])
 const total = ref<number>(0)
 const filterQuery = ref<string>('')
 const page = ref<number>(1)
-const pageSize = ref<number>(15)
+const pageSize = ref<number>(getInitialPageSize())
+const offset = ref<string | null>(null)
+const offsets = ref<Array<any>>([])
+const previousOffset = computed((): string | null => offsets.value[page.value - 1])
```
kaiarrowood, line 394:
```suggestion
const previousOffset = computed((): string | null => offsets.value[page.value - 1])
```

will it be a problem if `offsets.value` is null?

src/components/KCatalog/KCatalog.vue
```diff
@@ -393,11 +423,35 @@ const fetchData = async () => {
...
   data.value = res.data as CatalogItem[]
   total.value = props.paginationTotalItems || res.total || res.data?.length
 
+  if (props.paginationOffset) {
+    if (!res.pagination?.offset) {
+      nextOffset.value = null
+    } else {
+      nextOffset.value = res.pagination.offset
+
+      if (!offsets.value[page.value]) {
```
kaiarrowood, line 438: same question, need to make sure `offsets.value` is defined before checking for a specific page

PR #2427, opened 2024-10-01
feat(ktableview): add row-key prop [KHCP-13502]

portikM:
# Summary

Addresses:
https://konghq.atlassian.net/browse/KHCP-13502
https://konghq.atlassian.net/browse/KHCP-13404

Changes:
- adds `rowKey` prop in KTableView
- changes internal logic KTableView for bulk actions functionality: instead of copying `props.data` into internal ref `tableData` (removed) and adding a `selected` property for bulk actions, it uses an internal ref `dataSelectState` that's completely detached from `props.data` to eliminate chance of mutating passed data or returning unexpected values


docs/components/table-view.md
```diff
@@ -352,6 +366,30 @@ const getRowLink = (row: Record<string, any>): RowLink => ({
 </script>
 ```
 
+### rowKey
+
+Certain features of KTableView, such as bulk actions, require a unique identifier for each row of table data. The `rowKey` prop provides a way to define which property of the row object should serve as this unique identifier or to generate a custom unique value for each row. This prop accepts either a string or a function: `string | ((row: Record<string, any>) => string)`.
```
adamdehaven, line 371: docs: can we explicitly call out the features of the table that will utilize the `rowKey` prop instead of generalizing here?

adamdehaven, line 371:
```suggestion
Certain features of KTableView, such as bulk actions, require a unique identifier to utilize for each row of table data. 

The `rowKey` prop provides a way to define which property of the `row` object should serve as this unique identifier, or to generate a custom unique identifier for each row. 

The `rowKey` prop accepts either a unique string or a function that returns a unique string: `string | ((row: Record<string, any>) => string)`.
```

We should also note that the rowKey needs to be unique for the entire data set, e.g. not just the first page of data.

src/components/KTableView/KTableView.vue
```diff
@@ -407,6 +407,12 @@ enum TableViewHeaderKeys {
   BULK_ACTIONS = 'bulkActions',
 }
 
+interface DataSelectState {
+  rowKey: string
+  selected: boolean
+  disabled: boolean
+}
```
adamdehaven, line 414: nit: this should be moved to the types file

src/components/KTableView/KTableView.vue
```diff
@@ -547,6 +553,13 @@ const props = defineProps({
     type: Array as PropType<TableViewData>,
     default: () => [],
   },
+  /**
+   * A prop to pass row object key to be used as a unique identifier
```
adamdehaven, line 557: nit:
1. we don't need to explain it's a prop
2. reword
```suggestion
   * Provide the name of the data property keyto utilize as a unique identifier, or a function that receives the `row` object as a parameter that generates a unique identifier string for each row.
```

adamdehaven (changes requested):
Initial review, did not functionally test yet

docs/components/table-view.md
Justineo: I’ve been thinking that, if the row data do not already includes a unique key field, how can we generate a stable unique id from the object for the entire data set? (The case for passing a function)

adamdehaven: If you had a data set you wanted to take an action on (eg list of users) each user would inherently have to have a unique combination of some form of data (eg `firstName + lastName`) otherwise; how would your host app know which “thing” to act on in the first place? In this case, you could concatenate to generate a unique ID.

Alternatively, if your table data consisted of arbitrary, non-unique row data:

```js
[{ name: ‘foo’ }, { name: ‘foo’ }]
```

The host app would be required to do some sort of key generation itself, outside of the data object, to _make_ the rows unique.

src/components/KTableView/KTableView.vue
```diff
@@ -229,18 +229,18 @@
 
           <tbody>
             <template
-              v-for="(row, rowIndex) in tableData"
-              :key="`table-${tableId}-row-${rowIndex}`"
+              v-for="(row, rowIndex) in data"
+              :key="`table-${tableId}-row-${getRowKey(row) ?? useUniqueId()}`"
```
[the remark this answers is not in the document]
Justineo: Yes I think it should be stable within the same table which only backend API can guarantee. Generating keys on the frontend should always have its issues, either you are generating different keys for the same object (randomizing), or same keys for different object (using index).

docs/components/table-view.md
Leopoldthecoder: > I’ve been thinking that, if the row data do not already includes a unique key field, how can we generate a stable unique id from the object for the entire data set? (The case for passing a function)

It's a rare case but maybe we need to combine two or more fields of the row data to identify an item? e.g. `birthDate` and `name`, because people may have the same name but when combined with their date of birth, one can be uniquely determined.

Another case is that the unique key is nested, e.g. `row.xxx.id`, where we can utilize the function like this: `(row) => row.xxx.id`.

Justineo: Calculating an object’s hash (whether using its full name or other more complex cases) on the frontend cannot guarantee uniqueness, as we don’t have access to the source of truth, which typically resides in the database. Composite keys and nested keys are valid use cases, I agree. Other generated ids without predefined keys should be live with the row data itself and will only work if all the necessary data is available on the frontend (such as when data is created entirely on the frontend).

adamdehaven: > cannot guarantee uniqueness

correct, but this burden falls on the host application if it wants to take advantage of row actions; the same as the cache key uniqueness for swrv (in KTable)

Justineo: If the host app is going to use row actions there’s a great chance it may send request to backend and this requires the frontend to have the real unique key so I think it’s kind of inevitable. Generated ids are only useful for preserving pure frontend states, like expanded state, transition, etc. Maybe we should document that t host apps should always provide a unique row key in row data and fallback to a generated key if we’ve got nothing.

adamdehaven: > Maybe we should document that t host apps should always provide a unique row key in row data

This is what I actually suggested on the original PR 😅 ; essentially that each row is **required** to provide a unique property per row.

I could see where this would be overkill though for tables whereby the host app doesn't need to manipulate the data. e.g. if I just want to show a table of things, I probably don't want to have to mutate my data just to use the table component.

| Name | Count |
| :--- | :--- |
| Foo | 5 |
| Bar | 5 |

portikM: In tables that don't care about bulk actions we don't require a unique row identifier -- it's only required for bulk actions feature.

portikM: As for use-cases for function prop, apart from what Adam mentioned I can also envision a scenario where an `id` field in database is of type number: `[{ id: 0, name: 'foo' }, { id: 1, name: 'bar' }]` (not uncommon for SQL databases). In such scenario one might want to transform `id` a little: `:row-key="({ id }: Record<string, any>) => 'row-' + id"`.

portikM: > It would be needed for sorting as well and any other client action that manipulates row order

Why? Is this a requirement for this PR?

adamdehaven: > > It would be needed for sorting as well and any other client action that manipulates row order
> 
> Why? Is this a requirement for this PR?

https://vuejs.org/guide/essentials/list.html#maintaining-state-with-key

Justineo: In fact Vue is requiring the unique keys, if we try to hide that would leave footguns where someday we are adding row states. The row states can be beyond bulk actions. Like the original issue I demonstrated in our Slack discussion, row states can be anything do not coming from props, eg. transitions, local states of components inside each row, or even text selection or scroll positions.

adamdehaven: For row states that don't need to persist (e.g. tracking rows that are removed from the DOM as with bulk actions) I believe we _could_ actually rely on generated unique IDs, as long as those IDs do not change while the corresponding row is visible in the DOM.

Example, if we generate a unique ID per row when the data is loaded before the virtual DOM renders the HTML, and the ID doesn't change as long as that row is in the DOM, I think Vue tracking the state of the row via the `key` would function properly so long as there aren't duplicate IDs, and the ID of any given row does not change while it is present in the DOM

portikM: Oh, I thought the solution I pushed in [this commit](https://github.com/Kong/kongponents/pull/2427/commits/c72d08927e2830c8b7b3144a877f8ff5ad82812f) would be sufficient to not have to require `rowKey` for sorting. Do you foresee any issues with this approach?

```
:key="`table-${tableId}-row-${getRowKey(row) ?? useUniqueId()}`"
```

Justineo: If we are certain that the data is always static, not requiring unique ids is probably fine. However, in my experience, when I’m in a situation that a unique id is needed for row data, it is usually already provided by the backend API.

Justineo: If I understand correctly, `useUniqueId()` makes the key vary on each render so row state is always lost.

portikM: > If we are certain that the data is always static

Well I don't think we always are.. for example in KTableView, since it doesn't have fetcher logic, pagination logic is handled in the host app completely - which means when user navigates through pages the host app fetches the data and just swaps `props.data` under the hood. If `loading` state is not set during the re-fetch, the `<table>` element and all it's nested children doesn't even re-render. Should we at least provide a unique key to the table element to ensure it get's re-rendered each time props.data changes?

```
<table
  :key="`tableId-${tableKey}`"
  ...
  
const tableKey = ref<number>(0)

watch(() => props.data, () => {
 tableKey.value++
 })
```

adamdehaven: > If I understand correctly, `useUniqueId()` makes the key vary on each render so row state is always lost.

Yes, I was thinking the keys need to be generated before the rows are rendered in the DOM, tied to the data (but not in the data in all cases)

Justineo: And it only works for either 1. static tables 2. where frontend is the source of truth.

src/components/KTableView/KTableView.vue
```diff
@@ -266,13 +266,13 @@
...
                     <KTooltip
-                      v-else-if="header.key === TableViewHeaderKeys.BULK_ACTIONS"
+                      v-else-if="header.key === TableViewHeaderKeys.BULK_ACTIONS && getRowState(row)"
                       max-width="200"
                       placement="bottom-start"
                       :text="getRowBulkActionEnabled(row) ? undefined : getRowBulkActionTooltip(row)"
                     >
                       <KCheckbox
-                        v-model="row.selected"
+                        v-model="getRowState(row)!.selected"
```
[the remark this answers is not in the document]
portikM: It's not injecting properties into the data object. `getRowState` takes care of finding and returning a corresponding object within internal `dataSelectState` ref (that only reads from `props.data` but doesn't modify it in any way).

src/components/KTableView/KTableView.vue
portikM: This is an internal interface used by internal ref only -- we should not be exporting it (same deal as with `TableViewHeaderKeys` enum).

src/components/KTableView/KTableView.vue
```diff
@@ -603,6 +616,14 @@ const slots = useSlots()
 const tableId = useUniqueId()
 const { getSizeFromString } = useUtilities()
 
+const getRowKey = (row: Record<string, any>): string => {
+  if (typeof props.rowKey === 'function') {
```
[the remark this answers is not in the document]

src/components/KTableView/KTableView.vue
adamdehaven: our pattern is to define all types in the `/types` directory; even if it's internal, it doesn't hurt to export it.

portikM: In that case should the `TableViewHeaderKeys` enum be moved to the `/types` directory too?

docs/components/table-view.md
Justineo: > Should we at least provide a unique key to the table element to ensure it get's re-rendered each time props.data changes?

I’m sorry I missed this earlier. I think it’s ok to have generated keys as a last resort but we should also surface the issue to users so they can understand why providing keys is crucial in many cases, even when they might not realize it.

adamdehaven: I’m also not sure you’d want the table to rerender when the data changes every time

src/components/KTableView/KTableView.vue
```diff
@@ -603,6 +616,14 @@ const slots = useSlots()
...
+    return props.rowKey(row)
+  }
+
+  return row[props.rowKey] ? String(row[props.rowKey]) : ''
```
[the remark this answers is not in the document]

src/components/KTableView/KTableView.vue

src/components/KTableView/KTableView.vue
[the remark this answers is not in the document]
portikM: > What about using a WeakMap to hold the generated keys for the row objects that didn't change, so that we won't be reconstructing the rows at least when the same row objects are used here?

If I'm understanding you correctly, you suggest that we create a `WeakMap` that will hold all generated keys for each row (we'd still use `useUniqueId` function to make the keys unique if `rowKey` is not provided). Then in the `v-for` we reference the corresponding key from the `WeakMap` for each row? If so then I don't quite understand how would we benefit from this approach.

Scenario: `rowKey` is not provided, pagination page size is 15 so the table renders with 15 unique keys in `WeakMap`. User navigates to page 2 of table data which also contains 15 records so we need to generate new values for each row. Then user returns to page 1. We would have no way of identifying the keys that were generated for page 1 the first time the table was rendered since the keys don't contain a unique identifier associated with values sourced from data objects. So we would have to generate new keys for each row - effectively what we are doing right now.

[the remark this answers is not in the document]
Justineo: The `WeakMap` stores row object => generated key. We only generate new keys if different row objects are passed in. If I understand correctly, at lease we are not generating new keys if the re-render comes from props other than `data`, or the new `data` contains some of the row objects as in the previous value?

/cc @adamdehaven WDYT?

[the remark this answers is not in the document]
adamdehaven: > at lease we are not generating new keys if the re-render comes from props other than data

Yes that’s what I was thinking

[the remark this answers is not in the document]
portikM: Okay, let me try that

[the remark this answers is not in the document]
portikM: Implemented ([commit](https://github.com/Kong/kongponents/pull/2427/commits/493320adb4c52f865e3e4da49d92226ae296fbf5)), please review.

Justineo (approved):
LGTM 🚀

src/components/KTableView/KTableView.vue
```diff
@@ -603,6 +605,18 @@ const slots = useSlots()
...
+const getRowKey = (row: Record<string, any>): string => {
+  if (typeof props.rowKey === 'function' && typeof props.rowKey(row) === 'string') {
+    return props.rowKey(row)
+  }
+
+  if (typeof props.rowKey === 'string' && !!(props.rowKey in row) && typeof row[props.rowKey] === 'string') {
+    return row[props.rowKey]
+  }
+
+  return ''
```
adamdehaven, line 617: question/issue(?): there are some places where we compare the result of this `getRowKey` function.

e.g.

```ts
const row = newData.find((dataRow) => getRowKey(selectedRow) === getRowKey(dataRow))
```

If there is a scenario where both calls evaluate to an empty string, then usage like shown above will always evaluate to `true`.

- Is that going to cause other issues?
- Should we be checking that `getRowKey` doesn't evaluate to an empty string in some/all instances?

src/types/table.ts
```diff
@@ -76,3 +76,15 @@ export interface TablePaginationAttributes {
 }
 
 export type RowBulkAction = boolean | { enabled: boolean, disabledTooltip?: string }
+
+export interface DataSelectState {
```
adamdehaven, line 80: nit: should this be prefixed with `Table` like the other interfaces?

src/components/KTableView/KTableView.vue
portikM: That entire block is [wrapped in a condition check](https://github.com/Kong/kongponents/blob/493320adb4c52f865e3e4da49d92226ae296fbf5/src/components/KTableView/KTableView.vue#L1261) that accounts for checking whether `getRowKey(row)` returns a truthy value

src/types/table.ts
portikM: Updated (`DataSelectState` -> `TableViewSelectState`)

src/components/KTableView/KTableView.vue
adamdehaven: Unless I'm missing something, there are still other places that aren't wrapped?

e.g. `getRowState`

src/components/KTableView/KTableView.cy.ts
```diff
@@ -291,11 +291,30 @@ describe('KTableView', () => {
       cy.get('th').eq(options.headers.indexOf(options.headers.find((header => header.key === 'actions'))!)).find('.table-header-label').should('have.class', 'sr-only')
     })
 
+    it('bulk actions in not enabled when rowKey prop is not provided', () => {
```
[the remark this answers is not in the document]
adamdehaven: I'm not sure what KTruncate does here? The issue is specific to KTableView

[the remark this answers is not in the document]
adamdehaven: I'm talking about a test that ensures the table applies state to the correct row based on the keys like mentioned above in @Justineo's example

src/components/KTableView/KTableView.vue
portikM: In other places we don't care if `getRowKey` returns an empty string since in **all instances** we are comparing it to `dataSelectState[n].rowKey` which is never going to be an empty string

src/components/KTableView/KTableView.cy.ts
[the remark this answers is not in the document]
portikM: So what I am suggesting is replicating the scenario in the video to make sure that after removing a table row the table applies state to correct row.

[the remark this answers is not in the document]
adamdehaven: If you can do so as a KTableView test 👍🏼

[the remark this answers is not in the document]
portikM: Oh, I think you meant adding a test scenario to cover that? Sorry I misunderstood, cuz what I suggested is adding a scenario in sandbox to test that but maybe a test scenario is a better idea.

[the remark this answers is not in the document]
adamdehaven: > a component test scenario is a better idea
👍🏼

adamdehaven (approved):

PR #2433, opened 2024-10-02
fix(kmultiselect): arrow keys navigation [KHCP-13121]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-13121

Changes:
- allow navigating through KMultiselect items using keyboard up and down keys
- fix a bug in KSelect where navigating using up and down keys was working incorrectly when items were passed in indirect order


aanchalm01 (approved):
LGTM! 🚢

PR #2443, opened 2024-10-08
fix(kmultiselect): add item-creation-validator prop [KHCP-13541]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-13541

Adds `itemCreationValidator` prop for validating user input when `enableItemCreation` is `true`


filipgutica (dismissed):

kaiarrowood (approved):
LGTM! 🚀

PR #2446, opened 2024-10-09
fix(ktable): bulk actions checkboxes should be valign center

Justineo:
# Summary

Before:

image

After:

image

### DIFF
--- a/src/components/KCheckbox/KCheckbox.vue
+++ b/src/components/KCheckbox/KCheckbox.vue
@@ -192,7 +192,7 @@ export default {
     position: relative;
 
     &.has-label {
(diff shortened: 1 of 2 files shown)


src/components/KCheckbox/KCheckbox.vue
```diff
@@ -192,7 +192,7 @@ export default {
     position: relative;
 
     &.has-label {
-      margin-top: 3px; // align with label
+      margin-top: 2px; // align with label
```
Justineo, line 195: (20 - 16) / 2

adamdehaven (approved):

PR #2448, opened 2024-10-09
fix(kinput): add show-masking-toggle prop [KHCP-13466]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-13466

Add `showMaskingToggle` prop which allows toggling masking on and off when `type="password"`

Screenshot 2024-10-10 at 11 20 36 AM
Screenshot 2024-10-10 at 11 20 24 AM


Justineo (commented):
We may want to turn off the built-in password reveal for Edge in case its toggle button overlaps with ours. (Edge has a strategy that only shows the toggle button if you start typing from nothing.)

image

https://learn.microsoft.com/en-us/microsoft-edge/web-platform/password-reveal#remove-the-password-reveal-control

Justineo:
And long text is overlapping with the button.

image

src/components/KInput/KInput.vue
```diff
@@ -35,18 +35,37 @@
...
+          @click.stop="maskValue = !maskValue"
+        >
+          <VisibilityIcon
+            v-if="maskValue"
+            decorative
+          />
+          <VisibilityOffIcon
+            v-else
+            decorative
+          />
```
Justineo, line 62: Another question: should the icon reflect the current status or action? Current implementation is opposite to our login page's password input:

image

portikM:
> We may want to turn off the built-in password reveal for Edge in case its toggle button overlaps with ours. (Edge has a strategy that only shows the toggle button if you start typing from nothing.)
> 
> image
> https://learn.microsoft.com/en-us/microsoft-edge/web-platform/password-reveal#remove-the-password-reveal-control

Fixed

src/components/KInput/KInput.vue
portikM: Oh good catch -- I looked at some other website to look up how they handle that but aligning with our login page makes more sense. Fixed.

Justineo (approved):
LGTM 👍

adamdehaven (approved):

PR #2450, opened 2024-10-11
fix(kselect): accept null in v-model [KHCP-13685]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-13685

Typescript checks are getting tighter in v5.5.4, `v-model` now differentiates between `undefined` and `null` support. Since the backend APIs only recognize `null` as a value to unset an entry in the DB (particularly true for booleans) and not `undefined`, allow passing `null` as the `ModelValue` (treat passing a `null` the same as passing `undefined`)


kaiarrowood (dismissed):
LGTM! Thanks Max!

src/components/KInputSwitch/KInputSwitch.vue
```diff
@@ -57,7 +57,7 @@ const props = defineProps({
   * Sets whether or not toggle is checked
   */
   modelValue: {
-    type: Boolean,
+    type: [Boolean, null],
```
adamdehaven, line 60: If we made this change, it would need to be `Null` not `null`

Applies to all instances

adamdehaven (changes requested):
issue: we should not be updating OSS component behavior based on Kong APIs that will change over time. 

The host app should be responsible for handling sending the proper values to the API

portikM:
> * Should we consider foo a controlled prop if `typeof props.foo !== 'undefined'`?

But that's basically what we are already doing and doesn't help us at all, plus it doesn't work for all Kongponents (KSelect)

The problem with this approach is it doesn't save us any work, we still have to do the conversion from away from `null` with the  `GET` response and again convert back to `null` before we build the payload for `PUT`

adamdehaven:
> The problem with this approach is it doesn't save us any work, we still have to do the conversion from away from `null` with the `GET` response and again convert back to `null` before we build the payload for `PUT`

This should be expected then. 

We shouldn't be changing the component behavior based on the requirements of our APIs. Just as certain endpoints may require `null`, others may require `undefined` or have different logic, and that logic could change over time.

Mutating data to be acceptable by an external interface (e.g. Kong APIs) should be the responsibility of the host application.

Justineo:
> > Should we consider foo a controlled prop if typeof props.foo !== 'undefined'?
>
> But that's basically what we are already doing and doesn't help us at all, plus it doesn't work for all Kongponents (KSelect)

I wasn't specifically referring to how accepting null makes it easier to connect to our backend API. Rather, if we're using `undefined` to differentiate between controlled and uncontrolled props, then `null` might be necessary to represent an explicitly empty value. And we should think about what `null` value represents for each component. Like what does `null` even mean for Checkbox or Switch?

> We shouldn't be changing the component behavior based on the requirements of our APIs.

I agree. However, in this case, I was thinking less about it being a backend requirement and more about whether we should generally allow `null` values.

src/components/KInputSwitch/KInputSwitch.vue
portikM: It's `null` not `Null`

Screenshot 2024-10-16 at 1 32 47 PM

portikM:
Okay, so I've updated this PR to support `null` in KSelect and KMultiselect only. I've tested both components in sandbox and they looked good.

portikM:
Okay, so I've updated this PR to support `null` in KSelect and KMultiselect only. I've tested both components in sandbox and they looked good.

cc @kaiarrowood @Justineo @adamdehaven

portikM:
Closing it for now since it looks like we don't need it for Vue 3.5 upgrade like we thought we did

PR #2453, opened 2024-10-12
fix(kinput): prevent focus loss on icon/button click

Justineo:
# Summary

Fixed issue where clicking icons/buttons inside an input caused focus loss. This PR ensures the input remains focused.

Before:

Oct-13-2024 01-04-21
Oct-13-2024 01-04-36

After:

Oct-13-2024 01-04-27
Oct-13-2024 01-04-31

### DIFF
--- a/src/components/KInput/KInput.cy.ts
+++ b/src/components/KInput/KInput.cy.ts
@@ -188,11 +188,11 @@ describe('KInput', () => {
 
     cy.get('.k-input .mask-value-toggle-button').should('be.visible')
     cy.get('.k-input input').should('have.attr', 'type', 'password')
+    cy.get('.k-input input').focus()
     cy.get('.k-input .mask-value-toggle-button').click()
     cy.get('.k-input input').should('have.attr', 'type', 'text')
     cy.get('.k-input .mask-value-toggle-button').click()
-    cy.get('.k-input input').should('have.attr', 'type', 'password')
-
+    cy.get('.k-input input').should('have.attr', 'type', 'password').should('be.focused')
     // user-provided after slot should be rendered
     cy.get('.k-input').find(`[data-testid="${afterSlot}"]`).should('not.exist')
   })
--- a/src/components/KInput/KInput.vue
+++ b/src/components/KInput/KInput.vue
@@ -51,6 +51,8 @@
           :aria-label="`${maskValue ? 'Hide' : 'Show'} value`"
           class="mask-value-toggle-button"
           @click.stop="maskValue = !maskValue"
+          @mousedown.prevent
+          @mouseup.prevent
         >
           <VisibilityOffIcon
             v-if="maskValue"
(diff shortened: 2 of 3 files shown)


src/components/KInput/KInput.vue
```diff
@@ -51,6 +51,8 @@
           :aria-label="`${maskValue ? 'Hide' : 'Show'} value`"
           class="mask-value-toggle-button"
           @click.stop="maskValue = !maskValue"
+          @mousedown.prevent
```
Justineo, line 54: This makes the input stay focusd.

src/components/KInput/KInput.vue
```diff
@@ -51,6 +51,8 @@
...
+          @mouseup.prevent
```
Justineo, line 55: This keeps the cursor position.

src/components/KInput/KInput.cy.ts
```diff
@@ -188,11 +188,11 @@ describe('KInput', () => {
 
     cy.get('.k-input .mask-value-toggle-button').should('be.visible')
     cy.get('.k-input input').should('have.attr', 'type', 'password')
+    cy.get('.k-input input').focus()
     cy.get('.k-input .mask-value-toggle-button').click()
     cy.get('.k-input input').should('have.attr', 'type', 'text')
     cy.get('.k-input .mask-value-toggle-button').click()
-    cy.get('.k-input input').should('have.attr', 'type', 'password')
-
+    cy.get('.k-input input').should('have.attr', 'type', 'password').should('be.focused')
```
Justineo, line 195: Note: in real user interactions, `mouseup` moves the input cursor to the start position, but I couldn't reproduce this behavior in test cases, so this part of the test is not included.

src/components/KInput/KInput.vue
```diff
@@ -384,6 +386,12 @@ $kInputSlotSpacing: var(--kui-space-40, $kui-space-40); // $kSelectInputSlotSpac
         width: $kInputIconSize !important;
       }
 
+      // enhance the experience for most common cases that icon only slots should not
+      // prevent the input from being focused by click on the icon
+      &:has(> #{$kongponentsKongIconSelector}:not(button):not([role="button"]):only-child) {
+        pointer-events: none;
+      }
```
Justineo, line 393: I checked our usage of the `before`/`after` slot across codebases, in all cases for `before` slot we are using single icon and in `after` slot there's only [one place](https://github.com/Kong/public-ui-components/blob/83a06bbe5e57241d3fb4d93f95ec834c9f150572/packages/entities/entities-shared/src/components/entity-filter/EntityFilter.vue#L14-L20) we are using clickable icon that does not have a `role="button"` set. We should fix that first.

Justineo:
Temporarily blocked by [an unexpected usage](https://github.com/Kong/public-ui-components/blob/83a06bbe5e57241d3fb4d93f95ec834c9f150572/packages/entities/entities-shared/src/components/entity-filter/EntityFilter.vue#L14-L20) according to the [docs](https://kongponents.konghq.com/components/input.html#input:~:text=you%20must%20assign%20role%3D%22button%22%20and%20tabindex%3D%220%22%20attributes). Should fix that first.

> If you want to make an icon clickable, you **must** assign `role="button"` and `tabindex="0"` attributes to that element and bind an event handler. KBadge will take care of state styling (hover, focus, disabled).

PR #2469, opened 2024-10-18
feat(kbutton): add appearance="none" to button

Justineo:
# Summary

This PR adds a `none` value to the `<KButton>` `appearance` prop, enabling buttons to render with minimal styling similar to a `<div>`. This allows for fully custom styles while retaining `<KButton>` features like link rendering to links, cursor styles, and disabled states.

## Changes

- New Appearance: `none`
  - Renders `<KButton>` with minimal styles (`border-box`, `cursor`) for custom styling.

- Style Refactoring
  - `.k-button` no longer applies primary styles unless the `primary` class is present to make overriding easier.
  - Output `primary outline` classes to maintain backward compatibility when `appearance="outline"` (or any other invalid values).

- Disabled Link Behavior
  - Disabled links are now unfocusable and cannot be activated with the <kbd>enter</kbd> key.
  - Removed `pointer-events: none` to ensure `cursor: not-allowed` displays correctly for consistency.
  - Used JavaScript to prevent navigation and click events on disabled links, emulating native button behavior.

### DIFF
--- a/docs/components/button.md
+++ b/docs/components/button.md
@@ -21,6 +21,10 @@ The Button component can take 1 of 5 appearance values:
 - `danger`
 - `none`
 
+:::tip NOTE
+`appearance="none"` is used in scenarios where custom styles are needed while retaining the basic behavior of `KButton`. When enabled, the size and icon props will not be supported and do not need to be supported.
+:::
+
 <div class="spacing-container">
   <KButton appearance="primary">Primary</KButton>
   <KButton appearance="secondary">Secondary</KButton>
--- a/docs/components/button.md
+++ b/docs/components/button.md
@@ -13,18 +13,20 @@ and configuration options.
 
 ### appearance
 
-The Button component can take 1 of 4 appearance values:
+The Button component can take 1 of 5 appearance values:
 
 - `primary`
 - `secondary`
 - `tertiary`
 - `danger`
+- `none`
 
 <div class="spacing-container">
   <KButton appearance="primary">Primary</KButton>
   <KButton appearance="secondary">Secondary</KButton>
   <KButton appearance="tertiary">Tertiary</KButton>
   <KButton appearance="danger">Danger</KButton>
+  <KButton appearance="none">None</KButton>
 </div>
 
 ```html
@@ -32,6 +34,7 @@ The Button component can take 1 of 4 appearance values:
 <KButton appearance="secondary">Secondary</KButton>
 <KButton appearance="tertiary">Tertiary</KButton>
 <KButton appearance="danger">Danger</KButton>
+<KButton appearance="none">None</KButton>
 ```
 
 ### size
--- a/docs/components/button.md
+++ b/docs/components/button.md
@@ -22,7 +22,7 @@ The Button component can take 1 of 5 appearance values:
 - `none`
 
 :::tip NOTE
-`appearance="none"` is used in scenarios where custom styles are needed while retaining the basic behavior of `KButton`. When enabled, the size and icon props will not be supported and do not need to be supported.
+`appearance="none"` is used in scenarios where custom styles are needed while retaining the basic behavior of `KButton`. When enabled, the `size` and `icon` props will not be supported (and there's no need to support them).
 :::
 
 <div class="spacing-container">
--- a/docs/components/button.md
+++ b/docs/components/button.md
@@ -22,7 +22,9 @@ The Button component can take 1 of 5 appearance values:
 - `none`
 
 :::tip NOTE
-Use `appearance="none"` to apply custom styles while keeping KButton's basic functionality. When this is enabled, the `size` prop won’t work unless `icon` is set to `true`.
+Use `appearance="none"` to get an unstyled button, making it easier to customize from scratch. This removes the built-in styling but retains the button's functionality.
+
+When `appearance="none"` is set, the `size` prop only works if `icon` is `true`; otherwise, `size` has no effect.
 :::
 
 <div class="spacing-container">
--- a/docs/components/button.md
+++ b/docs/components/button.md
@@ -22,7 +22,7 @@ The Button component can take 1 of 5 appearance values:
 - `none`
 
 :::tip NOTE
-`appearance="none"` is used in scenarios where custom styles are needed while retaining the basic behavior of `KButton`. When enabled, the `size` and `icon` props will not be supported (and there's no need to support them).
+Use `appearance="none"` to apply custom styles while keeping KButton's basic functionality. When this is enabled, the `size` prop won’t work unless `icon` is set to `true`.
 :::
 
 <div class="spacing-container">
--- a/src/components/KButton/KButton.vue
+++ b/src/components/KButton/KButton.vue
@@ -2,7 +2,14 @@
   <component
     :is="buttonType"
     class="k-button"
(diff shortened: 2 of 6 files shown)


src/components/KButton/KButton.vue
```diff
@@ -192,23 +204,37 @@ export default {
...
+    }
+
+    &:active {
+      background-color: var(--kui-color-background-primary-strongest, $kui-color-background-primary-strongest);
+    }
+
+    &:disabled, &[disabled] {
+      background-color: var(--kui-color-background-disabled, $kui-color-background-disabled);
+      color: var(--kui-color-text-disabled, $kui-color-text-disabled);
+    }
```
Justineo, line 229: We're only moving the primary styles here to avoid having to override all of them for .none. Some of these styles have high specificity, which would make it difficult for the .none styles to take precedence, thus complicating user overrides.

The default size mixin doesn't need to change as we only add style to `.k-button` and will be overriden by a simple `all: unset`.

docs/components/button.md
```diff
@@ -13,25 +13,28 @@ and configuration options.
...
   <KButton appearance="danger">Danger</KButton>
+  <KButton appearance="none">None</KButton>
 </div>
 
 ```html
 <KButton appearance="primary">Primary</KButton>
 <KButton appearance="secondary">Secondary</KButton>
 <KButton appearance="tertiary">Tertiary</KButton>
 <KButton appearance="danger">Danger</KButton>
+<KButton appearance="none">None</KButton>
```
adamdehaven, line 37: Do we need a note here that using an appearance of "none" essentially disables other props like "size"?

src/components/KButton/KButton.vue
adamdehaven: I'm not sure I like moving the primary styles here out of the mixin.

Couldn't we change the mixin implementation to something like this?

```scss
.k-button {
  &:not(.none) {
    @include kButtonPrimaryAppearance;
  }
}
```

src/components/KButton/KButton.vue
```diff
@@ -282,6 +308,20 @@ export default {
     }
   }
 
+  // reset the style of a button to bare minimum
+  // using :where() here to make style override easier
+  &:where(.none) {
```
adamdehaven, line 313: I still think a `disabled` button with `appearance="none"` needs to have the disabled style applied whereby it looks "greyed out" etc.

image

The most common use case we have currently for `appearance="none"` is a button wrapping an icon, so we'd still want the disabled styles here.

src/components/KButton/KButton.vue
```diff
@@ -110,6 +130,25 @@ const strippedAttrs = computed((): typeof attrs => {
...
+  }
+
+  // Only prevent clicks on disabled links, emulating the same behavior as a disabled button
+  return {
+    clickCapture: stop,
+    dblclickCapture: stop,
+    mousedownCapture: stop,
+    mouseupCapture: stop,
+  }
+})
```
adamdehaven, line 150: I can still right-click on a disabled button in the sandbox with a `to` prop, is that intended?

image

docs/components/button.md
Justineo: I think so.

src/components/KButton/KButton.vue
adamdehaven: Actually... 

Now that you modified the buttonAppearance to always include a valid appearance (meaning, if undefined or an invalid appearance is added) I think this is fine as long as the fallback always applies these `primary` styles.

Going to resolve this comment.

Justineo: By moving the styles into `.primary`, they only output once. Do you think there are downsides by doing so?

adamdehaven: That should be fine. 

The original reason the `.k-button` had the `primary` mixin applied was to ensure that if a user passed an undefined or invalid `appearance` prop, the default `primary` styles would still be applied to the KButton element.

Since you updated the buttonAppearance to be computed and always fallback to `primary` it has the same effect.

cc @portikM

src/components/KButton/KButton.vue
Justineo: I see. What I was trying to do is create a `<button>` that looks like a plain `<div>` but still aligns with our core functionality and behaviors. If we want it to be styled out-of-the-box maybe we should take another approach? In my previous design language project we have text buttons and icon buttons, which can be used inside Table rows or as close buttons for Dialogs.

adamdehaven: I think the only exception would be here for `disabled` -- removing all other styles (e.g. size, etc.) is what I would expect.

adamdehaven: We also still want the `cursor: pointer` on `appearance="none"` as well if it's not there already

adamdehaven: ...and maybe the border radius for the `focus-visible` state 😬

src/components/KButton/KButton.vue
adamdehaven: If it's simply a button (e.g. with a `@click` handler) I think just preventing the click as you have it is sufficient.

For links, I agree, just removing the link itself may be more ideal; however, at that point rendering an `<a>` tag without the `href` is not semantically correct.

Instead, we could do something like this?

```html
<a 
  href="https://example.com" 
  aria-disabled="true" 
  @click.prevent="">Disabled Link</a>
```

Justineo: > Instead, we could do something like this?
> 
> ```
> <a 
>   href="https://example.com" 
>   aria-disabled="true" 
>   @click.prevent="">Disabled Link</a>
> ```

This leaves the URL visible, I think we should not expose it to user if it's unavailable:

image

Or shall we just follow ReactAria's approach? They seem to have put tons of efforts on A11Y.

adamdehaven: I don't think "exposing" the URL is an issue, it seems like the use-case is just not allowing them click it. If it was a security concern, you'd remove it from the DOM

src/components/KButton/KButton.vue
Justineo: Added `border-radius` (for other appearances it's related to `size` but here I just add a medium sized one) and disabled color.

docs/components/button.md
Justineo: Added a NOTE after this.

src/components/KButton/KButton.vue
Justineo: `<a role="link">` seem to be fine (at lease in Chrome's Accessibility Tree):

image

It's rendered the same way as `<a href>` in the a11y tree.

`<a href>` should be equal to `[role="link"]` as described [here](https://www.w3.org/TR/html-aria/#dfn-naming-prohibited:~:text=href-,role%3Dlink,-Roles%3A%20button).

adamdehaven:
When we utilize an icon button, or button with icon, and with `appearance="none"` I think we should still persist _some_ of the styling

image
image

As examples here, ideally, the button would still align with the other buttons in the row, just without the background color, etc.

We'd still want the gap between the icon + text, and the padding around the icon (probably?)

src/components/KButton/KButton.vue
Justineo: Done. Need a final review. @adamdehaven

portikM (dismissed):
LGTM 🚀

adamdehaven (approved):

PR #2476, opened 2024-10-23
refactor(kpop): improve kpop performance on page scrolling

2eha0:
# Summary
`autoUpdate` starts in the `KPop` component’s mounted hook, continuously calculating the position even when the pop-up is not visible, significant lag when there are many `Kpop` on the page

**Before:**

https://github.com/user-attachments/assets/9936e1b2-3ff3-480c-ba45-435447eca7d6

**After:**

https://github.com/user-attachments/assets/5c2704be-8dc0-44ba-be8a-365984727516

### DIFF
--- a/src/components/KPop/KPop.vue
+++ b/src/components/KPop/KPop.vue
@@ -185,6 +185,22 @@ const togglePopover = () => {
   }
 }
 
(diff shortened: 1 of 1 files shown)


Justineo (approved):
LGTM ⚡️

PR #2480, opened 2024-10-25
fix(ktabs): slotted anchor link

portikM:
# Summary

The issue:
When a link is slotted into anchor slot of a tab, when using keyboard nav, the user needs to press tab twice to navigate to next element (first time the wrapper element `div.tab-link` has focus, then the slotted link element).

Solution:
Add `anchorTabindex` prop that allows set tabindex for `div.tab-link` element. This allows to make only one element in tab focusable.

https://github.com/user-attachments/assets/8d37cb5a-548b-42fd-8652-465be6546780

### DIFF
--- a/docs/components/table-view.md
+++ b/docs/components/table-view.md
@@ -518,11 +518,28 @@ The `rowKey` prop accepts either a unique string or a function that returns a un
 If a string is provided which corresponds to a property of the `row`, the unique identifier will utilize the `row[rowKey]` as the unique identifier.
 
 ```html
-<KTableView
-  row-key="id"
-  :data="tableData"
-  :headers="headers"
-/>
+<template>
+  <KTableView
+    row-key="id"
+    :data="tableData"
+    :headers="headers"
+  />
+</template>
+
+<script setup lang="ts">
+import type { TableViewData } from '@kong/kongponents'
+
+const tableData: TableViewData = [
+  {
+    id: 'a70642b3-20f2-4658-b459-fe1cbcf9e315',
+    ...
+  },
+  {
+    id: '58c599a9-f453-41f3-9e64-0a7fc6caedad',
+    ...
+  }
+]
+</script>
 ```
 
 Alternatively, if a function is passed, it allows for the creation of a custom identifier based on the row data passed to the function.
--- a/docs/components/tabs.md
+++ b/docs/components/tabs.md
@@ -178,7 +178,7 @@ const tabChange = (hash: string): void => {
 
 ### anchorTabindex
 
-This prop allows setting a custom `tabindex` for the tab anchor element. It’s useful when passing a custom interactive element, like a link, through the [`anchor` slot](#anchor-panel), ensuring that only the slotted element is focusable by resetting the default anchor `tabindex`. Default value is `0`.
+This prop allows setting a custom `tabindex` for the tab anchor element. It’s useful when passing a custom interactive element, like a link, through the [`anchor` slot](#anchor-panel), ensuring that only the slotted element is focusable by resetting the default anchor `tabindex`. Default value is `0`. Accepted values are `0` and `-1`.
 
 #### Dynamic RouterView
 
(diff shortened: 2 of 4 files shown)


docs/components/table-view.md
```diff
@@ -518,11 +518,28 @@ The `rowKey` prop accepts either a unique string or a function that returns a un
...
+  {
+    id: 'a70642b3-20f2-4658-b459-fe1cbcf9e315',
+    ...
+  },
+  {
+    id: '58c599a9-f453-41f3-9e64-0a7fc6caedad',
+    ...
+  }
+]
+</script>
```
portikM, line 542: Just making an example  easier to understand

src/components/KTabs/KTabs.vue
```diff
@@ -75,6 +75,11 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+  anchorTabindex: {
+    type: String,
+    default: '0',
+    validator: (val: string): boolean => val === '0' || val === '-1',
```
adamdehaven, line 81: question: tabindex is always a number (even though it's passed as a string), why is this typed as a string? I think it might be easier to validate the provided value if the prop itself was a number.

e.g. if the user passes `my-wrong-value` as a string, the validator will error; however, you'd still be passing the invalid value directly to the HTML. I think this should be prevented.

Also, I'm not sure the validator makes sense either given the description of the prop. The logic of the validator prevents me from passing a tabindex of any value other than `0` or `-1`

src/components/KTabs/KTabs.vue
```diff
@@ -141,19 +156,20 @@ watch(() => props.modelValue, (newTabHash) => {
         }
 
         a, :deep(a) {
-          color: var(--kui-color-text, $kui-color-text);
+          color: var(--kui-color-text-neutral, $kui-color-text-neutral);
```
adamdehaven, line 159: why the change here?

src/components/KTabs/KTabs.vue
portikM: Addressed. Added a check to make sure the passed value is either `0` or `-1` and made the type more strict

adamdehaven: why can the value only be `0` or `-1`? if we have a prop to customize the tabindex, why can't we allow any value?

portikM: I forgot it can be other values 😅
Fixed now, accepts value `-1 <= x <= 32767` ([mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex))

docs/components/tabs.md
```diff
@@ -176,6 +176,10 @@ const tabChange = (hash: string): void => {
 </script>
 ```
 
+### anchorTabindex
+
+This prop allows setting a custom `tabindex` for the tab anchor element. It’s useful when passing a custom interactive element, like a link, through the [`anchor` slot](#anchor-panel), ensuring that only the slotted element is focusable by resetting the default anchor `tabindex`. Default value is `0`.
```
adamdehaven, line 181: The docs don't match the interface of the new prop. Here you're describing the default as a `number` `0` but the actual interface only allows for a string `'0'`

src/components/KTabs/KTabs.vue
```diff
@@ -90,6 +95,16 @@ const handleTabChange = (tab: string): void => {
...
 }
 
+const getTabSlotName = (tabHash: string): string => tabHash.replace('#', '')
+
+const getAnchorTabindex = (tab: Tab): string => {
+  if (tab.disabled) {
+    return '-1'
+  }
+
+  return props.anchorTabindex >= -1 && props.anchorTabindex <= 32767 ? String(props.anchorTabindex) : '0'
```
adamdehaven, line 105:
```suggestion
  return typeof props.anchorTabindex === 'number' && props.anchorTabindex >= -1 && props.anchorTabindex <= 32767 ? String(props.anchorTabindex) : '0'
```

adamdehaven (approved):
lgtm

PR #2484, opened 2024-10-28
docs(ktextarea): to clarify maximum characters of ktextarea for readers

Miltonbhowmick:
it gives hint of maximum characters for textarea by default.

# Summary
This PR enhances the docs of KTextArea by adding a note of default maximum 2048 characters limit.

### DIFF
--- a/docs/components/textarea.md
+++ b/docs/components/textarea.md
@@ -1,6 +1,9 @@
 # TextArea
 
 KTextArea provides a wrapper around textarea element, as well as contextual styling and states (error, focus, etc).
+:::tip NOTE
+The default maximum character limit is `2048 characters`.
+:::
 
 <KTextArea />
 


docs/components/textarea.md
```diff
@@ -1,6 +1,9 @@
 # TextArea
 
 KTextArea provides a wrapper around textarea element, as well as contextual styling and states (error, focus, etc).
+:::tip NOTE
```
adamdehaven, line 4: The limit is already documented in the `characterLimit` prop docs; is this sufficient? Or are you suggesting it needs to more prominent?

Justineo: Do you know why do we have this default limit? It seems to be uncommon so I think it’s reasonable to make it a little more prominent.

adamdehaven: IIRC it’s a limitation on some arbitrary endpoint we own or maybe our “normal” threshold. I don’t mind adding the extra note; we should just update this change to include a link to the corresponding prop in the note

Justineo: Maybe it’s better to remove this default limit in the future and make it injectable. Then we can provide meaningful defaults in our AppShell.

Miltonbhowmick: Thank you for the feedback, @adamdehaven and @Justineo.
I agree that making the character limit more prominent in the documentation could benefit users, especially since this limit might be unexpected. Even if it’s standard in this system or based on specific endpoint constraints, documenting it directly in the main description section (alongside the characterLimit prop link) would ensure users understand it early, without needing to search for details in the prop list.

Adding this information up front also helps make it clear that the default limit could be adjusted in the future, as discussed. This would set the stage for a more flexible limit if we later decide to make it configurable.

PR #2485, opened 2024-10-28
fix(kselect): add item-creation-validator prop [KHCP-13862]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-13862

Adds `itemCreationValidator` prop for validating user input when `enableItemCreation` is `true`


mihai-peteu (approved):

docs/components/select.md
```diff
@@ -372,6 +372,67 @@ When used in conjunction with `enableFiltering` set to `true`, KSelect will sugg
 <KSelect enable-item-creation enable-filtering placeholder="Try searching for 'service d'" :items="selectItems" />
 ```
 
+### itemCreationValidator
+
+Prop for passing a function for input validation when item creation is enabled. The function takes query input string as a single parameter and must return a `boolean` value. When a function passed through `itemCreationValidator` returns `false`, the _Add new value_ button will be disabled.
```
adamdehaven, line 377:
```suggestion
Prop for passing a function for input validation when item creation is enabled. 

The function takes the query input string as a single parameter and must return a `boolean` value. 

When the function passed through `itemCreationValidator` returns `false`, the "Add new value" button will be disabled.
```

src/components/KSelect/KSelect.vue
```diff
@@ -333,6 +333,13 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+  /**
+   * Validator function for item creation.
+   */
+  itemCreationValidator: {
+    type: Function,
```
adamdehaven, line 340:
```suggestion
    type: Function as PropType<() => boolean>,
```

adamdehaven (approved):

PR #2491, opened 2024-10-30
fix(kbutton): pass to as long as it's a router-link [KM-644]

TT1228:
# Summary

KM-644

`to` prop needs to be passed to `<router-link>`. So when `disabled` is passed, we still pass `to` prop to it and delete `href` attribute so that users cannot bypass this grey state by right clicking and open it in a new tab.

### DIFF
--- a/cypress/support/index.ts
+++ b/cypress/support/index.ts
@@ -1,5 +1,5 @@
 import { mount } from 'cypress/vue'
-//  vue-router only throws uncaught errors in production mode
+// vue-router only throws uncaught errors in production mode
 import { createMemoryHistory, createRouter } from 'vue-router/dist/vue-router.prod.cjs'
 import type { RouteRecordRaw } from 'vue-router'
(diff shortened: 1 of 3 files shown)


adamdehaven (commented):
Can you add test coverage to ensure disabled/non-disabled work and render as expected?

TT1228:
> Can you add test coverage to ensure disabled/non-disabled work and render as expected?

I'm working on it

cypress/support/index.ts
```diff
@@ -1,10 +1,23 @@
 import { mount } from 'cypress/vue'
+//  vue-router only throws uncaught errors in production mode
```
Justineo, line 2:
```suggestion
// vue-router only throws uncaught errors in production mode
```

Justineo (changes requested):

Justineo (approved):

Justineo (approved):
LGTM

adamdehaven (approved):

PR #2493, opened 2024-10-30
fix(kselect): don't emit query-change on select [KHCP-13895]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-13895

Selecting a value in KSelect should not trigger `@query-change` event

### DIFF
--- a/src/components/KSelect/KSelect.vue
+++ b/src/components/KSelect/KSelect.vue
@@ -717,7 +717,9 @@ watch(() => props.items, (newValue, oldValue) => {
 }, { deep: true, immediate: true })
 
 watch(filterQuery, (query: string) => {
-  if (!skipQueryChangeEmit.value || !query || uniqueFilterQuery.value) {
+  const isUnique = selectItems.value?.filter((item: SelectItem) => item.label === query)?.length
+
+  if (!skipQueryChangeEmit.value || !query || isUnique) {
     emit('query-change', query)
   }
   skipQueryChangeEmit.value = false
--- a/src/components/KSelect/KSelect.vue
+++ b/src/components/KSelect/KSelect.vue
@@ -717,11 +717,11 @@ watch(() => props.items, (newValue, oldValue) => {
 }, { deep: true, immediate: true })
 
 watch(filterQuery, (query: string) => {
-  if (!skipQueryChangeEmit.value) {
+  if (!skipQueryChangeEmit.value || !query || uniqueFilterQuery.value) {
     emit('query-change', query)
   }
   skipQueryChangeEmit.value = false
-}, { immediate: true })
+})
 
 watch(selectedItem, (newVal, oldVal) => {
   if (newVal && newVal !== oldVal) {
--- a/src/components/KSelect/KSelect.vue
+++ b/src/components/KSelect/KSelect.vue
@@ -379,6 +379,7 @@ const labelElement = ref<InstanceType<typeof KLabel> | null>(null)
 
 const strippedLabel = computed((): string => stripRequiredLabel(props.label, isRequired.value))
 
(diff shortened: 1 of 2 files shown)


filipgutica (approved):

filipgutica (dismissed):

src/components/KSelect/KSelect.vue
```diff
@@ -713,9 +715,12 @@ watch(() => props.items, (newValue, oldValue) => {
   }
 }, { deep: true, immediate: true })
 
-watch(filterQuery, (q: string) => {
-  emit('query-change', q)
-})
+watch(filterQuery, (query: string) => {
+  if (!skipQueryChangeEmit.value || !query) {
+    emit('query-change', query)
+  }
```
adamdehaven, line 721: question: why would we not emit when the query is emptied? 

If a user searches, emit. If they then clear the search, why wouldn't we emit that it changed again?

portikM: We do emit when it's empty

adamdehaven: Doesn't the if statement above say `|| !query` ? 

if skip is `true` it will emit, or if there's a query it will emit.

src/components/KSelect/KSelect.vue
```diff
@@ -379,6 +379,7 @@ const labelElement = ref<InstanceType<typeof KLabel> | null>(null)
 
 const strippedLabel = computed((): string => stripRequiredLabel(props.label, isRequired.value))
 
+const skipQueryChangeEmit = ref<boolean>(false)
```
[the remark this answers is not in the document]
portikM: Done

mihai-peteu (approved):

src/components/KSelect/KSelect.vue
portikM: No, if `!skip` or `!query` -> emit..

mihai-peteu (changes requested):
still pending comments

src/components/KSelect/KSelect.vue
portikM: Here is verifiation

https://github.com/user-attachments/assets/59229d33-f139-4b7e-b8e1-e607e98e67a2

src/components/KSelect/KSelect.vue
```diff
@@ -713,9 +716,12 @@ watch(() => props.items, (newValue, oldValue) => {
...
 
-watch(filterQuery, (q: string) => {
-  emit('query-change', q)
-})
+watch(filterQuery, (query: string) => {
+  if (!skipQueryChangeEmit.value) {
+    emit('query-change', query)
+  }
+  skipQueryChangeEmit.value = false
+}, { immediate: true })
```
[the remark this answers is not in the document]
portikM: Yes, that is more of backwards compatibility thing. In this PR I'm removing an emit on line 580 that used to fire every time input is in focus. What it was doing previously is fire this event with empty string (since we're resetting it) pretty much every time the dropdown opens. The component test caught this. That's why I added `immediate: true` to still have that "initial" emit. I was doubting between keeping it and removing it but then I thought knowing the initial value right away doesn't hurt?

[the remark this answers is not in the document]
adamdehaven: isn't an initial emit signaling that the query changed even though it didn't?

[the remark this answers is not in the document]
portikM: Well yeah I guess. I remove the `immediate: true` now but that created a problem where the component wasn't emitting when the value is refreshed on dropdown open so I had to tweak the condition in this watcher. Please re-review.

mihai-peteu (approved):

adamdehaven (approved):

PR #2504, opened 2024-11-05
fix(kradio): show-card-radio and card-orientation props [KHCP-13540]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-13540

Adds `cardOrientation` and `showCardRadio` props

Screenshot 2024-11-05 at 12 40 40 PM
Screenshot 2024-11-05 at 12 59 01 PM
Screenshot 2024-11-05 at 12 41 02 PM

### DIFF
--- a/docs/components/radio.md
+++ b/docs/components/radio.md
@@ -221,7 +221,7 @@ Accepted values are `vertical` (shown above) or `horizontal`. Defaults to `verti
 />
 ```
 
-### showCardRadio
+### cardRadioVisible
 
 Prop to show or hide radio button in card. Default value is `true`.
 
@@ -229,7 +229,7 @@ Prop to show or hide radio button in card. Default value is `true`.
   <KRadio
     v-model="showRadio"
     card
-    :card-show-radio="false"
+    :card-radio-visible="false"
     label="Vertical"
     description="Vertical radio card with hidden radio button."
     :selected-value="false"
@@ -238,7 +238,7 @@ Prop to show or hide radio button in card. Default value is `true`.
     v-model="showRadio"
     card
     card-orientation="horizontal"
-    :card-show-radio="false"
+    :card-radio-visible="false"
     label="Horizontal"
     description="Horizontal radio card with hidden radio button."
     :selected-value="true"
@@ -247,7 +247,7 @@ Prop to show or hide radio button in card. Default value is `true`.
 
 ```html
 <KRadio
-  :card-show-radio="false"
+  :card-radio-visible="false"
   v-model="radioCard"
   card
   label="Vertical"
@@ -255,7 +255,7 @@ Prop to show or hide radio button in card. Default value is `true`.
   :selected-value="false"
 />
 <KRadio
-  :card-show-radio="false"
+  :card-radio-visible="false"
   v-model="radioCard"
   card
   card-orientation="horizontal"
--- a/docs/components/radio.md
+++ b/docs/components/radio.md
@@ -215,8 +215,8 @@ Accepted values are `vertical` (shown above) or `horizontal`. Defaults to `verti
   card-orientation="horizontal"
   v-model="cardRadio"
   description="Choose this option if you want your APIs to be publicly accessible by anyone on the internet."
-  card
   label="Public"
+  card
   selected-value="public"
 />
 ```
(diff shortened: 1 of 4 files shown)


docs/components/radio.md
```diff
@@ -171,6 +171,98 @@ const cardRadio = ref<string>('')
 </script>
 ```
 
+### cardOrientation
+
+Prop for changing card orientation. Default value is `vertical`, the alternative to which is `horizontal`.
```
adamdehaven, line 176:
```suggestion
Used alongside the [`card` prop](/components/radio#card) to set the orientation of the card layout.

Accepted values are `vertical` (shown above) or `horizontal`. Defaults to `vertical`.
```

docs/components/radio.md
```diff
@@ -171,6 +171,98 @@ const cardRadio = ref<string>('')
...
+  </KBadge>
+</KRadio>
+<KRadio
+  v-model="horizontalCard"
+  card
+  card-orientation="horizontal"
+  description="Choose this option if you want your APIs to be publicly accessible by anyone on the internet."
+  label="Public"
+  selected-value="public"
+/>
```
adamdehaven, line 197: _(I'm not sure anything can be done here, just pointing it out)_

Using this docs layout as an example, I'm not sure it looks great that the text between the two horizontal cards doesn't line up when the Recommended badge is present
image

Notice the text in the second card extends further to the right. If I make the text longer it's more noticable:

image

docs/components/radio.md
```diff
@@ -171,6 +171,98 @@ const cardRadio = ref<string>('')
...
+  card-orientation="horizontal"
+  v-model="cardRadio"
+  description="Choose this option if you want your APIs to be publicly accessible by anyone on the internet."
+  card
+  label="Public"
+  selected-value="public"
+/>
+```
+
+### showCardRadio
```
adamdehaven, line 222: suggestion/nit: All of the other `card` props are prefixed with `card*` -- could we rename this prop to `cardRadioVisible` so that it is more similar?

src/components/KRadio/KRadio.vue
```diff
@@ -63,18 +64,23 @@
...
+          {{ label }}
+        </span>
+        <span
+          v-if="showCardDescription"
+          class="radio-description"
+        >
+          <slot name="description">
+            {{ description }}
+          </slot>
+        </span>
```
adamdehaven, line 83: For this whole section (unsure if I highlighted all the correct lines) should we utilize `div` instead of `span`? 

Since the user can slot any content into the `description` slot, we need to support block-level elements

src/components/KRadio/KRadio.vue
```diff
@@ -198,6 +212,8 @@ const kRadioClasses = computed((): Record<string, boolean> => {
     'input-error': props.error,
     checked: isChecked.value,
     'has-description': showDescription.value,
+    'card-horizontal': props.card && props.cardOrientation === 'horizontal',
+    'card-vertical': props.card && props.cardOrientation === 'vertical',
```
adamdehaven, line 216: issue: Since `vertical` is the default, you need to edit this logic a bit to ensure it is applied if the user passes an invalid value.
```suggestion
    'card-horizontal': props.card && props.cardOrientation === 'horizontal',
    // Add vertical class for `vertical` or an invalid prop value
    'card-vertical': props.card && props.cardOrientation !== 'horizontal',
```

adamdehaven (changes requested):

docs/components/radio.md
```diff
@@ -171,6 +171,98 @@ const cardRadio = ref<string>('')
...
+</KRadio>
+<KRadio
+  card-orientation="horizontal"
+  v-model="cardRadio"
+  description="Choose this option if you want your APIs to be publicly accessible by anyone on the internet."
+  card
+  label="Public"
+  selected-value="public"
+/>
+```
```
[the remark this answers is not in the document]
portikM: Yeah that's super nit
I'll keep it as is if you don't mind (this order pleases me aesthetically 😅)

src/components/KRadio/KRadio.vue
portikM: I don't think we should because here's what we say in the docs:

Screenshot 2024-11-07 at 3 47 55 PM

src/components/KRadio/KRadio.vue
```diff
@@ -379,17 +388,77 @@ $kRadioDotSize: 6px;
         transition: color $kongponentsTransitionDurTimingFunc;
       }
 
-      &.has-label, &.has-description {
-        .card-content-wrapper {
-          height: auto;
-          margin-bottom: var(--kui-space-40, $kui-space-40);
+      .card-label-container {
```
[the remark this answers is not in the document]
portikM: Good call, fixed now

docs/components/radio.md
portikM: Done

adamdehaven (approved):

PR #2512, opened 2024-11-12
fix(kselect): item creation logic [KHCP-13908]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-13908

Update KSelect item creation logic to allow creating an item for any unique string rather than only when search produces no results.


kaiarrowood (approved):
LGTM! 🚀

PR #2520, opened 2024-11-15
fix(ktableview, ktabledata): table updates event [KHCP-14032]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-14032

Fix `@update:table-preferences` event emits

### DIFF
--- a/src/components/KTableData/KTableData.vue
+++ b/src/components/KTableData/KTableData.vue
@@ -40,7 +40,7 @@
     @row-expand="($event: Record<string, any>) => emit('row-expand', $event)"
     @row-select="($event: Record<string, any>[]) => emit('row-select', $event)"
     @sort="sortHandler"
-    @update:table-preferences="tableViewPreferences = $event"
+    @update:table-preferences="tablePreferencesUpdateHandler"
   >
     <template
       v-if="$slots.toolbar && !hideToolbar"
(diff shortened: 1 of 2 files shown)


Leopoldthecoder (approved):
Manually tested and LGTM👍

src/components/KTableData/KTableData.vue
```diff
@@ -258,11 +258,11 @@ const hasInitialized = ref<boolean>(false)
...
-  page: 1,
-  query: '',
-  sortColumnKey: '',
-  sortColumnOrder: 'desc',
-  offset: null,
+  page: page.value,
+  query: filterQuery.value,
+  sortColumnKey: sortColumnKey.value,
+  sortColumnOrder: sortColumnOrder.value,
+  offset: offset.value,
```
adamdehaven, line 265: This object isn't reactive; I'm not sure I understand why we would always return the initial filter query?

portikM: Yes, it doesn't need to be reactive. We assign ref values to object properties here just so we only need to define values for those parameters once. We use this object to provide fallback values in case any are undefined in `props.initialFetcherParams` object.

src/components/KTableData/KTableData.vue
```diff
@@ -464,13 +461,25 @@ const pageSizeChangeHandler = ({ pageSize: newPageSize }: PageSizeChangeData) =>
   offset.value = null
   pageSize.value = newPageSize
   page.value = 1
+}
```
kaiarrowood, line 464: question: does it matter you are no longer calling `emitTablePreferences` in the `pageSizeChangeHandler`?

portikM: No because I'm calling it in `tablePreferencesUpdateHandler` which is triggered every time `KTableView` emits the `@update:table-preferences` event, which is emitted when page size is updated too.

kaiarrowood (approved):
LGTM! 🚀

PR #2527, opened 2024-11-26
fix(ktableview): row-expanded prop and event [KHCP-14243]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-14234

Changes:
* add `row-expanded` prop to provide an ability to render some rows expanded by default initially
* rename `row-expand` event to `update:row-expanded` and change event payload
  * instead of emitting event only when a row is expanded, event is emitted when a row is expanded OR collapsed
  * event payload changed from `row: Record<string, any>` object to `{ row: Record<string, any>, expanded: boolean }`

### DIFF
--- a/docs/components/table-view.md
+++ b/docs/components/table-view.md
@@ -754,6 +754,42 @@ Function for making a row expandable. The function receives row value object as 
 </KTableView>
 ```
 
+### rowExpanded
+
+Use this prop to specify initial expanded / collapsed state for each data row. The function receives row value object as an argument and should return a boolean value. Default value is `() => false`. When returned value is `true`, a row is rendered expanded.
+
+<KTableView
+  :data="userTypeData"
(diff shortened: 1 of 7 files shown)


docs/components/table-view.md
```diff
@@ -754,6 +754,42 @@ Function for making a row expandable. The function receives row value object as
 </KTableView>
 ```
 
+### rowExpanded
+
+Use this prop to specify initial expanded / collapsed state for each data row. The function receives row value object as an argument and should return a boolean value. Default value is `() => false`. When returned value is `true`, a row is rendered expanded.
```
kaiarrowood, line 759:
```suggestion
Use this prop to specifythe initial expanded / collapsed state for each data row. The function receives the row value object as an argument and should return a boolean value. Default value is `() => false`. When returned value is `true`, a row is rendered expanded.
```

kaiarrowood (approved):
One nit, but otherwise functional and code looks great 🚀

PR #2532, opened 2024-11-29
fix(ktabs): allow tab anchors to be links [KHCP-13866]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-13866

Add `to` property in `Tab` interface to support rendering tabs as links

### DIFF
--- a/docs/components/tabs.md
+++ b/docs/components/tabs.md
@@ -30,11 +30,17 @@ Required prop, which is an array of tab objects with the following interface:
 
 ```ts
 interface Tab {
-  hash: string // has to be unique, corresponds to the panel slot name
+  hash: string
   title: string
-  disabled?: boolean
+  disabled?: boolean,
+  to?: string | object
 }
 ```
+
+* `hash` - has to be unique, corresponds to the panel slot name
+* `title` - title to be displayed in the tab
+* `disabled` - whether or not tab is disabled
+* `to` - if present, tab will be rendered as either a `router-link` or an `a`
 
 <KTabs :tabs="tabsWithDisabled">
   <template #tab1>
@@ -176,41 +182,6 @@ const tabChange = (hash: string): void => {
 </script>
 ```
 
-### anchorTabindex
-
-This prop allows setting a custom `tabindex` for the tab anchor element. It’s useful when passing a custom interactive element, like a link, through the [`anchor` slot](#anchor-panel), ensuring that only the slotted element is focusable by resetting the default anchor `tabindex`. Default value is `0`.
-
-#### Dynamic RouterView
-
-Here's an example (code only) of utilizing a dynamic `router-view` component within the host app:
-
-```html
-<KTabs
-  hide-panels
-  :tabs="tabs"
->
-  <template
-    v-for="tab in tabs"
-    :key="`${tab.hash}-anchor`"
-    #[`${tab.hash}-anchor`]
-  >
-    <router-link
-      :to="{
-        name: tab.hash.split('?').shift(),
-        hash: `#${tab.hash.split('?').pop()}`,
-      }"
-    >
-      {{ tab.title }}
-    </router-link>
-  </template>
-</KTabs>
-
-<router-view v-slot="{ route }">
-  <h3>Router View content</h3>
-  <p>{{ route.path }}{{ route.hash }}</p>
-</router-view>
-```
-
 ## Slots
 
 ### anchor & panel
@@ -297,6 +268,43 @@ const tabs = ref<Tab[]>([
 </script>
 ```
 
+## Usage
+
+### Tab links
+
+Passing `to` property for each tab object allows to render tabs as links. If a string is passed, it will be used in `href` attribute in the rendered `a` element. If an object is passed, the tab will be rendered as a `router-link`.
+
+<KTabs :tabs="linkTabs" hide-panels v-model="linkTabValue" />
+
+{{ linkTabValue }}
+
+```vue
+<template>
+  <KTabs :tabs="linkTabs" hide-panels />
+
+  <router-view v-slot="{ route }">
+    {{ route.hash }}
+  </router-view>
+</template>
+
+<script setup lang="ts">
+import { Tab } from '@kong/kongponents'
+
+const linkTabs = ref<Tab[]>([
+  {
+    hash: '#tab1',
+    title: 'Tab 1',
+    to: '#tab-link-1'
+  },
+  {
+    hash: '#tab2',
+    title: 'Tab 2',
+    to: '#tab-link-2'
+  },
+])
+</script>
+```
+
 <script setup lang="ts">
 import { ref } from 'vue'
 import { KongIcon, InboxNotificationIcon, BookIcon } from '@kong/icons'
@@ -320,6 +328,20 @@ const slottedTabs = ref<Tab[]>([
   { hash: '#notifications', title: 'Notifications' },
   { hash: '#docs', title: 'Documentation' },
   { hash: '#disabled', title: 'Disabled', disabled: true }
+])
+
+const linkTabValue = ref<string>('#tab-link-1')
+const linkTabs = ref<Tab[]>([
+  {
+    hash: '#tab-link-1',
+    title: 'Tab 1',
+    to: '#tab-link-1',
+  },
+  {
+    hash: '#tab-link-2',
+    title: 'Tab 2',
+    to: '#tab-link-2',
+  },
 ])
 
 const panelsActiveHash = ref('#gateway')
--- a/docs/components/tabs.md
+++ b/docs/components/tabs.md
@@ -81,6 +81,45 @@ interface Tab {
 </KTabs>
 ```
 
+#### Tabs as links
+
+Passing the `to` property for each tab object enables rendering tabs as links. If a string is provided, it will be used as the `href` attribute in the rendered `a` element. If an object is provided, the tab will be rendered as a `router-link`.
+
+:::tip TIP
+When creating tab links, it is recommended to set the [`hidePanels` prop](#hidepanels) to `true`, as page changes typically do not involve the use of [`panel` slots](#slots).
+:::
+
+<KTabs :tabs="linkTabs" hide-panels v-model="linkTabValue" />
+
+{{ linkTabValue }}
+
+```vue
+<template>
+  <KTabs :tabs="linkTabs" hide-panels />
+
+  <router-view v-slot="{ route }">
+    {{ route.hash }}
+  </router-view>
+</template>
+
+<script setup lang="ts">
+import { Tab } from '@kong/kongponents'
+
+const linkTabs = ref<Tab[]>([
+  {
+    hash: '#tab1',
+    title: 'Tab 1',
+    to: '#tab-link-1'
+  },
+  {
+    hash: '#tab2',
+    title: 'Tab 2',
+    to: '#tab-link-2'
+  },
+])
+</script>
+```
+
 ### v-model
 
 KTabs will set the first tab in the `tabs` array as active. You can override this by passing in the hash of any other tab to be used with `v-model`.
@@ -264,43 +303,6 @@ import type { Tab } from '@kong/kongponents'
 const tabs = ref<Tab[]>([
   { hash: '#tab1', title: 'Tab 1' },
   { hash: '#tab2', title: 'Tab 2' },
-])
-</script>
-```
-
-## Usage
-
-### Tab links
-
-Passing `to` property for each tab object allows to render tabs as links. If a string is passed, it will be used in `href` attribute in the rendered `a` element. If an object is passed, the tab will be rendered as a `router-link`.
-
-<KTabs :tabs="linkTabs" hide-panels v-model="linkTabValue" />
-
-{{ linkTabValue }}
-
-```vue
-<template>
-  <KTabs :tabs="linkTabs" hide-panels />
-
-  <router-view v-slot="{ route }">
-    {{ route.hash }}
-  </router-view>
-</template>
-
-<script setup lang="ts">
-import { Tab } from '@kong/kongponents'
-
-const linkTabs = ref<Tab[]>([
-  {
-    hash: '#tab1',
-    title: 'Tab 1',
-    to: '#tab-link-1'
-  },
-  {
-    hash: '#tab2',
-    title: 'Tab 2',
-    to: '#tab-link-2'
-  },
 ])
 </script>
 ```
--- a/src/components/KCodeBlock/KCodeBlock.vue
+++ b/src/components/KCodeBlock/KCodeBlock.vue
@@ -397,7 +397,7 @@ const matchingLineNumbers = ref<number[]>([])
 const currentLineIndex = ref<null | number>(null)
 
 const totalLines = computed((): number[] => Array.from({ length: props.code?.split('\n').length }, (_, index) => index + 1))
-const maxLineNumberWidth = computed((): string => totalLines.value[totalLines.value.length - 1]?.toString().length + 'ch')
+const maxLineNumberWidth = computed((): string => totalLines.value[totalLines.value?.length - 1]?.toString().length + 'ch')
 const linePrefix = computed((): string => props.id.toLowerCase().replace(/\s+/g, '-'))
 const isProcessing = computed((): boolean => props.processing || isProcessingInternally.value)
 const isShowingFilteredCode = computed((): boolean => isFilterMode.value && filteredCode.value !== '')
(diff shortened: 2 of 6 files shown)


src/components/KTabs/KTabs.vue
```diff
@@ -75,11 +75,6 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
-  anchorTabindex: {
-    type: Number,
-    default: 0,
-    validator: (val: number): boolean => val >= -1 && val <= 32767,
-  },
```
adamdehaven, line 82: issue: isn't this a breaking change?

src/components/KTabs/KTabs.vue
```diff
@@ -11,21 +11,21 @@
...
           :aria-controls="hidePanels ? undefined : `panel-${tab.hash}`"
           :aria-selected="hidePanels ? undefined : (activeTab === tab.hash ? 'true' : 'false')"
           class="tab-link"
-          :class="{ 'has-panels': !hidePanels, disabled: tab.disabled }"
+          :class="{ disabled: tab.disabled }"
           role="tab"
-          :tabindex="getAnchorTabindex(tab)"
-          @click.prevent="!tab.disabled ? handleTabChange(tab.hash) : undefined"
-          @keydown.enter.prevent="!tab.disabled ? handleTabChange(tab.hash) : undefined"
-          @keydown.space.prevent="!tab.disabled ? handleTabChange(tab.hash) : undefined"
```
adamdehaven, line 23: question: why are we removing the keyboard events?

src/components/KTabs/KTabs.vue
```diff
@@ -11,21 +11,21 @@
...
+          :tabindex="!tab.disabled ? 0 : -1"
+          v-bind="tabComponent(tab).attributes"
+          @click="!tab.disabled ? handleTabChange(tab.hash) : undefined"
         >
           <slot :name="`${getTabSlotName(tab.hash)}-anchor`">
-            <span>{{ tab.title }}</span>
```
adamdehaven, line 26: question: why is this span tag being removed? I'm not sure I understand changing the structure.

src/components/KTabs/KTabs.vue
```diff
@@ -97,12 +92,16 @@ const handleTabChange = (tab: string): void => {
...
+  if (tab.to) {
+    if (typeof tab.to === 'string') {
+      return { tag: 'a', attributes: { href: tab.disabled ? undefined : tab.to } }
+    } else if (typeof tab.to === 'object') {
+      return { tag: 'router-link', attributes: { to: tab.disabled ? undefined : tab.to } }
+    }
   }
 
-  return typeof props.anchorTabindex === 'number' && props.anchorTabindex >= -1 && props.anchorTabindex <= 32767 ? String(props.anchorTabindex) : '0'
+  return { tag: 'button', attributes: { ...(tab.disabled && { disabled: true }) } }
```
adamdehaven, line 104: issue: this was previously a `div` and now you're returning a `button`

adamdehaven (changes requested):

src/components/KTabs/KTabs.vue
portikM: It was a `div` and now it's becoming either a `button` or an `a` - both elements support keyboard events natively

src/components/KTabs/KTabs.vue
portikM: Yes, we discussed this with @Justineo and decided this would not be considered a breaking change

src/components/KTabs/KTabs.vue
portikM: I decided to remove it as it's not really necessary

docs/components/tabs.md
```diff
@@ -176,41 +182,6 @@ const tabChange = (hash: string): void => {
 </script>
 ```
 
-### anchorTabindex
```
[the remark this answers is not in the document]
adamdehaven: It's still a breaking change. It should be deprecated instead so we don't have to do a breaking release.

src/components/KTabs/KTabs.vue
adamdehaven: This could have downstream impact that you may not be aware of. Let's leave in place.

docs/components/tabs.md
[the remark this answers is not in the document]
portikM: Okay, let me bring it back and deprecate it intead.

src/components/KTabs/KTabs.vue
adamdehaven: Let's have that discussion here -- changing the DOM structure technically _is_ a breaking change.

src/components/KTabs/KTabs.vue
Justineo: I believe we should discourage consumer applications from relying on the DOM structure of our core components, as this makes any upstream changes highly susceptible to breaking. It feels rather unreasonable if we can’t even implement such internal optimizations. 🫠

src/components/KTabs/KTabs.vue
Justineo: That depends on the public API we are providing. I believe this should be considered an implementation detail.

src/components/KTabs/KTabs.vue
adamdehaven: I'm more advocating for not changing the DOM structure if it isn't absolutely necessary. In this particular case, the "old" functionality (i.e. non-link tabs) doesn't require the change in this PR from `div` to `button`

src/components/KTabs/KTabs.vue
Justineo: I think the key lies in how we manage user expectations. We should convey to users as clearly as possible that relying on the DOM structure is unsafe, serves as an escape hatch, and carries the risk of breaking in the future.

In my previous work, as the provider of the component library, we would even periodically scan the code of consumer applications to identify potentially risky usage patterns (such as deep selectors) and provide reports to the users, further clarifying the associated risks.

adamdehaven: If you look at our own internal implementations of Kongponents, using `KButton` as an example, there are hundreds of customizations directly tied to the DOM structure.

src/components/KTabs/KTabs.vue
adamdehaven: I’m just being realistic 😅 

This will be exponentially more important with the new portal UI. Customers will likely be using CSS/selectors all over the place; it’s inevitable. 

Discouraging non-critical DOM structure changes is imperative for stability. 

> I decided to remove it as it's not really necessary

TBH this is not the right pattern for an OSS component library.

src/components/KTabs/KTabs.vue
Justineo: Kongponents might not be perfectly abstracted, but the risks are internal. For Portal, I think setting clear customization boundaries is key—unlimited customization makes future updates a lot harder.

adamdehaven: > unlimited customization makes future update

I agree; however, I still am against trivial changes that aren't related to a specific feature or critical fix. We shouldn't just be changing things "because we want to"

Justineo: > We shouldn't just be changing things "because we want to"

I agree with this. However, I also recognize that the changes being made here are not arbitrary. The intention is to improve the project.

> Yes, we discussed this with @Justineo and decided this would not be considered a breaking change

As for my earlier suggestion: rendering `<a>` elements for tabs with `to` shouldn’t be considered a breaking change, as `to` wasn’t supported before. However, changing a `<div>` to a `<button>` does align with the best practices outlined in [WAI's examples](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#examples). How about keeping this change? @adamdehaven

adamdehaven: The change to `button` is arbitrary if the proper `role` (and other) attributes are in place on the existing `div`, but yes, does require adding additional things you "get for free" with a button.

I'm just attempting to stress we need to be very stringent with DOM changes as it could very easily break downstream projects. I'll unblock the change if this is how it's preferred to go forward on this PR, but subjective changes should be pushed back on overall.

adamdehaven: > subjective changes

To be clear, using a button here is more proper, I'm speaking generally.

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -397,7 +397,7 @@ const matchingLineNumbers = ref<number[]>([])
 const currentLineIndex = ref<null | number>(null)
 
 const totalLines = computed((): number[] => Array.from({ length: props.code?.split('\n').length }, (_, index) => index + 1))
-const maxLineNumberWidth = computed((): string => totalLines.value[totalLines.value.length - 1]?.toString().length + 'ch')
+const maxLineNumberWidth = computed((): string => totalLines.value[totalLines.value?.length - 1]?.toString().length + 'ch')
```
johncowen, line 400: 👋 Drive-by comment: Is this a bug fix, and if so should it go in a different PR? If not should it be here?

src/components/KTabs/KTabs.vue
```diff
@@ -11,21 +11,21 @@
         class="tab-item"
         :class="{ active: activeTab === tab.hash }"
       >
-        <div
+        <component
```
johncowen, line 14: Not sure if this is something you'll want to address in this PR, but if all of the "tabs" are using anchors, it means its essentially a `role="navigation"` (or just a `nav`). It just happens to look like what is generally termed a "tab look and feel". (In regards to line 5 here which I can't make a comment specifically on that line because GH)

docs/components/tabs.md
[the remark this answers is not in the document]
portikM: Done.

src/components/KTabs/KTabs.vue
portikM: Brought the `span` back. I think my prior explanation was poorly phrased, I wanted to remove that `span` to optimize DOM structure in this component but your concern is valid.

src/components/KTabs/KTabs.vue
portikM: Changed the element back to div here.

src/components/KTabs/KTabs.vue
```diff
@@ -89,6 +95,8 @@ const emit = defineEmits<{
 
 const activeTab = ref<string>(props.modelValue ? props.modelValue : props.tabs[0]?.hash)
 
+const isNav = computed((): boolean => props.tabs.every(tab => tab.to))
```
adamdehaven, line 98: question: what happens if some tabs have a `to` prop and others don't?

src/components/KTabs/KTabs.vue
johncowen: If you are using the `role="navigation"` you'll probably want to let the user decide what the `aria-label` is.

Question: Would you consider a very simple `KNavigation` component that was just a `<nav><ul><li>` with some similar styling for the tabs (i.e. very minimal JS)?

I would guess a theoretical `KNavigation` would be an very simple component to write as you need about 5% of the JS you have in here. You would also then avoid complicating `KTabs`. I would guess that would also stop @adamdehaven sweating about possible breaking changes 😅 so a positive all round!

adamdehaven: I think this is getting out of scope for the PR. 

TBH the existing aria-label should be removed as it's generic and doesn't actually describe the content. "Tabs" doesn't label the type of content being displayed.

portikM: I agree that KNavigation is out of scope for this PR.
As for `aria-label`, it is an attribute required for elements with `role="tablist"` and `role="navigation"`. Removing it will bite us back in the upcoming a11y audit for sure. Instead I suggest that we bind `aria-label` (when provided) to this element instead of topmost wrapper `div` and add a paragraph in docs mentioning that it would be a good idea to provide a custom value, or otherwise the default ("Tabs") will be used?

adamdehaven: Again, I still think out of scope for this PR. 

Can create a ticket to discuss for a future update

src/components/KTabs/KTabs.vue
```diff
@@ -75,6 +78,9 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+  /**
+   * @deprecated
```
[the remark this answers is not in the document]
portikM: I removed this prop from docs (our standard practice for deprecated props is to remove them from Kongponents docs).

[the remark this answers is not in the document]
adamdehaven: > I removed this prop from docs (our standard practice for deprecated props is to remove them from Kongponents docs).

Standard is actually to mark as deprecated 

image

src/components/KTabs/KTabs.vue
portikM: Then `role="tablist"` will be used. In other words, means it's a tabs element with some tabs being the links (vs if all tabs are links then it's a navigation element `role="navigation"`).

src/components/KTabs/KTabs.vue
johncowen: Ok but if we are using `role=navigation` here make sure you remember to also remove all the other aria attributes and that aren't needed on `role=navigation`. Would also be good to remove or at least make sure that the custom events added to the `li`s here don't break any existing native functionality provided by using anchors.

Whilst discussing this (and this might also be out of scope so let me know if so), I think you might want to remove some of these for when "hide-panels" is true also, as in that case its not really a role="tablist" either.

Let me know the link for that ticket @adamdehaven and I'll catch up with that tomorrow 👍

src/components/KTabs/KTabs.vue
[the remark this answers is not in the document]
portikM: Well maybe we can't use a word "standard" here since we have different patterns:
* KModalFullscreen is deprecated and we don't have the component docs page
* KTable is deprecated but because it's such a crucial component internally we still have the page up
* `label` prop in KDropdown is deprecated and is not present in docs
* same applies to `isResizable` and `hasError` props in KTextArea
* in fact, same applies to every deprecated prop in any component

So how do we want to go about this one? In my opinion, we don't want to promote usage of it so we should remove it from docs.

src/components/KCodeBlock/KCodeBlock.vue
johncowen: Will there be a separate changelog entry for folks doing dependabot PRs and reading changelogs to see what has changed?

portikM: No but a question mark check is hardly a change anyone would be curious about I imagine

johncowen: Oh I thought it was a bug fix?

portikM: Well, DD error fix. Does not affect users.

johncowen: oki doki, feel free to resolve!

docs/components/tabs.md
```diff
@@ -297,6 +268,43 @@ const tabs = ref<Tab[]>([
 </script>
 ```
 
+## Usage
+
+### Tab links
```
adamdehaven, line 273: I'm not sure I understand the new "Usage" section here -- can we just document "Tabs as Links" under the Props instead (above the Slots section)

docs/components/tabs.md
```diff
@@ -297,6 +268,43 @@ const tabs = ref<Tab[]>([
...
+
+Passing `to` property for each tab object allows to render tabs as links. If a string is passed, it will be used in `href` attribute in the rendered `a` element. If an object is passed, the tab will be rendered as a `router-link`.
```
adamdehaven, line 275: Can you reword to explain this is an example of utilizing the `to` property along with the `hidePanels` prop?

The way it's currently written it sounds like you **must** use as links

src/components/KTabs/KTabs.vue
```diff
@@ -49,7 +52,7 @@
 
 <script lang="ts" setup>
 import type { PropType } from 'vue'
-import { ref, watch } from 'vue'
+import { computed, ref, watch } from 'vue'
```
adamdehaven, line 55: remove unused import

docs/components/tabs.md
portikM: Rewritten

docs/components/tabs.md
portikM: https://deploy-preview-2532--kongponents.netlify.app/components/tabs.html#tabs-as-links

adamdehaven (approved):

PR #2540, opened 2024-12-13
fix(kalert): default slotted code styles [KHCP-14406]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-14406

Add default styles for slotted code in KAlert

Screenshot 2024-12-13 at 12 05 54 PM


adamdehaven (approved):
lgtm

PR #2541, opened 2024-12-13
fix(kselect): remove altering slected item key [KHCP-13749]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-13749

We were appending selected item key with `-selected` which is not a good practice as makes it hard for browser to track element state


kaiarrowood (approved):
LGTM! 🚀

PR #2542, opened 2024-12-16
fix(kpagination,ktableview): add missing import

Justineo:
# Summary

- Add missing imports in `PaginationOffset` and `ColumnVisibilityMenu`.
- Use `TableDataHeader` instead of the deprecated `TableHeader`.


adamdehaven (approved):
lgtm

PR #2547, opened 2024-12-19
fix(kbutton): add text-align

adamdehaven:
# Summary

Sometimes the button `display` property is overriden to `block`, and in these scenarios we want to ensure the text-align is set to center to properly align the button text.


kaiarrowood (approved):
LGTM! 🚀

PR #2555, opened 2025-01-03
feat(kcodeblock): add highlighted line numbers prop

kaiarrowood:
# Summary

Add new `highlightedLineNumbers` prop to manually set highlighting without requiring search.

image

### DIFF
--- a/docs/components/codeblock.md
+++ b/docs/components/codeblock.md
@@ -98,7 +98,9 @@ Boolean to control whether to show or hide the search bar and related action but
 
 ### highlightedLineNumbers
 
-An array of the line numbers to highlight by default in the codeblock. If search is enabled, search results will override the default highlighted lines and be cleared once the search query is cleared
+An array of line numbers to initially highlight in the code block. 
+
+If search is enabled, matching results will be highlighted in place of the provided line numbers, and all highlighted lines will be cleared once the search query is cleared.
 
 <ClientOnly>
   <KCodeBlock
--- a/docs/components/codeblock.md
+++ b/docs/components/codeblock.md
@@ -96,6 +96,28 @@ Boolean to control whether to show or hide the search bar and related action but
 />
 ```
 
+### highlightedLineNumbers
+
+An array of the line numbers to highlight by default in the codeblock.
+
+<ClientOnly>
+  <KCodeBlock
+    id="code-block-highlight"
+    :code="code"
+    language="json"
+    :highlighted-line-numbers="[2,4,6]"
+  />
+</ClientOnly>
+
+```html
+<KCodeBlock
+  id="code-block-highlight"
+  :code="code"
+  language="json"
+  :highlighted-line-numbers="[2,4,6]"
+/>
+```
+
 ### singleLine
 
 Whether the code passed into the component should be displayed on a single line. Defaults to `false`.
--- a/docs/components/codeblock.md
+++ b/docs/components/codeblock.md
@@ -98,7 +98,7 @@ Boolean to control whether to show or hide the search bar and related action but
 
 ### highlightedLineNumbers
 
-An array of the line numbers to highlight by default in the codeblock.
+An array of the line numbers to highlight by default in the codeblock. If search is enabled, search results will override the default highlighted lines and be cleared once the search query is cleared
 
 <ClientOnly>
   <KCodeBlock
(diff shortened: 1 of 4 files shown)


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -577,6 +595,18 @@ function handleSearchInputValue(): void {
...
+  regExpError.value = null
+
+  matchingLineNumbers.value = Array.from(new Set(props.highlightedLineNumbers))
+  numberOfMatches.value = matchingLineNumbers.value.length
+
+  emitMatchingLinesChangeEvent()
+
+  isProcessingInternally.value = false
+}
+
```
adamdehaven, line 609: question: If you want to pass in a range (e.g. highlight lines 2-8) could this be accounted for instead of having to pass in each line number individually?

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -290,6 +290,15 @@ const props = defineProps({
...
   },
 
+  /**
+   * The line numbers for the lines to highlight by default. **Default: `[]`**.
+   */
+  highlightedLineNumbers: {
+    type: Array as PropType<number[]>,
+    required: false,
+    default: () => [],
+  },
```
adamdehaven, line 300: can you add test coverage for the new prop?

adamdehaven (commented):
Can you please update the sandbox page as well with an example for the new prop, as well as enabling search at the same time? How does the component respond if line numbers are passed in and search is also available?

kaiarrowood:
> How does the component respond if line numbers are passed in and search is also available?

I will update the sandbox 👍🏻 , this is setting the "default" highlighted lines. If you enable search too your default highlighted lines will be overridden by your search results. I'll update the docs

src/components/KCodeBlock/KCodeBlock.vue
kaiarrowood: We will stick to an array of line numbers for now (since that is what is expected by the highlighter) and revisit if a need for a more elegant solution arises.

docs/components/codeblock.md
```diff
@@ -96,6 +96,28 @@ Boolean to control whether to show or hide the search bar and related action but
 />
 ```
 
+### highlightedLineNumbers
+
+An array of the line numbers to highlight by default in the codeblock. If search is enabled, search results will override the default highlighted lines and be cleared once the search query is cleared
```
adamdehaven, line 101: nit: reword
```suggestion
An array of line numbers to initially highlight in the code block. 

If search is enabled, matching results will be highlighted in place of the provided line numbers, and all highlighted lines will be cleared once the search query is cleared.
```

adamdehaven (approved):
lgtm

docs/components/codeblock.md
```diff
@@ -96,6 +96,30 @@ Boolean to control whether to show or hide the search bar and related action but
 />
 ```
 
+### highlightedLineNumbers
+
+An array of line numbers to initially highlight in the code block. 
+
+If search is enabled, matching results will be highlighted in place of the provided line numbers, and all highlighted lines will be cleared once the search query is cleared.
```
Justineo, line 103: Why do we need to cancel line highlighting when the query is cleared? I think it's better to emit an event for this and let users to decide whether to cancel line highlighting.

docs/components/codeblock.md
```diff
@@ -96,6 +96,30 @@ Boolean to control whether to show or hide the search bar and related action but
...
+    :highlighted-line-numbers="[2,4,6]"
+  />
+</ClientOnly>
+
+```html
+<KCodeBlock
+  id="code-block-highlight"
+  :code="code"
+  language="json"
+  :highlighted-line-numbers="[2,4,6]"
```
Justineo, line 119: Shall we also support the syntax that [VitePress](https://vitepress.dev/guide/markdown#line-highlighting-in-code-blocks) and [Gatsby](https://www.gatsbyjs.com/plugins/gatsby-remark-highlight-code/#lines-highlighting) supports (`4,7-13,16,23-27,40`) so that it's easier to highlight a range of lines?

kaiarrowood: Tracking in https://konghq.atlassian.net/browse/KHCP-14552

PR #2558, opened 2025-01-07
feat(kcodeblock): support line highlight syntax

Justineo:
# Summary

[KM-886](https://konghq.atlassian.net/browse/KM-886)
[KHCP-14552](https://konghq.atlassian.net/browse/KHCP-14552)

This PR:

1. Introduces support for line highlighting syntax (e.g., `"2,4,6-12"`) for the `<KCodeBlock>` component’s `highlighted-line-numbers` prop. I recommend using this approach over arrays, as primitive values are less likely to trigger unnecessary re-renders. This is particularly advantageous compared to binding array literals like `:highlighted-line-numbers="[2, 4, 6]"`.

2. Improves the line navigation behavior to prevent scrolling each time the “Next match” or “Previous match” button is clicked, which reduce the pain for consecutive clicks.

3. Enhances line matching performance, particularly when dealing with multiple matches for a lengthy section of code.

[KM-886]: https://konghq.atlassian.net/browse/KM-886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KHCP-14552]: https://konghq.atlassian.net/browse/KHCP-14552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KCodeBlock/KCodeBlock.cy.ts
+++ b/src/components/KCodeBlock/KCodeBlock.cy.ts
@@ -7,6 +7,39 @@ const code = `{
   "key3": [1, 2, 3]
 }`
 
+const longCode = `{
+  "key1": "string value",
+  "key2": 5681,
+  "key3": [1, 2, 3],
+  "key4": "string value",
+  "key5": 5681,
+  "key6": [1, 2, 3],
+  "key7": "string value",
+  "key8": 5681,
+  "key9": [1, 2, 3],
+  "key10": "string value",
+  "key11": 5681,
+  "key12": [1, 2, 3],
+  "key13": "string value",
+  "key14": 5681,
+  "key15": [1, 2, 3],
+  "key16": "string value",
+  "key17": 5681,
+  "key18": [1, 2, 3],
+  "key19": "string value",
+  "key20": 5681,
+  "key21": [1, 2, 3],
+  "key22": "string value",
+  "key23": 5681,
+  "key24": [1, 2, 3],
+  "key25": "string value",
+  "key26": 5681,
+  "key27": [1, 2, 3],
+  "key28": "string value",
+  "key29": 5681,
+  "key30": [1, 2, 3]
+}`
+
 function renderComponent(props = {}) {
   return cy.mount(KCodeBlock, {
     props: {
@@ -91,6 +124,35 @@ describe('KCodeBlock', () => {
     for (const lineNumber of expectedLineNumbers) {
       cy.get('[data-testid="k-code-block"]').trigger('keydown', { code: 'F3', bubbles: true })
       cy.get(`.line-is-highlighted-match .line-anchor#${id}-L${lineNumber}`).should('be.visible')
+    }
+  })
+
+  it('can highlight matching lines when initialized with highlightedLineNumbers in range expressions', () => {
+    const id = 'code-block'
+    const expectedLineNumbers = [1, 3, 4, 5, 6, 7, 8, 9, 10, 12, 13, 30, 31, 32]
+    renderComponent({
+      id,
+      code: longCode,
+      highlightedLineNumbers: '10-3,4,6,12,1,13-13,30-34',
+    })
+
+    for (const lineNumber of expectedLineNumbers) {
+      console.log(lineNumber)
+      cy.get('.line').eq(lineNumber - 1).should('have.class', 'line-is-match')
+    }
+  })
+
+  it('can highlight matching lines when initialized with highlightedLineNumbers in range expressions', () => {
+    const id = 'code-block'
+    const expectedLineNumbers = [1, 3, 4, 5, 6, 7, 8, 9, 10, 12, 13, 30, 31, 32]
+    renderComponent({
+      id,
+      code: longCode,
+      highlightedLineNumbers: [[10, 3], 4, 6, 12, 1, [13, 13], [30, 34]],
+    })
+
+    for (const lineNumber of expectedLineNumbers) {
+      cy.get('.line').eq(lineNumber - 1).should('have.class', 'line-is-match')
     }
   })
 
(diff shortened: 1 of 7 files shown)


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -216,6 +217,12 @@
   </div>
 </template>
 
+<script lang="ts">
+const IS_MAYBE_MAC = typeof window !== 'undefined' && window.navigator.platform.toLowerCase().includes('mac')
```
adamdehaven, line 221: suggestion: a little better useragent check
```suggestion
const IS_MAYBE_MAC = typeof window !== 'undefined' && /macOS|Mac|iPhone|iPod|iPad/i.test(navigator?.userAgentData?.platform)
```

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -294,9 +298,25 @@ const props = defineProps({
    * The line numbers for the lines to highlight by default. **Default: `[]`**.
    */
   highlightedLineNumbers: {
-    type: Array as PropType<number[]>,
-    required: false,
+    type: [String, Array] as PropType<string | (number | [number, number])[]>,
     default: () => [],
+    validator: (value: string | (number | [number, number])[]) => {
```
adamdehaven, line 303: nit
```suggestion
    validator: (value: string | (number | [number, number])[]): boolean => {
```

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -216,6 +217,12 @@
   </div>
 </template>
 
+<script lang="ts">
+const IS_MAYBE_MAC = typeof window !== 'undefined' && window.navigator.platform.toLowerCase().includes('mac')
+const ALT_SHORTCUT_LABEL = IS_MAYBE_MAC ? 'Option' : 'Alt'
+const LINE_NUMBER_EXPRESSION_RE = /^\d+(-\d+)?(,\d+(-\d+)?)*$/
```
adamdehaven, line 223: suggestion: naming
```suggestion
const LINE_NUMBER_EXPRESSION_REGEX = /^\d+(-\d+)?(,\d+(-\d+)?)*$/
```

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -294,9 +298,25 @@ const props = defineProps({
    * The line numbers for the lines to highlight by default. **Default: `[]`**.
    */
   highlightedLineNumbers: {
-    type: Array as PropType<number[]>,
-    required: false,
+    type: [String, Array] as PropType<string | (number | [number, number])[]>,
     default: () => [],
+    validator: (value: string | (number | [number, number])[]) => {
+      if (typeof value === 'string') {
```
adamdehaven, line 304: question: do we also need to handle an empty string?

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -744,6 +767,40 @@ async function copyCode(event: Event): Promise<void> {
...
+  }).sort((a, b) => a - b)
+
+  return Array.from(new Set(lines))
+}
+
+function normalizeHighlightedLines(lines: string | (number | [number, number])[]): number[] {
+  return typeof lines === 'string'
+    ? expressionToLines(lines, totalLines.value.length)
+    : rangesToLines(lines, totalLines.value.length)
+}
```
adamdehaven, line 803: suggestion: can you please add a few comments in the code

src/components/KCodeBlock/KCodeBlock.cy.ts
```diff
@@ -94,6 +127,35 @@ describe('KCodeBlock', () => {
...
+    const id = 'code-block'
+    const expectedLineNumbers = [1, 3, 4, 5, 6, 7, 8, 9, 10, 12, 13, 30, 31, 32]
+    renderComponent({
+      id,
+      code: longCode,
+      highlightedLineNumbers: '10-3,4,6,12,1,13-13,30-34',
+    })
+
+    for (const lineNumber of expectedLineNumbers) {
+      console.log(lineNumber)
```
kaiarrowood, line 140:
```suggestion
```

nit

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -744,6 +770,48 @@ async function copyCode(event: Event): Promise<void> {
...
+
+  // Ensure no duplicates
+  return Array.from(new Set(lines))
+}
+
+function normalizeHighlightedLines(lines: string | (number | [number, number])[]): number[] {
+  return typeof lines === 'string'
+    ? expressionToLines(lines, totalLines.value.length)
+    : rangesToLines(lines, totalLines.value.length)
+}
```
kaiarrowood, line 814: suggestion: this component is pretty beefy, it might help to move these conversion helpers to a composable?

kaiarrowood (commented):
I know it would be a breaking change, but is there still value in supporting an array of numbers if we also support passing a string that can do the same thing? I think I'm the only one currently using this functionality, I'm not sure continuing to maintain support for arrays makes sense.

adamdehaven:
> I know it would be a breaking change, but is there still value in supporting an array of numbers if we also support passing a string that can do the same thing? I think I'm the only one currently using this functionality, I'm not sure continuing to maintain support for arrays makes sense.

I think changing the implementation as in this PR is ideal, and we can just update your usage once merged? Allowing ranges (and using the standard format) is ideal

src/components/KCodeBlock/KCodeBlock.vue
adamdehaven: ~~composable~~ utility function(s)

Justineo:
If you already have an array derived from some computation, I think there’s no need to convert it back into a string.

adamdehaven (approved):
lgtm

PR #2562, opened 2025-01-09
fix(kcodeblock): improve matching perf

Justineo:
# Summary

[KM-888](https://konghq.atlassian.net/browse/KM-888)

> An earlier optimization was implemented in #2558, where line matching now uses a `Set` instead of looping through an array, reducing the operation’s time complexity to O(1).

This PR introduces further performance enhancements to the `<KCodeBlock>` component while addressing multiple bugs.

## Key improvements

1. **Correctly escape HTML and RegExp**

    Resolved that when highlighting matched characters in filtered code:

    - It did not escape HTML, potentially causing incorrect output.
    - It did not escape when creating `RegExp` in exact match mode, eg. `.` was actually highlighting all characters.

2. **Reduce redundant `updateMatchingLineNumbers` calls**

    Prevented calls to `updateMatchingLineNumbers` when `isShowingFilteredCode` changes, avoiding unnecessary processing.

3. **Optimized line matching logic**

    - Previously, for each match index, the implementation split all text from the beginning to `pos` or `match.index` in order to get the line number for each match, leading to high computational costs (`O(n*m)`, where `n` is code length and `m` is the number of matches).

    - The new approach precomputes line offsets and uses binary search to determine line numbers, significantly reducing overhead.

4. **Support for multi-line regular expressions**

    Fixed a bug preventing regular expression patterns from matching across multiple lines.

5. **Optimized character highlighting in filter mode**

    Escaping HTML for highlight output now involves additional work compared to the old implementation, but further optimizations have made the process significantly faster in most cases.

   - **Efficient Highlight Processing**: The previous approach split the filtered code into individual lines, highlighting each one separately. It not only failed to properly escape the input code but was also creating a new `RegExp` instance during every iteration. The new approach processes the entire code at once, using `RegExp.prototype.exec` to identify matches and perform both matching and escaping in a single pass.

    - **Selective escaping**: Escapes only `<` and `&` since the output HTML is isolated (`v-html`), reducing escape complexity (compared to escaping a broader set of characters like `<`, `>`, `'`, `"`, and `&`).

    - **Fast path for non-escaping cases**: Introduced a quick pre-check using `/[<&]/` to determine if escaping is necessary, enabling a fast path for cases without special characters. The new `escapeIfNeeded` function adopts this optimized approach, replacing the original escape implementation that executed upon component render. The previous method, which applied `replaceAll` to the entire input for each character in the escape list, was sub-optimal. While the performance difference is negligible for short inputs, the new approach significantly improves highlighting speed in most use cases while maintaining correctness.

    - **Merging adjacent matches**: For short queries or patterns like `/./` that generate many consecutive matches, adjacent match marks are merged. This reduces the number of extra elements and accelerates both calculation and rendering. This is achieved by wrapping the RegExp in a grouping construct `(:...)+`. This improves speed for both calculation and rendering.

6. **Significant improvement in Rendering Performance**

    ~~Following the optimization of the matching and highlighting algorithms, the primary bottleneck shifted to the rendering process. Using Chrome’s Performance Profiler, we identified that the most time-consuming operation was “Recalculate Style.” To address this, we replaced `display: grid`, which can be inefficient for a large number of elements, with a more static layout. Additionally, `content-visibility` was applied for line number elements. These changes resulted in a substantial boost in rendering performance.~~

    `content-visibility: auto` hides elements from the render tree and only renders them when they scroll into view or when their metrics are accessed programmatically (e.g., `getBoundingClientRect`, `offsetHeight`, etc.). However, this can introduce unpredictable issues.

    One issue I encountered involves the interaction between our current DOM structure and 1Password’s heuristics. 1Password attempts to identify labels associated with form inputs and iterates through all line number elements and retrieves their layout information one by one. This process forces the browser to repeatedly calculate layouts for each previously hidden element styled with `content-visibility: auto`.

    In the case of a code input with a large number of lines, this behavior causes the page to freeze when focusing on the search input within the code block for the first time.

    Ultimately, I introduced a new dependency, `virtua`, which provides virtual scrolling primitives to fully resolve the issue. And now the bottleneck switched back to the matching algorithm but it's already fast enough for most our use cases.

## Benchmarks

Input code:

```ts
const code = `{
  "compilerOptions": {
    "markup": "<h1>true</h1>",
    "target": "es2020",
    "module": "esnext",
    "moduleResolution": "node",
    "allowUnreachableCode": false,
    "exactOptionalPropertyTypes": true,
    "noFallthroughCasesInSwitch": true,
    "noImplicitReturns": true,
    "noUncheckedIndexedAccess": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "strict": true,
    "jsx": "preserve"
  },
  "include": [
    "./src",
    "./types",
    "./particularly-long-value-that-will-inadvertently-cause-scrolling-for-narrower-containers"
  ]
}
`.repeat(100)
```

| Task | Query | Fast path<br><sup>Doesn't contain `<` or `&`</sup> | Original<br><sup>(Ops/s)</sup> | Optimized<br><sup>(Ops/s)</sup> | 🚀 |
| -- | -- | :--: | --: | --: | --: |
| [Highlight](https://jsbm.dev/2w4giA8RM4Scd) | `"true"` | | 8,015 | 10,778 | 1.3× |
| [Highlight](https://jsbm.dev/N0ROwN0sRiAJZ) | `" "` | | 1,923 | 2,730 | 1.4× |
| [Highlight](https://jsbm.dev/lwvyCvxGaRTLA) | `"true"` | ✓ | 9,460 | 31,690 | 3.3× |
| [Highlight](https://jsbm.dev/ixyrVZS0zSLH3) | `" "` | ✓ | 2,045 | 6,996 | 3.4× |
| [Highlight](https://jsbm.dev/GmYUhhjnrLHIr) | `/true/` | | 8,323 | 10,982 | 1.3× |
| [Highlight](https://jsbm.dev/FTAfF40CsQFCL) | `/./` | | 505 | 41,814 | 82.8× |
| [Highlight](https://jsbm.dev/nWRtl8E1RlDkk) | `/true/` | ✓ | 9,460 | 32,103 | 3.4× |
| [Highlight](https://jsbm.dev/s6JSElAH9DDzl) | `/./` | ✓ | 533 | 64,362 | 120.8× |
| [Match](https://jsbm.dev/HyJSVQeLTYFlm)† | `"true"` | - | 9,135 | 462,782 | 50.7× |
| [Match](https://jsbm.dev/y5j82ccR4zHZf)† | `" "` | - | 790 | 441,588 | 559.0× |
| [Match](https://jsbm.dev/Lu94bgn7iqfjJ)† | `/true/` | - | 8,840 | 250,267 | 28.3× |
| [Match](https://jsbm.dev/ZQBwXOqfdGa0I)† | `/./` | - | 124 | 6,016 | 48.5× |

> † The original implementation is too slow so only used 1/10 of the input code.

[KM-888]: https://konghq.atlassian.net/browse/KM-888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.vscode/settings.json
```diff
@@ -44,5 +44,6 @@
   "scss.scannerExclude": [
     "**/.git",
     "**/bower_components"
-  ]
+  ],
+  "eslint.useFlatConfig": true
```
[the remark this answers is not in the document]
Justineo: Yes it's needed because I was rejected by ESLint in prepush hooks but the linting error isn't showing up in VSCode. After I added this config the error shows up in the editor.

src/utilities/codeBlockHelpers.ts
```diff
@@ -0,0 +1,215 @@
...
+    return getAllMatchingLineNumbersByRegExp(code, query)
+  } else {
+    return getAllMatchingLineNumbersByExactMatch(code, query)
+  }
+}
+
+// ================================
+// Highlighting matching characters
+// ================================
+const wrapMark = (match: string) => `<mark class="matched-term">${match}</mark>`
```
Justineo, line 127: Use `<mark>` instead of `<span>` for semantic correctness.

src/utilities/codeBlockHelpers.ts
```diff
@@ -0,0 +1,215 @@
...
+  let high = lineOffsets.length - 1
+
+  while (low < high) {
+    const mid = (low + high + 1) >>> 1
+    if (lineOffsets[mid] <= offset) {
+      low = mid
+    } else {
+      high = mid - 1
+    }
+  }
```
adamdehaven, line 37: nit: add some comments plz

src/utilities/codeBlockHelpers.ts
```diff
@@ -0,0 +1,215 @@
...
+// especially useful for large bodies of text where the overhead of escaping is non-trivial
+// It also allows to only escape necessary characters.
+export function escapeHTMLIfNeeded(raw: string, regExp = ESCAPE_REGEX, escape = escapeInnerHTML): string {
+  return regExp.test(raw) ? escape(raw) : raw
+}
+
+export function highlightMatchingChars(code: string, query: string, isRegExpMode: boolean): string {
+  let regExp: RegExp
+  try {
+    // This RegExp mus have a capture group to work here
```
adamdehaven, line 147: nit
```suggestion
    // This RegExp must have a capture group to work here
```

src/utilities/codeBlockHelpers.ts
```diff
@@ -0,0 +1,215 @@
...
+    regExp = new RegExp(`((?:${isRegExpMode ? query : escapeRegExp(query)})+)`, 'sgi')
```
adamdehaven, line 148: question: I don't have the reference on hand, but doesn't Safari have issues with lookbehind?

src/utilities/codeBlockHelpers.ts
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  return result.join('')
+}
+
+// ========================
+// Line highlighting syntax
+// ========================
+
+export const LINE_NUMBER_EXPRESSION_REGEX = /^\d+(?:-\d+)?(?:,\d+(?:-\d+)?)*$/
```
adamdehaven, line 173: suggestion: please add comments to explain the expression

src/utilities/codeBlockHelpers.cy.ts
```diff
@@ -0,0 +1,229 @@
...
+
+describe('highlightMatchingChars', () => {
+  it('wraps matched characters with <mark class="matched-term"> matched by exact match', () => {
+    expect(highlightMatchingChars(code, 'true', false)).eq(`{
+  "compilerOptions": {
+    "target": "es2020",
+    "module": "esnext",
+    "moduleResolution": "node",
+    "allowUnreachableCode": false,
+    "exactOptionalPropertyTypes": <mark class="matched-term">true</mark>,
```
adamdehaven, line 87: suggestion: can you add the mark HTML into a util function here in the test file so that you can call the function with the expected content and output rather than repeating the HTML throughout all of the tests? 

Ideally you could just import your wrap function from the actual util file in case the actual wrapping HTML changes, it won't break the tests.

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -656,47 +654,8 @@ function updateMatchingLineNumbers(): void {
...
-    if (match.index !== undefined) {
-      const lineNumber = code.substring(0, match.index)?.split('\n').length
-      totalMatchingLineNumbers.push(lineNumber)
-    }
-  }
-
-  return totalMatchingLineNumbers
-}
+// Debounce the update function so that repeated calls within the same task will only trigger a single update.
+const queueUpdateMatchingLineNumbers = debounce(updateMatchingLineNumbers, 0)
```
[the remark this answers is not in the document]
Justineo: Yes this is intentional but I removed this later as we only need to remove the unnecessary update upon `isShowingFilteredCode` change. This debouncing became unnecessary after that.

src/utilities/codeBlockHelpers.ts
Justineo: It's a non-capture group, not a look behind expression.

src/utilities/codeBlockHelpers.cy.ts
Justineo: Because though the major part of the code stays the same, the highlight function modifies the initial code in a non-trivial manner so it's kind of hard to extract a "render function" for this piece of code input.

src/utilities/codeBlockHelpers.ts
Justineo: Done.

src/utilities/codeBlockHelpers.ts
Justineo: Done.

src/utilities/codeBlockHelpers.ts
Justineo: Done.

src/utilities/codeBlockHelpers.cy.ts
```diff
@@ -0,0 +1,229 @@
...
+})
+
+describe('escapeHTMLIfNeeded', () => {
+  it('escapes only < and &', () => {
+    const regex = /<&>/
+    const escape = cy.spy((v: string) => v)
+
+    expect(escapeHTMLIfNeeded('{ foo: "bar" }', regex, escape)).eq('{ foo: "bar" }')
+    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+    expect(escape).not.have.been.called
```
johncowen, line 68: Just thinking out loud more than anything, but bit annoying there's an API here to make things nicer/easier/shorter to write, but then the linter makes you also add a large comment above it 🤷 😆 

(half noting here as this is related to a completely different piece of work we are doing over in Kuma and it might be handy to link to this at some point, feel free to resolve)

src/utilities/codeBlockHelpers.cy.ts
johncowen, line 1: Based on the `.cy.` here it looks like you chose to use Cypress to test this out (did I understand that right?) because there seems to be no existing CLI runner in kongponents, was that the reason?

johncowen (commented):
Thanks for digging into this! Looks great!

We spoke offline, and I mentioned I'm gonna have a play with this in place (we/kuma-gui originally reported a performance issues with KCodeBlock in Slack).

I had a relatively quick scan of the code, but I for one am happy you've done this with some significant care, so I'm pretty confident it'll be a big improvement judging by the description and the overall approach (also thanks taking the time with the detailed description!)

I left a couple of Q's mainly for myself, so let me know on those. I'll try and give this a spin in our application today and pop back later 👍 , could end up being after the weekend but hopefully today.

johncowen:
>  I'll try and give this a spin in our application today and pop back later 👍 , could end up being after the weekend but hopefully today.

P.S. But don't wait for me if you get an approval from elsewhere and want to go ahead and merge

src/utilities/codeBlockHelpers.cy.ts
Justineo: Yes as it's the only unit test file for now I think I'd just use the existing test framework.

src/utilities/codeBlockHelpers.cy.ts
Justineo: I think I'd just leave this as-is for now and may change this if we switch to Vitest at some point.

src/utilities/codeBlockHelpers.cy.ts
adamdehaven: I was just suggesting something like this on the lines that should be highlighted

```ts
import { wrapMark } from './codeBlockHelpers'

// ... test code

"noFallthroughCasesInSwitch": `${wrapMark(true)}`,
```

Justineo: Oh I misread this. That makes sense.

src/utilities/codeBlockHelpers.cy.ts
johncowen: Oh yeah sorry, totally leave as is! It was more of a self-note so I can link to this later for my own purposes.

src/utilities/codeBlockHelpers.cy.ts
johncowen: Gotcha 👍 , makes sense, thanks!

Justineo:
In addition to optimizing the matching and highlighting algorithms, I implemented several CSS enhancements to significantly improve rendering performance when loading large files (e.g., files with more than 20,000 lines). These changes are detailed in https://github.com/Kong/kongponents/pull/2562/commits/b20c7aff0a73203a456463df689f8de3767476c7, with `content-visibility: auto` providing the most substantial performance boost.

I have also added two 21,000-line demos to the sandbox: one under “# programmatic highlightedLineNumbers & syntax highlighting” and the other under “maxHeight”.

johncowen:
Hey 👋 I gave this a spin in our app with a ~18,000 line config file (not totally sure what realistic size we are expecting in our app 🤷 ).

I then tried searching for the letter `c` and timing the old school human way, i.e. stopwatch, to time the freeze/delay. I then did this for `main` and this PR: `feat/codeblock-perf`

- `main`: several seconds (~4)
- `feat/codeblock-perf`: barely noticeable

So all in all with my completely non-scientific testing, this seems great! 🎉 

BTW I haven't tried this with the subsequent CSS changes you've made which also sound great 🙌 

Couple of random thoughts whilst I was doing this:

- I noticed that if I copy and pasted the word `config` straight into the search, instead of typing a character at a time I didn't notice any significant difference or freeze between the two PRs.
- I up-ed the line count of the config file significantly (> 100,000) and managed to get it to freeze (multiple seconds) via a single keypress even with this PR (note: without the new CSS changes). I guess we might not ever get config files with 100,000 lines, but I guess it simulates an less powerful computer?

Anyway, one tiny suggestion would be (maybe not for this PR): Do you think it would be worthwhile adding a debounce to the input field? Maybe even as a parameter to the component so the consumer of the component can choose whether they'd like to add a debounce or not (or maybe the other way around, add debounce by default, but give folks the option to turn it off)

---

On a completely different note. I found it difficult to get this PR installed using the `@kong/kongponents@pr-2562` version, but maybe I was just holding it wrong, or suffering from Monday morning syndrome 😆 

In the end I built this PR and copied the result over to test with, adding a console log to the top of the entire file to verify I was running the code I was expecting to run.

Couple of other things off the back of that that are maybe for @adamdehaven 

Running `pnpm build` gave me this:

Screenshot 2025-01-13 at 09 51 40

I removed the `build:cli` bit from the `build` script, and that got me past that. I then encountered this:

Screenshot 2025-01-13 at 09 53 14

Again, might just be me/Monday morning, but might be worth someone else checking that its just me.

Justineo:
> Anyway, one tiny suggestion would be (maybe not for this PR): Do you think it would be worthwhile adding a debounce to the input field? Maybe even as a parameter to the component so the consumer of the component can choose whether they'd like to add a debounce or not (or maybe the other way around, add debounce by default, but give folks the option to turn it off)

The old implementation already denounced the query input with a delay of 150ms.

Justineo:
A long debounce only makes it feel slow. I think current delay is find and I think in most cases it should be fast enough now. Also the speed depends on multiple factors, eg. Whether there are a lot of matches or does the code contain characters that need escaping, etc.

johncowen:
Yep no prob 👍

adamdehaven:
> Couple of other things off the back of that that are maybe for @adamdehaven
> 
> Running `pnpm build` gave me this:
> 
> Screenshot 2025-01-13 at 09 51 40

Thanks @johncowen this has been fixed

johncowen:
> In https://github.com/Kong/kongponents/commit/58dec25039913b51be20a218a1a69324f1708fbc I added virtua to alleviate the potential performance issues of rendering a large number of line number elements. As a result, the search experience should now be much smoother. Read more https://github.com/Kong/kongponents/pull/2562#:~:text=Significant%20improvement%20in%20Rendering%20Performance.

Ah perfect! So this is to only render the bits we need to see which would result in less DOM nodes on the page at one time?

Lemme take this one for a spin!

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -213,23 +164,87 @@
...
+        class="filtered-code-block"
+        data-testid="filtered-code-block"
+      >
+        <Virtualizer
+          v-if="!singleLine"
+          v-slot="{ item: line }"
+          v-bind="getVirtualizerProps(true)"
+        >
+          <a
+            :id="`${linePrefix}-L${line}`"
```
johncowen, line 187: I noticed the `id` usage here and it reminded me of something.

This point is not at all related to the performance scope of this PR, but while we are here could we either:

- remove the component `props.id` entirely and use `uniqueId` from inside the component
- or at least default `props.id` to a `uniqueId` from inside the component

That way the consumer doesn't have to provide a `id` even though its only used for internal functionality.

I would guess the first option there would be prefered choice.

FYI: We (over in kuma) already do this via one of our wrappers and our uniqueId `++` helper (I think theres a similar native one since Vue3.5 you can use) https://github.com/kumahq/kuma-gui/blob/fb104ddbd6b9fb7e140f27446f54e6cfd3046c8d/packages/kuma-gui/src/app/x/components/x-code-block/XCodeBlock.vue#L57 but we could remove this extra stuff if it was removed here. It would probably help a lot of other consumers also.

(lmk if I missed that `id` is used for other things, but it doesn't look like it is)

johncowen:
I had a quick look here at this next iteration, and I'm not totally sure if its any better or not. I couldn't see any DOM recycling happening when looking in Web Inspector, but I might have misunderstood where I should expect that. How can I validate the the DOM recycling/virtual scrolling thing is happening? Is it only in certain cases in certain areas? I'd understood it was only for the line numbers? Is that right?

Justineo:
> I had a quick look here at this next iteration, and I'm not totally sure if its any better or not. I couldn't see any DOM recycling happening when looking in Web Inspector, but I might have misunderstood where I should expect that. How can I validate the the DOM recycling/virtual scrolling thing is happening? Is it only in certain cases in certain areas? I'd understood it was only for the line numbers? Is that right?

Where exactly did you look? You can verify this in the [preview sandbox](https://deploy-preview-2562--kongponents-sandbox.netlify.app/#/codeblock).

If DOM recycling is functioning properly, you should see the line elements updating as you scroll.

[recycle.webm](https://github.com/user-attachments/assets/09fe988e-8a1e-47ab-b24f-de730da67b38)

johncowen:
> Where exactly did you look?

In our application not the sandbox

> If DOM recycling is functioning properly, you should see the line elements updating as you scroll.

Ah awesome yeah, that was what I was expecting to see but didn't 🤔. 

Lemme try again in a bit

Justineo:
> In our application not the sandbox

Did you reinstalled the preview package (and make sure there's no cache or sth)?

johncowen:
I previously had problems just installing via `@kong/kongponents@pr-2562` and weirdly file linking (see above somewhere). 

So I hack-ily checked out this branch separately, built it, copied the resulting source from `/dist` over into my applications node_modules folder, added a `console.log()` to the top of the copied source to verify I was running it, I also checked for several things that are added in this PR via Web Inspector (such as the autocomplete attribute). For this second iteration I also had to manually add the new deps you added to my applications package.json (i did this first so I didn't install over the hacky kongponents copy)

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -135,66 +136,16 @@
...
+    <div
+      ref="codeBlockContent"
+      class="code-block-content"
+    >
+      <!--
+        This was moved up before the lines because 1Password's (yes, 1Password!) heuristics will cause a delay because
+        it will walk through all line number elements one by one to calculate layout metrics, this will cause the
+        browser to schedule layout recalculations after each element is processed.
+        Though we are using virtual scrolling now, this still helps with the overall performance.
+      -->
```
adamdehaven, line 148: nit: we need to remove these inline HTML comments before merge; we don't want to ship these. If we want the record, you can add block comments in the script section, etc.

src/components/KCodeBlock/KCodeBlock.vue
adamdehaven: Let's please move this to a new discussion; not in scope for this PR

adamdehaven, line 187: suggestion: you utilize the `${linePrefix}-L${line}` in a few places in the loop. Could we extract this to a getter?

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -213,23 +164,87 @@
...
+          ref="codeBlockLineNumbers"
+          v-slot="{ item: line }"
+          v-bind="getVirtualizerProps(false)"
+        >
+          <a
+            :id="`${linePrefix}-L${line}`"
+            class="line-anchor"
+            :class="{ 'hide-links': !showLineNumberLinks }"
+            :href="showLineNumberLinks ? `#${linePrefix}-L${line}` : undefined"
+          >{{ line }}</a>
```
adamdehaven, line 218: suggestion: should the anchor here have an `aria` attribute(s) to indicate what the link represents?

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -240,7 +255,10 @@ const IS_MAYBE_MAC = typeof navigator !== 'undefined' &&
 const ALT_SHORTCUT_LABEL = IS_MAYBE_MAC ? 'Option' : 'Alt'
 
 // Debounces the search handler which ensures that we don’t trigger several searches while the user is still typing.
-const debouncedHandleSearchInputValue = debounce(handleSearchInputValue, 150)
+// Use Lodash’s debounce function as it supports leading and trailing options.
+// Adding `leading: true` to the options ensures that the search is triggered immediately when the user types the first character,
+// mak
```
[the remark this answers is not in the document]
adamdehaven: I think this would be ideal if you'd like to create a separate JIRA. We'd just need to check the usage in all components and ensure we aren't doing anything special (and that it's not exported)

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -480,21 +509,20 @@ watch(() => props.highlightedLineNumbers, function() {
   setDefaultMatchingLineNumbers()
 }, { immediate: true, deep: true })
 
-watch(() => isShowingFilteredCode.value, async function() {
+watch(() => isShowingFilteredCode.value, function(value) {
```
adamdehaven, line 512: nit: since this is a ref, you don't need the getter. Also, can we rename the param something besides `value` (there's enough "value" in Vue already 😅 )
```suggestion
watch(isShowingFilteredCode, function(isFilteredCode) {
```

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -743,13 +737,17 @@ function jumpToMatch(direction: number): void {
     return
   }
 
-  const line = codeBlock.value.querySelector(`#${linePrefix.value}-L${lineNumber}`)
-  if (line instanceof HTMLElement) {
+  const line = codeBlock.value.querySelector<HTMLElement>(`#${linePrefix.value}-L${lineNumber}`)
```
adamdehaven, line 740: nit: here's another place using a getter for this combination would be useful

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -768,6 +766,31 @@ async function copyCode(event: Event): Promise<void> {
 }
 
 const getIconColor = computed(() => props.theme === 'light' ? KUI_COLOR_TEXT_NEUTRAL_STRONG : KUI_COLOR_TEXT_INVERSE)
+
+type VirtualizerProps = InstanceType<typeof Virtualizer>['$props']
```
adamdehaven, line 770: nit: we typically define all types separately in the `types/` directory, or at least at the top of the script

May make sense here as it's next to the function; just resolve the comment if you'd prefer to keep it here

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -768,6 +766,31 @@ async function copyCode(event: Event): Promise<void> {
...
+    }),
+    overscan: 8,
+    itemSize: parseInt(KUI_LINE_HEIGHT_30, 10),
+    scrollRef: codeBlockContent.value ?? undefined,
+    style: {
+      position: 'absolute',
+      width: 'auto',
+    },
+  }
+}
```
adamdehaven, line 793: suggestion: please add some code comments here to explain the logic & usage

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "@vue/tsconfig/tsconfig.json",
   "compilerOptions": {
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
adamdehaven, line 4: This definitely isn't related to the PR, but yes 👍🏼

src/components/KCodeBlock/KCodeBlock.vue
johncowen: See https://github.com/Kong/kongponents/issues/2575

adamdehaven (approved):
lgtm

PR #2563, opened 2025-01-10
fix(kinput): add type button to mask toggle

lukewhchen:
# Summary
To fix the issue when clicking the mask-toggle button triggers a form submission.
slack: https://kongstrong.slack.com/archives/C05UE909XU5/p1736469294982699

After fix: 

https://github.com/user-attachments/assets/ec8d9751-f031-4b39-a7cd-9481c7e39d37

### DIFF
--- a/src/components/KInput/KInput.vue
+++ b/src/components/KInput/KInput.vue
@@ -51,6 +51,7 @@
           v-if="type === 'password' && showPasswordMaskToggle"
           :aria-label="`${maskValue ? 'Hide' : 'Show'} value`"
           class="mask-value-toggle-button"
+          type="button"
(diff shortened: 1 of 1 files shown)


src/components/KInput/KInput.vue
```diff
@@ -51,6 +51,7 @@
           v-if="type === 'password' && showPasswordMaskToggle"
           :aria-label="`${maskValue ? 'Hide' : 'Show'} value`"
           class="mask-value-toggle-button"
+          type="button"
```
adamdehaven, line 54: good catch

lukewhchen: I thought you wouldn’t notice since the PR is a draft...

PR #2569, opened 2025-01-14
fix(kmultiselect): select jump glitch [KHCP-14366]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-14366

Add `min-height: 24px;` on selected badges element to prevent it from jumping once the first item is selected


kaiarrowood (approved):
LGTM! 🚀

PR #2570, opened 2025-01-15
refactor(kcodeblock): secondary actions as sticky position [KHCP-14618]

afrzhussain:
Jira ticket : https://konghq.atlassian.net/browse/KHCP-14618
Updated the secondary actions position as sticky to make sure actions are visible when scrolled to the bottom

### DIFF
--- a/src/components/KCodeBlock/KCodeBlock.vue
+++ b/src/components/KCodeBlock/KCodeBlock.vue
@@ -183,6 +183,7 @@
         </span>
         <code v-html="filteredCode" />
       </pre>
+
       <pre
         v-else
         class="highlighted-code-block"
@@ -820,17 +821,13 @@ $kCodeBlockDarkLineMatchBackgroundColor: rgba(255, 255, 255, 0.12); // we don't 
     gap: var(--kui-space-40, $kui-space-40);
   }
 
-
-
-
-
   @media (min-width: $kui-breakpoint-laptop) {
-    .secondary-actions-wrapper {
+    .code-block-content .secondary-actions-wrapper {
       opacity: 0;
       transition: opacity $kongponentsTransitionDurTimingFunc, border $kongponentsTransitionDurTimingFunc;
     }
 
-    .secondary-actions-wrapper:focus-within,
+    .code-block-content .secondary-actions-wrapper:focus-within,
     .code-block-content:hover .secondary-actions-wrapper {
       opacity: 1;
     }
@@ -849,8 +846,8 @@ $kCodeBlockDarkLineMatchBackgroundColor: rgba(255, 255, 255, 0.12); // we don't 
       top: 0px;
       z-index: 2;
 
-
       .code-block-secondary-actions {
+        display: flex;
         gap: var(--kui-space-40, $kui-space-40);
         position: absolute;
         right: 0;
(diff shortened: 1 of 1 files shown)


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -897,10 +896,11 @@ $kCodeBlockDarkLineMatchBackgroundColor: rgba(255, 255, 255, 0.12); // we don't
 
     .code-block-secondary-actions {
       display: flex;
+      float: right;
       gap: var(--kui-space-40, $kui-space-40);
       margin-right: var(--kui-space-40, $kui-space-40);
       margin-top: var(--kui-space-40, $kui-space-40);
```
Justineo, line 902: I found that if we remove these two lines, the copy button will fit perfectly in the right place. I checked the Figma component and it seems the spacing is correct.

image

Justineo: Oh I found that this margin is also used by the single line version, maybe we should keep them for that.

afrzhussain: Screenshot 2025-01-15 at 5 17 59 PM
It looks fine in the single line mode without the margins

Justineo: 👌

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -897,10 +896,9 @@ $kCodeBlockDarkLineMatchBackgroundColor: rgba(255, 255, 255, 0.12); // we don't
 
     .code-block-secondary-actions {
       display: flex;
+      float: right;
```
mptap, line 899: We tend to avoid using `float`.

Justineo: +1

`float` will cause the code area to shrink:

image

afrzhussain: Added the changes, changed the structure a bit, by nesting the code-content and also added a sticky wrapper.

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -134,9 +134,28 @@
         </KCodeBlockIconButton>
       </div>
     </div>
-
     <div class="code-block-content">
       <!-- eslint-disable vue/no-v-html -->
```
portikM, line 138: nit: can you please move this eslint disable opening block to line 159? So that it wraps `<pre>` element only?

portikM (changes requested):
One small nit, the rest looks good!

src/components/KCodeBlock/KCodeBlock.vue
adamdehaven: @Justineo I believe this PR and yours may have some overlap

Justineo: Yeah I believe this one should be merged first and then I'll merge this

afrzhussain: Added changes @portikM

portikM (approved):
LGTM 🚀

Justineo (approved):
LGTM

PR #2571, opened 2025-01-15
fix(kalert): kbutton text color within kalerts [KHCP-14565]

afrzhussain:
JIRA ticket : https://konghq.atlassian.net/browse/KHCP-14565
Fixed Kalert modifying Kbutton text color when the button is of type external link( :to is sent in prop)

### DIFF
--- a/src/components/KAlert/KAlert.vue
+++ b/src/components/KAlert/KAlert.vue
@@ -128,8 +128,6 @@ const getAlertIcon = computed((): AlertIcon => {
             color: $hoverColor;
           }
         }
-
-
       }
 
       :deep(code),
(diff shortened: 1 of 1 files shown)


kaiarrowood (commented):
question: the links are blue regardless of alert appearance (which I think is the point of this ticket), but it doesn't have a hover effect. Should the link text darken on hover?

afrzhussain:
> question: the links are blue regardless of alert appearance (which I think is the point of this ticket), but it doesn't have a hover effect. Should the link text darken on hover?
It does have a small hover effect already

src/components/KAlert/KAlert.vue
```diff
@@ -121,11 +121,15 @@ const getAlertIcon = computed((): AlertIcon => {
...
+          color: $textColor;
 
-        &:hover {
-          color: $hoverColor;
+          &:hover {
+            color: $hoverColor;
+          }
         }
+
+
```
kaiarrowood, line 132:
```suggestion
```

nit

kaiarrowood (approved):
LGTM! 🚀

PR #2577, opened 2025-01-17
fix(kcodeblock): updated copy button data-test-id [KHCP-14752]

afrzhussain:
JIRA ticket : https://konghq.atlassian.net/browse/KHCP-14752
Added Unique Id for copy button data-test-id

### DIFF
--- a/src/components/KCodeBlock/KCodeBlock.vue
+++ b/src/components/KCodeBlock/KCodeBlock.vue
@@ -156,7 +156,7 @@
             :aria-label="`Copy (${ALT_SHORTCUT_LABEL}+C)`"
             class="code-block-copy-button"
             :copy-tooltip="`Copy (${ALT_SHORTCUT_LABEL}+C)`"
-            data-testid="code-block-copy-button"
+            :data-testid="`code-block-copy-button${id.length ? '-' + id : '' }`"
             :theme="theme"
             @click="copyCode"
           >


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -156,7 +156,7 @@
             :aria-label="`Copy (${ALT_SHORTCUT_LABEL}+C)`"
             class="code-block-copy-button"
             :copy-tooltip="`Copy (${ALT_SHORTCUT_LABEL}+C)`"
-            data-testid="code-block-copy-button"
+            :data-testid="`code-block-copy-button${id.length ? '-' + id : '' }`"
```
portikM, line 159:
```suggestion
            :data-testid="`code-block-copy-button${id ? '-' + id : '' }`"
```

suggestion: no need to check length here

afrzhussain: Added the changes

src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -156,7 +156,7 @@
             :aria-label="`Copy (${ALT_SHORTCUT_LABEL}+C)`"
             class="code-block-copy-button"
             :copy-tooltip="`Copy (${ALT_SHORTCUT_LABEL}+C)`"
-            data-testid="code-block-copy-button"
+            :data-testid="`code-block-copy-button${id ? '-' + id : '' }`"
```
adamdehaven, line 159: if id is a required prop, how would it be empty...?

kaiarrowood: `''` is considered falsey for prop `{ type: String, required: true }`

afrzhussain: yeah, i'll just remove the logic

PR #2583, opened 2025-01-20
feat(ktreelist): add collapse/expand option [KHCP-14716]

DariaYeremina:
# Summary

https://konghq.atlassian.net/browse/KHCP-14716

- add `collapsible` prop to enable collapse/expand feature.
- add `collapseAll` prop to allow initial collapse/expand for all items.
- add docs
- add component tests

image

### DIFF
--- a/docs/components/tree-list.md
+++ b/docs/components/tree-list.md
@@ -215,24 +215,24 @@ Boolean to hide icons. Defaults to `false`.
 <KTreeList hide-icons :items="items" />
 ```
 
-### collapsable
+### collapsible
 
 Boolean to enable/disable collapse feature. Defaults to `false`.
 
-<KTreeList :collapsable="true" :items="collapsableItems" />
+<KTreeList :collapsible="true" :items="collapsibleItems" />
 
 ```html
-<KTreeList :collapsable="true" :items="items" />
+<KTreeList :collapsible="true" :items="items" />
 ```
 
 ### collapseAll
 
-Boolean to collapse/expand all TreeListItems. Should be used only with `collapsable` prop. Defaults to `false`.
+Boolean to collapse/expand all TreeListItems. Should be used only with `collapsible` prop. Defaults to `false`.
 
-<KTreeList :collapsable="true" :collapse-all="true" :items="collapseAllItems" />
+<KTreeList :collapsible="true" :collapse-all="true" :items="collapseAllItems" />
 
 ```html
-<KTreeList :collapsable="true" :collapse-all="true" :items="items" />
+<KTreeList :collapsible="true" :collapse-all="true" :items="items" />
 ```
 
 ## Slots
@@ -405,7 +405,7 @@ const widthItems = ref<TreeListItem[]>(JSON.parse(JSON.stringify(items)))
 
 const hideIconsItems = ref<TreeListItem[]>(JSON.parse(JSON.stringify(items)))
 
-const collapsableItems = ref<TreeListItem[]>(JSON.parse(JSON.stringify(items)))
+const collapsibleItems = ref<TreeListItem[]>(JSON.parse(JSON.stringify(items)))
 
 const collapseAllItems = ref<TreeListItem[]>(JSON.parse(JSON.stringify(items)))
 
--- a/docs/components/tree-list.md
+++ b/docs/components/tree-list.md
@@ -229,7 +229,7 @@ Boolean to enable/disable collapse feature. Defaults to `false`.
 
 Boolean to collapse/expand all TreeListItems. Should be used only with `collapsable` prop. Defaults to `false`.
 
-<KTreeList :collapsable="true" :collapse-all="true" :items="collapsableItems" />
+<KTreeList :collapsable="true" :collapse-all="true" :items="collapseAllItems" />
 
 ```html
 <KTreeList :collapsable="true" :collapse-all="true" :items="items" />
@@ -406,6 +406,8 @@ const widthItems = ref<TreeListItem[]>(JSON.parse(JSON.stringify(items)))
 const hideIconsItems = ref<TreeListItem[]>(JSON.parse(JSON.stringify(items)))
 
 const collapsableItems = ref<TreeListItem[]>(JSON.parse(JSON.stringify(items)))
+
+const collapseAllItems = ref<TreeListItem[]>(JSON.parse(JSON.stringify(items)))
 
 const slotItems = ref<TreeListItem[]>(JSON.parse(JSON.stringify(items)))
 
--- a/docs/components/tree-list.md
+++ b/docs/components/tree-list.md
@@ -225,14 +225,14 @@ Boolean to enable/disable collapse feature. Defaults to `false`.
 <KTreeList :collapsible="true" :items="items" />
 ```
 
-### collapseAllInitially
+### initialCollapseAll
 
 Boolean to initially collapse/expand all TreeListItems. Should be used only with `collapsible` prop. Defaults to `false`.
 
-<KTreeList :collapsible="true" :collapse-all-initially="true" :items="collapseAllItems" />
+<KTreeList :collapsible="true" :initial-collapse-all="true" :items="collapseAllItems" />
 
 ```html
-<KTreeList :collapsible="true" :collapse-all-initially="true" :items="items" />
+<KTreeList :collapsible="true" :initial-collapse-all="true" :items="items" />
 ```
 
 ## Slots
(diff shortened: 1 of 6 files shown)


src/components/KTreeList/KTreeItem.vue
```diff
@@ -83,6 +104,31 @@ const handleClick = (event: any) => {
 
   emit('selected', props.item)
 }
+
+const isExpanded = ref<boolean>(true)
+
+const expandedIcon = shallowRef(ChevronUpIcon)
+
+const setExpandedIcon = (): void => {
+  expandedIcon.value = isExpanded.value ? ChevronUpIcon : ChevronDownIcon
```
arashsheyda, line 113: would it make sense to use a css class and rotate the chevron icon?

adamdehaven, line 113: When the item is collapsed, we want the chevron to be pointed to the right rather than up

docs/components/tree-list.md
```diff
@@ -215,6 +215,26 @@ Boolean to hide icons. Defaults to `false`.
...
+
+Boolean to enable/disable collapse feature. Defaults to `false`.
+
+<KTreeList :collapsable="true" :items="collapsableItems" />
+
+```html
+<KTreeList :collapsable="true" :items="items" />
+```
+
+### collapseAll
```
adamdehaven, line 228: issue: when I select the first item from the example above, the first item in the example below is also selected.
image


Is this just a case that can happen on the docs site, or is there another bug here that would potentially impact 2 components on the same page?

docs/components/tree-list.md
```diff
@@ -215,6 +215,26 @@ Boolean to hide icons. Defaults to `false`.
...
+
+Boolean to collapse/expand all TreeListItems. Should be used only with `collapsable` prop. Defaults to `false`.
```
adamdehaven, line 230: question: does this mean just "initially collapse" all?

src/components/KTreeList/KTreeDraggable.vue
```diff
@@ -184,6 +200,16 @@ const handleSelectionEvent = (item: TreeListItem): void => {
   emit('selected', item)
 }
 
+const childrenVisibilityMap = ref(new Map<string, boolean>())
```
adamdehaven, line 203: suggestion: could you use `shallowRef` here instead?

src/components/KTreeList/KTreeDraggable.vue
```diff
@@ -184,6 +200,16 @@ const handleSelectionEvent = (item: TreeListItem): void => {
...
+
+const handleExpandedEvent = (value: boolean, id: string): void => {
+  childrenVisibilityMap.value.set(id, value)
```
adamdehaven, line 206: nit: can you rename the first param something other than "value" since that's already used everywhere in Vue?

src/components/KTreeList/KTreeItem.vue
```diff
@@ -140,5 +186,10 @@ const handleClick = (event: any) => {
   &.not-draggable {
     cursor: pointer;
   }
+
+  &-expanded-icon {
+    cursor: pointer;
+    z-index: 10;
```
adamdehaven, line 192: question: why `10` here? Can you add a comment?

src/components/KTreeList/KTreeItem.vue
```diff
@@ -3,14 +3,25 @@
...
+      'collapsed': collapsable && !isExpanded
     }"
     :data-testid="`tree-item-${item.id}`"
     data-tree-item-trigger="true"
     draggable="false"
     type="button"
     @click.prevent="handleClick"
   >
+    <component
+      :is="expandedIcon"
```
adamdehaven, line 17: image

We need the icon to appear outside of the item, closer to as in the Figma

src/components/KTreeList/KTreeItem.vue
```diff
@@ -3,14 +3,25 @@
...
+      v-if="collapsable"
+      class="tree-item-expanded-icon"
+      data-testid="tree-item-expanded-icon"
+      role="button"
```
adamdehaven, line 21: We should add a dynamic `aria-label` for "Collapse" and "Expand," as well as `aria-controls` and `aria-expanded` (you'll need to generate a unique ID with `useId()`

adamdehaven, line 21: We also need to be able to keyboard navigate to this button, so ensure it has the proper tabindex and focus-visible styles

docs/components/tree-list.md
DariaYeremina: found the reason: I used the same array for these 2 examples. Changed to different arrays and there's no issue

src/components/KTreeList/KTreeDraggable.vue
DariaYeremina: it doesn't work as expected if I use `shallowRef`

src/components/KTreeList/KTreeItem.vue
DariaYeremina: it makes the button click easier on draggable wrapper

docs/components/tree-list.md
Justineo: I think that’s a bit odd because it’s neither a default value nor a controlled prop.

If it’s meant to be a default value, it should be overridden by internal state changes (e.g., when users expand or collapse nodes).
If it’s a controlled prop, then `:collapse-all="true"` should ensure all nodes are collapsed.

docs/components/tree-list.md
```diff
@@ -215,6 +215,26 @@ Boolean to hide icons. Defaults to `false`.
 <KTreeList hide-icons :items="items" />
 ```
 
+### collapsable
```
adamdehaven, line 218: issue: spelling; whole PR 

`collapsable` => `collapsible`

src/components/KTreeList/KTreeItem.vue
```diff
@@ -1,41 +1,64 @@
...
+      v-if="collapsable"
+      :aria-controls="`tree-list-draggable-${controlsId}`"
+      :aria-expanded="isExpanded"
+      :aria-label="isExpanded ? 'Collapse' : 'Expand'"
+      class="tree-item-expanded-icon"
+      :class="{
+        'collapsed': !isExpanded,
+        'expanded': isExpanded
+      }"
+      data-testid="tree-item-expanded-icon"
```
adamdehaven, line 16: Since we have the `aria-label` we need to add `decorative` prop to the icon

adamdehaven:
image


The indentation here needs to be tweaked so that all of the top-level items are aligned on the left, and the children are indented visually to the right of the parent

docs/components/tree-list.md
adamdehaven: @DariaYeremina can we add two new methods in the component, `expandAll()` (expands all collapsable items) and `collapseAll()` (collapses all collapsable items), add them to `defineExpose` and the documentation?

Then, the prop name `collapseAll` could be renamed to `collapse-all-initial` (or something) and remove the watcher?

docs/components/tree-list.md
```diff
@@ -215,6 +215,26 @@ Boolean to hide icons. Defaults to `false`.
 <KTreeList hide-icons :items="items" />
 ```
 
+### collapsible
+
+Boolean to enable/disable collapse feature. Defaults to `false`.
+
+<KTreeList :collapsible="true" :items="collapsibleItems" />
```
adamdehaven, line 222: can you update the example to also include an item with no children as the second item so we can see how it displays alongside the collapsible items?

docs/components/tree-list.md
```diff
@@ -215,6 +215,26 @@ Boolean to hide icons. Defaults to `false`.
...
+
+```html
+<KTreeList :collapsible="true" :items="items" />
+```
+
+### collapseAllInitially
```
adamdehaven, line 228: Trying to name consistent to other "initial" props we have, can we rename to `initial-collapse-all`

src/components/KTreeList/KTreeItem.vue
```diff
@@ -1,41 +1,64 @@
...
-    @click.prevent="handleClick"
+  <div
+    class="tree-item-wrapper"
+    :data-testid="`tree-item-wrapper-${item.id}`"
   >
-    <div
-      v-if="hasIcon"
-      class="tree-item-icon"
-      data-testid="tree-item-icon"
+    <ChevronDownIcon
```
adamdehaven, line 6: A few issues:

1. The chevron should point "down" when expanded, and "right" when collapsed.
2. Clicking the button seems to toggle the icon direction.
3. After reordering the items, but not touching collapse icon button, the icon direction seems to change on the docs site.
4. After reordering, I can no longer expand/collapse the items on the first click consistently. Sometimes it takes several clicks.

adamdehaven (changes requested):

adamdehaven:
image


For a child node that itself is not expandable, can we remove the "gap" on the left between the item and the line?

src/components/KTreeList/KTreeDraggable.vue
```diff
@@ -222,11 +277,18 @@ const onStartDrag = (draggedItem: any): void => {
...
   setDragCursor(false)
 
   key.value++
+
+  // set previous icon state after DOM was re-rendered
+  await nextTick(() => {
+    treeItems.value?.forEach((item: typeof KTreeItem) => {
+      item?.setExpandedValue(childrenVisibilityMap.value.get(item?.id))
+    })
+  })
```
adamdehaven, line 291: nit: can we update the syntax here to match the other usage in the repo?
```suggestion
  // set previous icon state after DOM was re-rendered
  await nextTick()
  treeItems.value?.forEach((item: typeof KTreeItem) => {
    item?.setExpandedValue(childrenVisibilityMap.value.get(item?.id))
  })
```

src/components/KTreeList/KTreeDraggable.vue
```diff
@@ -251,15 +312,30 @@ watch(() => props.items, (newValue, oldValue) => {
...
       item.children = []
     }
   })
+
+  // Handles correct initial collapsed/expanded state
+  await nextTick(() => {
+    if (props.collapsible) {
+      triggerCollapse()
+    }
+  })
```
adamdehaven, line 335: ah can we update the syntax here as well?

adamdehaven (approved):
lgtm

PR #2586, opened 2025-01-21
fix(kcodeblock): pre double horizontal scroll

portikM:
# Summary

Before
Screenshot 2025-01-21 at 3 18 17 PM

After
Screenshot 2025-01-21 at 3 37 13 PM

### DIFF
--- a/src/components/KCodeBlock/KCodeBlock.vue
+++ b/src/components/KCodeBlock/KCodeBlock.vue
@@ -898,7 +898,6 @@ $kCodeBlockDarkLineMatchBackgroundColor: rgba(255, 255, 255, 0.12); // we don't 
       display: grid;
       margin: var(--kui-space-0, $kui-space-0);
       min-height: 32px; // ensures that scroll bar doesn't show up when there's only one line of content (not to confuse with single line mode)
-      overflow-x: hidden;
       position: relative;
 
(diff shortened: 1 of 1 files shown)


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -965,6 +965,11 @@ $kCodeBlockDarkLineMatchBackgroundColor: rgba(255, 255, 255, 0.12); // we don't
         min-width: 0;
         overflow-x: auto;
         z-index: 1;
+
+        // making sure code block doesn't have double horizontal scrollbars
+        pre {
+          overflow: unset;
```
adamdehaven, line 971: can you add a test (or at least a sandbox example)?

Justineo:
> @Justineo and I would like to see if this is a result of incorrect Kongponents styling from the beginning, or if something in the host app caused this.

As discussed on Slack, the refactored code block is impacted by some base styles in Konnect, so we need to reset it here.

Justineo:
@portikM I missed this PR and made another one (https://github.com/Kong/kongponents/pull/2592) to fix the same issue. But I believe the changes here doesn't seem to be the correct fix. (You can verify this by adding a `pre { overflow: auto }` style to the sandbox page.)

PR #2587, opened 2025-01-22
feat(ktextarea): add `autosize` for `<KTextArea>`

Justineo:
# Summary

- Introduced a new `autosize` prop that dynamically adjusts the height of the textarea based on its content.
- Added an enhancement for `resizable` textareas where double-clicking the drag handle resets the default sizing (which Firefox supports natively)
- Added a new browser utilities module for handling user-agent detection and ensuring browser compatibility.

Jan-22-2025 10-22-48

### DIFF
--- a/docs/components/textarea.md
+++ b/docs/components/textarea.md
@@ -54,10 +54,22 @@ You can specify `rows` to adjust the height of the textarea.
 
 Boolean value to allow vertically resizing using the drag handle in the right-hand corner of the textarea.
 
+Double-clicking the drag handle will reset the textarea to its default height.
+
 <KTextArea resizable />
 
 ```html
 <KTextArea resizable />
+```
+
+### autosize
+
+Boolean value to automatically adjust the height of the textarea based on its content.
+
+<KTextArea autosize />
+
+```html
+<KTextArea autosize />
 ```
 
 ### help
--- a/src/components/KTextArea/KTextArea.cy.ts
+++ b/src/components/KTextArea/KTextArea.cy.ts
@@ -1,10 +1,13 @@
 import KTextArea from '@/components/KTextArea/KTextArea.vue'
 
+// Helper function to get the number of rendered rows in a textarea
 function getRenderedRows(el: HTMLElement) {
-  const { paddingTop, paddingBottom, lineHeight } = window.getComputedStyle(el)
-  const padding = parseFloat(paddingTop) + parseFloat(paddingBottom)
+  const { paddingTop, paddingBottom, borderTopWidth, borderBottomWidth, lineHeight } = window.getComputedStyle(el)
+  const extraSpacing = [paddingTop, paddingBottom, borderTopWidth, borderBottomWidth]
+    .map((value) => parseFloat(value))
+    .reduce((acc, val) => acc + val, 0)
   const lineHeightValue = parseFloat(lineHeight)
-  return Math.round((el.offsetHeight - padding) / lineHeightValue)
+  return Math.round((el.offsetHeight - extraSpacing) / lineHeightValue)
 }
 
 describe('KTextArea', () => {
@@ -148,7 +151,7 @@ describe('KTextArea', () => {
       },
     })
 
-    cy.get('.input-wrapper').should('have.class', 'autosize')
+    cy.get('.input-textarea-wrapper').should('have.class', 'autosize')
 
     cy.get('textarea')
       .type('1\n2\n3\n4').then(($el) => {
@@ -159,6 +162,23 @@ describe('KTextArea', () => {
       })
       .type('{backspace}'.repeat(6)).then(($el) => {
         expect(getRenderedRows($el[0])).to.eq(3)
+      })
+  })
+
+  it('should not auto-adjust height when `autosize` is not provided', () => {
+    cy.mount(KTextArea, {
+      props: {
+        rows: 2,
+      },
+    })
+
+    cy.get('.input-textarea-wrapper').should('not.have.class', 'autosize')
+
+    cy.get('textarea')
+      .should('have.attr', 'rows', '2')
+      .type('1\n2\n3\n4')
+      .then(($el) => {
+        expect(getRenderedRows($el[0])).to.eq(2)
       })
   })
 })
--- a/src/components/KTextArea/KTextArea.cy.ts
+++ b/src/components/KTextArea/KTextArea.cy.ts
@@ -132,4 +132,23 @@ describe('KTextArea', () => {
 
     cy.get('textarea').should('have.class', 'resizable')
   })
+
+  it('should handle `autosize` prop correctly', () => {
+    cy.mount(KTextArea, {
+      props: {
+        autosize: true,
+        rows: 2,
+      },
+    })
+
+    cy.get('.input-wrapper').should('have.class', 'autosize')
+
+    cy.get('textarea').type('1\n2\n3\n4').then(($el) => {
+      const textarea = $el[0]
+      const { paddingTop, paddingBottom, lineHeight } = window.getComputedStyle(textarea)
+      const padding = parseFloat(paddingTop) + parseFloat(paddingBottom)
+      const lineHeightValue = parseFloat(lineHeight)
+      expect(textarea.offsetHeight).to.eq(4 * lineHeightValue + padding)
+    })
+  })
 })
--- a/src/components/KTextArea/KTextArea.cy.ts
+++ b/src/components/KTextArea/KTextArea.cy.ts
@@ -1,5 +1,12 @@
 import KTextArea from '@/components/KTextArea/KTextArea.vue'
 
+function getRenderedRows(el: HTMLElement) {
+  const { paddingTop, paddingBottom, lineHeight } = window.getComputedStyle(el)
+  const padding = parseFloat(paddingTop) + parseFloat(paddingBottom)
+  const lineHeightValue = parseFloat(lineHeight)
+  return Math.round((el.offsetHeight - padding) / lineHeightValue)
+}
+
 describe('KTextArea', () => {
   it('renders text when value is passed', () => {
     const value = 'Howdy!'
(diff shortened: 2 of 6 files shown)


src/utilities/browser.ts
```diff
@@ -0,0 +1,32 @@
+export const IS_MAYBE_MAC = typeof navigator !== 'undefined' &&
```
adamdehaven, line 1: issue: the utils in this file need to first check they are in a browser environment so that they don't cause issues in SSR

src/components/KTextArea/KTextArea.vue
```diff
@@ -18,16 +18,28 @@
...
-      @input="inputHandler"
-    />
+    <div
+      class="input-wrapper"
+      :class="{
+        autosize,
+        legacy: !SUPPORT_FIELD_SIZING_CONTENT
+      }"
+      :data-value="SUPPORT_FIELD_SIZING_CONTENT ? null : currValue"
+    >
```
adamdehaven, line 28: question: does the `input-wrapper` class inherit styles from anywhere as it didn't exist before?

src/components/KTextArea/KTextArea.vue
```diff
@@ -277,15 +312,44 @@ $kTextAreaPaddingY: var(--kui-space-40, $kui-space-40); // corresponds to mixin,
...
+  .input-wrapper.legacy::after,
   .input-textarea {
     @include inputDefaults;
 
-    // fixing mixed-decls deprecation: https://sass-lang.com/d/mixed-decls
     // stylelint-disable-next-line no-duplicate-selectors
     & {
-      min-height: calc(($kTextAreaLineHeight * 2) + ($kTextAreaPaddingY * 2)); // 2 lines + padding
-      resize: none;
+      grid-area: 1 / 1 / 2 / 2;
```
adamdehaven, line 346: question: why are we suddenly using grid properties?

src/components/KTextArea/KTextArea.vue
```diff
@@ -277,15 +312,44 @@ $kTextAreaPaddingY: var(--kui-space-40, $kui-space-40); // corresponds to mixin,
     }
   }
 
+  .input-wrapper {
+    display: contents;
+
+    &.autosize.legacy {
+      display: grid;
```
adamdehaven, line 319: question: why do we need to change the display here? Couldn't this have downstream impact?

src/components/KTextArea/KTextArea.vue
```diff
@@ -18,16 +18,28 @@
...
-      :value="getValue()"
-      @input="inputHandler"
-    />
+    <div
+      class="input-wrapper"
+      :class="{
+        autosize,
+        legacy: !SUPPORT_FIELD_SIZING_CONTENT
+      }"
+      :data-value="SUPPORT_FIELD_SIZING_CONTENT ? null : currValue"
```
adamdehaven, line 27: question: what's the purpose of `data-value` here?

adamdehaven (changes requested):

src/utilities/browser.ts
adamdehaven: I started reviewing textarea PR on mobile but on second look I don’t think SSR is an issue since you first check that the properties are not undefined; we can verify before merge 👍

src/components/KTextArea/KTextArea.vue
Justineo: https://chriscoyier.net/2023/09/29/css-solves-auto-expanding-textareas-probably-eventually/

This is required by a common hack before `field-sizing: content` is supported.

I'll add the reference link to the comments.

src/components/KTextArea/KTextArea.vue
Justineo: ^ Same as above

src/components/KTextArea/KTextArea.vue
Justineo: It will be referenced by the `attr(data-value)` below.

src/components/KTextArea/KTextArea.vue
Justineo: Didn't notice any of that in the sandbox, I'll put up a preview PR in Konnect.

src/utilities/browser.ts
Justineo: Yes I'm expecting the checks here to be enough.

src/components/KTextArea/KTextArea.vue
adamdehaven: There are still a few issues with `display: contents` in Safari; we should make sure they don't have an impact here

src/components/KTextArea/KTextArea.vue
adamdehaven: Just a bit concerned about the disconnect between the otherwise unused attribute having an actual purpose here in the style block. 

I'm thinking at a minimum we add context for the relationship somewhere (eg comments, etc.)

src/components/KTextArea/KTextArea.vue
Justineo: Or we can simply remove it.

src/components/KTextArea/KTextArea.vue
Justineo: Added.

src/components/KTextArea/KTextArea.vue
Justineo: Removed `display: contents`.

src/components/KTextArea/KTextArea.vue
```diff
@@ -268,24 +309,47 @@ $kTextAreaPaddingY: var(--kui-space-40, $kui-space-40); // corresponds to mixin,
...
-    & {
-      // reset default margin from browser
-      margin: 0;
-      margin-top: var(--kui-space-40, $kui-space-40) !important; // need important to override some overrides of default p margin in other components
+    // reset default margin from browser
+    margin: 0;
+    margin-top: var(--kui-space-40, $kui-space-40) !important; // need important to override some overrides of default p margin in other components
+  }
+
+  .input-wrapper {
```
adamdehaven, line 317: nit: can we rename to `input-textarea-wrapper` to match the naming below?

src/components/KTextArea/KTextArea.vue
```diff
@@ -49,8 +67,12 @@ import { ref, computed, watch, useAttrs, useSlots, useId } from 'vue'
 import useUtilities from '@/composables/useUtilities'
 import KLabel from '@/components/KLabel/KLabel.vue'
 import type { TextAreaLimitExceed } from '@/types'
+import { cssSupports, getScrollbarSize, IS_MAYBE_FIREFOX } from '@/utilities/browser'
 
 const DEFAULT_CHARACTER_LIMIT = 2048
+// Firefox has the double click feature built in, so we don't need to calculate the size
+const RESIZE_HANDLE_SIZE = IS_MAYBE_FIREFOX ? 0 : getScrollbarSize()
+const SUPPORT_FIELD_SIZING_CONTENT = cssSupports('field-sizing', 'content')
```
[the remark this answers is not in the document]
Justineo: > Wouldn't this break the functionality on a page that renders a `<KTextarea autosize />` component on the server?

I believe the auto sizing is a client-only feature as it depends on the render result.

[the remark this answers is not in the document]
adamdehaven: Correct, but I'm asking if the component is initially rendered on the Server, and the utility functions evaluate to false (since there is no `navigator` etc.) wouldn't the autosize functionality on the page not work on first render? (I can test this by pulling into the Nuxt app and adding one to the page if you'd like)

docs/components/textarea.md
```diff
@@ -54,12 +54,24 @@ You can specify `rows` to adjust the height of the textarea.
...
 
+Double-clicking the drag handle will reset the textarea to its default height.
+
 <KTextArea resizable />
 
 ```html
 <KTextArea resizable />
 ```
 
+### autosize
```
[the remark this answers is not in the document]
Justineo: Or we can directly add a `max-height` like GitHub does to the comment box (and I think maybe that's enough):

image

src/components/KTextArea/KTextArea.cy.ts
```diff
@@ -132,4 +139,26 @@ describe('KTextArea', () => {
 
     cy.get('textarea').should('have.class', 'resizable')
   })
+
+  it('should handle `autosize` prop correctly', () => {
```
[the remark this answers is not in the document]
Justineo: Added

src/components/KTextArea/KTextArea.vue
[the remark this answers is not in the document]
Justineo: > wouldn't the autosize functionality on the page not work on first render

No, it won't work on first render. And I believe this is expected.

src/components/KTextArea/KTextArea.vue
Justineo: renamed

docs/components/textarea.md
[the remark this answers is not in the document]
adamdehaven: I guess I'm suggesting the host app my way to control the max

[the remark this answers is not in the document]
Justineo: The built-in `max-height` serves as a fallback and should work in most cases. However, host applications can override it with a more appropriate `max-height` to suit their specific use cases.

src/components/KTextArea/KTextArea.vue
```diff
@@ -18,16 +18,34 @@
...
+      The wrapper element is required for the grid fallback for browsers that don’t support `field-sizing: content`.
+      The data-value attribute is used here to render the textarea content in the grid fallback. It’s rendered via
+      a CSS pseudo-element with `content: attr(data-value) " "`, ensuring it’s at least one line tall.
+      See https://chriscoyier.net/2023/09/29/css-solves-auto-expanding-textareas-probably-eventually/
+    -->
+    <div
+      class="input-textarea-wrapper"
+      :class="{
+        autosize,
+        legacy: autosize && !SUPPORT_FIELD_SIZING_CONTENT
```
adamdehaven, line 31: image
issue: the current implementation functionally works in SSR; however, it causes hydration errors

adamdehaven (changes requested):

src/components/KTextArea/KTextArea.vue
Justineo: The mismatch is expected so I think we'd just allow it. I'll add  a `data-allow-mismatch="class"` annotation here.

adamdehaven: Works for me

Justineo:
@adamdehaven All comments are resolved. Please take a look again.

adamdehaven (approved):

PR #2589, opened 2025-01-22
fix(ktreelist): collapse handle click behavior

portikM:
# Summary

Add `filter` prop in KTreeList to make sure item cannot be dragged by grabbing the expand / collapse handle so that clicks on caret never trigger unintentional dragging


adamdehaven (approved):
Perfect 👨🏼‍🍳

PR #2592, opened 2025-01-23
fix(kcodeblock): prevent unexpected horizontal scrollbar

Justineo:
# Summary

[KM-919](https://konghq.atlassian.net/browse/KM-919)

[KM-919]: https://konghq.atlassian.net/browse/KM-919?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KCodeBlock/KCodeBlock.vue
+++ b/src/components/KCodeBlock/KCodeBlock.vue
@@ -898,6 +898,7 @@ $kCodeBlockDarkLineMatchBackgroundColor: rgba(255, 255, 255, 0.12); // we don't 
       display: grid;
       margin: var(--kui-space-0, $kui-space-0);
       min-height: 32px; // ensures that scroll bar doesn't show up when there's only one line of content (not to confuse with single line mode)
+      overflow: visible;
       position: relative;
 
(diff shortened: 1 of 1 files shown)


portikM:
Two issues I found related to this fix:
1. Sometimes when the scroll bar is at the very bottom it looks as if code block isn't rounded at the bottom - maybe that is okay but if we could control which element is scrolled horizontally maybe we could fix it?
2. When code block is scrollable vertically, the scroll bar ends up at the very bottom which I think isn't ideal


Screenshot 2025-01-23 at 12 59 53 PM
Screenshot 2025-01-23 at 1 04 04 PM

Justineo:
I reviewed the old implementation, and it appears to have been this way even before the recent changes.

`<KTextArea>` has a similar issue, if not a more noticeable one:

image

I believe this can be mitigated by generally improving our scrollbar styles.

Leopoldthecoder (approved):
LGTM

adamdehaven (approved):

PR #2610, opened 2025-02-10
feat(ktreelist): add prop to filter draggable element selectors

adamdehaven:
# Summary

Add a new `ignoreDragSelectors` prop that can be utilized to disable dragging on slotted elements within the `KTreeList`


DariaYeremina (approved):
lgtm

PR #2614, opened 2025-02-13
fix(ktabledata): `loading` prop should override internal loading state

Justineo:
# Summary

Currently, our products usually follow the approach described in the Kongponents documentation, where revalidation is triggered by incrementing the `fetcherCacheKey` (although strictly speaking, this is not the correct method, as it actually caches data for the same query at different locations, which is loading new data instead of revalidation).

After adding the `isLoading` state for SWRV, `<KTableData>` also triggers the loading state when the `fetcherCacheKey` changes (which is technically correct, but the semantics of incrementing the `fetcherCacheKey` are slightly different). This is generally acceptable (for example, after adding or removing list items).

However, in certain scenarios like polling, we actually don’t want to trigger the loading state. But since we haven’t exposed the revalidate-related primitives, users can only achieve this by incrementing the `fetcherCacheKey`, which may lead to unexpected loading states. A quick workaround is for users to control the `loading` prop themselves to prevent the loading state during polling.

This PR implements a solution where, if the user passes the `loading` prop, it will always override the internal loading state of the component (which is also the expected behavior for the `loading` prop), helping users with a quick workaround.

Also updated Cypress to fix https://github.com/cypress-io/cypress/issues/30715.

### DIFF
--- a/src/components/KTableData/KTableData.vue
+++ b/src/components/KTableData/KTableData.vue
@@ -19,7 +19,7 @@
     :hide-pagination="hidePagination || !showPagination"
     :hide-pagination-when-optional="false"
     :hide-toolbar="hideToolbar"
-    :loading="loading || fetcherIsLoading"
+    :loading="loading ?? fetcherIsLoading"
     :max-height="maxHeight"
     :nested="nested"
     :pagination-attributes="tablePaginationAttributes"
@@ -188,7 +188,7 @@ const props = withDefaults(defineProps<TableDataProps>(), {
   rowBulkActionEnabled: () => true,
   rowKey: '',
   cellAttrs: () => ({}),
-  loading: false,
+  loading: undefined,
   emptyStateTitle: 'No Data',
   emptyStateMessage: 'There is no data to display.',
   emptyStateActionMessage: '',
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "commitizen": "^4.3.1",
     "copyfiles": "^2.4.1",
     "cross-env": "^7.0.3",
-    "cypress": "^13.15.2",
+    "cypress": "^13.17.0",
     "cypress-fail-fast": "^7.1.1",
     "cz-conventional-changelog": "3.3.0",
     "eslint": "^9.19.0",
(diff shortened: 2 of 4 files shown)


src/components/KTableData/KTableData.vue
```diff
@@ -19,7 +19,7 @@
     :hide-pagination="hidePagination || !showPagination"
     :hide-pagination-when-optional="false"
     :hide-toolbar="hideToolbar"
-    :loading="loading || fetcherIsLoading"
+    :loading="loading ?? fetcherIsLoading"
```
adamdehaven, line 22: https://github.com/Kong/konnect-ui-apps/blob/37492e0dcbc48d8376d4cc06b1fdc0132203f0ea/apps/gateway-manager/src/pages/GatewayLogsPage.vue#L74

It appears that it’s common practice across our tables in Konnect to initialize a `ref<boolean>` that defaults to `false` for many of our `KTableData` instances. 

Wouldn’t this usage pattern prevent the table’s built-in loading state from showing when it actually _does_ need to initially fetch the data when a component mounts?

Justineo: In the linked implementation, each time `fetcher` is called, it triggers `clearLoadingAndError`, which sets `isLoading` to `true`. The logic seems a bit convoluted, but I assume the behavior remains consistent. We can confirm this by testing it in a preview PR.

Actually, I believe we’re misusing `fetcherCacheKey` to trigger revalidation like this. Instead, we should directly include the fetcher parameters in the cache key.

For the current change, if users explicitly set the `loading` prop to `false`, we should consider respecting that.

Justineo: The behavior of the loading prop isn’t the core of the problem we’re facing. The key issue is that we haven’t exposed a “revalidate only” interface on the Table. The auto-incrementing cache key actually implies loading new data that hasn’t been loaded before, which, technically, is not accurate. Another viable solution is to `defineExpose` the `revalidate` method from `<KTableData>`, and also, for the scenario we’re currently encountering (polling), expose the `refreshInterval` configuration of SWRV as a prop. This way, we can trigger polling revalidation in a more declarative manner. WDYT?

PR #2617, opened 2025-02-14
fix(ktreelist): increase fallback tolerance

portikM:
# Summary

Increase `fallbackTolerance` to make selecting an item easier


adamdehaven (approved):
lgtm

PR #2635, opened 2025-03-05
fix(kmodal): kmodal not scrollable when content slot used [KHCP-15360]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-15360

Fix issue in KModal where it's not scrollable vertically when `content` slot is used


adamdehaven (approved):
lgtm

PR #2640, opened 2025-03-11
fix(klabel): required indicator icon not aligned properly [KM-1034]

jw-foss:
Fix the required indicator bottom value did not include  $kLabelRequiredDotSize to calculate

# Summary

The value of the red dot `bottom` did not include the `$kLabelRequiredDotSize` into its calculation, this PR simply brings the value to the equation so that the red dot can be properly aligned with the label text.

## Screenshots

| Before | After |
| ------ | ----- |
| CleanShot 2025-03-11 at 17 48 52@2x |  CleanShot 2025-03-11 at 17 49 27@2x |


Justineo (approved):
LGTM!

PR #2641, opened 2025-03-11
fix(types): popover attributes allow width

portikM:
# Summary

Allow `width` property in `PopoverAttributes` type for backward compatibility


arashsheyda (approved):
lgtm!

PR #2643, opened 2025-03-11
fix(ktableview): row links tabindex [KHCP-15202]

portikM:
# Summary

Don't set `tabindex` on rows when `rowLink` prop is provided


mptap (approved):
lgtm

PR #2644, opened 2025-03-11
feat(ktabs): add before change prop

portikM:
# Summary

Adds `beforeChange` prop in KTabs

### DIFF
--- a/docs/components/tabs.md
+++ b/docs/components/tabs.md
@@ -221,6 +221,80 @@ const tabChange = (hash: string): void => {
 </script>
 ```
 
+### beforeChange
+
+Prop that takes a function return value of which determines whether tab change should be skipped. Useful when some information needs to me brought to user's attention before they leave the content in the currently displayed tab.
+
+The function receives new tab value as an argument and returns a boolean value, `false` will prevent tab change. Defaults to `() => true`.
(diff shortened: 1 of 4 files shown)


docs/components/tabs.md
```diff
@@ -221,6 +221,80 @@ const tabChange = (hash: string): void => {
 </script>
 ```
 
+### beforeChange
+
+Prop that takes a function return value of which determines whether tab change should be skipped. Useful when some information needs to me brought to user's attention before they leave the content in the currently displayed tab.
+
+The function receives new tab value as an argument and returns a boolean value, `false` will prevent tab change. Defaults to `() => true`.
```
Justineo, line 228: I would suggest we also add support for async operations:

```ts
async function onBeforeTabchange() {
  return await checkSomeCondition()
}
```

Or when we add programatical use for modals:

async function onBeforeTabchange() {
  return await showConfirmModal()
}
```

adamdehaven: I think it's reasonable to assume the callback might need to be asynchronous

portikM: Updated

DariaYeremina (approved):
lgtm

PR #2646, opened 2025-03-12
feat(kpop): add target prop and teleport [KHCP-15449]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-15449

Add `target` prop in KPop to specify a selector for the element where the popover will be teleported. Also expose it through `kpopAttributes` prop for KTooltip

### DIFF
--- a/src/components/KSelect/KSelect.vue
+++ b/src/components/KSelect/KSelect.vue
@@ -431,6 +431,7 @@ const createKPopAttributes = computed(() => {
   return {
     ...defaultKPopAttributes,
     ...props.kpopAttributes,
+    target: '', // unset target in case it was passed in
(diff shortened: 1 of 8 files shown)


src/components/KSelect/KSelect.vue
```diff
@@ -431,6 +431,7 @@ const createKPopAttributes = computed(() => {
   return {
     ...defaultKPopAttributes,
     ...props.kpopAttributes,
+    target: '', // unset target in case it was passed in
```
adamdehaven, line 434: suggestion: what's the default in this case? can we update the comment to explain _why_ we don't want to allow it to be passed in here? (don't need to document the default here)

portikM: Yes. This is in draft for now as it needs more testing, just fyi

PR #2649, opened 2025-03-14
fix(ktableview): make header non-sticky in firefox

portikM:
# Summary

Before
Screenshot 2025-03-14 at 3 45 26 PM

After
Screenshot 2025-03-14 at 3 46 13 PM

### DIFF
--- a/src/styles/mixins/_tables.scss
+++ b/src/styles/mixins/_tables.scss
@@ -72,6 +72,10 @@
           top: 0;
           z-index: 2;
 
+          @-moz-document url-prefix() {
+            position: unset; // make header non-sticky in Firefox because otherwise tooltips are cut off by the table boundaries
+          }
+
           &.is-scrolled {
             background-color: var(--kui-color-background, $kui-color-background);
           }


aanchalm01 (approved):
lgtm!

src/styles/mixins/_tables.scss
```diff
@@ -72,6 +72,10 @@
           top: 0;
           z-index: 2;
 
+          @-moz-document url-prefix() {
+            position: unset; // make header non-sticky in Firefox because otherwise tooltips are cut off by the table boundaries
```
adamdehaven, line 76: So we just aren't going to support sticky headers in firefox? Shouldn't we fix the tooltip positioning instead?

adamdehaven: tbh I'm not sure this change should have been merged in; this degrades the experience in a specific browser instead of providing a working solution

portikM: I couldn't find another solution that would solve the issue. I thought we should fix it now (I would argue that not heaving sticky header does not degrade experience that much - especially if you take into account the fact that only time you get sticky header is when `maxHeight` prop is provided) and revisit once we support teleporting tooltips.

adamdehaven: Would you mind please creating a ticket in the backlog to fix the sticky headers in Firefox? You can link to this PR

PR #2650, opened 2025-03-14
feat(kpop): add target prop and conditional teleport [KHCP-15449]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-15449

Add `target` prop in KPop to specify a selector for the element where the popover will be teleported. Also expose it through `kpopAttributes` prop for KTooltip

### DIFF
--- a/docs/components/popover.md
+++ b/docs/components/popover.md
@@ -396,7 +396,7 @@ Prop for specifying popover offset (string). Default value is `16px`.
 
 ### target
 
-In certain scenarios, you may need to render the popover above other elements. To achieve this, use the `target` prop to specify a selector for the element where the popover will be teleported. When falsy value us passed, the teleport will be disabled, therefore popover won't be [teleported](https://vuejs.org/guide/built-ins/teleport). Defaults to `null`.
+In certain scenarios, you may want to attach the popover to other target elements. To achieve this, use the `target` prop to specify a selector for the element where the popover will be teleported. When falsy value us passed, the teleport will be disabled, therefore popover won't be [teleported](https://vuejs.org/guide/built-ins/teleport). Defaults to `null`.
 
 <KPop
   target="body"
--- a/docs/components/popover.md
+++ b/docs/components/popover.md
@@ -394,6 +394,30 @@ Prop for specifying popover offset (string). Default value is `16px`.
 </KPop>
 ```
 
+### target
+
+In certain scenarios, you may need to render the popover above other elements. To achieve this, use the `target` prop to specify a selector for the element where the popover will be teleported. When falsy value us passed, the teleport will be disabled, therefore popover won't be [teleported](https://vuejs.org/guide/built-ins/teleport). Defaults to `null`.
+
+<KPop
+  target="body"
+  button-text="Open popover"
+>
+  <template #content>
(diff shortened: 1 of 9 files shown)


src/components/KPop/KPop.vue
```diff
@@ -382,224 +392,225 @@ $kPopCaretOffset: 16px;
     display: inline-flex;
     width: 100%;
   }
+}
```
kaiarrowood, line 395: @portikM was this added intentionally?

src/types/popover.ts
```diff
@@ -27,5 +27,8 @@ export interface PopoverAttributes {
   popoverClasses?: string
   offset?: string
   width?: string
+  /** Not supported in KDropdown, KSelect and KMultiselect */
+  target?: string | null
+  trigger?: PopTrigger
```
adamdehaven, line 32: If these are not supported in the three mentioned components, should the `kPopAttributes` props for those components be updated to something like this:

```ts
kPopAttributes: {
  type: Object as PropType<Omit<PopoverAttributes, 'target' | 'trigger'>>,
}
```

src/components/KPop/KPop.vue
```diff
@@ -18,67 +18,72 @@
...
-        :id="popoverId"
-        ref="popoverElement"
-        :aria-labelledby="$slots.title || title ? titleId : undefined"
-        class="popover"
-        :class="popoverClassesObj"
-        role="dialog"
-        :style="floatingStyles"
-        :x-placement="calculatedPlacement"
-      >
+    <KPopTeleportWrapper
```
adamdehaven, line 21: nit: can you point out what was changed in this file? I'm thinking just a wrapper mainly?

adamdehaven (dismissed):
I'd like to verify in the Portal app and docs site before merge

src/components/KPop/KPop.vue
portikM: Yes, in template just wrapper and bound styles (lines 38-39)

src/types/popover.ts
portikM: Oh good idea! Done

docs/components/popover.md
```diff
@@ -394,6 +394,30 @@ Prop for specifying popover offset (string). Default value is `16px`.
 </KPop>
 ```
 
+### target
+
+In certain scenarios, you may need to render the popover above other elements. To achieve this, use the `target` prop to specify a selector for the element where the popover will be teleported. When falsy value us passed, the teleport will be disabled, therefore popover won't be [teleported](https://vuejs.org/guide/built-ins/teleport). Defaults to `null`.
```
adamdehaven, line 399:
```suggestion
In certain scenarios, you may want to attach the popover to other target elements. To achieve this, use the `target` prop to specify a selector for the element where the popover will be teleported. When falsy value us passed, the teleport will be disabled, therefore popover won't be [teleported](https://vuejs.org/guide/built-ins/teleport). Defaults to `null`.
```

kaiarrowood (approved):
LGTM! Also functionally verified Konnect again. Thanks @portikM 🚀 🚀

PR #2654, opened 2025-03-18
fix(types): allow data-testid in popover attributes

portikM:
# Summary

Allow `data-testid` in `PopoverAttributes` type


kaiarrowood (approved):
LGTM! 🚀

Justineo:
Should all valid HTML global attributes be allowed here?

PR #2656, opened 2025-03-19
feat(kcodeblock): add show-line-numbers

Justineo:
# Summary

Add a `show-line-numbers` prop to control the display of line numbers.

### DIFF
--- a/docs/components/codeblock.md
+++ b/docs/components/codeblock.md
@@ -6,14 +6,12 @@ Searching highlights matching lines within the code while filtering shows only m
 
 Searching and filtering can happen by exact matches (the default) and by matching regular expressions. The search-related UI controls can also be interacted with [using keyboard shortcuts](#default-shortcuts).
 
-<ClientOnly>
-  <KCodeBlock
-    id="code-block-default"
-    :code="code"
-    language="json"
-    searchable
-  />
-</ClientOnly>
+<KCodeBlock
+  id="code-block-default"
+  :code="code"
+  language="json"
+  searchable
+/>
 
 ```html
 <KCodeBlock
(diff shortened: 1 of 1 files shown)


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -364,6 +368,15 @@ const props = defineProps({
     default: true,
   },
 
+  /**
+   * Controls whether to show line numbers. **Default: `true`**.
+   */
+  showLineNumbers: {
```
adamdehaven, line 374: nit: when we went through the reskin effort last year we tried to get into the practice of removing prefixes from props like “has” and “is” and “show”

Could we rename the prop `lineNumbers`?

Justineo: I would have named it as `line-numbers` if not for the existing `show-line-number-links` and `show-copy-button`…removing the `show` prefix feels weird now.

adamdehaven: oof yep 🫣

docs/components/codeblock.md
```diff
@@ -245,6 +245,19 @@ Controls whether to show a copy button which copies the original code (the value
 />
 ```
 
+### showLineNumbers
+
+Controls whether to show line numbers. Defaults to `true`.
+
+<ClientOnly>
```
adamdehaven, line 252: Why do we need client only here?

Justineo: I removed all `<ClientOnly>` usages from the `<KCodeBlock>` docs, and everything seems to be working fine.

adamdehaven (approved):

PR #2662, opened 2025-03-24
fix(kpop): close on esc key press [KHCP-15519]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-15519

Close KPop on esc key press

### DIFF
--- a/src/components/KPop/KPop.vue
+++ b/src/components/KPop/KPop.vue
@@ -3,6 +3,7 @@
     :is="tag"
     ref="kPopoverElement"
     class="k-popover"
+    @keydown.esc="onEscapePress"
   >
     <div
       ref="triggerWrapperElement"
@@ -316,6 +317,11 @@ const popoverClassesObj = computed(() => [props.popoverClasses, { 'hide-caret': 
 
 const floatingUpdates = ref<() => void>()
 
+const onEscapePress = (event: Event) => {
+  hidePopover()
+  event.stopPropagation()
+}
+
 defineExpose({
   hidePopover,
   showPopover,
@@ -326,12 +332,6 @@ onMounted(() => {
     // handle various click events to determine how to handle the click event in a generic clickHandler function
     // we don't set any other click event listeners on purpose to avoid conflict of event listeners
     useEventListener(document, 'click', clickHandler)
-
-    useEventListener(document, 'keydown', (event: KeyboardEvent) => {
-      if (event.key === 'Escape') {
-        hidePopover()
-      }
-    })
 
     if (popoverTrigger.value) {
       // determine the element to bind aria-controls attribute to
--- a/src/components/KPop/KPop.vue
+++ b/src/components/KPop/KPop.vue
@@ -98,6 +98,7 @@ import useUtilities from '@/composables/useUtilities'
 import { CloseIcon } from '@kong/icons'
 import { KUI_ICON_SIZE_30, KUI_SPACE_60 } from '@kong/design-tokens'
 import KPopTeleportWrapper from './KPopTeleportWrapper.vue'
+import { useEventListener } from '@vueuse/core'
 
 const props = defineProps({
   buttonText: {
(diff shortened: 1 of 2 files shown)


src/components/KPop/KPop.vue
```diff
@@ -324,7 +325,13 @@ onMounted(() => {
...
     // handle various click events to determine how to handle the click event in a generic clickHandler function
     // we don't set any other click event listeners on purpose to avoid conflict of event listeners
-    document?.addEventListener('click', clickHandler)
+    useEventListener(document, 'click', clickHandler)
+
+    useEventListener(document, 'keydown', (event: KeyboardEvent) => {
+      if (event.key === 'Escape') {
+        hidePopover()
+      }
+    })
```
jw-foss, line 334: **Question:** do you think if it is necessary for us to bind the event to the `document` after the popover is triggered. Or we should do a delegation to the document so that we do ended up binding a ton of events to the document like this. Something in my mind was:

```ts
import { onMounted, onBeforeUnmount } from 'vue'
import type { Ref } from 'vue'

// ignore the name, just showcasing
type UseESCPressedParams = {
  indicator: Ref<boolean>
  handle: () => void
}
const indicatorSet = new Set<UseESCPressedParams>()

export function useESCPressed(params: {
  indicator: Ref<boolean>
  handle: () => void // handle means to close the corresponding popup
}) {
  onMounted(() => {
    indicatorSet.add(params)
  })

  onBeforeUnmount(() => {
    indicatorSet.delete(params)
  })
}

document.addEventListener('keydown', (e) => {
  if (e.key === 'Escape') {
    indicatorSet.forEach(({
      handle,
      indicator,
    }) => {
      if (indicator.value) {
        handle()
      }
    })
  }
})
```

portikM: Sorry I'm not sure I understand what benefit would this composable give us. If your concern is that we bind too many events to the `document` from the KPop components I don't think it's costing us any performance, as long as we clean them up properly.

jw-foss: > If your concern is that we bind too many events to the document from the KPop components

About this I would like to know, would there be many popups showing up at the same time? If true, would the `ESC` key pressing close them all?

portikM: In 99% of use-cases, the dropdowns we use close automatically on click outside - which means it's almost impossible to have two open KPop components on the page at the same time. The 1% falls on custom use cases where we programatically open the popover - I don't know a single use case like that but I guess technically it is possible. So yes, technically pressing escape will close all open popovers but we shouldn't be in situation where we have multiple open popovers at the same time to begin with.

portikM: Reworked my solution so that esc key press only closes popover when it has focus.

src/components/KPop/KPop.vue
```diff
@@ -335,41 +337,23 @@ onMounted(() => {
...
 
       if (props.trigger === 'hover') {
-        popoverTrigger.value.addEventListener('mouseenter', showPopover)
-        popoverTrigger.value.addEventListener('focus', showPopover)
-        popoverTrigger.value.addEventListener('mouseleave', hidePopover)
-        popoverTrigger.value.addEventListener('blur', hidePopover)
+        useEventListener(popoverTrigger.value, 'mouseenter', showPopover)
+        useEventListener(popoverTrigger.value, 'focus', showPopover)
+        useEventListener(popoverTrigger.value, 'mouseleave', hidePopover)
+        useEventListener(popoverTrigger.value, 'blur', hidePopover)
```
Justineo, line 343:
```suggestion
        useEventListener(popoverTrigger.value, 'mouseenter', showPopover)
        useEventListener(popoverTrigger.value, 'focusin', showPopover)
        useEventListener(popoverTrigger.value, 'mouseleave', hidePopover)
        useEventListener(popoverTrigger.value, 'focusout', hidePopover)
```

Should this also use `focusin` and `focusout`? eg. the trigger element is not focusable itself but its descendant is. Like an `<input>` with a wrapper.

Then we can simplify the code further as:

```ts
if (props.trigger === 'hover') {
  [popoverTrigger.value, popoverElement.value].forEach((el) => {
    if (!el) {
      return
    }
    useEventListener(el, 'mouseenter', showPopover)
    useEventListener(el, 'focusin', showPopover)
    useEventListener(el, 'mouseleave', hidePopover)
    useEventListener(el, 'focusout', hidePopover)
  })
}
```

portikM: Good call, addressed.

jw-foss (approved):
LGTM 👍

Justineo (approved):
LGTM 👍

PR #2669, opened 2025-03-28
refactor(kbutton): migrate to type-based defineProps

Justineo:
# Summary

[KM-1094](https://konghq.atlassian.net/browse/KM-1094)

[KM-1094]: https://konghq.atlassian.net/browse/KM-1094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KButton/KButton.vue
+++ b/src/components/KButton/KButton.vue
@@ -17,71 +17,33 @@
 </template>
 
 <script setup lang="ts">
-import type { PropType } from 'vue'
-import { computed, onMounted, useAttrs, useSlots } from 'vue'
-import { ButtonAppearances, ButtonSizes } from '@/types'
-import type { ButtonAppearance, ButtonSize } from '@/types'
+import { computed, onMounted, useAttrs, useSlots, watch } from 'vue'
+import { ButtonAppearances } from '@/types'
+import type { ButtonAppearance, ButtonProps, ButtonSize } from '@/types'
 
-const props = defineProps({
-  /**
-  * Base styling of the button
-  * One of ['primary', 'secondary', 'tertiary', 'danger']
-  */
-  appearance: {
-    type: String as PropType<ButtonAppearance>,
-    default: 'primary',
-    validator: (value: ButtonAppearance): boolean => {
-      return Object.values(ButtonAppearances).indexOf(value) !== -1
-    },
-  },
-  /**
-  * Size variations
-  * One of ['small', 'medium', 'large' ]
-  */
-  size: {
-    type: String as PropType<ButtonSize>,
-    default: 'medium',
-    validator: (value: ButtonSize): boolean => {
-      return Object.values(ButtonSizes).indexOf(value) !== -1
-    },
-  },
-  /**
-  * Route object or path. If object will render <router-link>, if string will render <a>
-  */
-  to: {
-    type: [Object, String],
-    default: null,
-  },
-  type: {
-    type: String,
-    default: 'button',
-  },
-  disabled: {
-    type: Boolean,
-    default: false,
-  },
-  icon: {
-    type: Boolean,
-    default: false,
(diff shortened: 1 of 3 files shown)


src/components/KButton/KButton.vue
```diff
@@ -17,71 +17,33 @@
...
-})
+import { computed, onMounted, useAttrs, useSlots, watch } from 'vue'
+import { ButtonAppearances } from '@/types'
+import type { ButtonAppearance, ButtonProps, ButtonSize } from '@/types'
+
+const {
+  appearance = 'primary',
+  size = 'medium',
+  to = null,
+  type = 'button',
```
adamdehaven, line 28: This `type` prop doesn’t seem to be documented anywhere

src/types/button.ts
```diff
@@ -1,11 +1,21 @@
 import type { Replace } from '@/types/utils'
+import type { RouterLinkProps } from 'vue-router'
 
 export type ButtonAppearance = 'primary' | 'secondary' | 'tertiary' | 'danger' | 'none'
 export type ButtonAppearanceRecord = Record<Replace<ButtonAppearance, '-', ''>, ButtonAppearance>
 
 export type ButtonSize = 'large' | 'medium' | 'small'
 export type ButtonSizeRecord = Record<ButtonSize, ButtonSize>
 
+export type ButtonProps = {
```
adamdehaven, line 10: suggestion: if we can add TS block comments here for each prop, it would expose them should we ever be integrated with vue-component-meta

Justineo: Good idea!

src/components/KButton/KButton.vue
Justineo: Do you mean we should add this to the docs? I’m a bit confused about how we should document it—some components include it under the "HTML attributes" section, while others don't mention it at all.

adamdehaven: I haven't looked how it's used here; but if it's a valid prop, we should document it.

adamdehaven: Here I believe it's a pass through for the HTML attribute so I guess it could be documented similarly

Justineo: Updated.

src/types/button.ts
Justineo: Added.

src/components/KButton/KButton.vue
adamdehaven: I meant should we update on the docs site?

src/types/button.ts
```diff
@@ -1,11 +1,52 @@
...
+  size?: ButtonSize
+
+  /**
+   * Route object or path. If object will render <router-link>, if string will render <a>
+   * @default null
+   */
+  to?: RouterLinkProps['to'] | null
+
+  /**
+   * Button type
```
adamdehaven, line 32: nit:
```suggestion
   * HTML Button type attribute
```

src/types/button.ts
```diff
@@ -1,11 +1,52 @@
...
+   * @default 'button'
+   */
+  type?: HTMLButtonElement['type']
+
+  /**
+   * Disables the button
```
adamdehaven, line 38:
```suggestion
   * When true, disables the button
```

src/types/button.ts
```diff
@@ -1,11 +1,52 @@
...
+   * @default false
+   */
+  disabled?: boolean
+
+  /**
+   * Icon only button
```
adamdehaven, line 44:
```suggestion
   * When true, an icon-only button
```

adamdehaven (approved):

adamdehaven (approved):
lgtm

PR #2690, opened 2025-04-09
feat(composables): introduce useDeprecated for deprecated APIs

jw-foss:
# Summary

While we are doing the `typescript` improvements, I noticed that we have quite a few places were using `validator` for deprecated APIs, since we are on `Vue 3.5` I think it is fair to migrate from that to a composition API. 

In this PR, I added a composable named `useDeprecated` and updated `KButton` as an example, I also included a test file against the newly added composable.

[KM-1149]

[KM-1149]: https://konghq.atlassian.net/browse/KM-1149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KButton/KButton.vue
+++ b/src/components/KButton/KButton.vue
@@ -17,8 +17,9 @@
 </template>
 
 <script setup lang="ts">
-import { computed, onMounted, useAttrs, watch } from 'vue'
+import { computed, onMounted, useAttrs } from 'vue'
 import { ButtonAppearances } from '@/types'
+import { useDeprecated } from '@/composables/useDeprecated'
 import type { ButtonAppearance, ButtonProps, ButtonSize, ButtonSlots } from '@/types'
 
 const {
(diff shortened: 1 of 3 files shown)


adamdehaven (commented):
I'm not sure this change is really necessary. We may not want to always watch and continue to log the deprecations, and the logging was only put in place to aid in migration without needing to push breaking changes to the repository.

If we had gone with an approach similar to this originally I think it would have been fine, but I'm not sure it's necessary to refactor at this time

jw-foss:
> I'm not sure this change is really necessary. We may not want to always watch and continue to log the deprecations, and the logging was only put in place to aid in migration without needing to push breaking changes to the repository.

Understood, for the part "we may not want to watch and continue to log the deprecation" I can provide a `once` watcher for this.

> If we had gone with an approach similar to this originally I think it would have been fine, but I'm not sure it's necessary to refactor at this time

For sure, I'll close this PR if this is what we wanted.

Justineo:
Makes sense either way. I was thinking a consistent way to handle the fraction might be helpful too, if it ends up being reused.

jw-foss:
Closing due to staled

PR #2701, opened 2025-04-16
fix(clipboardprovider): update method signature to prevent downstream…

jw-foss:
Right now, the updated typescript definition breaks the CI since the signature passed to consumers did not match, in this PR we wish to introduce a proxy layer for allowing tolerances. 

See: https://github.com/Kong/konnect-ui-apps/actions/runs/14481370936/job/40618905147?pr=6489

# Summary

| Before | After |
|--------|--------|
|  CleanShot 2025-04-16 at 11 32 51@2x |  CleanShot 2025-04-16 at 11 32 55@2x |

### DIFF
--- a/docs/components/renderless/clipboard-provider.md
+++ b/docs/components/renderless/clipboard-provider.md
@@ -3,14 +3,12 @@
 KClipboardProvider provides clipboard functionality to components.
 
 <KCard>
-  <template #body>
     <KInput :model-value="dataToCopy" @input="newValue => dataToCopy = newValue" type="text" class="vertical-spacing" />
     <KClipboardProvider v-slot="{ copyToClipboard }">
       <KButton @click="() => { if (copyToClipboard(dataToCopy)) alert(`Copied: '${dataToCopy}'`) }">
         Copy to Clipboard
       </KButton>
     </KClipboardProvider>
-  </template>
 </KCard>
 
 ```html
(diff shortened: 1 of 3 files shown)


Justineo:
Most of our consuming apps currently follow this pattern:

```vue
<KClipboardProvider v-slot="{ copyToClipboard }">
  <CopyIcon
    class="copy-icon"
    :color="KUI_COLOR_TEXT_NEUTRAL_WEAK"
    data-testid="copy-to-clipboard"
    :size="KUI_ICON_SIZE_50"
    @click="handleCopyToClipboard(copyToClipboard)"
  />
</KClipboardProvider>
```

```ts
const handleCopyToClipboard = (copyToClipboard: (arg0: string | undefined) => any) => {
  if (!copyToClipboard(selectedEntry.value?.request_uri)) {
    notify({
      appearance: 'danger',
      message: i18n.t('general.copy_to_clipboard_failed'),
    })
  } else {
    notify({
      appearance: 'success',
      message: i18n.t('general.copy_to_clipboard'),
    })
  }
}
```

Given this usage across apps, we think it’s necessary to provide a compatibility layer for now to prevent potential breakages.

We’ve also noticed that in most cases, consumer implementations are incorrect—`copyToClipboard` is asynchronous, so the conditional check always mistakenly takes the success path regardless of whether the copy actually succeeded.

Justineo (approved):
LGTM

Justineo (approved):
Doc changes LGTM.

adamdehaven (commented):
n/a

adamdehaven (approved):
I'm curious why the original PR was merged without first testing the implementation?

Justineo:
> I'm curious why the original PR was merged without first testing the implementation?

There was indeed a preview PR earlier, and I didn’t notice any issues at the time（maybe the pipeline didn’t get triggered due to a conflict, so it slipped through）. Will be more mindful going forward.

PR #2706, opened 2025-04-17
fix(k-input): typescript improvements for k-input

jw-foss:
# Summary

[KM-1139]

TypeScript improvements for `KInput` component.

- [x] Preview at https://github.com/Kong/konnect-ui-apps/pull/6515
  - [x] CI passed

[KM-1139]: https://konghq.atlassian.net/browse/KM-1139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KInput/KInput.vue
+++ b/src/components/KInput/KInput.vue
@@ -91,81 +91,39 @@
 
 <script setup lang="ts">
 import { computed, ref, watch, useSlots, useAttrs, onMounted, nextTick, useId } from 'vue'
-import type { PropType } from 'vue'
-import type { LabelAttributes, LimitExceededData } from '@/types'
+import type { InputProps, InputEmits, InputSlots } from '@/types'
 import useUtilities from '@/composables/useUtilities'
 import KLabel from '@/components/KLabel/KLabel.vue'
 import { KUI_ICON_SIZE_40 } from '@kong/design-tokens'
 import { VisibilityIcon, VisibilityOffIcon } from '@kong/icons'
 
-const props = defineProps({
-  modelValue: {
-    type: [String, Number],
-    default: '',
-  },
-  label: {
-    type: String,
-    default: '',
-  },
-  labelAttributes: {
-    type: Object as PropType<LabelAttributes>,
-    default: () => ({}),
-    validator: (value: LabelAttributes): boolean => {
-      if (value.help) {
-        console.warn('KInput: `help` property of `labelAttributes` prop is deprecated. Please use `info` prop instead. See the migration guide for more details: https://kongponents.konghq.com/guide/migrating-to-version-9.html#klabel')
-      }
+const {
+  modelValue = '',
+  label = '',
+  labelAttributes = {},
+  help = '',
+  error = false,
+  errorMessage = '',
+  characterLimit = null,
+  hasError = false,
+  type = 'text',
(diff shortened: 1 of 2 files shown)


src/types/input.ts
```diff
@@ -8,6 +8,104 @@ export interface LabelAttributes {
...
+   * @defaults false
+   */
+  errorMessage?: string
+
+  /**
+   * Use this prop to specify a character limit for the input.
+   * See the @char-limit-exceeded event for more details.
+   * @defaults null
+   */
+  characterLimit?: number
```
Justineo, line 55: Shall we remove the default value `null` for this one? Otherwise it should be `number | null` here.

src/components/KInput/KInput.vue
```diff
@@ -237,21 +195,21 @@ const helpText = computed((): string => {
...
+  return help
 })
 
 watch(charLimitExceeded, (newVal, oldVal) => {
   if (newVal !== oldVal) {
     emit('char-limit-exceeded', {
       value: currValue.value,
       length: currValue.value.length,
-      characterLimit: props.characterLimit,
+      characterLimit: characterLimit!,
```
Justineo, line 212: Could this be `undefined` or `null`?

jw-foss: @Justineo could be, you can see in the screenshot, if the value `characterLimit` is not a valid value, this watcher will not be triggered.
CleanShot 2025-04-17 at 14 36 13@2x

src/types/input.ts
jw-foss: IMHO, the characterLimit default value should be `0`, when it is `0`, meaning that the limit is `Infinite` or maybe we should use `Number.Infinity` here instead. But since `null` has been used here for this matter, maybe it is safer to keep it this way.

src/components/KInput/KInput.vue
```diff
@@ -91,81 +91,39 @@
...
+  modelValue = '',
+  label = '',
+  labelAttributes = {},
+  help = '',
+  error = false,
+  errorMessage = '',
+  characterLimit = null,
+  hasError = false,
+  type = 'text',
+  showPasswordMaskToggle = false,
```
adamdehaven, line 110: to be like the other PRs, we shouldn't need to define the `false` prop defaults here for boolean props, right?

Justineo (approved):

src/components/KInput/KInput.vue
```diff
@@ -91,81 +91,39 @@
...
-   */
-  hasError: {
-    type: Boolean,
-    default: false,
-    validator: (value: boolean): boolean => {
-      if (value) {
-        console.warn('KInput: `hasError` prop is deprecated. Please use `error` prop instead. See the migration guide for more details: https://kongponents.konghq.com/guide/migrating-to-version-9.html#kinput')
-      }
+const {
+  modelValue = '',
```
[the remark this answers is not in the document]
Justineo: Migrating to `defineModel` would require a bit more refactoring than our current changes. Given our current priorities, I think it's okay to keep the changes minimal for this round.

adamdehaven (approved):

PR #2717, opened 2025-04-29
fix(kfileupload): ts improvements for kfileupload

Justineo:
# Summary

[KM-1138](https://konghq.atlassian.net/browse/KM-1138)

- [x] Preview PR: https://github.com/Kong/konnect-ui-apps/pull/6624
  - [x] CI passed

[KM-1138]: https://konghq.atlassian.net/browse/KM-1138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/components/file-upload.md
```diff
@@ -114,7 +114,7 @@ This is the text that will be displayed on the button that triggers the click on
 
 ### maxFileSize
 
-Use this prop to customize the maximize size of file that can be uploaded. Default value for `image` upload is `1MB` and for `file` is `5.25MB`.
+Use this prop to customize the maximize size of file that can be uploaded. Default value is `5.25MB` if no valid value is provided.
```
Justineo, line 117: I noticed that the current implementation incorrectly calculates the KB/MB size from bytes. By default, we are defining the `maxFileSize` as 5,250,000 bytes, which the current implementation converts to 5.25MB. However, since 1KB should equal 1024 bytes, 5,250,000 bytes is actually about 5.007MB. This makes more sense if we intend to set a default max size of 5MB, but due to the incorrect assumption that 1KB equals 1000 bytes, it results in 5.25MB. Shall we fix the implementation and make the default value 5MB? @adamdehaven @portikM

portikM: Yeah I think we should, not sure how we ended up rounding 1KB to 1000 bytes here..

portikM (approved):
LGTM 🚀

PR #2723, opened 2025-04-30
chore: bump node and pnpm

portikM:
# Summary

Bump pnpm & node versions

https://docs.cypress.io/app/get-started/install-cypress#pnpm-Configuration


pnpm-workspace.yaml
portikM, line 1: https://docs.cypress.io/app/get-started/install-cypress#pnpm-Configuration

adamdehaven (approved):
lgtm

PR #2724, opened 2025-04-30
fix: link main page to correct url for theming and address SSR build [MA-3892]

Chalks:
# Summary

Addresses [MA-3892](https://konghq.atlassian.net/browse/MA-3892). Depends on #2722 

I noticed that the "Themeable ... read more" tile on the docs homepage links to the wrong page. While I fixed that (which was just a minor route change) I fixed some issues with the Vitepress build. Because Vitepress is SSR by default, it was exposing a few build issues with the components when used in that context.

**Page 404 issue**

https://github.com/user-attachments/assets/bf21269c-3b5b-44c7-83ff-1cd55d092a7e

**BEFORE - Build warnings when Vitepress builds the docs**

https://github.com/user-attachments/assets/3885552e-30c5-408c-86b3-cafd0ebdc960

**AFTER - Almost all build warnings are gone when Vitepress builds the docs**

https://github.com/user-attachments/assets/f5e16289-2935-45de-b4d2-78bae14b75b3


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -385,6 +385,10 @@ watch(() => highlightedLineNumbers, function() {
 }, { immediate: true, deep: true })
 
 watch(() => isShowingFilteredCode.value, function(value) {
+  if (typeof document === 'undefined') {
```
adamdehaven, line 388: question: why not just check `document` here as well?

src/components/KModalFullscreen/KModalFullscreen.vue
```diff
@@ -217,17 +221,23 @@ const proceed = () => {
 }
 
 onMounted(() => {
-  document.addEventListener('keydown', handleKeydown)
+  if (typeof document !== 'undefined') {
+    document.addEventListener('keydown', handleKeydown)
+  }
```
adamdehaven, line 226: `onMounted` is only ever called in the client, so this check shouldn't be needed

src/components/KModalFullscreen/KModalFullscreen.vue
```diff
@@ -217,17 +221,23 @@ const proceed = () => {
...
+  if (typeof document !== 'undefined') {
+    document.removeEventListener('keydown', handleKeydown)
+  }
 })
 
 onUnmounted(() => {
-  document.body.style.overflow = ''
+  if (typeof document !== 'undefined') {
+    document.body.style.overflow = ''
+  }
```
adamdehaven, line 240: `onBeforeUnmount` and `onUnmounted` are only ever called in the client, so these checks shouldn't be needed

src/components/KCodeBlock/KCodeBlock.vue
Chalks: Mostly because I felt like it made the comment less clear when this was part of the condition. Partly because this was the pattern used in a few other files and I thought it was reasonable to keep it consistent.

I don't have a strong opinion though, happy to change it.

src/components/KModalFullscreen/KModalFullscreen.vue
Chalks: yep, fixed. this was cruft that I forgot to remove. Same with the sandbox router change which is no longer here.

adamdehaven (approved):

PR #2734, opened 2025-05-06
feat(ktooltip): add disabled prop [KHCP-14941]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-14941

Add `disabled` prop in KTooltip

### DIFF
--- a/src/components/KTooltip/KTooltip.vue
+++ b/src/components/KTooltip/KTooltip.vue
@@ -53,11 +53,12 @@ const {
   tooltipId = '',
   zIndex = 9999,
   kpopAttributes = {},
+  disabled = false,
 } = defineProps<TooltipProps>()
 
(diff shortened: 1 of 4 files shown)


src/components/KTooltip/KTooltip.vue
```diff
@@ -53,11 +53,12 @@ const {
   tooltipId = '',
   zIndex = 9999,
   kpopAttributes = {},
+  disabled = false,
```
Justineo, line 56:
```suggestion
  disabled,
```

Our current convention is to omit `false` default value for boolean props as Vue will automatically coerce `undefined` to `false`.

portikM: Fixed

Justineo (approved):
LGTM

arashsheyda (approved):

PR #2743, opened 2025-05-08
fix(kstepper): ts improvements for kstepper

Justineo:
# Summary

[KM-1252](https://konghq.atlassian.net/browse/KM-1252)

- [ ] ~~Preview PR~~
  - [ ] ~~CI passed~~

*Not used in Konnect/KM.*

[KM-1252]: https://konghq.atlassian.net/browse/KM-1252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KStepper/KStep.vue
+++ b/src/components/KStepper/KStep.vue
@@ -40,35 +40,23 @@
 </template>
 
 <script lang="ts" setup>
-import type { PropType } from 'vue'
 import { computed } from 'vue'
 import useUtilities from '@/composables/useUtilities'
-import type { StepperState } from '@/types'
-import { StepperStateArray } from '@/types'
+import type { StepProps } from '@/types'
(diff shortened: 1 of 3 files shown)


src/components/KStepper/KStep.vue
```diff
@@ -40,35 +40,23 @@
...
-  },
-  maxLabelWidth: {
-    type: String,
-    default: '170',
-  },
-})
+const {
+  label,
+  state = 'default',
+  maxLabelWidth,
```
adamdehaven, line 54: issue: it looks like the default value has been removed?

Justineo: Since this is an internal component and always receives its value from `<KStepper>`, the default value isn’t needed. I removed it to keep a single source of truth.

adamdehaven (approved):

PR #2746, opened 2025-05-12
fix(ktable): ts improvements for ktable*

Justineo:
# Summary

[KM-1253](https://konghq.atlassian.net/browse/KM-1253)
[KM-1262](https://konghq.atlassian.net/browse/KM-1262)

- [ ] Preivew PR
  - [ ] CI passed

### DIFF
--- a/src/components/KTable/KTable.vue
+++ b/src/components/KTable/KTable.vue
@@ -544,9 +544,9 @@ const hasColumnVisibilityMenu = computed((): boolean => {
 // columns whose visibility can be toggled
 const visibilityColumns = computed((): TableHeader[] => tableHeaders.value.filter((header: TableHeader) => header.hidable))
 // visibility preferences from the host app (initialized by app)
-const visibilityPreferences = computed((): Record<string, boolean> => hasColumnVisibilityMenu.value ? props.tablePreferences.columnVisibility || {} : {})
+const visibilityPreferences = computed((): Partial<Record<string, boolean>> => hasColumnVisibilityMenu.value ? props.tablePreferences.columnVisibility || {} : {})
 // current column visibility state
-const columnVisibility = ref<Record<string, boolean>>(hasColumnVisibilityMenu.value ? props.tablePreferences.columnVisibility || {} : {})
+const columnVisibility = ref<Record<string, boolean>>(hasColumnVisibilityMenu.value ? props.tablePreferences.columnVisibility as Record<string, boolean> || {} : {})
 const total = ref(0)
 const isScrolled = ref(false)
 const page = ref(1)
@@ -680,7 +680,7 @@ const tdlisteners = computed((): any => {
   }
 })
 
-const columnWidths = ref<Record<string, number>>(props.resizeColumns ? props.tablePreferences.columnWidths || {} : {})
+const columnWidths = ref<Record<string, number>>(props.resizeColumns ? props.tablePreferences.columnWidths as Record<string, number> || {} : {})
 const columnStyles = computed(() => {
   const styles: Record<string, any> = {}
   for (const colKey in columnWidths.value) {
@@ -1133,7 +1133,7 @@ watch([query, page, pageSize], async (newData, oldData) => {
 // because hasColumnVisibilityMenu also accounts for error/loading/empty state, we need to watch it
 watch(hasColumnVisibilityMenu, (newVal) => {
   if (newVal) {
-    columnVisibility.value = props.tablePreferences.columnVisibility || {}
+    columnVisibility.value = props.tablePreferences.columnVisibility as Record<string, boolean> || {}
   }
 }, { immediate: true })
 
--- a/src/components/KTableData/KTableData.vue
+++ b/src/components/KTableData/KTableData.vue
@@ -29,9 +29,11 @@
     :row-expandable="rowExpandable"
     :row-expanded="rowExpanded"
     :row-hover="rowHover"
-    :row-key="rowKey"
+    :row-key="rowKeyProp"
     :row-link="rowLink"
     :table-preferences="tablePreferences"
+    :tooltip-target="tooltipTarget"
+    v-bind="restProps"
     @empty-state-action-click="emit('empty-state-action-click')"
     @error-action-click="emit('error-action-click')"
     @get-next-offset="getNextOffsetHandler"
@@ -39,13 +41,13 @@
     @page-change="pageChangeHandler"
     @page-size-change="pageSizeChangeHandler"
     @row-actions-toggle="($event) => $emit('row-actions-toggle', $event)"
-    @row-select="($event: Record<string, any>[]) => emit('row-select', $event)"
+    @row-select="($event) => emit('row-select', $event)"
     @sort="sortHandler"
-    @update:row-expanded="($event: RowExpandPayload) => emit('update:row-expanded', $event)"
+    @update:row-expanded="($event) => emit('update:row-expanded', $event)"
     @update:table-preferences="tablePreferencesUpdateHandler"
   >
     <template
-      v-if="$slots.toolbar && !hideToolbar"
+      v-if="slots.toolbar && !hideToolbar"
       #toolbar
     >
       <slot
@@ -55,41 +57,41 @@
     </template>
 
     <template
-      v-if="$slots['bulk-actions']"
-      #bulk-actions="{ selectedRows }"
+      v-if="slots['bulk-actions']"
+      #bulk-actions="slotProps"
     >
       <slot
         name="bulk-actions"
-        :selected-rows="selectedRows"
+        v-bind="slotProps"
       />
     </template>
 
     <template
-      v-if="$slots['bulk-action-items']"
-      #bulk-action-items="{ selectedRows }"
+      v-if="slots['bulk-action-items']"
+      #bulk-action-items="slotProps "
     >
       <slot
         name="bulk-action-items"
-        :selected-rows="selectedRows"
+        v-bind="slotProps"
       />
     </template>
 
     <template
-      v-if="$slots['error-state']"
+      v-if="slots['error-state']"
       #error-state
     >
       <slot name="error-state" />
     </template>
 
     <template
-      v-if="$slots['empty-state']"
+      v-if="slots['empty-state']"
       #empty-state
     >
       <slot name="empty-state" />
     </template>
 
     <template
-      v-if="$slots['empty-state-action-icon']"
+      v-if="slots['empty-state-action-icon']"
       #empty-state-action-icon
     >
       <slot name="empty-state-action-icon" />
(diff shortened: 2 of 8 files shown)


src/components/KTableView/ColumnVisibilityMenu.vue
```diff
@@ -114,28 +113,14 @@ import KInput from '@/components/KInput/KInput.vue'
...
-  },
-  disabled: {
-    type: Boolean,
-    default: false,
-  },
-})
+const {
+  columns,
+  tableId,
+  visibilityPreferences,
```
adamdehaven, line 119: it looks like we lost the default here which could cause an error to be thrown when accessing properties below that aren't optionally chained?

src/components/KTableData/KTableData.vue
```diff
@@ -172,91 +170,90 @@ import type {
...
-  fetcherCacheKey: '',
-  cacheIdentifier: '',
-  searchInput: '',
-  initialFetcherParams: () => ({}),
-  clientSort: false,
-  sortable: true,
-})
+type ColumnKey = TableColumnKey<H>
+type ColumnVisibility = TableColumnVisibility<H>
+type ColumnWidths = TableColumnWidths<H>
```
adamdehaven, line 186: nit: can these be moved and imported from the table types file?

src/components/KTableData/KTableData.vue
```diff
@@ -98,65 +100,61 @@
...
-        :nested-headers="nestedHeaders"
-        :row="row"
+        v-bind="slotProps"
       />
     </template>
   </KTableView>
 </template>
 
-<script setup lang="ts">
+<script setup lang="ts" generic="const H extends TableDataHeader = TableDataHeader, D extends readonly Record<string, any>[] = readonly Record<string, any>[], O extends string | number = string | number">
```
adamdehaven, line 155: nit: could we name the generics a bit more specifically so it's easier to read down below?

src/components/KTableData/KTableData.vue
```diff
@@ -172,91 +170,90 @@ import type {
...
-const filterQuery = ref<string>(props.searchInput ?? '')
-const sortColumnKey = ref<string>('')
+const pageSize = ref<number>(getInitialPageSize(tablePreferences, paginationAttributes))
+const filterQuery = ref<string>(searchInput ?? '')
+const sortColumnKey = ref('') as Ref<ColumnKey>
 const sortColumnOrder = ref<SortColumnOrder>('desc')
-const offset: Ref<string | null> = ref(null)
-const offsets: Ref<Array<any>> = ref([])
+const offset = ref(null) as Ref<O | null>
+const offsets = ref([]) as Ref<(O | null)[]>
```
adamdehaven, line 256: The types here changed significantly - mind dropping the explanation for the change here?

src/components/KTableData/KTableData.vue
```diff
@@ -275,39 +272,43 @@ const tablePaginationAttributes = computed((): TablePaginationAttributes => ({
...
 
-function isTableColumnTooltipSlotName(slot: string): slot is TableColumnTooltipSlotName {
+function isTableColumnTooltipSlotName(slot: string): slot is TableColumnTooltipSlotName<ColumnKey> {
   return slot.startsWith('tooltip-')
 }
 
-const getHeaderSlots = computed((): TableColumnSlotName[] => {
+function isTableColumnKey(slot: string): slot is ColumnKey {
+  return tableHeaders.value.some((header) => header.key === slot)
+}
```
adamdehaven, line 288: nit: can we stick with a single syntax to declare functions?

src/components/KTableData/KTableData.vue
```diff
@@ -275,39 +272,43 @@ const tablePaginationAttributes = computed((): TablePaginationAttributes => ({
...
+const getHeaderTooltipSlots = computed(() => {
   if (!slots) {
     return []
   }
 
   return Object.keys(slots).filter(isTableColumnTooltipSlotName)
 })
 
-const getCellSlots = computed((): string[] => {
+const getCellSlots = computed(() => {
```
adamdehaven, line 306: If we don't define a more explicit return type here for these computed functions I worry that over time with future edits they may change unexpectedly. Can we add return types?

src/components/KTableData/KTableData.vue
```diff
@@ -370,25 +371,25 @@ const initData = () => {
   hasInitialized.value = true
 }
 
-const previousOffset = computed((): string | null => offsets.value[page.value - 1])
-const nextOffset = ref<string | null>(null)
+const previousOffset = computed(() => offsets.value[page.value - 1])
```
adamdehaven, line 374: nit: add return type

src/components/KTableData/KTableData.vue
```diff
@@ -524,16 +525,16 @@ const showPagination = computed((): boolean => {
...
 
-watch(fetcherResponse, (res: Record<string, any>) => {
+watch(fetcherResponse, (res) => {
   if (!res?.data) {
     return
   }
 
-  tableData.value = res.data as Record<string, any>[]
-  total.value = props.paginationAttributes?.totalCount || res.total || res.data?.length || 0
+  tableData.value = [...res.data]
```
adamdehaven, line 533: Much nicer

src/components/KTableView/KTableView.vue
```diff
@@ -404,17 +404,17 @@
...
+        @page-change="emit('page-change', $event)"
         @page-size-change="onPaginationPageSizeChange"
       />
     </div>
   </div>
 </template>
 
-<script setup lang="ts">
-import { ref, watch, computed, useAttrs, useSlots, nextTick, useId, useTemplateRef } from 'vue'
+<script setup lang="ts" generic="const H extends TableViewHeader = TableViewHeader, D extends readonly Record<string, any>[] = readonly Record<string, any>[]">
```
adamdehaven, line 416: Same comment on generic naming

src/components/KTableView/KTableView.vue
```diff
@@ -542,36 +536,36 @@ const hasColumnVisibilityMenu = computed((): boolean => {
...
-const columnVisibility = ref<Record<string, boolean>>(hasColumnVisibilityMenu.value ? props.tablePreferences.columnVisibility || {} : {})
+const columnVisibility = ref(hasColumnVisibilityMenu.value ? tablePreferences.columnVisibility || {} : {}) as Ref<ColumnVisibility>
 
 const tableWrapperHeight = ref<string>('100%')
 const isScrollableVertically = ref<boolean>(false)
 const isScrolledVertically = ref<boolean>(false)
 const isScrolledHorizontally = ref<boolean>(false)
 const isScrollableRight = ref<boolean>(false)
-const sortColumnKey = ref('')
+const sortColumnKey = ref('') as Ref<ColumnKey | ''>
```
adamdehaven, line 554: Why do we need to cast rather than type? (Some question for a few refs in the PR)

src/components/KTableView/KTableView.vue
```diff
@@ -612,35 +606,26 @@ const getColumnSlotName = (columnKey: string): TableColumnSlotName => {
...
+const pluckListeners = (prefix: 'onRow:' | 'onCell:', listenerProps: Record<string, unknown>): any => {
+  return (entity: unknown, type: 'row' | 'cell') => {
     const onRE = /^on[^a-z]/
     const listeners = {} as any
 
-    for (const property in attrs) {
-      if (onRE.test(property) && !!attrs[property]) {
-        listeners[property] = attrs[property]
+    for (const property in listenerProps) {
+      if (onRE.test(property) && !!listenerProps[property]) {
```
adamdehaven, line 627: nit: can you rename `onRe` to something more readable (I realize it's out of scope for PR)

src/components/KTableView/KTableView.vue
```diff
@@ -769,7 +756,7 @@ const onHeaderClick = (column: TableViewHeader) => {
       prevKey: sortColumnKey.value,
       sortColumnKey: column.key,
       sortColumnOrder: newSortColumnOrder,
-    })
+    } as TableSortPayload<ColumnKey>)
```
adamdehaven, line 759: Can you use `satisfies` instead of casting?

src/components/KTableView/KTableView.vue
Justineo: It's probably copied from Vue's old implementation: https://github.com/vuejs/core/blob/5a7a1b8293822219283d6e267496bec02234b0bc/packages/shared/src/index.ts#L40-L41 😅

src/components/KTable/KTable.vue
```diff
@@ -544,9 +544,9 @@ const hasColumnVisibilityMenu = computed((): boolean => {
 // columns whose visibility can be toggled
 const visibilityColumns = computed((): TableHeader[] => tableHeaders.value.filter((header: TableHeader) => header.hidable))
 // visibility preferences from the host app (initialized by app)
-const visibilityPreferences = computed((): Record<string, boolean> => hasColumnVisibilityMenu.value ? props.tablePreferences.columnVisibility || {} : {})
+const visibilityPreferences = computed((): Partial<Record<string, boolean>> => hasColumnVisibilityMenu.value ? props.tablePreferences.columnVisibility || {} : {})
```
adamdehaven, line 547: I'm not sure I understand `Partial<Record<string, boolean>>` here

Isn't `Record<string, boolean>` equivalent?

src/components/KTableData/KTableData.vue
```diff
@@ -172,91 +170,90 @@ import type {
...
 
-const slots = useSlots()
+const {
+  resizeColumns,
+  tablePreferences = {},
+  rowHover = true,
+  rowAttrs = () => ({}),
+  rowLink = () => ({} as RowLink),
+  rowBulkActionEnabled = () => true,
+  rowKey: rowKeyProp,
```
adamdehaven, line 195: I'm probably missing something obvious, but where does `rowKeyProp` come from here?

src/components/KTableData/KTableData.vue
```diff
@@ -172,91 +170,90 @@ import type {
...
+  fetcherCacheKey = '',
+  cacheIdentifier = '',
+  searchInput = '',
+  initialFetcherParams = {},
+  clientSort,
+  sortable = true,
+  fetcher,
+  sortHandlerFunction,
+  tooltipTarget,
+  ...restProps
```
adamdehaven, line 228: issue(?): This seems a bit opaque - how to know what other props are being applied here and in the `v-bind`? 

It may be a bit confusing to know what is being swallowed up here; if other props were added later in the TS file down the line they would automatically be applied via the `v-bind` in the template if the developer isn't careful to explicitly extract it.

src/components/KTableData/KTableData.vue
```diff
@@ -172,91 +170,90 @@ import type {
...
-  (e: 'row-select', data: Record<string, any>[]): void
-  (e: 'update:row-expanded', data: RowExpandPayload): void
-  (e: 'row-actions-toggle', data: RowActionsTogglePayload): void
-}>()
-
 const tableId = useId()
 
-const tableData = ref<Record<string, any>[]>([])
-const tableHeaders = computed((): TableDataHeader[] => props.sortable ? props.headers : props.headers.map((header) => ({ ...header, sortable: false })))
+const tableData = ref([]) as Ref<Data[number][]>
```
adamdehaven, line 239: nit: it might be useful to leave comments in the PR as to why we are casting so that it's clear for future devs?

src/components/KTableData/KTableData.vue
```diff
@@ -275,39 +272,43 @@ const tablePaginationAttributes = computed((): TablePaginationAttributes => ({
   currentPage: page.value,
   offsetPreviousButtonDisabled: !previousOffset.value,
   offsetNextButtonDisabled: !nextOffset.value || !hasNextPage.value,
-  ...props.paginationAttributes,
+  ...paginationAttributes,
 }))
 
-function isTableColumnSlotName(slot: string): slot is TableColumnSlotName {
+const isTableColumnSlotName = (slot: string): slot is TableColumnSlotName<ColumnKey> => {
```
adamdehaven, line 278: nit: more comments in the code as to what different constants are for may be more valuable now that the types are more complex

src/components/KTableData/KTableData.vue
Justineo: There seems to be no need to rename here. Changed.

src/components/KTableData/KTableData.vue
Justineo: These are event handlers for `@cell:<event>` / `@row:<event>` events. We need to declare them in props otherwise we cannot get the listeners as we already declared them in `defineEmits`. I'll add a comment here.

adamdehaven: But wouldn't future props need to be added/extracted carefully?

src/components/KTableData/KTableData.vue
Justineo: Added.

src/components/KTableData/KTableData.vue
Justineo: Added.

src/components/KTableData/KTableData.vue
Justineo: Updated.

adamdehaven: I'm not sure I see what was changed here?

PR #2739, opened 2025-05-08
fix(kradio): ts improvements for kradio

Justineo:
# Summary

[KM-1230](https://konghq.atlassian.net/browse/KM-1230)

- [x] Preview PR: https://github.com/Kong/konnect-ui-apps/pull/6713
  - [x] CI passed

[KM-1230]: https://konghq.atlassian.net/browse/KM-1230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2749, opened 2025-05-13
fix(k-radio): lint error

vaibhavrajsingh2001:
# Description

#2747 was merged without rebasing with the changes in #2739, which removed the `const props = defineProps` from the file and caused lint errors


johncowen:
Thanks @vaibhavrajsingh2001 I was just about to put a PR up myself 👍 

We are hitting some problems https://github.com/kumahq/kuma-gui/pull/3882 and I'm hoping this is the root of it

adamdehaven (approved):

PR #2752, opened 2025-05-14
chore(types): type check all files including tests

johncowen:
TBD

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "docs:dev": "vitepress dev docs",
     "docs:serve": "vitepress serve docs",
     "docs:preview": "vitepress preview docs --port 8080",
-    "typecheck": "vue-tsc -p './tsconfig.build.json' --noEmit",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "docs:dev": "vitepress dev docs",
     "docs:serve": "vitepress serve docs",
     "docs:preview": "vitepress preview docs --port 8080",
-    "typecheck": "vue-tsc -p './tsconfig.build.json' --noEmit",
+    "typecheck": "vue-tsc --noEmit",
```
adamdehaven, line 46: We shouldn't exclude the `tsconfig.build.json` here, but rather update it to include all types we want to check.

There are some types in the project that likely _should_ be excluded from checking during the build/release process.

johncowen: Yep I'm fine whichever. I guess whichever option means we get typechecking on `.ts` files.

Do you have an example of which files likely should be excluded? (that are a different set to those that are in tsconfig.json)

PR #2761, opened 2025-05-16
fix(ktabledata): prioritize bound pagination attributes

portikM:
# Summary

Bound `paginationAttributes` are getting overridden by the component which is incorrect.


Justineo (approved):
LGTM

PR #2773, opened 2025-05-28
feat: add kslider component [KHCP-16430]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-16430

Create KSlider component

Screenshot 2025-05-28 at 4 27 30 PM

Screenshot 2025-05-28 at 4 27 55 PM

### DIFF
--- a/docs/components/slider.md
+++ b/docs/components/slider.md
@@ -182,7 +182,6 @@ Boolean that controls whether KSlider should display a popover showing the input
 <KSlider
   :show-value="false"
   v-model="vModel8"
-  label="Volume"
 />
 
 
@@ -190,7 +189,6 @@ Boolean that controls whether KSlider should display a popover showing the input
 <KSlider
   :show-value="false"
   v-model="vModel8"
-  label="Volume"
 />
 ```
 


docs/components/slider.md
```diff
@@ -0,0 +1,220 @@
...
+```html
+<KSlider
+  show-marks
+  v-model="sliderValue"
+  :max="50"
+  :step="5"
+/>
+```
+
+### marks
```
adamdehaven, line 115: IMO "marks" is a bit confusing; there's not an inherit relationship to "steps" here

docs/components/slider.md
```diff
@@ -0,0 +1,220 @@
...
+/>
+```
+
+### labelAttributes
+
+Use the `labelAttributes` prop to configure the KLabel's [props](/components/label) when using the `label` prop.
+
+## Events
+
+### change
```
adamdehaven, line 190: question: why do we need the `change` event when there is already the default `update:model-value` event fired?

src/types/slider.ts
```diff
@@ -0,0 +1,71 @@
...
+export interface SliderEmits {
+  /**
+   * Fired on change, returns the new value of the slider.
+   */
+  change: [value: number]
+
+  /**
+   * Fired on change, returns the new value of the slider.
+   */
+  'update:modelValue': [modelValue: number]
```
adamdehaven, line 68: This event isn't documented on the docs page

adamdehaven:
How is a user to know what value they have selected when there's only a min and max label on the element? If I'm picking a value between 1 and 10, there's nothing in the UI to confirm the value that I selected. e.g. How do I know if I picked 3 or 4? 

The issue is compounded a bit on mobile where the control is much more condensed.

adamdehaven:
image

With this example from the sandbox, there appear to only be 3 valid selections but I can move the slider anywhere from left to right. 

And again, I don't seem to have a way of knowing which is selected

docs/components/slider.md
Justineo: Initially Max named it as `labels` but we both agree it may cause confusion as we already have `label`.

Justineo: They are called mark texts according to [Open UI](https://open-ui.org/components/slider.research.parts/). But we do not have those little ticks on the track (which I believe are the "marks") so marks can be confusing as well…

docs/components/slider.md
portikM: In case we need to hook up some logic to run once new value is selected - IMO slightly better than a watcher + it was @kaiarrowood's explicit request to have that event.

src/types/slider.ts
portikM: `v-model` is documented on the docs page, do we still need to document `update:modelValue` too?

portikM:
> How is a user to know what value they have selected when there's only a min and max label on the element? If I'm picking a value between 1 and 10, there's nothing in the UI to confirm the value that I selected. e.g. How do I know if I picked 3 or 4?
> 
> The issue is compounded a bit on mobile where the control is much more condensed.

input[type="range"] isn't meant to be precise:

> [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input) elements of type range let the user specify a numeric value which must be no less than a given value, and no more than another given value. The precise value, however, is not considered important. This is typically represented using a slider or dial control rather than a text entry box like the [number](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number) input type.
> 
> Because this kind of widget is imprecise, it should only be used if the control's exact value isn't important.

[[MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range)]

I will add a sentence in the docs clarifying that KSlider should only be used in scenarios when imprecise value is acceptable - otherwise `input[type="number"]` is most appropriate.

docs/components/slider.md
adamdehaven: Is there any difference between the emitted native event and change event?

If yes, why?

portikM:
> image
> 
> With this example from the sandbox, there appear to only be 3 valid selections but I can move the slider anywhere from left to right.
> 
> And again, I don't seem to have a way of knowing which is selected

That is so by design. Sometimes we only want to show marks for certain steps, but not for all of them.

docs/components/slider.md
```diff
@@ -0,0 +1,228 @@
...
+```html
+<KSlider
+  v-model="nodesCount"
+  label="How many nodes do you want to run in this cluster?"
+  :min="1"
+/>
+```
+
+:::tip NOTE
+Since KSlider doesn't always convey the exact selected value, it's best suited for use cases where precision isn't critical—such as adjusting volume or rating satisfaction. For scenarios requiring precise input, use [KInput type="number"](/components/input#attribute-binding) instead.
```
kaiarrowood, line 20:
```suggestion
KSlider is best used in scenarios such as specifying volume or rating satisfaction, where the ranges are generally small and not all ticks/marks may be labelled. For scenarios requiring precise input in a large range, use [KInput type="number"](/components/input#attribute-binding) instead.
```

suggestion

docs/components/slider.md
kaiarrowood: You could call it `markLabels` or `tickLabels`. `tick` may be a little more well known than `mark`?

adamdehaven: > I suggest we leave it as is as all other options feel same as this one - "a little confusing / unintuitive" 😅

I'm not sure I agree. 

This needs to imply "available option labels" 

I think the current implementation is a bit confusing, especially considering some of the issues in identifying the selected value I commented separately.

portikM:
> https://deploy-preview-2773--kongponents-sandbox.netlify.app

Do you think adding an option to show a tooltip over the thumb with selected value would solve this issue?

adamdehaven:
> I think we can solve that problem in next iteration

TBH I don't understand not solving this before releasing the component.

docs/components/slider.md
```diff
@@ -0,0 +1,228 @@
...
+
+Minimum value of the range. Default value is `0`.
+
+### max
+
+Minimum value of the range. Default value is `10`.
+
+### step
+
+Step value. Default value is `1`.
```
adamdehaven, line 77: docs: shouldn't we document the type of these props?

kaiarrowood (commented):
Code and functional looks good to me 🚀

sandbox/pages/SandboxSlider.vue
```diff
@@ -0,0 +1,98 @@
...
+  {
+    label: 'Very likely',
+    value: 0,
+  },
+  {
+    label: 'Extremely likely',
+    value: 5,
+  },
+  {
+    label: 'Epically likely',
```
kaiarrowood, line 89: Epic!

docs/components/slider.md
portikM: So are we good to proceed with `marks` and `showMarks` or do we want to change these prop names?

kaiarrowood (approved):
LGTM! 🚀

PR #2780, opened 2025-06-04
fix(kselect): keyboard operation issues

jw-foss:
* Fixes the issue that `tab` focus on the select trigger automatically opens the select dropdown.
* Fixes the issue when `selection` occurred, the dropdown gets closed, the focus won't restore back to the trigger.

# Summary

### Root cause

As the screencast shows, when we tried to navigate via keyboards, the `Tab` key comes into play, whenever we `Tab` onto a select trigger, the dropdown automatically opens, and it does not attach focus state to the select items. According to the W3C standard https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/ the `Tab` action should only set focus onto the trigger element (input in this case), with `open keys` such as `space bar`, `enter`, `arrow down` and `arrow up` (`Home` and `End` was not included in this PR), to open the popup, and sets focus on the selected item if any or first item in the selection list.

https://github.com/user-attachments/assets/2b7196d4-a670-4995-8762-10eb759d39f7

### Patch made

After this patch, the behavior should be aligned with the standard to allow user to set focus onto the trigger element, and they can use the `open keys` here to open the popup(dropdown), see the screenshot down below :point_down:

https://github.com/user-attachments/assets/0dca96e8-1aa9-4eba-b1aa-2c6bc2d9d89a

/cc @portikM 
[KHCP-15521]

[KHCP-15521]: https://konghq.atlassian.net/browse/KHCP-15521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KInput/KInput.vue
+++ b/src/components/KInput/KInput.vue
@@ -140,7 +140,7 @@ const inputId = computed((): string => attrs.id ? String(attrs.id) : defaultId)
 const helpTextId = useId()
 const strippedLabel = computed((): string => stripRequiredLabel(label, isRequired.value))
 const hasLabelTooltip = computed((): boolean => !!(labelAttributes?.info || slots['label-tooltip']))
-const input$ = useTemplateRef('inputRef')
+const inputRef = useTemplateRef('inputRef')
 
 // we need this so we can create a watcher for programmatic changes to the modelValue
 const value = computed({
@@ -248,11 +248,11 @@ const getValue = (): string | number => {
 }
 
 const focus = () => {
-  input$.value?.focus?.()
+  inputRef.value?.focus?.()
 }
 
 const blur = () => {
-  input$.value?.blur?.()
+  inputRef.value?.blur?.()
 }
 
 defineExpose({
--- a/src/components/KSelect/KSelectItems.vue
+++ b/src/components/KSelect/KSelectItems.vue
@@ -117,13 +117,7 @@ const moveItemFocus = (direction: 'down' | 'up' | 'current'): void => {
   }
 }
 
-const onKeyPress = (e : KeyboardEvent) => {
-  const { target, key } = e
-  if (key === 'Enter') {
-    e.stopPropagation()
-    e.stopImmediatePropagation()
-  }
-
+const onKeyPress = ({ target, key } : KeyboardEvent) => {
   if (key === 'ArrowDown' || key === 'ArrowUp') {
     // all selectable items
     const selectableItems = itemsContainerRef.value?.querySelectorAll<HTMLButtonElement>(SELECTABLE_ITEM_SELECTOR)


src/components/KInput/KInput.vue
```diff
@@ -139,6 +140,7 @@ const inputId = computed((): string => attrs.id ? String(attrs.id) : defaultId)
 const helpTextId = useId()
 const strippedLabel = computed((): string => stripRequiredLabel(label, isRequired.value))
 const hasLabelTooltip = computed((): boolean => !!(labelAttributes?.info || slots['label-tooltip']))
+const input$ = useTemplateRef('inputRef')
```
Justineo: I think we don't have this naming conventions in the project and it would be better to not include this here.

src/components/KSelect/KSelect.vue
```diff
@@ -448,20 +448,32 @@ const clearSelection = (): void => {
...
+const NAVIGATION_KEYS = ['ArrowDown', 'ArrowUp']
+const OPEN_KEYS = [' ', 'Enter', ...NAVIGATION_KEYS]
 
-const triggerFocus = (evt: any, isToggled: Ref<boolean>): void => {
+const triggerFocus = (evt: KeyboardEvent, isToggled: Ref<boolean>): void => {
   // Ignore `esc` key
-  if (evt.keyCode === 27) {
+  if (evt.key === 'Escape' && isToggled.value) {
     isToggled.value = false
+    popperRef.value?.hidePopover?.()
```
Justineo, line 458: Shouldn't setting `isToggled.value = false` already closing the popper?

src/components/KSelect/KSelectItems.vue
```diff
@@ -110,7 +117,13 @@ const moveItemFocus = (direction: 'down' | 'up'): void => {
   }
 }
 
-const onKeyPress = ({ target, key } : KeyboardEvent) => {
+const onKeyPress = (e : KeyboardEvent) => {
+  const { target, key } = e
+  if (key === 'Enter') {
+    e.stopPropagation()
+    e.stopImmediatePropagation()
+  }
```
Justineo: What would happen if we didn’t stop here? Might be worth adding a comment.

src/components/KSelect/KSelect.vue
jw-foss: Actually the `isToggled` does not control the state of `popper` this gets tricky when someone wants to mutate the state of the `popper` it's like we are controlling two states at the same time.

`KToggle` does not provide any control over `KPopper`, because the initial state of `KToggle` and `KPopper` were both `false`, it is our liability to sync the state.

src/components/KSelect/KSelectItems.vue
jw-foss: This should be dismissed, I only added this because previously we are using `keyup` on the trigger element `KInput`, and `keydown` on the `SelectItem` component. Removed.

src/components/KInput/KInput.vue
jw-foss: Updated

jw-foss:
> I noticed that if I clicked outside the select menu to close it, the focus still returns to the select itself, is this intentional?


Looks like a bug, converting this to a draft for fixing minor issues.

src/components/KInput/KInput.vue
```diff
@@ -245,6 +247,19 @@ const getValue = (): string | number => {
...
+}
+
+const blur = () => {
+  inputRef.value?.blur?.()
+}
+
+defineExpose({
+  focus,
+  blur,
+})
```
adamdehaven, line 261: suggestion: can we just expose the input ref rather than focus and blur methods?

src/components/KSelect/KSelect.vue
```diff
@@ -448,20 +448,32 @@ const clearSelection = (): void => {
...
-  const inputElem = selectWrapperRef.value?.children[0] as HTMLInputElement
-  if (!isToggled.value && inputElem) { // simulate click to trigger dropdown open
-    inputElem.click()
+  // per https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/
+  if (!isToggled.value && OPEN_KEYS.includes(evt.key)) { // simulate click to trigger dropdown open
+    if (NAVIGATION_KEYS.includes(evt.key)) {
+      evt.preventDefault()
+    }
+    // TODO: when it is opened by 'Home', 'End', 'ArrowUp', 'ArrowDown' keys, it should handle
+    // setting focus to different item
```
adamdehaven, line 468: I'm not sure I understand this comment.

src/components/KSelect/KSelect.vue
```diff
@@ -448,20 +448,32 @@ const clearSelection = (): void => {
...
+    popperRef.value?.showPopover?.()
+    nextTick(() => selectItemsRef.value?.setFocus('current'))
   }
 
-  if ((evt.key === 'ArrowDown' || evt.key === 'ArrowUp') && isToggled.value) {
+  // TODO: we need to remove this part since the popover should trap the focus
```
adamdehaven, line 473: How do you then escape the popover?

src/components/KSelect/KSelectItems.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-  <div ref="itemsContainer">
+  <div
+    ref="itemsContainer"
+  >
```
adamdehaven, line 4: nit: this formatting change isn't necessary

src/components/KSelect/KSelectItems.vue
```diff
@@ -89,7 +91,7 @@ const getGroupItems = (group: string) => items?.filter(item => item.group === gr
 
 const itemsContainerRef = useTemplateRef('itemsContainer')
 
-const moveItemFocus = (direction: 'down' | 'up'): void => {
+const moveItemFocus = (direction: 'down' | 'up' | 'current'): void => {
```
adamdehaven, line 94: `current` being a value for a `direction` param is a bit confusing?

PR #2783, opened 2025-06-09
fix(ktextarea): ts improvements for ktextarea

Justineo:
# Summary

[KM-1338](https://konghq.atlassian.net/browse/KM-1338)

### DIFF
--- a/src/components/KTextArea/KTextArea.vue
+++ b/src/components/KTextArea/KTextArea.vue
@@ -70,104 +70,54 @@
 </template>
 
 <script lang="ts">
-import { ref, computed, watch, useAttrs, useSlots, useId } from 'vue'
-import useUtilities from '@/composables/useUtilities'
-import KLabel from '@/components/KLabel/KLabel.vue'
-import type { TextAreaLimitExceed } from '@/types'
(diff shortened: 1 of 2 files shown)


src/components/KTextArea/KTextArea.vue
```diff
@@ -70,123 +70,73 @@
...
-   * @deprecated in favor of `error` prop
-   */
-  hasError: {
-    type: Boolean,
-    default: false,
-    validator: (value: boolean): boolean => {
-      if (value) {
-        console.warn('KTextArea: the `hasError` prop is deprecated in favor of the `error` prop. See the migration guide for more details: https://kongponents.konghq.com/guide/migrating-to-version-9.html#ktextarea')
-      }
+import { ref, computed, watch, useAttrs, useId } from 'vue'
```
adamdehaven, line 82: Do we still actually need the two script sections now?

Justineo: I put some feature detection code in a separate `<script>` block so that they are evaluated at top level instead of repeatedly evaluated in every `setup` call.

PR #2784, opened 2025-06-09
feat(*): add number support for css lengths

Justineo:
# Summary

[KM-1361](https://konghq.atlassian.net/browse/KM-1361)

This PR:

- Make all props accepting CSS length as values accept number
- Rename `getSizeFromString` to `normalizeSize` and move it from `composables` to `utilities/css`

Preview PR: https://github.com/Kong/konnect-ui-apps/pull/7061

[KM-1361]: https://konghq.atlassian.net/browse/KM-1361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KBadge/KBadge.vue
+++ b/src/components/KBadge/KBadge.vue
@@ -39,7 +39,7 @@ const {
   appearance = 'info',
   tooltip = '',
   truncationTooltip = false,
-  maxWidth: maxWidthProp = '200px',
+  maxWidth = '200px',
   iconBefore = true,
   tooltipAttributes = {},
 } = defineProps<BadgeProps>()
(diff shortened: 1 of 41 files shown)


src/components/KBadge/KBadge.vue
```diff
@@ -31,23 +31,21 @@
...
+import { normalizeSize } from '@/utilities/css'
 
 import type { BadgeProps, BadgeSlots } from '@/types'
 
 const {
   appearance = 'info',
   tooltip = '',
   truncationTooltip = false,
-  maxWidth: maxWidthProp = '200',
+  maxWidth: maxWidthProp = '200px',
```
adamdehaven: (applies to whole PR)
I personally don't like renaming the props here -- can we instead rename the computed variable that is only used internally to something like `computedMaxWidth` instead? 

Renaming the props may cause confusion throughout the template rather than renaming a computed property or something only used internal.

Justineo: Good point. I'll open a PR to update all similar cases.

adamdehaven: For renaming computed, if they're just internal you could also just do `_maxWidth` (lead with underscore) to show it's internal 🤷🏼 (less verbose)

Justineo: Sure. We can just use this convention.

PR #2791, opened 2025-06-13
fix(kcodeblock): fix navigation among query matches

Justineo:
# Summary

[KM-1388](https://konghq.atlassian.net/browse/KM-1388)


adamdehaven (approved):
lgtm

PR #2794, opened 2025-06-17
fix(kinput): expose input element [KHCP-16490]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-16490

Expose KInput's input element through define expose

### DIFF
--- a/docs/components/input.md
+++ b/docs/components/input.md
@@ -344,7 +344,7 @@ KInput allows you to listen to DOM events:
 
 ### input
 
-KInput exposes HTML input element through `input` property for ease of performing manipulations.
+KInput exposes its internal HTML `input` element `ref` via an exposed `input` property to allow access to the underlying element.
 
 <div class="input-container">
   <KInput 
@@ -363,7 +363,7 @@ KInput exposes HTML input element through `input` property for ease of performin
 ```vue
 <template>
   <KInput 
-    ref="kInput"
+    ref="myInput"
     placeholder="Focusable input"
   />
   <KButton
@@ -375,9 +375,9 @@ KInput exposes HTML input element through `input` property for ease of performin
 </template>
 
 <script setup lang="ts">
-const kInputRef = useTemplateRef('kInput')
+const myInputRef = useTemplateRef('myInput')
 const setFocus = () => {
-  kInputRef.value?.input?.focus()
+  myInputRef.value?.input?.focus()
 }
 </script>
 ```


src/components/KInput/KInput.vue
```diff
@@ -33,11 +33,12 @@
 
       <input
         :id="inputId"
+        v-bind="modifiedAttrs"
```
adamdehaven: question: why move this up in the chain? It will then potentially be overridden by any attribute defined below it

src/components/KInput/KInput.vue
```diff
@@ -273,12 +280,10 @@ onMounted(async () => {
...
 })
-</script>
 
-<script lang="ts">
-export default {
-  inheritAttrs: false,
-}
+defineExpose({
+  input: inputElementRef,
+})
```
Chalks, line 286: **suggestion, don't know if this is a blocker or not**

update [documentation](https://github.com/Kong/kongponents/blob/main/docs/components/input.md) for this? maybe a new section titled "Input" or "Elements"?

src/components/KInput/KInput.vue
portikM: Oh good catch. I think linter did it automatically for me. Moved back to it's place.

Chalks (dismissed):

src/components/KInput/KInput.vue
adamdehaven: Since we utilize "Vue" terminology in the docs, could we just have a section `Expose` (or similar)

portikM: Yeah I thought we'd benefit from documenting this but was unsure where to put it...

portikM: Ok added Expose section in the component docs. Please feel free to suggest improvements to the sentence.

src/components/KInput/KInput.vue
```diff
@@ -125,6 +130,8 @@ watch(() => labelAttributes.help, (help) => {
 const emit = defineEmits<InputEmits>()
 defineSlots<InputSlots>()
 
+const inputElementRef = useTemplateRef('inputElement')
```
Chalks, line 133: **question**

if you do this instead

```suggestion
const inputElementRef = useTemplateRef<HTMLInputElement>('inputElement')
```

do you still have to cast it like you do in your examples `(focusableKInputRef.value?.input as HTMLInputElement).focus()`?

portikM: That would be a mistake, `focusableKInputRef` is not of type `HTMLInputElement`, it's property `input` is.

Chalks: I think we might be misunderstanding one another, I'm not suggesting that the `focusableKInputRef` should be typed differently. Basically I noticed that in your examples you cast `input` to `HTMLInputElement`. I think that is unnecessary if you type it before it's ever exposed.

That is to say, if in `<KInput>`:

```
<input ref="inputElement" />

...

const inputElementRef = useTemplateRef<HTMLInputElement>('inputElement')

...

defineExpose({
  input: inputElementRef,
})
```

then anything that uses `<KInput>` can do:

```
<KInput ref="focusableKInput" />

...

const focusableKInputRef = useTemplateRef('focusableKInput')

const setFocus = () => {
  focusableKInputRef.value?.input.focus()
}
```

src/components/KInput/KInput.vue
Chalks: lgtm

docs/components/input.md
```diff
@@ -340,9 +340,51 @@ KInput allows you to listen to DOM events:
 </KComponent>
 ```
 
+## Expose
+
+### input
+
+KInput exposes HTML input element through `input` property for ease of performing manipulations.
```
adamdehaven:
```suggestion
KInput exposes its internal HTML `input` element `ref` via an exposed `input` property to allow access to the underlying element.
```

docs/components/input.md
```diff
@@ -340,9 +340,51 @@ KInput allows you to listen to DOM events:
...
+    @click="setFocus"
+  >
+    Set focus
+  </KButton>
+</div>
+
+```vue
+<template>
+  <KInput 
+    ref="kInput"
```
adamdehaven: nit: can we rename the template ref here something other than `kInput` just to avoid some confusion?
```suggestion
    ref="myInputRef"
```

docs/components/input.md
```diff
@@ -340,9 +340,51 @@ KInput allows you to listen to DOM events:
...
+    @click="setFocus"
+  >
+    Set focus
+  </KButton>
+</template>
+
 <script setup lang="ts">
+const kInputRef = useTemplateRef('kInput')
+const setFocus = () => {
+  (kInputRef.value?.input as HTMLInputElement).focus()
```
adamdehaven: You'd need to name these as well

src/components/KInput/KInput.vue
portikM: Ohh sorry I misunderstood you. According to Vue docs `useTemplateRef` infers the type of the element so in theory, specifying the type like so: `useTemplateRef<HTMLInputElement>(...` shouldn't be necessary. I checked and realized the only reason I had to cast the type was because `input` could be `null` so I fixed that with a question mark and now we don't need to cast type there anymore, you were right. Pushed my changes up.

docs/components/input.md
portikM: Done

docs/components/input.md
portikM: Done

Chalks (approved):

PR #2798, opened 2025-06-19
fix(kdatetimepicker): add readonly prop [KHCP-16750]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-16750

Add `readonly` prop in KDateTimePicker


arashsheyda:
suggestion: should the cursor not be `pointer` when readonly?

filipgutica (approved):

arashsheyda (approved):
lgtm!

PR #2812, opened 2025-07-01
fix(ktreelist): ts improvements for ktreelist

Justineo:
# Summary

[KM-1340](https://konghq.atlassian.net/browse/KM-1340)

- [x] Preview PR
  - [x] CI passed

[KM-1340]: https://konghq.atlassian.net/browse/KM-1340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KTreeList/KTreeDraggable.vue
+++ b/src/components/KTreeList/KTreeDraggable.vue
@@ -94,12 +94,6 @@
 </template>
 
 <script lang="ts">
-import type { PropType } from 'vue'
-import { computed, ref, watch, onMounted } from 'vue'
-import { VueDraggableNext } from 'vue-draggable-next'
-import type { TreeListItem, TreeListChangeEvent, TreeListChildChangeEvent } from '@/types'
-import KTreeItem from '@/components/KTreeList/KTreeItem.vue'
-
 /**
  * Recursive check to get the maximum depth of an object.
  * Documented here: https://stackoverflow.com/a/48505969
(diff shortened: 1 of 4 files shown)


src/components/KTreeList/KTreeDraggable.vue
```diff
@@ -110,61 +104,25 @@ export const getMaximumDepth = ({ children = [] }): number => {
...
+  items,
+  disableDrag,
+  filter,
+  maxDepth,
+  level = 0,
+  parentId,
+  hideIcons,
+  group,
+  collapsible,
+  initialCollapseAll,
```
Justineo, line 122: Removed default values from the private component’s props, as they’re already defined in the public parent. This avoids redundancy and keeps a single source of truth.

src/components/KTreeList/KTreeItem.vue
```diff
@@ -67,64 +67,31 @@
...
-  },
-})
+const {
+  item,
+  disabled,
+  hideIcons,
+  collapsible,
+  initialCollapse,
+  hasChildren,
+  controlsId,
```
Justineo, line 87: Same as `KTreeDraggable`.

src/components/KTreeList/KTreeItem.vue
```diff
@@ -67,64 +67,31 @@
...
 </template>
 
-<script lang="ts">
-import type { PropType } from 'vue'
-import { computed, useSlots } from 'vue'
-import type { TreeListItem } from '@/types'
-
-export const itemsHaveRequiredProps = (items: TreeListItem[]): boolean => {
-  return items.every(i => i.name !== undefined && i.id !== undefined && (!i.children?.length || itemsHaveRequiredProps(i.children)))
-}
```
Justineo, line 77: This was removed because TypeScript already performs this check. Recursively verifying it in every node would be excessive.

src/components/KTreeList/KTreeList.vue
```diff
@@ -75,73 +74,39 @@ const itemsWithinMaximumDepth = (items: TreeListItem[], maxDepth: number): boole
...
-</script>
-
-<script setup lang="ts">
-import { useTemplateRef } from 'vue'
+const {
+  modelValue = null,
+  items = null,
+  disableDrag,
+  ignoreDragSelectors = '',
+  maxDepth = 3,
```
adamdehaven, line 82: we lose the validator on the maxDepth here; can we still enforce it in code? This is particularly important for the nested trees

Justineo: ~~It is validated in `onMounted` as was before. But we should indeed move them to immediate watchers as others.~~
Ah, I know what you mean now, nice catch!

PR #2819, opened 2025-07-07
Fix/khcp 17101 table sort issues

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-17101

### DIFF
--- a/src/components/KTableView/KTableView.vue
+++ b/src/components/KTableView/KTableView.vue
@@ -549,8 +549,8 @@ const isScrollableVertically = ref<boolean>(false)
 const isScrolledVertically = ref<boolean>(false)
 const isScrolledHorizontally = ref<boolean>(false)
 const isScrollableRight = ref<boolean>(false)
-const sortColumnKey = ref('') as Ref<ColumnKey | ''>
-const sortColumnOrder = ref<SortColumnOrder>('desc')
+const sortColumnKey = ref(tablePreferences.sortColumnKey || '') as Ref<ColumnKey | ''>
+const sortColumnOrder = ref<SortColumnOrder>(tablePreferences.sortColumnOrder || 'desc')
 const isClickable = ref(false)
 const hasToolbarSlot = computed((): boolean => !hideToolbar && !nested && (!!slots.toolbar || hasColumnVisibilityMenu.value || showBulkActionsToolbar.value))
 const isActionsDropdownHovered = ref<boolean>(false)
--- a/sandbox/pages/SandboxTableData.vue
+++ b/sandbox/pages/SandboxTableData.vue
@@ -32,6 +32,7 @@
             client-sort
             :fetcher="fetcher"
             :headers="headers(false, true)"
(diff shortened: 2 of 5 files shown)


src/components/KTableView/KTableView.cy.ts
```diff
@@ -481,6 +481,22 @@ describe('KTableView', () => {
         })
       })
     })
+
+    it('should respect initial sort order from table preferences', () => {
```
adamdehaven, line 485: Can you assert the table is actually sorted correctly here as well by the column?

Would also be nice to have negative path tests, e.g. table shows correct icon when no initial sort is applied, etc.

src/components/KTableView/KTableView.vue
```diff
@@ -212,12 +212,23 @@
                     </template>
                   </KTooltip>
 
-                  <ArrowDownIcon
-                    v-if="!column.hideLabel && column.sortable && column.key !== TableViewHeaderKeys.BULK_ACTIONS && column.key !== TableViewHeaderKeys.ACTIONS"
-                    class="sort-icon"
-                    :color="`var(--kui-color-text-neutral, ${KUI_COLOR_TEXT_NEUTRAL})`"
-                    :size="KUI_ICON_SIZE_30"
-                  />
+                  <template v-if="!column.hideLabel && column.sortable && column.key !== TableViewHeaderKeys.BULK_ACTIONS && column.key !== TableViewHeaderKeys.ACTIONS">
```
adamdehaven: Let's move this to a computed property if possible?

src/components/KTableView/KTableView.vue
```diff
@@ -212,12 +212,23 @@
...
+                    />
+
+                    <SwapSortIcon
+                      v-else
+                      class="sort-icon"
+                      :color="`var(--kui-color-text-neutral-weak, ${KUI_COLOR_TEXT_NEUTRAL_WEAK})`"
+                      decorative
+                      :size="KUI_ICON_SIZE_30"
+                    />
+                  </template>
```
adamdehaven: Should these icons have aria-attributes or something? (also for the ArrowUpIcon if it's in here somewhere)

src/components/KTableView/KTableView.cy.ts
portikM: Updated one of the existing tests to check for that

src/components/KTableView/KTableView.vue
portikM: Done (a function, not a computed)

src/components/KTableView/KTableView.vue
portikM: The problem here is we handle sorting incorrectly here from the a11y step-point. We trigger sorting directly on `th` click, while we really should have a `button` element within the `th` to handle any interaction. However, fixing that here now would be a breaking change so for now I'll add a `title` on the swap sort icon (the active sort icon doesn't need any attributes as `aria-sort` on the parent `th` element handles what we need here).

Screenshot 2025-07-14 at 11 03 53 AM

Chalks (dismissed):
looks good, works well. nice!

PR #2831, opened 2025-07-21
fix(ktabledata): incorrect page size on table preferences update

portikM:
# Summary

Fix incorrect page size on `update:table-preferences` event

### DIFF
--- a/src/components/KTableData/KTableData.vue
+++ b/src/components/KTableData/KTableData.vue
@@ -502,12 +502,11 @@ const tablePreferencesUpdateHandler = ({ columnWidths: newColumnWidth, columnVis
 const tableViewColumnWidths = ref<ColumnWidths | undefined>({})
 const tableViewColumnVisibility = ref<ColumnVisibility | undefined>({})
 const tableDataPreferences = computed((): TablePreferences<Header['key']> => ({
+  pageSize: pageSize.value,
   sortColumnKey: sortColumnKey.value,
   sortColumnOrder: sortColumnOrder.value,
   ...(tableViewColumnWidths.value ? { columnWidths: tableViewColumnWidths.value } : {}),
   ...(tableViewColumnVisibility.value ? { columnVisibility: tableViewColumnVisibility.value } : {}),
-  ...tablePreferences,
-  pageSize: pageSize.value,
 }))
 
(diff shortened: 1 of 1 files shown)


src/components/KTableData/KTableData.vue
```diff
@@ -502,12 +502,12 @@ const tablePreferencesUpdateHandler = ({ columnWidths: newColumnWidth, columnVis
 const tableViewColumnWidths = ref<ColumnWidths | undefined>({})
 const tableViewColumnVisibility = ref<ColumnVisibility | undefined>({})
 const tableDataPreferences = computed((): TablePreferences<Header['key']> => ({
-  pageSize: pageSize.value,
   sortColumnKey: sortColumnKey.value,
   sortColumnOrder: sortColumnOrder.value,
   ...(tableViewColumnWidths.value ? { columnWidths: tableViewColumnWidths.value } : {}),
   ...(tableViewColumnVisibility.value ? { columnVisibility: tableViewColumnVisibility.value } : {}),
   ...tablePreferences,
+  pageSize: pageSize.value,
```
Leopoldthecoder: I believe the values in `tablePreferences` are still overriding other fields. For instance, if `tablePreferences` includes `sortColumnOrder: 'desc'`, no matter how users change the sort order, the `update:table-preferences` event always emits with `sortColumnOrder: 'desc'`.

https://github.com/user-attachments/assets/4f3c8681-e611-4501-893e-9b5a3e95a8a1

Why do we want to merge `tablePreferences` into the computed `tableDataPreferences` in the first place?

portikM: You're right, I didn't realize we don't need it anymore. I removed that line. Initially I added it because I needed a way to pass `sortColumnKey` and `sortColumnOrde` values down to KTableView when those are passed through table preferences, but then I handled that differently by assigning proper initial values to those refs ([here](https://github.com/Kong/kongponents/pull/2819/files#diff-c00cf2df267951c4bb9bed11cc757d1f6e22f0f2d825c2bc7a312bda1785f8c2R263-R264)) and didn't realize `tablePreferences` is no longer needed in `tableDataPreferences`.

Leopoldthecoder: Thanks for the context!

PR #2832, opened 2025-07-23
fix(kdatetimepicker): add validation for start and end time

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-4122

Adds validation to prevent users from selecting a start time that exceeds the end time when using the custom time range picker.

#### Changes 

- Added hasCalendarError reactive flag and bound it to the submit button to prevent submission on invalid ranges
- Introduced helpText to surface validation errors to the user (e.g., "Start time cannot exceed end time")
- Added input-error class styling for invalid inputs
- Extended watchers on startTimeValue and endTimeValue to validate the full range on each update
- Corrected a bug in the endTimeParts assignment (was reading startTimeValue)

image

### DIFF
--- a/src/components/KDateTimePicker/CalendarWrapper.vue
+++ b/src/components/KDateTimePicker/CalendarWrapper.vue
@@ -78,7 +78,7 @@
 </template>
 
 <script lang="ts" setup>
-import { computed, onMounted, ref, watch } from 'vue'
+import { computed, onMounted, ref, useId, watch } from 'vue'
 import { DatePicker } from 'v-calendar'
 import type { DatePickerModel, DatePickerRangeObject, DateTimePickerMode } from '@/types'
 import { format } from 'date-fns'
@@ -98,7 +98,7 @@ const calendarVModel = defineModel<DatePickerModel>({ required: true })
 const hasError = defineModel<boolean>('error', { default: false })
 const startTimeValue = ref<string>(format(new Date(), 'HH:mm:ss'))
 const endTimeValue = ref<string>(format(new Date(), 'HH:mm:ss'))
-const componentId = ref<string>(crypto.randomUUID())
+const componentId = useId()
 
 const showTime = computed(() => {
   return ['time', 'dateTime', 'relativeDateTime'].includes(props.kDatePickerMode)
--- a/src/components/KDateTimePicker/CalendarWrapper.vue
+++ b/src/components/KDateTimePicker/CalendarWrapper.vue
@@ -58,12 +58,12 @@
         mode="out-in"
         name="kongponents-fade-transition"
       >
-        <p
-          v-if="helpText"
+        <div
+          v-if="hasError && errorMessage"
           class="help-text"
         >
-          {{ helpText }}
-        </p>
+          {{ errorMessage }}
+        </div>
       </Transition>
     </div>
   </div>
@@ -75,12 +75,17 @@ import { DatePicker } from 'v-calendar'
 import type { DatePickerModel, DatePickerRangeObject, DateTimePickerMode } from '@/types'
 import { format } from 'date-fns'
 
-const props = defineProps<{
+const props = withDefaults(defineProps<{
   isRange: boolean
   kDatePickerMode: DateTimePickerMode
   maxDate?: Date
   minDate?: Date
-}>()
+  errorMessage?: string
+}>(), {
+  maxDate: undefined,
+  minDate: undefined,
+  errorMessage: 'Start time cannot exceed end time.',
+})
 const calendarVModel = defineModel<DatePickerModel>({ required: true })
 const hasError = defineModel<boolean>('error', { default: false })
 const startTimeValue = ref<string>(format(new Date(), 'HH:mm:ss'))
@@ -88,13 +93,6 @@ const endTimeValue = ref<string>(format(new Date(), 'HH:mm:ss'))
 
 const showTime = computed(() => {
   return ['time', 'dateTime', 'relativeDateTime'].includes(props.kDatePickerMode)
-})
-
-const helpText = computed(() => {
-  if (hasError.value) {
-    return 'Start time cannot exceed end time.'
-  }
-  return ''
 })
 
 const isInvalidRange = (start: Date, end: Date): boolean => {
@@ -187,7 +185,7 @@ watch(() => calendarVModel.value, () => {
 
     calendarVModel.value.setHours(startTime.getHours(), startTime.getMinutes(), startTime.getSeconds(), 0)
   }
-})
+}, { immediate: true })
 
 const showRange = (rangeType: 'start' | 'end') => {
   const value = calendarVModel.value as { start: Date, end: Date }
--- a/src/components/KDateTimePicker/CalendarWrapper.vue
+++ b/src/components/KDateTimePicker/CalendarWrapper.vue
@@ -33,6 +33,7 @@
         <input
           v-model="startTimeValue"
           class="time-input-start"
+          :class="{ 'input-error': hasError }"
           data-testid="time-input-start"
           :step="60"
           type="time"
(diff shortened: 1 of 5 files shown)


src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -146,12 +173,13 @@ watch(() => calendarVModel.value, () => {
     const startTime = new Date()
     const endTime = new Date()
     const startTimeParts = startTimeValue.value.split(':')
-    const endTimeParts = startTimeValue.value.split(':')
+    const endTimeParts = endTimeValue.value.split(':')
```
filipgutica, line 183: This was a bug

src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -47,11 +48,23 @@
...
           data-testid="time-input-end"
           :step="60"
           type="time"
         >
       </div>
+      <Transition
+        mode="out-in"
+        name="kongponents-fade-transition"
+      >
+        <p
```
adamdehaven: nit: this should likely be a `div` or `span` instead of `p`

src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -69,13 +82,25 @@ const props = defineProps<{
...
 const startTimeValue = ref<string>(format(new Date(), 'HH:mm:ss'))
 const endTimeValue = ref<string>(format(new Date(), 'HH:mm:ss'))
 
 const showTime = computed(() => {
   return ['time', 'dateTime', 'relativeDateTime'].includes(props.kDatePickerMode)
 })
 
+const helpText = computed(() => {
+  if (hasError.value) {
+    return 'Start time cannot exceed end time.'
```
adamdehaven: suggestion: I'd map this error message to a prop, but this value can be the default

portikM: In that case we'd need to expose a prop for setting `hasError` too (doesn't make sense to pass message text when the component still only displays it in certain scenario - when start time exceeds the end time). Do we invision having other validation logic here? I think for now since we don't have any other limits we can treat it not as generic error message functionality but rather specific to case when start time exceeds the end time. So I think we don't need a prop for this.

adamdehaven: > I think for now since we don't have any other limits we can treat it not as generic error message functionality but rather specific to case when start time exceeds the end time. So I think we don't need a prop for this.

Non-blocking, but it would be a problem for translation (eventually), but also what if they needed it phrased differently for some reason?

portikM: > what if they needed it phrased differently for some reason?

In that case we'd need to name the prop appropriately so that it's specific to this particular validation logic. Something like `invalidTimeErrorMessage`?

adamdehaven: Prob fine for now; can revisit 👍🏼

src/components/KDateTimePicker/CalendarWrapper.vue
filipgutica: done

src/components/KDateTimePicker/CalendarWrapper.vue
filipgutica: done

src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -33,6 +33,7 @@
         <input
           v-model="startTimeValue"
           class="time-input-start"
+          :class="{ 'input-error': hasError }"
```
adamdehaven, line 40: suggestion: while we're editing these inputs, can we add proper attributes to link the input with its corresponding labels?

e.g.

```vue
<input :id="time-input-start-${componentId}">
...

<script setup lang="ts">
const componentId = useId()
</script>
```

Then you can use:

```vue
<label :for="time-input-start-${componentId}">
```

instead of the `span` elements? 

This is pseudo-code but you get what I'm saying?

src/types/date-time-picker.ts
```diff
@@ -186,6 +186,13 @@ export interface DateTimePickerProps {
    * @default 'bottom-start'
    */
   popoverPlacement?: PopPlacement
+
+  /**
+   * String to be displayed as an error message.
+   * If not provided, the default message is 'Start time cannot exceed end time.'
+   * @default undefined
+   */
+  errorMessage?: string
```
portikM: suggestion: if we do want to add a prop for this can we name it appropriately so that it's apparent it's error message for that specific validation logic and not a generic component error state message? e.g. `invalidTimeErrorMessage`

src/components/KDateTimePicker/CalendarWrapper.vue
filipgutica: done

filipgutica: Oh I didn't see the `useId()` part, didn't know that's a thing.

src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -62,20 +83,31 @@ import { DatePicker } from 'v-calendar'
...
+}>(), {
+  maxDate: undefined,
+  minDate: undefined,
+  errorMessage: 'Start time cannot exceed end time.',
+})
 const calendarVModel = defineModel<DatePickerModel>({ required: true })
+const hasError = defineModel<boolean>('error', { default: false })
 const startTimeValue = ref<string>(format(new Date(), 'HH:mm:ss'))
 const endTimeValue = ref<string>(format(new Date(), 'HH:mm:ss'))
+const componentId = ref<string>(crypto.randomUUID())
```
adamdehaven: issue: you don't need this to be a ref, and don't need to use crypto; Vue has a built-in for this
```suggestion
const componentId = useId()
```

filipgutica: done

src/types/date-time-picker.ts
filipgutica: done

src/components/KDateTimePicker/CalendarWrapper.vue
adamdehaven: https://vuejs.org/api/composition-api-helpers.html#useid

And it's SSR compatible so the ID persists in server-side rendering

src/components/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -157,7 +157,7 @@ const {
   disabled,
   readonly,
   popoverPlacement = 'bottom-start',
-  errorMessage = undefined,
+  invalidTimeErrorMessage = undefined,
```
portikM: nit: can you please set the default value here rather than in CalendarWrapper component and have `errorMessage = undefined` there instead? The reason is so that the default value is passed down and so KDateTimePicker is the source of truth

portikM (approved):
LGTM 🚀

src/types/date-time-picker.ts
```diff
@@ -186,6 +186,13 @@ export interface DateTimePickerProps {
    * @default 'bottom-start'
    */
   popoverPlacement?: PopPlacement
+
+  /**
+   * String to be displayed as an error message.
+   * If not provided, the default message is 'Start time cannot exceed end time.'
+   * @default undefined
```
Justineo, line 193: nit: this should be `@default 'Start time cannot exceed end time.'` and we can simplify the description.

src/types/date-time-picker.ts
```diff
@@ -186,6 +186,13 @@ export interface DateTimePickerProps {
    * @default 'bottom-start'
    */
   popoverPlacement?: PopPlacement
+
+  /**
+   * String to be displayed as an error message.
```
Justineo, line 191: nit: this is a little bit inaccurate.

PR #2833, opened 2025-07-23
fix(ktableview): make row actions column sticky [KHCP-17263]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-17263

Make row actions column sticky when table is scrollable horizontally


src/styles/mixins/_tables.scss
```diff
@@ -95,10 +103,10 @@
             }
 
             &.is-scrolled {
-              &:after {
+              &::after {
                 border-bottom: var(--kui-border-width-10, $kui-border-width-10) solid
                   var(--kui-color-border, $kui-color-border);
-                box-shadow: var(--kui-shadow, $kui-shadow);
+                height: calc(100% - 1px); // removes the tiny gap between the bottom border and the :after layer
```
jw-foss: shall we use `var(--kui-border-width-10, $kui-border-width-10)` instead of magic number here?

jw-foss (dismissed):

src/styles/mixins/_tables.scss
portikM: I was debating that same question with myself. I guess yes, let's use the token here. Updated my code, thanks!

src/components/KTableView/KTableView.vue
```diff
@@ -264,113 +271,130 @@
...
+                      'row-link': !!rowLink(row).to,
+                      'sticky-column': (header.key === TableViewHeaderKeys.BULK_ACTIONS || header.key === TableViewHeaderKeys.EXPANDABLE) && isScrolledHorizontally,
+                      'second-sticky-column': header.key === TableViewHeaderKeys.BULK_ACTIONS && hasExpandableRows,
+                      'has-row-scroll-overlay': isLastStickyColumn(header.key),
+                      'actions-column': header.key === TableViewHeaderKeys.ACTIONS,
+                    }"
+                    :style="{
+                      ...columnStyles[header.key],
+                      ...(cellHelperData.isRowActionsDropdownOpen ? { zIndex: 2 } : {}),
+                    }"
```
adamdehaven: issue: there's way too much logic in the template here. Can you create computed properties to handle the mapping, and maybe add some comments?

src/components/KTableView/KTableView.vue
```diff
@@ -264,113 +271,130 @@
...
-                        />
-                      </KButton>
+                        {{ row[header.key] }}
+                      </slot>
 
-                      <template #items>
-                        <slot
-                          name="action-items"
-                          :row="row"
+                      <KTooltip
```
adamdehaven, line 314: Is most of the code from here down actually changing, or just indentation, etc?

Can you point out what actually changed in the PR so it's easier to review?

adamdehaven (changes requested):
The PR is pretty hard to review as it appears the entirety of the `KTableView.vue` component has changed. If that isn't the case, can you point out what actually changed in the code?

Also, I don't see any tests being added for the new exposed methods or new functionality?

src/components/KTableView/KTableView.vue
Justineo: This could help:

Image

adamdehaven: I was referring to all of the computed classes.

But overall yes, once I hide whitespace the PR's actual changes are much easier to go through

src/components/KTableView/KTableView.vue
portikM: I'm guessing without whitespace these changes don't require as much explanation? What I've typed up before @Justineo posted his comment and made reviewing PRs muuuch easier for me (thanks for the tip!) is this:
Here it's 99% indentation. In the template section of this component, all that changed is I've wrapped the `td` in KComponent, added `actions-column` in `td` element bound classes and `zIndex` in bound styles.

src/components/KTableView/KTableView.vue
portikM: Yes, I will simplify template logic here by breaking down some of these conditions into computed and functions.

src/components/KTableView/KTableView.vue
adamdehaven: Yep after hiding the whitespace this is resolved

src/components/KTableView/KTableView.vue
portikM: Addressed, I've simplified the template logic

portikM:
> The PR is pretty hard to review as it appears the entirety of the `KTableView.vue` component has changed. If that isn't the case, can you point out what actually changed in the code?
> 
> Also, I don't see any tests being added for the new exposed methods or new functionality?

I've added component tests for newly exposed methods in KDropdown.

adamdehaven (dismissed):

src/styles/mixins/_tables.scss
```diff
@@ -430,8 +447,13 @@
     }
 
     &.right {
-      background: linear-gradient(to left, rgba(#000, 0.1), rgba(#000, 0));
+      background: linear-gradient(to left, rgba(#000, 0.08), rgba(#000, 0));
       right: 0;
+
+      &.row-overlay {
+        left: calc(-1 * $scrollOverlayWidth);
```
adamdehaven: nit: can we just do this?

```suggestion
        left: -#{$scrollOverlayWidth};
```

portikM: Turns out we can! Done.

PR #2836, opened 2025-07-26
fix(kdatetimepicker): Update internal state on modelValue changed

filipgutica:
# Summary

https://konghq.atlassian.net/jira/software/c/projects/MA/boards/109

This PR replaces the initialization approach in KDateTimePicker from `onMounted` to  a `watch` on `modelValue` with `{ immediate: true }`. This ensures the component properly initializes and also reacts to external model value changes consistently.

#### Changes
- Set the time input values correctly when v calendar model updates
- Use `watch` with `{ immediate: true }` instead of `onMounted` for both initializing the state and reacting to external changes to `modelValue`
- Added a reset time range button in the sandbox to simulate external updates to the model value

#### Original issue
- When updating the time range externally (e.g. via zoom), the time portion was incorrectly reset to the current time instead of preserving the intended value.
- Updating the time inputs using the time picker would also reset them to the current time after applying.

### DIFF
--- a/src/components/KDateTimePicker/CalendarWrapper.vue
+++ b/src/components/KDateTimePicker/CalendarWrapper.vue
@@ -142,30 +142,25 @@ const calendarSelectAttributes = {
 
 watch(() => startTimeValue.value, (newTime) => {
   if (calendarVModel.value && props.isRange && 'start' in calendarVModel.value && calendarVModel.value.start instanceof Date && 'end' in calendarVModel.value && calendarVModel.value.end instanceof Date) {
-
     const startTime = new Date()
     const timeParts = newTime.split(':')
     startTime.setHours(parseInt(timeParts[0], 10), parseInt(timeParts[1], 10), 0, 0)
-
-    calendarVModel.value.start.setHours(startTime.getHours(), startTime.getMinutes(), startTime.getSeconds(), 0)
+    calendarVModel.value.start.setHours(startTime.getHours(), startTime.getMinutes(), 0, 0)
     hasError.value = isInvalidRange(calendarVModel.value.start, calendarVModel.value.end)
   } else if (calendarVModel.value instanceof Date) {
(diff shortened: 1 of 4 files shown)


src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -176,25 +171,13 @@ watch(() => calendarVModel.value, () => {
...
-    const endTime = new Date()
-    const startTimeParts = startTimeValue.value.split(':')
-    const endTimeParts = endTimeValue.value.split(':')
-    startTime.setHours(parseInt(startTimeParts[0], 10), parseInt(startTimeParts[1], 10), 0, 0)
-    endTime.setHours(parseInt(endTimeParts[0], 10), parseInt(endTimeParts[1], 10), 0, 0)
-
-    calendarVModel.value.start.setHours(startTime.getHours(), startTime.getMinutes(), startTime.getSeconds(), 0)
-    calendarVModel.value.end.setHours(endTime.getHours(), endTime.getMinutes(), endTime.getSeconds(), 0)
+    startTimeValue.value = format(calendarVModel.value.start, 'HH:mm')
+    endTimeValue.value = format(calendarVModel.value.end, 'HH:mm')
```
portikM, line 175: ✨ This is much more neat than what we had, thanks!

portikM (changes requested):
Looks good mostly, however I've noticed one issue where the calendar still displays the old selection even after I've rest it by clicking the reset button (I'm guessing maybe you need to re-render the v-calendar component?). In the video you can see how I'm selecting a different time range and then after I click the reset button the KDateTimePicker value changes, but in v-calendar I'm still seeing the old date range.

https://github.com/user-attachments/assets/48b2aa6c-4c9b-4f13-803d-12aa8172160c

portikM (approved):
LGTM 🚀

PR #2838, opened 2025-07-29
fix(ktabledata): fetcher firing twice when initial sorting provided

portikM:
# Summary

Fixing the issue in KTableData where `fetcher` is firing twice on component mount when `clientSort` is `false`


lukewhchen (approved):
LGTM! Confirmed the fix works!

Issue #2753, opened 2025-05-14
Several swallowed typescript issues


johncowen:
We noticed here https://github.com/kumahq/kuma-gui/pull/3883 that there are some issues with KRadio's types and we are subsequently unable to upgrade to the latest minor version of Kongponents.

The bigger issue is that there seem to be quite a few typescript issues that do not block a release due to typechecking not being used for tests (see https://github.com/Kong/kongponents/pull/828) and these issues are essentially being swallowed.

https://github.com/Kong/kongponents/blob/686eb360ae4b2f213b8838d8b35cc992a94a8389/tsconfig.build.json#L4

<img width="646" alt="Image" src="https://github.com/user-attachments/assets/636cf1ea-ad49-4dc8-b289-27be77077ab1" />

Typechecking the tests would help prevent engineers from publishing code that causes issues downstream, also see https://github.com/Kong/kongponents/pull/2752

Justineo:
There’s an ongoing epic to improve type accuracy for Kongponents: https://konghq.atlassian.net/browse/KM-1093. Based on the error message, the issue you encountered was likely introduced in this PR: https://github.com/Kong/kongponents/pull/2739.

It seems we can only declare slots as optional with `defineSlots` here (though the official docs declare slots as required fields). I'll put up a PR to fix to test against `kuma-gui`.

---

To minimize the risk of breaking existing code, we’re currently opening preview PRs in `konnect-ui-apps`. Type-checking the tests also sounds like a good idea if you plan to continue working on #828.

adamdehaven:
Is this issue now resolved?

johncowen:
Hey sorry for the late response, feel free to ping me (here or in Slack) if you ever get no response from me! I must have missed this PR comment, sorry!

---

I'm running https://github.com/Kong/kongponents/commit/2f62c929700b6d0fe7f3af93c18b87b6f3dc736e and this issue is still seems un-resolved. If I remove the `"src/**/*.cy.ts", ` exclude mentioned above and run `pnpm run typecheck` I still get the following:

<img width="646" height="416" alt="Image" src="https://github.com/user-attachments/assets/333efcae-e976-43db-a463-6cc41ffe911f" />

I also tried removing the similar exclude line from the eslint config and running `pnpm run lint` gives me (plus more):

<img width="777" height="1177" alt="Image" src="https://github.com/user-attachments/assets/ff6a4d0f-9cc6-488d-8601-d7eb315a187c" />

[...]

PR #2844, opened 2025-08-01
feat(kemptystate): add new props and slots [KHCP-17264]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-17264

Add new props and slots to make KEmptyState compatible with [EntityEmptyState](https://github.com/Kong/public-ui-components/blob/main/packages/entities/entities-shared/docs/entity-empty-state.md) component

[Component Figma](https://www.figma.com/design/Yze0SWXl5nKjR0rFdilljK/Components?node-id=2020-11252)

Screenshot 2025-08-01 at 9 05 56 AM


docs/components/empty-state.md
```diff
@@ -174,6 +174,63 @@ Accepted values:
...
+  :features="features"
+  message="Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla nec purus feugiat, molestie ipsum et, consequat nibh."
+  title="Empty State With Features"
+/>
+
+```html
+<KEmptyState
+  :features="[
+    {
+      title: 'Feature 1',
```
Justineo, line 223: Shall we start the index from 0, since related slots are named that way?

docs/components/empty-state.md
```diff
@@ -313,9 +513,44 @@ const onActionClick = (): void => {
...
+import { KongIcon, AddCircleIcon, WavingHandIcon, SparklesIcon, RocketIcon, DesignIcon } from '@kong/icons'
+import type { EmptyStateFeature } from '@/types'
 
 const onActionClick = (): void => {
   alert('Action button clicked!')
 }
+
+const features: EmptyStateFeature[] = [
+  {
+    title: 'Feature 1',
```
Justineo, line 515: Same as above.

src/types/empty-state.ts
```diff
@@ -71,6 +88,26 @@ export interface EmptyStateSlots {
...
+
+  /**
+   * Slot for a secondary message or additional information (e.g. pricing).
+   */
+  'supporting-text'?(): any
+
+  /**
+   * Slot for custom icon of the empty state feature card.
+   */
+  [key: `feature-${number}-icon`]: () => any
```
Justineo:
```suggestion
  [key: `feature-${number}-icon`]?: () => any
```

Justineo: I think maybe we can restrict the number according to the length of the `features` prop. But we can do it in another PR.

docs/components/empty-state.md
```diff
@@ -174,6 +174,63 @@ Accepted values:
...
+  icon-variant="kong"
+  action-button-text="Action"
+  message="Lorem ipsum dolor sit amet..."
+  title="Empty State Decorative Icon"
+/>
+```
+
+### features
+
+If provided, will display card for each feature below the action button, along with an icon slot, a title and a short description. You can also use [`feature-icon` slot](#featrue-icon) to display an icon in each feature card.
```
Justineo: There are several `feature`s spelled as `featrue` like this.

Justineo (changes requested):

docs/components/empty-state.md
portikM: Good call, done.

docs/components/empty-state.md
portikM: Good call, done.

docs/components/empty-state.md
portikM: Oops good catch, thanks!

src/types/empty-state.ts
portikM: Yeah could be a small improvement for future. You suggestion failed the lint step though so instead I tried
```
[key: `feature-${number}-icon`]: () => any | undefined
```
-- will that work?

Justineo: We had an issue earlier that non-optional slots might cause some trouble:
https://github.com/Kong/kongponents/issues/2753#issuecomment-2879648008

But it seems that you cannot use a `?` modifier after index signature. Let's just remove it first and see if we'll encounter any problems.

docs/components/empty-state.md
```diff
@@ -174,6 +174,63 @@ Accepted values:
...
+<KEmptyState
+  icon-background
+  icon-variant="kong"
+  action-button-text="Action"
+  message="Lorem ipsum dolor sit amet..."
+  title="Empty State Decorative Icon"
+/>
+```
+
+### features
```
adamdehaven, line 199: issue: we need to document the interface, etc.

docs/components/empty-state.md
```diff
@@ -223,6 +280,30 @@ Slot for passing title text.
...
+```html
+<KEmptyState
+  message="Lorem ipsum dolor sit amet..."
+  title="Empty State With Supporting Text"
+>
+  <template #supporting-text>
+    <b>Pricing:</b> starting at $30 per user per month.
+  </template>
+</KEmptyState>
+```
```
adamdehaven: question: it seems like this could just be part of the default slot. Why does it need to be separate?

nit: if we are going to keep it, would a slot name of `footer` be more appropriate?

docs/components/empty-state.md
```diff
@@ -281,8 +390,99 @@ Slot for providing your custom action button.
 </KEmptyState>
 ```
 
-## Events
+### feature-icon
```
adamdehaven, line 382: issue: it's a bit unintuitive to slot icons via index (a `number`) - is it possible to require each `feature` have a unique `key` instead, and then the `key` is used in the slot name?

docs/components/empty-state.md
```diff
@@ -281,8 +390,99 @@ Slot for providing your custom action button.
...
+  <template #feature-2-icon>
+    <RocketIcon />
+  </template>
+  <template #feature-3-icon>
+    <DesignIcon />
+  </template>
+</KEmptyState>
+```
+
+### bottom
```
adamdehaven: issue: this slot is confusing with the existence of the `supportingText` slot as well.

Can some of these slots be consolidated? It seems soooo specific to a certain use-case

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -4,14 +4,23 @@
...
-            decorative
-            :size="KUI_ICON_SIZE_60"
-          />
+      <div
+        :aria-hidden="$slots.image ? undefined : 'true'"
+        :class="{
+          'empty-state-icon': !$slots.image,
+          'has-background': !slots.image && iconBackground,
+          'empty-state-image': $slots.image,
+        }"
```
adamdehaven: suggestion: move to a computed property

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -4,14 +4,23 @@
...
+      >
+        <slot name="image">
```
adamdehaven, line 11: with these slots nested, how would you utilize the `icon` slot without first utilizing the `image` slot?

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -32,31 +41,73 @@
...
+            </div>
+
+            <div class="card-header">
+              {{ feature.title }}
+            </div>
+          </template>
+
+          <div :title="feature.description">
+            {{ feature.description }}
+          </div>
```
adamdehaven, line 80: question: is there logic that guarantees `title` and `description` are defined? If not, there should be conditionals on these elements so they do not render if empty

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -125,6 +182,23 @@ const getIconColor = computed((): string => {
...
+
+        :deep(#{$kongponentsKongIconSelector}) {
+          height: var(--kui-icon-size-50, $kui-icon-size-50) !important;
+          width: var(--kui-icon-size-50, $kui-icon-size-50) !important;
+        }
+      }
+    }
+
+    .empty-state-image {
+      max-width: 640px;
```
adamdehaven: what if the viewport is smaller?

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -146,12 +221,68 @@ const getIconColor = computed((): string => {
...
   }
 
-  .empty-state-action {
-    align-items: center;
+  .empty-state-features-container {
+    display: grid;
+    gap: var(--kui-space-70, $kui-space-70);
+    grid-template-columns: 1fr;
+    justify-content: space-around;
+    max-width: 650px;
```
adamdehaven: question: what if the viewport is smaller?

adamdehaven (changes requested):

docs/components/empty-state.md
portikM: I agree that it should just be a default slot. We had it in EntityEmptyState so I thought I'd carry it over but it's just a submessage under the default slot - I can just tweak the default slot wrapper styles to add flexbox spacing and we'll be able to achieve the same result. Will remove this slot.

docs/components/empty-state.md
portikM: Agree, updated.

docs/components/empty-state.md
portikM: I removed the `supporting-text` slot and when I did I felt like more appropriate name for this slot would be `footer`. Renamed it accordingly - I think it's much more straight-forward now.

src/components/KEmptyState/KEmptyState.vue
portikM: Done

src/components/KEmptyState/KEmptyState.vue
portikM: It's one or the other. If you use `image` slot you can't use the icon (since we don't want to have both an image and an icon in an empty state). If you don't use the image you can use the icon slot normally.

src/components/KEmptyState/KEmptyState.vue
portikM: The `title` and `description` are required properties in the `EmptyStateFeature` interface - you'll get type error when trying to pass an object without one or the other, I feel like that should be enough.

src/components/KEmptyState/KEmptyState.vue
portikM: It will take up all available width but won't stretch more than 640px on larger viewports. Looks something like this on small:
Screenshot 2025-08-05 at 11 31 10 AM

src/components/KEmptyState/KEmptyState.vue
portikM: Features stack:

Screenshot 2025-08-05 at 11 32 32 AM

docs/components/empty-state.md
portikM: Done

docs/components/empty-state.md
```diff
@@ -174,6 +174,75 @@ Accepted values:
...
+/>
+```
+
+### features
+
+If provided, will display card for each feature below the action button, along with an icon slot, a title and a short description. You can also use [`feature-icon` slot](#feature-icon) to display an icon in each feature card. Expects an array of objects of type `EmptyStateFeature`:
+
+```ts
+interface EmptyStateFeature {
+  key?: string
```
adamdehaven, line 206: Do we need to document the key must be unique per item?

docs/components/empty-state.md
```diff
@@ -249,6 +318,34 @@ Slot for providing custom icon.
...
+</KEmptyState>
+
+```html
+<KEmptyState
+  message="Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla nec purus feugiat, molestie ipsum et, consequat nibh."
+  title="Empty State With Image"
+>
+  <template #image>
+    <img src="https://picsum.photos/640/300">
+  </template>
```
adamdehaven, line 342: Should this be a slot, or a prop that accepts the image source string so that we have control over the image display? (Or is there a reason you went with a slot?)

docs/components/empty-state.md
```diff
@@ -281,8 +378,99 @@ Slot for providing your custom action button.
...
+### feature-icon
 
+For each feature provided through the [`features` prop](#features), you can pass a custom icon through the `feature-key-icon` slot where `key` is the value of the `key` property in the corrsponding feature object.
+
+<KEmptyState
+  :features="features"
+  message="Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla nec purus feugiat, molestie ipsum et, consequat nibh."
+  title="Empty State With Features"
+>
+  <template #feature-wave-icon>
```
adamdehaven: I would expect the key to be at the end of the dynamic slot name, not in the middle. 

@Justineo thoughts?

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -32,31 +37,67 @@
...
-          :disabled="actionButtonDisabled"
-          type="button"
-          @click="emit('click-action')"
-        >
-          {{ actionButtonText }}
-        </KButton>
-      </slot>
+      <template
+        v-for="feature in features"
+        :key="feature"
```
adamdehaven: issue: should the key here be `feature.title`?

src/components/KEmptyState/KEmptyState.vue
```diff
@@ -32,31 +37,67 @@
...
+      >
+        <KCard class="empty-state-feature-card">
+          <template #title>
+            <div
+              v-if="feature.key && $slots[`feature-${feature.key}-icon`]"
+              aria-hidden="true"
+              class="feature-icon"
+            >
+              <slot :name="`feature-${feature.key}-icon`" />
```
adamdehaven: Same comment on slot naming, I think the key should be at the end.

Also, what if the key has invalid characters (e.g. space, etc.)?

adamdehaven (commented):
I noticed the lock file changed; I think this is unintentional?

docs/components/empty-state.md
Justineo: I think we can have both.

adamdehaven: I was thinking about how a string would be interpreted in a host app; e.g. would it be able to resolve the image if it's in `assets` etc.?

portikM: I think it's easier with a slot because what if the image you want to slot here is an svg?

docs/components/empty-state.md
Justineo: I was gonna say let’s stick to our current convention but I found that currently it’s inconsistent. `<KTabs>` has `${key}-anchor` while `<KTableView>` has `tooltip-${key}`…

adamdehaven: Considering the key could have invalid characters, IMO it makes sense to put the "most unstable" piece at the end so that the uncertainty is at the end instead of in the middle.

e.g. 

```ts
feature-icon-{unknown}

feature-{unknown}-icon
```

Meaning if we have to do text replacements on the key, at least we just change the end and not create something less than ideal like `feature-my-key-had-weird-characters-icon`

docs/components/empty-state.md
portikM: Done

docs/components/empty-state.md
portikM: Ok, pushed a commit changing it so that the `key` comes in the end.

src/components/KEmptyState/KEmptyState.vue
portikM: Oops this is meant to be `feature.key || idx`. Fixed now.

src/components/KEmptyState/KEmptyState.vue
portikM: Done.

portikM:
> I noticed the lock file changed; I think this is unintentional?

Yes, bad rebase - fixed now.

adamdehaven (approved):

PR #2848, opened 2025-08-06
fix(kslider): floating point rounding error

Leopoldthecoder:
# Summary

This PR fixes a bug where the component couldn't work when `min`, `max` and/or `step` is a floating-point number.

| Before | After |
| - | - |
| https://github.com/user-attachments/assets/7b6aa149-476b-4a22-8870-92775ffbffd5 | https://github.com/user-attachments/assets/a02d26a7-16f5-40b5-b8f9-6ca37f7225e3 |

### DIFF
--- a/src/components/KSlider/KSlider.vue
+++ b/src/components/KSlider/KSlider.vue
@@ -143,14 +143,27 @@ const inputValue = defineModel({
   type: Number,
 })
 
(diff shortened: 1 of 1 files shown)


portikM (approved):
LGTM 🚀

PR #2851, opened 2025-08-06
fix(kselect): hide caret in readonly [KHCP-17489]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-17489

Hide caret in readonly mode in KSelect

### DIFF
--- a/src/components/KSelect/KSelect.cy.ts
+++ b/src/components/KSelect/KSelect.cy.ts
@@ -21,7 +21,8 @@ describe('KSelect', () => {
       },
     })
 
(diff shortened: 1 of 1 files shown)


src/components/KSelect/KSelect.vue
```diff
@@ -82,6 +82,7 @@
                 <CloseIcon decorative />
               </button>
               <ChevronDownIcon
+                v-if="!isReadonly"
```
adamdehaven, line 85: can you add a simple component test?

portikM: Done

PR #2859, opened 2025-08-11
fix(ktabledata): presist table preferences

portikM:
# Summary

Fix issue where KTableData doesn't persist column width and column visibility table preferences

### DIFF
--- a/src/components/KTableData/KTableData.cy.ts
+++ b/src/components/KTableData/KTableData.cy.ts
@@ -871,6 +871,18 @@ describe('KTableData', () => {
         }
       })
     })
+
+    it.skip('does not apply page size, sort column key and order preferences when not passed', () => {
+      // TODO: KHCP-17564 expand table preferences test coverage
+    })
+
+    it.skip('applies page size, sort column key and order preferences when passed', () => {
+      // TODO: KHCP-17564 expand table preferences test coverage
+    })
+
+    it.skip('emits update:table-preferences event when table preferences are updated', () => {
+      // TODO: KHCP-17564 expand table preferences test coverage
+    })
   })
 
   describe('misc', () => {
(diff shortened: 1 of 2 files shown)


src/components/KTableView/KTableView.vue
```diff
@@ -752,7 +752,10 @@ const expandableColumnWidth = (parseInt(KUI_SPACE_60) * 2) + parseInt(KUI_ICON_S
  */
 const DEFAULT_COLUMN_WIDTHS: Record<TableReservedColumnKey, number> = { expandable: expandableColumnWidth, bulkActions: 48, actions: 54 }
 const DEFAULT_COLUMN_WIDTHS_PX: Record<TableReservedColumnKey, string> = mapValues(DEFAULT_COLUMN_WIDTHS, (width) => `${width}px`)
-const columnWidths = ref(tablePreferences?.columnWidths || DEFAULT_COLUMN_WIDTHS) as Ref<ColumnWidths>
+const columnWidths = ref({
+  ...tablePreferences.columnWidths || {},
+  ...DEFAULT_COLUMN_WIDTHS, // override special columns with default widths
+}) as Ref<ColumnWidths>
```
Chalks, line 758: **Question, maybe a blocker**

I'm slightly concerned about this change. Just from reading the code, it looks like that if someone had previously set the `actions` (or `bulkActions` or `expandable`) column width to 200 via `tablePreferences`, that will no longer work. Are there reasonable use cases for doing such a thing, and am I interpreting this correctly?

portikM: You are correct, special columns (expandable row controls, bulk actions and row actions) are assigned set width and not allowed to be resized. It's a design / UX decision really but there is no real need to have those columns wider (or narrower) than they need to be. Those columns cannot be resized (the resize handler doesn't appear on special columns) so really we're just ensuring consistency here.

Chalks (approved):

adamdehaven (approved):

PR #2869, opened 2025-08-19
fix: better error handling for time inputs

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-4240

When clearing the time input calendar v-model was ending up as an "Invalid Date" -> date-fns was throwing error when attempting to format the invalid date.

Changes
- Check if new time value is defined in watcher
- Wrap date-fns format() fn in a try - catch and console.warn the error

### DIFF
--- a/src/components/KDateTimePicker/CalendarWrapper.vue
+++ b/src/components/KDateTimePicker/CalendarWrapper.vue
@@ -27,10 +27,10 @@
         >
           <span v-if="showRange('start')">
             <!-- @vue-ignore: typeguard in showRange -->
-            {{ format(calendarVModel.start, 'EEE MMM d yyyy') }}
+            {{ formatDateDisplay(calendarVModel.start) }}
           </span>
           <span v-else-if="(calendarVModel && calendarVModel instanceof Date)">
-            {{ format(calendarVModel, 'EEE MMM d yyyy') }}
+            {{ formatDateDisplay(calendarVModel) }}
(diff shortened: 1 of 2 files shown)


adorack (commented):
Is it possible to add a test to KDateTimePicker.cy.ts?

adorack (approved):

src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -200,6 +206,14 @@ const resetTime = () => {
   endTimeValue.value = originalTimeValues.value.end
 }
 
+const formatDateDisplay = (date: Date) => {
+  try {
+    return format(date, 'EEE MMM d yyyy')
+  } catch (error) {
+    console.warn('Error formatting date:', error)
+  }
```
adamdehaven, line 214: should we not expose the error anywhere in the UI? How would the user know how to recover or that there is an issue?

filipgutica: There's nothing really to expose in the UI in this case. If somehow an invalid date ends up here we will see a warning and can address the issue without polluting datadog with errors.

adamdehaven: Sounds good, thanks for the context 👍🏼

adamdehaven (approved):

PR #2874, opened 2025-08-22
fix(ktableview): table preferences test coverage [KHCP-17564]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-17564

Minor fix in KTableView, also add test coverage for `tablePreferences` prop in KTableView and KTableData

### DIFF
--- a/src/components/KTableView/KTableView.cy.ts
+++ b/src/components/KTableView/KTableView.cy.ts
@@ -1,6 +1,7 @@
 import { h } from 'vue'
 import KTableView from '@/components/KTableView/KTableView.vue'
 import type { TableViewHeader, RowBulkAction } from '@/types'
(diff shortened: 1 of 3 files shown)


src/components/KTableView/KTableView.cy.ts
```diff
@@ -595,16 +596,73 @@ describe('KTableView', () => {
       })
     })
 
-    it.skip('does not apply page size, sort column key and order preferences when not passed', () => {
-      // TODO: KHCP-17564 expand table preferences test coverage
+    it('does not apply page size, sort column key and order preferences when not passed', () => {
```
kaiarrowood:
```suggestion
    it('correctly handles when page size, sort column key and order preferences are not passed', () => {
```

kaiarrowood (approved):
LGTM! 🚀

PR #2881, opened 2025-08-27
feat(kfileupload): allow drag & drop, dropzone appearance [KHCP-17516]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-17516

* Add `appearance="dropzone"` prop to support dropzone appearance in KFileUpload ([Figma](https://www.figma.com/design/TKG7r9QQSMDsBRfR0Gphpc/Drop-zone?node-id=0-1&p=f&t=4q58NCsIsnAMjThc-0))

Screenshot 2025-09-02 at 9 29 22 AM

### DIFF
--- a/docs/components/file-upload.md
+++ b/docs/components/file-upload.md
@@ -1,6 +1,6 @@
 # File Upload
 
-KFileUpload provides a wrapper around an `input` element with `type=file` for file upload.
+KFileUpload provides a wrapper around an `input` element with `type="file"` for file upload.
 
 <ClientOnly>
   <KFileUpload label="File upload" :label-attributes="{ info: `Accepted file types: ${acceptedFileType.join(', ')}` }" help="File size must be less than 1MB." :accept="acceptedFileType" />
@@ -15,6 +15,37 @@ KFileUpload provides a wrapper around an `input` element with `type=file` for fi
 ```
 
 ## Props
+
+### appearance
+
+KFileUpload comes in two variations: `input` and `dropzone`. Default value is `input`. Both appearances support drag-and-drop file upload functionality which is the default for `dropzone` appearance and configurable for `input` appearance through [`allowDragAndDrop` prop](#allowdraganddrop).
+
+<ClientOnly>
+  <div class="vertical-spacing-container">
+    <KFileUpload
+      label="Input file upload"
+      :accept="acceptedFileType"
+    />
+    <KFileUpload
+      appearance="dropzone"
+      label="Dropzone file upload"
+      :accept="acceptedFileType"
+    />
+  </div>
+</ClientOnly>
+
+```html
+<KFileUpload
+  label="Input file upload"
+  :accept="acceptedFileType"
+/>
+
+<KFileUpload
+  appearance="dropzone"
+  label="Dropzone file upload"
+  :accept="acceptedFileType"
+/>
+```
 
 ### accept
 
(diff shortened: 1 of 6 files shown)


docs/components/file-upload.md
```diff
@@ -128,22 +242,53 @@ Use this prop to customize the maximize size of file that can be uploaded. Defau
 By default KFileUpload will display the error state with a generic error message when selected file exceeds specified maximum file size. You can use [`errorMessage` prop](#errormessage) in conjunction with the [`error` event](#events) to display a custom error message.
 :::
 
+### allowDragAndDrop
+
+Boolean to control whether to allow drag and drop functionality when `appearance="input"`. Defaults to `true`.
```
Justineo: Why do we need this prop to disable the drag and drop behavior? Native file inputs allow drag and drop too.

src/components/KFileUpload/KFileUpload.vue
```diff
@@ -64,6 +78,65 @@
...
+    >
+      <div
+        ref="dropzone"
+        :aria-disabled="disabled ? 'true' : undefined"
+        :aria-labelledby="label ? `file-upload-dropzone-label-${fileInputId}` : undefined"
+        class="dropzone"
+        :class="{
+          'dragging': isDragging && !disabled,
+          'disabled': disabled,
+          'error': hasError,
```
Justineo: q: do we prefer always add quotes around object keys in Kongponents?

src/components/KFileUpload/KFileUpload.vue
```diff
@@ -64,6 +78,65 @@
...
+        }"
+        data-testid="file-upload-dropzone"
+        role="button"
+        :tabindex="!disabled ? 0 : -1"
```
Justineo:
```suggestion
        :tabindex="disabled ? -1 : 0"
```

src/components/KFileUpload/KFileUpload.vue
```diff
@@ -64,6 +78,65 @@
...
+        v-if="hasError || help"
+        class="help-text"
+        :class="{ 'error': hasError }"
+      >
+        <template v-if="hasError">
+          {{ errorMessage || invalidFileTypeErrorMessage || fileSizeErrorMessage || help }}
+        </template>
+        <template v-else>
+          {{ help }}
+        </template>
```
Justineo:
```suggestion
        {{ hasError ? errorMessage || invalidFileTypeErrorMessage || fileSizeErrorMessage || help : help }}
```

src/components/KFileUpload/KFileUpload.vue
```diff
@@ -271,6 +422,76 @@ $kFileUploadInputPaddingY: var(--kui-space-40, $kui-space-40); // corresponds to
...
+      }
+
+      &:hover,
+      &.dragging {
+        border-color: var(--kui-color-border-primary-weak, $kui-color-border-primary-weak);
+        border-style: solid;
+      }
+
+      &:focus:not(.disabled),
+      &:focus-visible {
```
Justineo, line 468: Why do we need both `:focus` and `:focus-visible`?

docs/components/file-upload.md
portikM: Fair question, I'm not sure what use-case for disabling drag and drop in input appearance would be but this requirement is coming from the [component brief](https://docs.google.com/document/d/1zKi51zd3MgUPWfvGWNOmKTQglE7jFjQOKQhEe1ZVLuU/edit?disco=AAABfRn6B-k). Maybe @adamdehaven has an opinion. I'm totally fine with removing `allowDragAndDrop` prop and have KFileUpload support drag and drop by default.

src/components/KFileUpload/KFileUpload.vue
portikM: Actually I prefer no quotes for single-word keys, this was just copy-pasta. Updated now.

src/components/KFileUpload/KFileUpload.vue
portikM: Because we're setting focus on a `div[role="button"]` manually when file uploaded. I checked, without `:focus` it doesn't look how it's supposed to.

docs/components/file-upload.md
adamdehaven: IIRC my original comment there was to have a prop that controlled the look of the component between a regular file input and the "drop zone" UX.

adamdehaven: > IIRC my original comment there was to have a prop that controlled the look of the component between a regular file input and the "drop zone" UX.

If this is accomplished with `appearance="dropzone"` then I don't think we need the `allowDragAndDrop` prop 👍🏼

portikM: Removed `allowDragAndDrop` prop

src/components/KFileUpload/KFileUpload.vue
```diff
@@ -64,6 +78,62 @@
         </template>
       </KInput>
     </div>
+    <div
+      v-if="isAppearanceDropzone"
+      class="file-upload-dropzone-wrapper"
+    >
+      <div
+        ref="dropzone"
+        :aria-disabled="disabled ? 'true' : undefined"
```
kaiarrowood, line 87:
```suggestion
        :aria-disabled="disabled ? 'true' : undefined"
```
should it be `'true'` or `true`?

portikM: Doesn't matter, it will still be rendered as a string even if boolean is passed as it's an attribute

PR #2883, opened 2025-08-30
fix(kpop): fix event handlers to allow stopPropagation work outside the popover

Justineo:
# Summary

[KM-1684](https://konghq.atlassian.net/browse/KM-1684)

[Preview PR on Konnect](https://github.com/Kong/konnect-ui-apps/pull/8154)

`<KPop>` previously attached a single `click` event listener on the `document` and determined behavior based on the event’s origin.

This caused an issue when a parent element used `stopPropagation` in its `click` handler, since the global listener never fired, causing the `<KPop>` component inside to stop working.

```vue
<div @click.stop> <!-- This breaks the internal <KPop> -->
  ...
    <KPop>...</KPop>
  ...
</div>
```

We cannot simply switch the global event listener to use the capture phase, as this would break the existing behavior where stopping propagation inside the popover trigger prevents the popover from being toggled.

So this PR separates the logic into three independent handlers, each responsible for its own task.  
In addition, a `capture` flag was added to the click-outside handler so it continues to work regardless of event propagation inside or outside `<KPop>`.

[KM-1684]: https://konghq.atlassian.net/browse/KM-1684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KPop/KPop.vue
+++ b/src/components/KPop/KPop.vue
@@ -1,13 +1,13 @@
 <template>
   <component
     :is="tag"
-    ref="kPopoverElement"
     class="k-popover"
     @keydown.esc.stop="hidePopover"
   >
     <div
       ref="triggerWrapperElement"
       class="popover-trigger-wrapper"
+      @click="triggerClickHandler"
     >
       <slot>
         <KButton
@@ -34,6 +34,7 @@
           role="dialog"
           :style="popoverStyles"
           :x-placement="calculatedPlacement"
+          @click="popoverClickHandler"
         >
           <div
             class="popover-container"
@@ -124,7 +125,6 @@ const slots = defineSlots<PopSlots>()
 
 const popoverId = useId()
 const titleId = useId()
-const kPopoverElement = ref<HTMLElement | null>(null)
 const triggerWrapperElement = ref<HTMLElement | null>(null)
 const popoverElement = ref<HTMLElement | null>(null)
 const isVisible = ref<boolean>(false)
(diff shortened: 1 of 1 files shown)


src/components/KPop/KPop.vue
```diff
@@ -1,13 +1,13 @@
...
   <component
     :is="tag"
-    ref="kPopoverElement"
     class="k-popover"
     @keydown.esc.stop="hidePopover"
   >
     <div
       ref="triggerWrapperElement"
       class="popover-trigger-wrapper"
+      @click="triggerClickHandler"
```
adamdehaven, line 10: If this is now clickable, shouldn't it also have a tabindex? or other aria or role attributes?

src/components/KPop/KPop.vue
```diff
@@ -34,6 +34,7 @@
           role="dialog"
           :style="popoverStyles"
           :x-placement="calculatedPlacement"
+          @click="popoverClickHandler"
```
adamdehaven, line 37: If this is now clickable, shouldn't it also have a tabindex? or other aria or role attributes?

src/components/KPop/KPop.vue
Justineo: The behavior remains the same as before so I don’t think we should change this. We’re still expecting a clickable trigger element inside.

adamdehaven: Sounds good

src/components/KPop/KPop.vue
Justineo: Same

adamdehaven (approved):

PR #2892, opened 2025-09-06
feat(*): support specifying target for dropdown/select/multiselect popovers

Justineo:
# Summary

[KM-1721](https://konghq.atlassian.net/browse/KM-1721)

Currently, the `target` option is not supported in the `kpop-attributes` prop for `<KDropdown>`, `<KSelect>`, and `<KMultiselect>` at the type level.

This PR updates the approach by applying styles to the popover contents within a global `<style>`, ensuring teleportation works as intended.

image

For example, we may need to teleport the dropdown popover to `body` or other top-level container to make the dropdown appear above everything else.

[KM-1721]: https://konghq.atlassian.net/browse/KM-1721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KDropdown/KDropdown.vue
+++ b/src/components/KDropdown/KDropdown.vue
@@ -117,7 +117,7 @@ const tooltipComponent = computed(() => disabledTooltip ? KTooltip : 'div')
 const kPopRef = useTemplateRef('kPop')
 const defaultKPopAttributes: PopoverAttributes = {
   hideCaret: true,
-  popoverClasses: 'dropdown-popover',
+  popoverClasses: 'k-dropdown-popover dropdown-popover',
   popoverTimeout: 0,
   placement: 'bottom-start',
 }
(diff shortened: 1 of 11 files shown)


src/components/KDropdown/KDropdown.vue
```diff
@@ -186,16 +186,21 @@ defineExpose({
...
+// The selector might look unusual, but it’s intentional: keeping the same specificity
+// as before supporting teleportation ensures backward compatibility and avoids style conflicts.
+.k-dropdown-popover.dropdown-popover.popover > .popover-container {
+  border: var(--kui-border-width-10, $kui-border-width-10) solid var(--kui-color-border, $kui-color-border);
+  border-radius: var(--kui-border-radius-30, $kui-border-radius-30);
+  padding: var(--kui-space-0, $kui-space-0);
+
+  .dropdown-list {
+    margin: 0;
+    padding: var(--kui-space-20, $kui-space-20) 0;
```
adamdehaven, line 203: Could we utilize [global selectors](https://vuejs.org/api/sfc-css-features#global-selectors) here instead of needing the separate style block? (Applies to whole PR)

Justineo: `:global` needs to be applied to the whole selector which makes it not very friendly for nested usage.

adamdehaven: I was just suggesting moving your existing changes back into the single scoped style block as-is, but wrapped with :global

Justineo: Yes it’s what I’m replying to 😅

PR #2894, opened 2025-09-08
fix(kmodal): remove body overflow class on unmount

portikM:
# Summary

Rework body scroll lock implementation when KModal is open, make sure it's unlocked when component unmounts

### DIFF
--- a/src/components/KModal/KModal.cy.ts
+++ b/src/components/KModal/KModal.cy.ts
@@ -358,4 +358,18 @@ describe('KModal', () => {
       })
     })
   })
+
+  it('unlocks body scroll when modal is unmounted', () => {
+    cy.mount(KModal, {
+      props: {
+        visible: true,
+      },
+    }).then(() => {
+      cy.get('body').should('have.css', 'overflow', 'hidden').then(() => {
+        Cypress.vueWrapper.unmount()
+
+        cy.get('body').should('have.css', 'overflow', 'visible')
+      })
+    })
+  })
 })


src/components/KModal/KModal.vue
```diff
@@ -238,6 +238,7 @@ watch(() => inputAutofocus, async (inputAutofocus: boolean): Promise<void> => {
 onBeforeUnmount(async () => {
   await toggleFocusTrap(false)
   toggleEventListeners(false)
+  document?.body?.classList?.remove('k-modal-overflow-hidden')
```
adamdehaven: Do we need to first check if any other modal is open?

Justineo:
Shall we move to use VueUse's [`useScrollLock`](https://vueuse.org/core/useScrollLock/) instead?

src/components/KModal/KModal.vue
portikM: Good call, I did not think of that. Reworked my solution a bit - in order to make sure we only remove this KModal instance styles, the class has to be unique.

Justineo:
> I don't think we can, it requires a template ref while we use body element. Am I not seeing something here?

You can directly pass in the body element.

src/components/KModal/KModal.vue
```diff
@@ -132,6 +132,8 @@ const attrs = useAttrs()
 
 const textSelection = useTextSelection()
 
+const isBodyScrollLocked = useScrollLock(document?.body, false)
```
adamdehaven, line 135: issue: @portikM this is breaking SSR in portal. Can you please refactor and check that the preview passes the build in the portal repo?

PR #2898, opened 2025-09-09
feat(kpop): dismiss popover upon mousedown outside instead of click

Justineo:
# Summary

[KM-1734](https://konghq.atlassian.net/browse/KM-1734)

### DIFF
--- a/src/components/KPop/KPop.vue
+++ b/src/components/KPop/KPop.vue
@@ -274,31 +274,29 @@ defineExpose({
 })
 
 onMounted(() => {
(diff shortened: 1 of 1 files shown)


src/components/KPop/KPop.vue
```diff
@@ -274,31 +274,29 @@ defineExpose({
 })
 
 onMounted(() => {
-  if (document) {
```
Justineo, line 277: This check is unnecessary as `onMounted` only runs on the client side.

PR #2916, opened 2025-09-25
fix(kmodal): screen lock ssr issue [KHCP-18120]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-18120

Refactor implementation of scroll lock in KModal as using [useScrollLock](https://vueuse.org/core/useScrollLock/) causes build to fail in SSR

### DIFF
--- a/src/components/KModal/KModal.vue
+++ b/src/components/KModal/KModal.vue
@@ -91,9 +91,9 @@
 </template>
 
 <script lang="ts" setup>
-import { computed, nextTick, onBeforeUnmount, ref, useAttrs, watch } from 'vue'
+import { computed, nextTick, onBeforeUnmount, onMounted, onUnmounted, ref, useAttrs, useId, watch } from 'vue'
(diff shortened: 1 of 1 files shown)


Justineo:
Why not `useScrollLock` in `onMounted`?

src/components/KModal/KModal.vue
```diff
@@ -210,13 +211,13 @@ watch(() => visible, async (visible: boolean): Promise<void> => {
...
     }
-    isBodyScrollLocked.value = true
     toggleEventListeners(true)
   } else {
     await toggleFocusTrap(false)
-    isBodyScrollLocked.value = false
     toggleEventListeners(false)
   }
+
+  document?.body?.classList?.toggle(KMODAL_BODY_OVERFLOW_HIDDEN_CLASS, visible)
```
adamdehaven, line 220: @portikM this may still cause an issue during SSR since it's referencing document immediately. We need to refactor this.

PR #2918, opened 2025-09-26
docs(kcodeblock): add example for shiki

kingyue737:
# Summary

KM-1815

This PR adds a best-practice example for integrating with Shiki. By applying `{ structure: 'inline' }`, we can avoid manually removing the extra wrapper and background that Shiki renders.

### DIFF
--- a/docs/components/codeblock.md
+++ b/docs/components/codeblock.md
@@ -421,13 +421,11 @@ This component has a few shortcuts for interacting with its search and filter fe
 
 ## Syntax highlighting
 
-The `KCodeBlock` component does not have syntax highlighting built in; however, your project can provide its own implementation.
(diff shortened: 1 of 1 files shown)


docs/components/codeblock.md
```diff
@@ -504,6 +501,59 @@ function highlight({ preElement, codeElement, language, code }) {
 </script>
 ```
 
+### Shiki
+
+Below is an integration example for [Shiki](https://shiki.matsu.io/guide/), a beautiful and powerful syntax highlighter.
```
Justineo:
```suggestion
Below is an integration example for [Shiki](http://shiki.style/), a beautiful and powerful syntax highlighter.
```

PR #2921, opened 2025-09-26
fix(ktableview, ktabledata): table preference wathcers

portikM:
# Summary

Add watchers for `tablePreferences` in KTableView and KTableData in case `tablePreferences` are updated later after component has mounted (issue discovered in https://github.com/Kong/konnect-ui-apps/pull/8561 where `tablePreferences` fetched from the API didn't work correctly in tables).


kaiarrowood (dismissed):
LGTM! 🚀

PR #2925, opened 2025-10-02
fix(kdatetimepicker): remove start and end for relative time ranges

filipgutica:
# Summary


src/components/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -180,20 +173,19 @@ const calendarRemountKey = ref<number>(0)
...
  */
-const calendarSingleDate = ref<Date | null>(modelValue.start)
+const calendarSingleDate = ref<Date | null>(modelValue.start || null)
 const calendarRange = ref<TimeRange>(modelValue || defaultTimeRange)
 const calendarVModel = isSingleDatepicker.value
   ? calendarSingleDate as DatePickerModel
   : calendarRange as DatePickerModel
 
-const calendarWrapperRef = useTemplateRef('calendarWrapperRef')
+const calendarWrapper = useTemplateRef('calendarWrapperRef')
```
filipgutica, line 195: Renamed to avoid readonly warning caused when useTemplateRef() variable name matches the template ref name in Vue versions < 3.5.2.
See: https://github.com/vuejs/core/commit/bc63df01992fdbf0b6749ad234153725697ed896 https://github.com/vuejs/core/commit/9b7797d0d1fc773e979e042673d5b9b3151c40fc

(Observed in shared-ui sandbox running Vue 3.5.11)

src/components/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -57,13 +57,6 @@
           ]"
           @click="(selected: string) => state.tabName = selected"
         />
-        <!-- Time range readout -->
```
filipgutica, line 60: Removed time range display since it's not visible because the popover disappears immediately after selecting a relative time range.

src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -104,6 +100,32 @@ const originalTimeValues = ref<{ start: string, end: string }>({
 })
 const componentId = useId()
 
+const initialPage = computed(() => {
```
adorack, line 103: Why do you need to set this now?  Is it fixing a bug or somehow adapting to the other changes you mentioned in the PR?

src/types/date-time-picker.ts
```diff
@@ -17,20 +17,18 @@ export type DatePickerRangeObject = {
...
+  start?: Date | null
+  end?: Date | null
 }
 
 export interface TimePeriod {
   /** Unique identifier */
   key: string
   display: string
   timeframeText: string
-  timeframeLength: () => string
```
adorack, line 31: To avoid a breaking change, I would mark these as optional (`?`) rather than removing them entirely.

src/components/KDateTimePicker/CalendarWrapper.vue
filipgutica: So this kind of fixes a "bug" that was uncovered now because we removed "start" and "end" from relative time ranges. 

When switching the tab to custom from relative, since there's no time stamps, by default, v-calendar initializes the calendar at the "minimum date". 

I put this change to make sure that the initial page the calendar is rendered at is the current month.

src/components/KDateTimePicker/KDateTimePicker.vue
```diff
@@ -261,18 +259,19 @@ const changeCalendarRange = (vCalValue: TimeRange | null): void => {
...
   }
 
-  state.fullRangeDisplay = formatDisplayDate(state.selectedRange, false)
+  if (state.selectedTimeframe.start && state.selectedTimeframe.end) {
+    const start = new Date(state.selectedTimeframe.start())
+    const end = new Date(state.selectedTimeframe.end())
+    state.fullRangeDisplay = formatDisplayDate({ start, end }, false)
+    state.selectedRange.start = start
+    state.selectedRange.end = end
+  }
```
adorack, line 276: There should probably be an `else` block here to set `fullRangeDisplay` back to `''` if the conditions are not met.  Better that it's empty rather than incorrect.

filipgutica: ah, good call.

src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -104,6 +100,32 @@ const originalTimeValues = ref<{ start: string, end: string }>({
...
+  if (props.isRange && calendarVModel.value && 'start' in calendarVModel.value && calendarVModel.value.start instanceof Date) {
+    return { year: calendarVModel.value.start.getFullYear(), month: calendarVModel.value.start.getMonth() + 1 }
+  } else if (calendarVModel.value instanceof Date) {
+    return { year: calendarVModel.value.getFullYear(), month: calendarVModel.value.getMonth() + 1 }
+  } else if (props.maxDate) {
+    return { year: props.maxDate.getFullYear(), month: props.maxDate.getMonth() + 1 }
```
Justineo, line 109: I think we should always prefer current date over `maxDate` unless it's already beyond `maxDate`.

filipgutica: I agree, pushed a change

PR #2929, opened 2025-10-08
fix(ktableview): table actions dropdown teleport [KHCP-18164]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-18164

### DIFF
--- a/src/components/KTableView/KTableView.vue
+++ b/src/components/KTableView/KTableView.vue
@@ -332,7 +332,7 @@
                         ref="actionsDropdown"
                         class="actions-dropdown"
                         data-testid="actions-dropdown"
-                        :kpop-attributes="{ placement: 'bottom-end' }"
+                        :kpop-attributes="{ placement: 'bottom-end', target: 'body' }"
                         @toggle-dropdown="($event: boolean) => onRowActionsToggle(row, $event, cellHelperData)"
                       >
                         <KButton


src/types/popover.ts
```diff
@@ -196,6 +202,7 @@ export interface PopSlots {
 export type PopoverAttributes = Pick<PopProps,
   | 'placement'
   | 'popoverClasses'
+  | 'popoverAttributes'
```
adamdehaven: I'm a bit confused.

The `PopoverAttributes` type has a property that lets you pass in `popoverAttributes`? This seems... odd?

src/components/KTableView/KTableView.vue
```diff
@@ -332,7 +332,7 @@
                         ref="actionsDropdown"
                         class="actions-dropdown"
                         data-testid="actions-dropdown"
-                        :kpop-attributes="{ placement: 'bottom-end' }"
+                        :kpop-attributes="{ placement: 'bottom-end', target: 'body', popoverAttributes: { 'data-testid': rowAttrs(row)['data-testid'] ? `${rowAttrs(row)['data-testid']}-actions-dropdown-popover` : `row-${rowIndex}-actions-dropdown-popover` } }"
```
adamdehaven: suggestion: can you write a small utility function to compute this attribute value instead of all of the inline logic?

src/components/KPop/KPop.vue
```diff
@@ -114,6 +115,7 @@ const {
   maxWidth = 'none',
   maxHeight = 'none',
   popoverClasses = '',
+  popoverAttributes = {},
```
adamdehaven: question: why an empty object rather than `undefined`?

src/components/KPop/KPop.vue
```diff
@@ -26,6 +26,7 @@
       <Transition name="kongponents-fade-transition">
         <div
           v-show="isVisible"
+          v-bind="popoverAttributes"
```
adamdehaven: issue: this seems a bit confusing. If you were to apply actual `PopoverAttributes` here, meaning something like `placement`, `hideCaret`, etc. they would not have the same functionality as if you set these properties on another component and passed them to the popover.

portikM: The prop that maps object properties to KPop props you are talking about is called `kpopAttributes` (`placement`, `hideCaret`, etc). `popoverAttributes` prop can be passed in `kpopAttributes` to set specific attributes on a popover element.

src/components/KPop/KPop.vue
portikM: Just keeping it consistent with other similar props in other components (`kpopAttributes` in KDropdown, KTooltip, KSelect and KMultiselect, `modalAttributes` in KPrompt).

src/components/KTableView/KTableView.vue

src/types/popover.ts
portikM: Yes I realize the naming is weird here but `popoverAttributes` is the only suitable name for this prop as we have another prop `popoverClasses` that binds classes to the same element the new prop binds attributes to.

src/types/popover.ts
```diff
@@ -127,6 +127,12 @@ export interface PopProps {
    */
   popoverClasses?: string
 
+  /**
+   * Additional attributes to be passed to the popover container element.
```
adamdehaven:
```suggestion
   * Attributes to be applied to the popover container element.
```

src/types/popover.ts
```diff
@@ -127,6 +127,12 @@ export interface PopProps {
...
+   * @default {}
+   */
+  popoverAttributes?: Record<string, unknown>
```
adamdehaven: popoverElementAttributes

docs/components/popover.md
```diff
@@ -368,6 +368,23 @@ List of class names you want to assign to `.k-popover` element.
 </KPop>
 ```
 
+### popoverAttributes
```
adamdehaven: rename

src/components/KPop/KPop.vue
adamdehaven: - Can you verify if you pass a class in popoverAttributes and NOT in popoverClasses what happens?
- Can you verify if you pass a class in popoverClasses and ALSO in popoverAttributes what happens? Needs to be added to docs

adamdehaven: and if possible add test(s)

src/types/popover.ts
portikM: Done

src/types/popover.ts
portikM: Done

src/components/KPop/KPop.vue
portikM: Done

docs/components/popover.md
portikM: Done

adamdehaven (approved):
lgtm

PR #2947, opened 2025-10-17
feat(klabel): improve prevent default logic to allow tooltip links

Justineo:
# Summary

[KM-1891]

Only call `preventDefault` if the event target is neither an interactive element nor inside one, since the activation behavior there is defined as a no-op by HTML spec. Otherwise we’ll end up blocking anchor clicks.

[KM-1891]: https://konghq.atlassian.net/browse/KM-1891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KLabel/KLabel.cy.ts
+++ b/src/components/KLabel/KLabel.cy.ts
@@ -88,7 +88,9 @@ describe('KLabel', () => {
             id: 'label-input',
             type: 'checkbox',
           }),
-          'Full Name',
+          h('span', {
+            id: 'label-text',
+          }, 'Full Name'),
         ],
         tooltip: () => h('a', {
           id: 'link',
@@ -106,7 +108,7 @@ describe('KLabel', () => {
     cy.location('hash').should('eq', '#docs-link')
     cy.get('#label-input').should('not.be.checked')
 
-    cy.get('.k-label').click()
-    cy.get('#label-input').should('be.checked')
+    cy.get('#label-text').click()
+    cy.get('input[type="checkbox"]').should('be.checked')
   })
 })
--- a/src/components/KLabel/KLabel.vue
+++ b/src/components/KLabel/KLabel.vue
@@ -54,31 +54,45 @@ const tooltipId = useId()
 
 const labelEl = useTemplateRef('label')
 
-function closestInteractiveElement(element: Element): HTMLElement | null {
-  const interactiveSelectors = [
-    'a[href]',
-    'audio[controls]',
-    'button',
-    'details',
-    'embed',
-    'iframe',
-    'img[usemap]',
-    'input:not([type="hidden"])',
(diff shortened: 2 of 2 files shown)


src/components/KLabel/KLabel.vue
```diff
@@ -50,6 +51,38 @@ const slots = defineSlots<LabelSlots>()
 const hasTooltip = computed((): boolean => !!(help || info || slots.tooltip))
 
 const tooltipId = useId()
+
+const labelEl = useTemplateRef('label')
+
+function closestInteractiveElement(element: Element): HTMLElement | null {
+  const interactiveSelectors = [
```
adamdehaven: What if for example the next closest element is a clickable `div` e.g. with `tabindex="0"` etc.?

Justineo: `<div>`s don't have default behavior on clicks so it doesn't matter.

src/components/KLabel/KLabel.cy.ts
```diff
@@ -79,4 +79,34 @@ describe('KLabel', () => {
 
     cy.get('.k-label').invoke('attr', 'for').should('eq', id)
   })
+
+  it('does not toggle the associated control when clicking the tooltip trigger', () => {
```
adamdehaven, line 83: suggestion: can you add a test for the other path, where clicking a clickable element should toggle the associated control?

src/components/KLabel/KLabel.vue
```diff
@@ -50,6 +51,38 @@ const slots = defineSlots<LabelSlots>()
...
+    'button',
+    'details',
+    'embed',
+    'iframe',
+    'img[usemap]',
+    'input:not([type="hidden"])',
+    'label',
+    'select',
+    'textarea',
+    'video[controls]',
```
adamdehaven: There are other clickable elements:

- `summary`
- `area`
- `object`
- `svg` (with tabindex or embedded `<a>`

Do these need to be accounted for?

src/components/KLabel/KLabel.cy.ts
Justineo: Done.

src/components/KLabel/KLabel.vue
Justineo: There are two types of elements that do not trigger a label’s activation behavior:

1. Interactive content (see list)
2. Elements with their own activation behavior (`a[href]`, `area[href]`, `summary`)

We don’t need to handle elements like `object` or `svg`, and `summary` is already covered by `details`. Since `a` is already considered interactive content, the only case we need to add is `area[href]`.

PR #2950, opened 2025-10-21
feat(k-code-block): add new api for always showing the action buttons

jw-foss:
* Add new API `always-show-action-buttons` to always show the action buttons without needing to hover when on large screens.

# Summary

- Add a new API to allow "action buttons" to always be visible to the users, default to `false`

https://github.com/user-attachments/assets/e723ad33-0e79-4120-b11d-fc5200651034


docs/components/codeblock.md
```diff
@@ -228,6 +228,32 @@ Controls whether to show a copy button which copies the original code (the value
...
 ```
 
+### alwaysShowActionButtons
+
+Action buttons section has following trait:
+
+- Action buttons **will be hidden on larger display** by default and only shown when the user hovers the code block.
+- Action buttons **will be always shown** on smaller displays (mobile devices).
+
+Controls whether to show the action button(s) no matter whether the user has hovered the code block or not. Defaults to `false`. By turning this on, the action buttons will **always** be visible regardless of hover state and the screen size.
```
Leopoldthecoder: nit: asked OpenAI to improve the language
```suggestion
The action buttons behave as follows:

- On larger screens, they are **hidden by default** and appear only when the user hovers over the code block.
- On smaller screens (e.g., mobile devices), they are **always visible**.

This property controls whether the action buttons should be shown at all times, regardless of hover state or screen size. Defaults to `false`. When enabled, the action buttons remain **permanently** visible.
```

kingyue737:
Currently, the copy button is not shown when `alwaysShowActionButtons` is true but `showCopyButton` is set to false. This behavior seems inconsistent with the "always", as it doesn't actually guarantee that the button will always be visible.

What about modifying the existing `showCopyButton` prop? 
```
showCopyButton:  boolean | 'always'
```

docs/components/codeblock.md
```diff
@@ -228,6 +230,30 @@ Controls whether to show a copy button which copies the original code (the value
 />
 ```
 
+#### Always mode
+
+The action buttons behave as follows:
+- On larger screens (breakpoint >= 1280px), they are **hidden by default** and appear only when the user hovers over the code block.
+- On smaller screens (e.g., mobile devices), they are **always visible**.
+This property controls whether the action buttons should be shown at all times, regardless of hover state or screen size. Defaults to `false`. When enabled, the action buttons remain **permanently** visible.
```
kingyue737: Description here also needs to be updated.

jw-foss: Updated!

src/components/KCodeBlock/KCodeBlock.cy.ts
```diff
@@ -65,6 +65,16 @@ describe('KCodeBlock', () => {
     cy.get('[data-testid="code-block-copy-button-code-block"]').should('not.exist')
   })
 
+  it('always show the copy button without needing to hover when props.alwaysShowActionButtons is true', () => {
```
portikM:
```suggestion
  it('always show the copy button without needing to hover when showCopyButton is always', () => {
```

wrong prop name in the test scenario name

src/components/KCodeBlock/KCodeBlock.cy.ts
```diff
@@ -65,6 +65,16 @@ describe('KCodeBlock', () => {
...
+    cy.viewport(1281, 800)
+
+    renderComponent({ id: 'code-block' })
+    cy.get('[data-testid="code-block-copy-button-code-block"]').should('not.be.visible')
```
portikM, line 72: Could you please also add a check here to make sure the button becomes visible when hovering the component?

Justineo: I believe if we are using CSS to show the action buttons on hover than we cannot emulate that with JS (with current Cypress setup).

portikM: Ah true - never mind then.

PR #2953, opened 2025-10-21
feat: add official nuxt integration module for kongponents [KHCP-18358]

arashsheyda:
# Summary

this PR adds support of kongponents for a Nuxt app

### usage:

1. install kongponents
```bash
pnpm i @kong/kongponents
```

2. register the module `'@kong/kongponents/nuxt'` in `nuxt.config`
```ts
export default defineNuxtConfig({
  modules: ['@kong/kongponents/nuxt'],

  kongponents: {
    // optional config
  },
}
```

3. start using the components:
```vue
<template>
	<KButton>Hi</KButton>
</template>
```

ticket: https://konghq.atlassian.net/browse/KHCP-18358

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -76,7 +76,7 @@ jobs:
           uses: ./.github/actions/setup-pnpm-with-dependencies/
 
         - name: Download Build Artifacts
-          uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+          uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
           with:
             name: kongponents-ci-build-output-artifact
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kong/kongponents",
-  "version": "9.42.11",
+  "version": "9.44.0",
   "description": "Kong Component library",
   "type": "module",
   "repository": {
@@ -76,7 +76,7 @@
   },
   "peerDependencies": {
     "vue": ">= 3.5.0 < 4",
-    "vue-router": "^4.6.0"
+    "vue-router": "^4.6.3"
   },
   "devDependencies": {
     "@babel/types": "^7.28.4",
@@ -85,8 +85,8 @@
     "@cypress/vite-dev-server": "^6.0.3",
     "@digitalroute/cz-conventional-changelog-for-jira": "^8.0.1",
     "@evilmartians/lefthook": "^1.13.6",
-    "@kong-ui-public/sandbox-layout": "^2.3.4",
-    "@kong/design-tokens": "^1.18.0",
+    "@kong-ui-public/sandbox-layout": "^2.3.6",
+    "@kong/design-tokens": "^1.18.1",
     "@kong/eslint-config-kong-ui": "1.5.1",
     "@nuxt/kit": "^4.1.3",
     "@semantic-release/changelog": "^6.0.3",
@@ -94,7 +94,7 @@
     "@stylistic/stylelint-plugin": "^4.0.0",
     "@types/inquirer": "^9.0.7",
     "@types/lodash-es": "^4.17.12",
-    "@types/node": "^22.18.10",
+    "@types/node": "^22.18.11",
     "@types/sortablejs": "^1.15.8",
     "@vitejs/plugin-vue": "^5.2.4",
     "@vue/compiler-core": "^3.5.22",
@@ -118,7 +118,7 @@
     "postcss-custom-properties": "^14.0.6",
     "postcss-html": "^1.8.0",
     "rimraf": "^6.0.1",
-    "rollup-plugin-visualizer": "^6.0.4",
+    "rollup-plugin-visualizer": "^6.0.5",
     "sass": "^1.93.2",
     "semantic-release": "^24.2.9",
     "shiki": "^3.13.0",
@@ -130,11 +130,11 @@
     "stylelint-order": "^7.0.0",
     "tsc-alias": "^1.8.16",
     "typescript": "^5.9.3",
-    "vite": "^6.3.6",
+    "vite": "^6.3.7",
     "vite-plugin-vue-devtools": "^7.7.7",
     "vitepress": "^1.6.4",
     "vue": "^3.5.22",
-    "vue-router": "^4.6.0",
+    "vue-router": "^4.6.3",
     "vue-tsc": "^2.2.12"
   },
   "resolutions": {
@@ -194,10 +194,10 @@
     "node": ">=v16.20.2 || >=18.12.1 || >=20.14.0 || >=22.14.0",
     "pnpm": ">=9.11.0 || >=10.10.0"
   },
-  "packageManager": "pnpm@10.18.2",
+  "packageManager": "pnpm@10.18.3",
   "volta": {
     "node": "24.10.0",
-    "pnpm": "10.18.2"
+    "pnpm": "10.18.3"
   },
   "config": {
     "commitizen": {
--- a/.github/actions/setup-pnpm-with-dependencies/action.yaml
+++ b/.github/actions/setup-pnpm-with-dependencies/action.yaml
@@ -37,7 +37,7 @@ runs:
         echo "pnpm-version=${voltaPnpmVersion}">> $GITHUB_OUTPUT
 
     - name: Setup Node
-      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+      uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
       with:
         node-version: ${{ steps.node-version.outputs.node-version }}
         registry-url: https://registry.npmjs.org
@@ -50,7 +50,7 @@ runs:
 
     - name: Dependency Cache
       id: dependency-cache
-      uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+      uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
       with:
         path: |
           ~/.cache/Cypress
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,36 @@
+# [9.44.0](https://github.com/Kong/kongponents/compare/v9.43.2...v9.44.0) (2025-10-28)
+
+
+### Bug Fixes
+
+* **deps:** bump @kong/design-tokens ([#2961](https://github.com/Kong/kongponents/issues/2961)) ([3e2d48c](https://github.com/Kong/kongponents/commit/3e2d48cb5cb8f4602c915e8c12bd13e1375e6ffc))
+
+
+### Features
+
+* **k-code-block:** add new api for always showing the action buttons ([#2950](https://github.com/Kong/kongponents/issues/2950)) ([fe872d7](https://github.com/Kong/kongponents/commit/fe872d7dfe9525e67c1248f0e187d7fbb55e9ef5)), closes [#2951](https://github.com/Kong/kongponents/issues/2951) [#2946](https://github.com/Kong/kongponents/issues/2946) [#2947](https://github.com/Kong/kongponents/issues/2947) [#2947](https://github.com/Kong/kongponents/issues/2947) [#2955](https://github.com/Kong/kongponents/issues/2955) [#2955](https://github.com/Kong/kongponents/issues/2955) [#2952](https://github.com/Kong/kongponents/issues/2952) [#2952](https://github.com/Kong/kongponents/issues/2952) [#2958](https://github.com/Kong/kongponents/issues/2958)
+
+## [9.43.2](https://github.com/Kong/kongponents/compare/v9.43.1...v9.43.2) (2025-10-24)
+
+
+### Bug Fixes
+
+* **ktableview:** overlay should be hidden when no room to scroll right ([#2952](https://github.com/Kong/kongponents/issues/2952)) ([b29072e](https://github.com/Kong/kongponents/commit/b29072e961c197373e00ca6e7b99438712a8f0ec))
+
+## [9.43.1](https://github.com/Kong/kongponents/compare/v9.43.0...v9.43.1) (2025-10-23)
+
+
+### Bug Fixes
+
+* **kmodal:** full screen modal content height [KHCP-18359] ([#2955](https://github.com/Kong/kongponents/issues/2955)) ([9f8169c](https://github.com/Kong/kongponents/commit/9f8169c5bff1987a6d519657131645d4395f6119))
+
+# [9.43.0](https://github.com/Kong/kongponents/compare/v9.42.11...v9.43.0) (2025-10-23)
+
+
+### Features
+
+* **klabel:** improve prevent default logic to allow tooltip links ([#2947](https://github.com/Kong/kongponents/issues/2947)) ([2f2ec42](https://github.com/Kong/kongponents/commit/2f2ec42ca767261ec53c892c1367dedb432fef65))
+
 ## [9.42.11](https://github.com/Kong/kongponents/compare/v9.42.10...v9.42.11) (2025-10-20)
 
 
--- a/docs/components/codeblock.md
+++ b/docs/components/codeblock.md
@@ -210,21 +210,44 @@ Used as the initial value of the search input. Can be used to initialize a code 
 
 ### showCopyButton
 
-Controls whether to show a copy button which copies the original code (the value of `props.code`) to the clipboard. Defaults to `true`.
+Controls whether to show the copy button (along with any other secondary action buttons passed through [`secondary actions` slot](#secondary-actions)) which copies the original code (the value of [`code` prop](#code)) to the clipboard. Defaults to `true`.
+
+The secondary action buttons behave as follows:
+
+- On larger screens (breakpoint >= 1280px), they are **hidden by default** and appear only when the user hovers over the code block.
+- On smaller screens (e.g., mobile devices), they are **always visible**.
 
 <KCodeBlock
+  :show-copy-button="false"
   id="code-block-show-copy-button"
   :code="code"
   language="json"
-  :show-copy-button="false"
 />
 
 ```html
 <KCodeBlock
+  :show-copy-button="false"
   id="code-block-show-copy-button"
   :code="code"
   language="json"
-  :show-copy-button="false"
+/>
+```
+
+Passing `'always'` will render secondary action buttons **always visible**, regardless of screen size or hover state.
+
+<KCodeBlock
+  show-copy-button="always"
+  id="code-block-show-copy-button"
+  :code="code"
+  language="json"
+/>
+
+```html
+<KCodeBlock
+  show-copy-button="always"
+  id="code-block-show-copy-button"
+  :code="code"
+  language="json"
 />
 ```
 
--- a/docs/guide/migrating-to-version-9.md
+++ b/docs/guide/migrating-to-version-9.md
@@ -62,7 +62,7 @@ Kongponents styles are no longer designed to be utilized standalone, separately 
 * `isBordered` prop has been removed
 * `shape` prop has been removed (now KBadge only comes in one what used to be `rectangular` shape)
 * `borderColor` prop has been removed
-* `backgroundColor` prop has beed removed
+* `backgroundColor` prop has been removed
 * `color` prop has been removed
 * `dismissable` prop has been removed. You can now pass a clickable close icon (remember to add `role="button"` and `tabindex="0"` attributes) through the `icon` slot and handle KBadge showing/hiding from your host app
 * `hoverColor` prop has been removed
--- a/lefthook.yaml
+++ b/lefthook.yaml
@@ -1,5 +1,5 @@
 # Reference:
-# https://github.com/evilmartians/lefthook/blob/master/docs/full_guide.md
+# https://lefthook.dev/configuration/
 
 pre-push:
   parallel: true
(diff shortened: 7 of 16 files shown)


src/components/KToaster/ToastManager.ts
```diff
@@ -43,7 +43,7 @@ export default class ToastManager {
   }
 
   setTimer(key: string, timeout: number): number {
-    return setTimeout(() => this.close(key), timeout)
+    return window.setTimeout(() => this.close(key), timeout)
```
arashsheyda: type error, setTimeout type is `NodeTimeout` but `window.setTimeout` type is `number`

tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "importHelpers": true,
     "sourceMap": false,
     "noEmit": false,
-    "baseUrl": ".",
+    "allowImportingTsExtensions": false,
```
arashsheyda, line 10: `baseUrl` is  deprecated in new TS version and `compilerOptions` requires `allowImportingTsExtensions` and `noEmit` to be set

adamdehaven (changes requested):
Just blocking merge until we can verify in host applications.

.github/workflows/publish.yaml
```diff
@@ -80,6 +80,9 @@ jobs:
           with:
             name: kongponents-ci-build-output-artifact
 
+        - name: Build Nuxt package
+          run: pnpm build:nuxt
```
adamdehaven: issue: Shouldn't the singular build step cover building the existing kongponents package as well as the nuxt exports? Otherwise, the package previews in the `test.yaml` job wouldn't even build the nuxt exports into the preview package.

The build flow should follow the existing build flow and pass the assets from the test.yaml job to here. We must build during the test phase just like the regular package to ensure things work _before_ the code is merged to main. This means the nuxt build job should be incorporated into the existing `build:kongponents` script

docs/guide/usage.md
```diff
@@ -31,27 +31,26 @@ app.mount('#app')
...
+// nuxt.config.ts
+export default defineNuxtConfig({
+  modules: ['@kong/kongponents/nuxt'],
+
+  kongponents: {
+    /**
+     * Optional prefix for component names
+     * Default: 'K'
+     */
+    prefix: 'K',
```
adamdehaven: issue: we do not want to support customizing the prefix or component names

nuxt/runtime/plugin.ts
```diff
@@ -0,0 +1,18 @@
+// @ts-ignore - import path exists
+import { defineNuxtPlugin } from '#app'
+
+// @ts-ignore - import path exists
```
adamdehaven: issue: these ignores shouldn't be needed should they?

nuxt/index.ts
```diff
@@ -0,0 +1,71 @@
+import { addPlugin, defineNuxtModule, createResolver, addComponent, useLogger } from '@nuxt/kit'
+
+import { components } from '@kong/kongponents'
+
+export interface ModuleOptions {
+  /**
+   * Prefix to use for registered components
+   * @default 'K'
+   */
+  prefix?: string
```
adamdehaven: Same comment on removing

nuxt/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * List of components to exclude from automatic registration
+   * @default []
+   */
+  exclude?: Array<Exclude<keyof typeof components, 'ToastManager' | 'KTable'>>
+}
+
+export default defineNuxtModule<ModuleOptions>({
+  meta: {
+    name: '@kong/kongponents/nuxt',
```
adamdehaven: question: why would we want the module name to not just be "kongponents"?

nuxt/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    configKey: 'kongponents',
+  },
+  defaults: {
+    prefix: 'K',
```
adamdehaven: remove

nuxt/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  setup(options, nuxt) {
+    const { resolve } = createResolver(import.meta.url)
+    const logger = useLogger()
+
+    // Register the module's plugin (can be used for global styles, etc.)
+    addPlugin(resolve('./runtime/plugin'))
+
+    // Define a blacklist of components that should never be auto-registered.
+    // Includes user-specified exclusions merged with a fixed internal list.
+    const blacklist = ['ToastManager', 'KTable', ...(options.exclude || [])]
```
adamdehaven: The term "blacklist" and "whitelist" have negative connotations. 

Let's use "includeList" and "excludeList" (or something similar) instead.

nuxt/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    Object.entries(components).forEach(([name]) => {
+      if (!name || blacklist.includes(name)) return
+
+      const componentName = `${options.prefix}${name.startsWith('K') ? name.slice(1) : name}`
+
+      addComponent({
+        name: componentName,
+        export: name,
+        filePath: '@kong/kongponents',
+        global: true,
```
adamdehaven: Adding a comment
```suggestion
        // !IMPORTANT: Components must be registered globally
        global: true,
```

nuxt/index.ts
```diff
@@ -0,0 +1,71 @@
...
+        mode: 'all',
```
adamdehaven, line 84: Add a comment here on why it's set to 'all'

nuxt/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    name: '@kong/kongponents/nuxt',
+    configKey: 'kongponents',
+  },
+  defaults: {
+    prefix: 'K',
+    exclude: [],
+  },
+  setup(options, nuxt) {
+    const { resolve } = createResolver(import.meta.url)
+    const logger = useLogger()
```
adamdehaven: I believe you can pass in a prefix here (check the docs please)
```ts
const logger = useLogger('kongponents')
```

package.json
```diff
@@ -30,8 +34,9 @@
     "build:cli": "rimraf ./bin && tsc --project ./cli/tsconfig.json && chmod u+x ./bin/index.js",
     "build:docs": "vitepress build docs",
     "build:kongponents": "pnpm typecheck && vite build && vue-tsc -p './tsconfig.build.json' --emitDeclarationOnly && tsc-alias -p './tsconfig.build.json'",
+    "build:nuxt": "tsc -p ./tsconfig.build.nuxt.json --outDir dist/nuxt",
     "build:visualize": "cross-env-shell BUILD_VISUALIZER=true vite build -m production",
-    "build": "pnpm build:cli && pnpm stylelint && pnpm lint && pnpm build:kongponents && pnpm build:docs",
+    "build": "pnpm build:cli && pnpm stylelint && pnpm lint && pnpm build:kongponents && pnpm build:nuxt && pnpm build:docs",
```
adamdehaven: refactor: the nuxt build should be incorporated into the `build:kongponents` command instead

tsconfig.json
adamdehaven: > in new TS version

I'm not sure I understand; it doesn't appear you're bumping TS in this PR

.github/workflows/publish.yaml
arashsheyda: yeah I didn't spend timing checking the workflows, I needed to add this here to include the nuxt build in preview

docs/guide/usage.md
arashsheyda: :sad: why don't we want give the user the ability to customize our lib?

nuxt/runtime/plugin.ts
arashsheyda: they were needed otherwise the build would fail

nuxt/index.ts
arashsheyda: it could be, but it's better to be more unique (if one day someone else has a module named `kongponents` things will crash)

portikM: IMO `@kong/kongponents` is most intuitive

tsconfig.json
arashsheyda: yes, but if you're TS locally is different the build would fail. and I saw this as an opportunity to update (since this would work in other TS versions as well)

nuxt/index.ts
portikM: There are more deprecated components that should never be auto-registered: `KModalFullscreen`, `KDropdownMenu`

src/components/KToaster/ToastManager.ts
portikM:
```suggestion
    return window?.setTimeout(() => this.close(key), timeout)
```

docs/guide/usage.md
adamdehaven: - All of our documentation utilizes `K` prefix
- If they end up here from Portal, they will only be exposed as `K` prefixed
- I don't think it makes sense to make something more complex that will likely only be utilized by us internally. It's better to keep "speaking the same language"
- I could argue it's also part of our branding

.github/workflows/publish.yaml
adamdehaven: I can pair with you on this tomorrow if you're unsure -- happy to give you more context

nuxt/runtime/plugin.ts
adamdehaven: You should be able to augment the types then so it thinks they exist

nuxt/index.ts
adamdehaven: What are the result(s) of the module name here?

src/components/KToaster/ToastManager.ts
arashsheyda: but if it was optional it means it could return undefined so not a number ?

.github/workflows/publish.yaml
arashsheyda: yeah thanks! I think including the `build:nuxt` in `build:kongponents` solved the issue but a call would be nice just to over the stuff togeather

docs/guide/usage.md
arashsheyda: Ah yes, I've removed it thanks

.github/workflows/publish.yaml
adamdehaven: just make sure when multiple builds run it doesn't clear out the dist assets of the other one

nuxt/index.ts
arashsheyda: ToastManager would break the module, that's why it's excluded. and I only saw KTable deprecated in the docs sidebar. could you give me the list tomorrow?

adamdehaven: `ToastManager` has to be initialized as we do in portal separately. It would be nice to add this example to the Nuxt docs. See this file in the `Kong/portal` repo: `layers/core/app/composables/useToast.ts`

nuxt/index.ts
arashsheyda: this doesn't do anything with the module name, this meta.name is just for nuxt(for example in the devtools module section, the module name comes from this meta.name). the name of the module is in package.json + directory of dist (nuxt). is this what you're asking?

nuxt/index.ts
arashsheyda: Ah in that case we can include that composable here. thanks

nuxt/index.ts
adamdehaven: If for display purposes like in devtools, etc. it should be "Kongponents" not the npm package name, right?

.github/workflows/publish.yaml
arashsheyda: `build:nuxt` only clears itself, but `build:kongponents` clear all, including `build:nuxt` after `build:kongponents` should be fine

nuxt/index.ts
adamdehaven: You mean in the docs, right? or as part of the module?

nuxt/index.ts
arashsheyda: it's not just for display

> Using a unique and descriptive name is recommended, especially if you plan to publish your module or use it across multiple projects. The name is also required for certain features, such as lifecycle hooks like onInstall and onUpgrade [Module Definition](https://nuxt.com/docs/4.x/guide/going-further/modules#module-definition).

adamdehaven: There we go. 

I just kept commenting so you'd link the docs 😉

nuxt/index.ts
arashsheyda: as part of the module. we could have the `useToast` composable which makes sense to me (and document it in the Nuxt usage?)

adamdehaven: Works for me

docs/guide/usage.md
```diff
@@ -31,27 +31,26 @@ app.mount('#app')
...
-export default defineNuxtPlugin({
-  name: 'kongponents',
-  setup(nuxtApp) {
-    // Inject the crypto module into the global scope if it is not already available
-    if (import.meta.server && typeof globalThis?.crypto === 'undefined') {
-      globalThis.crypto = globalThis.crypto || crypto
-    }
-    // Initialize the Kongponents Vue plugin
-    nuxtApp.vueApp.use(Kongponents)
+Kongponents now provides a first-class Nuxt module for seamless integration and automatic component registration.
```
adamdehaven:
```suggestion
Kongponents provides a first-class [Nuxt module](https://nuxt.com/docs/4.x/guide/concepts/modules) for seamless integration and automatic component registration.
```

docs/guide/usage.md
adamdehaven: Please remove this prefix setting

docs/guide/usage.md
```diff
@@ -31,27 +31,26 @@ app.mount('#app')
...
+    /**
+     * Optional prefix for component names
+     * Default: 'K'
+     */
+    prefix: 'K',
+
+    /**
+     * Optional list of components to exclude from auto-registration
+     */
+    exclude: [],
```
adamdehaven: suggestion: could we change this to an object interface:

```ts
components: {
  /**
   * Optional list of components to include in auto-registration. If unset or empty, all components will be included.
   */
  include: [],
  /**
   * Optional list of components to exclude from auto-registration
   */
  exclude: [],
}
```

nuxt/index.ts
```diff
@@ -0,0 +1,65 @@
+import { addPlugin, defineNuxtModule, createResolver, addComponent, useLogger } from '@nuxt/kit'
+
+import { components } from '@kong/kongponents'
+
+export interface ModuleOptions {
+  /**
+   * List of components to exclude from automatic registration
+   * @default []
+   */
+  exclude?: Array<Exclude<keyof typeof components, 'ToastManager' | 'KTable'>>
```
adamdehaven: suggestion: move the excluded component keys to another type instead of inline

adamdehaven (changes requested):

docs/guide/usage.md
arashsheyda: so if I had `include: ['KButton']` and `exclude: ['KButton']` we want to exclude it right?

ps: I feel like these two kinda serve the same thing... so maybe we should just have `exclude`

adamdehaven: > I feel like these two kinda serve the same thing... so maybe we should just have `exclude`

What if a user only wants to register `KAlert` and `KButton`? They'd have to find the name of every other component to add to the `exclude` array

nuxt/index.ts
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * List of components to exclude from automatic registration
+     * @default []
+     */
+    exclude?: ExcludedComponentKeys[]
+  }
+}
+
+// Components that are deprecated or not meant for auto-registration
+const DEPRECATED_COMPONENTS = ['ToastManager', 'KTable']
```
portikM:
```suggestion
const DEPRECATED_COMPONENTS = ['ToastManager', 'KTable', 'KModalFullScreen', 'KDropdownMenu']
```

issue: full list of deprecated components.. also since `ToastManager` is not deprecated, should this be named differently?

nuxt/index.ts
```diff
@@ -0,0 +1,101 @@
+import { addPlugin, defineNuxtModule, createResolver, addComponent, useLogger, addImportsDir } from '@nuxt/kit'
```
adamdehaven: you have an unused import here

nuxt/index.ts
```diff
@@ -0,0 +1,101 @@
...
+  },
+  setup(options, nuxt) {
+    const { resolve } = createResolver(import.meta.url)
+    const logger = useLogger('kongponents')
+
+    // Register the styles
+    nuxt.options.css.push('@kong/kongponents/dist/style.css')
+
+    // Register the module's plugin (can be used for global styles, etc.)
+    // addPlugin(resolve('./runtime/plugins/kongponents'))
```
adamdehaven: should this still be commented out?

nuxt/index.ts
arashsheyda: testing something

nuxt/index.ts
arashsheyda: testing something

docs/guide/usage.md
```diff
@@ -56,6 +60,37 @@ export default defineNuxtPlugin({
...
+
+```ts [nuxt.config.ts]
+// nuxt.config.ts
+export default defineNuxtConfig({
+  modules: ['@kong/kongponents/nuxt'],
+
+  kongponents: {
+    components: {
+      /**
+       * Optional list of components to include in auto-registration. If unset or empty, all components will be included.
```
adamdehaven: nit: break this into multiple lines so it's readable on the docs site
```suggestion
       * Optional list of component names to include in auto-registration. 
       * If unset or empty, all components will be included.
```

docs/guide/usage.md
```diff
@@ -56,6 +60,37 @@ export default defineNuxtPlugin({
...
+       */
+      include: [],
+      /**
+       * Optional list of components to exclude from auto-registration
+       */
+      exclude: [],
+    },
+    /**
+     * Whether to register composables globally
+     * you can use composables like `useToast`
```
adamdehaven: nit
```suggestion
     * Whether to register provided composables.
     * For example, you can access the included `useToast` composable.
```

docs/guide/usage.md
```diff
@@ -56,6 +60,37 @@ export default defineNuxtPlugin({
...
+  modules: ['@kong/kongponents/nuxt'],
+
+  kongponents: {
+    components: {
+      /**
+       * Optional list of components to include in auto-registration. If unset or empty, all components will be included.
+       */
+      include: [],
+      /**
+       * Optional list of components to exclude from auto-registration
```
adamdehaven:
```suggestion
       * Optional list of component names to exclude from auto-registration
```

docs/guide/usage.md
```diff
@@ -56,6 +60,37 @@ export default defineNuxtPlugin({
...
+    },
+    /**
+     * Whether to register composables globally
+     * you can use composables like `useToast`
+     */
+    composables: true,
+  },
+})
+```
+
```
adamdehaven, line 106: In this same Nuxt documentation section, we should:

1. Document that components will be auto-imported by Nuxt
2. Document the `useToast` composable's functionality, and how to use it

nuxt/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    // Define a list of components that should never be auto-registered.
+    const excludeList = [...ALWAYS_EXCLUDE_COMPONENTS, ...(options.components?.exclude || [])]
+
+    const filteredComponents = Object.keys(components).filter((name) => {
+      // If include list is set, only register those
+      if (includeList.length > 0 && !includeList.includes(name as ComponentKeys)) return false
+      // Skip excluded or deprecated components
+      if (excludeList.includes(name)) return false
+      return true
+    })
```
adamdehaven: suggestion: just as a safety precaution for future components, should we do a check here like `if (name.startsWith('K')`

nuxt/index.ts
```diff
@@ -0,0 +1,103 @@
+import { defineNuxtModule, createResolver, addComponent, useLogger, addImportsDir } from '@nuxt/kit'
+import { components } from '@kong/kongponents'
```
adamdehaven, line 2: I think this is correct, but asking: should this be imported from here or the local files? Same for other imports from this same repo

arashsheyda: I don't think there's a difference, as this is like an alias. I went with this because it's cleaner. if we were to import from local files the entry would start from `~/dist/nuxt/index.js` meaning it would look like something like `../kongponents.es.js` or '../types/components/' and it could be confusing because we're in `~/nuxt/index.ts` there's no such a path

docs/guide/usage.md
```diff
@@ -56,6 +72,68 @@ export default defineNuxtPlugin({
...
+### composables
+
+If you enable composables in the module options, you can access utilities like useToast directly in your Nuxt app.
+
+#### UseToast
```
adamdehaven:
```suggestion
#### useToast
```

docs/guide/usage.md
```diff
@@ -56,6 +72,68 @@ export default defineNuxtPlugin({
...
+
+| Option               | Type                                | Description                                    |
+|----------------------|-------------------------------------|------------------------------------------------|
+| `key`                | `string`                            | Unique identifier for the toast                |
+| `title`              | `string`                            | Title text displayed above the message         |
+| `message`            | `string`                            | The main content of the toast                  |
+| `appearance`         | `'info' \| 'success' \| 'warning' \| 'error'` | The visual style of the toast        |
+| `timeoutMilliseconds`| `number`                            | Duration before the toast automatically closes |
+| `timer`              | `number`                            | Optional delay before the toast appears        |
+
```
adamdehaven: Let's add a typescript interface with comments here instead of a table

nuxt/index.ts
```diff
@@ -0,0 +1,107 @@
...
+      // If an include list is provided, only keep those in the list
+      && (includeList.length === 0 || includeList.includes(name as ComponentKeys))
+      // Exclude any components explicitly listed in the exclude list
+      && !excludeList.includes(name),
+    )
+
+
+    if (filteredComponents.length === 0) {
+      logger.warn('⚠️ No Kongponents components were registered. Check your include/exclude config.')
+    }
```
adamdehaven, line 74: if this evaluates to true, shouldn't we skip all of the logic below?

arashsheyda: we could, the loop will return early because of `0` length and I didn't return because wanted to still see the devtools tab anyways (since it's just a documentation)

adamdehaven: ok

portikM:
Functionally all components look and behave correctly in sandbox and docs previews.

PR #2956, opened 2025-10-24
fix(ktableview): update overlay visibility when the table is resized

kingyue737:
# Summary

[KM-1906]

This PR fixes that the scrollable overlay is not updated when the table is resized.

### Before

https://github.com/user-attachments/assets/d4dbe5eb-b358-464e-85a4-8e330118654d

[KM-1906]: https://konghq.atlassian.net/browse/KM-1906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KTableView/KTableView.vue
+++ b/src/components/KTableView/KTableView.vue
@@ -118,6 +118,7 @@
         @scroll.passive="scrollHandler"
       >
         <table
+          ref="table"
           class="table"
           :class="{
             'has-hover': rowHover && !isActionsDropdownHovered,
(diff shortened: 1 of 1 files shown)


src/components/KTableView/KTableView.vue
```diff
@@ -1384,8 +1386,8 @@ watch(() => tablePreferences, (newVal) => {
   columnVisibility.value = newVal?.columnVisibility ? newVal.columnVisibility : columnVisibility.value
 })
 
-useResizeObserver(tableWrapperRef, (entries) => {
-  const el = entries[0]?.target
+useResizeObserver([tableWrapperRef, tableRef], () => {
+  const el = tableWrapperRef.value
```
adamdehaven, line 1390: You should still reference the element from the watch params

kingyue737: The params array won’t include the `tableWrapper` element if only the `table` itself changes size.

PR #2962, opened 2025-10-28
fix(ktableview): 3 state sort interface [KHCP-18152]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-18152

Currently once user sorts table data by any column, there is no way to reset sorting in the table. Implement a 3-state sorting interface: clicking on the column heading will sort it in ascending order, clicking for the second time will switch order to descending, clicking for the third time will reset the sorting.

Functional testing: the very first example in [KTableView sandbox](https://deploy-preview-2962--kongponents-sandbox.netlify.app/#/table-view) allows toggling initial sorting on and off to test sorting behaviour.

### DIFF
--- a/docs/components/table-data.md
+++ b/docs/components/table-data.md
@@ -153,6 +153,12 @@ const sortHandlerFunction = ({ key, sortColumnOrder, data }) => {
 
 :::tip NOTE
 Notice in the example above the _Username_ column is `sortable`. In this example sorting is handled client-side; however, should you want to handle sorting differently (e.g. make a call to the back-end to apply server-side logic), you can bind your logic to the [`sort` event](#sort) and perform re-fetching of data as needed.
+
+Sorting cycles through three states: ascending, descending, and unset.
+
+1. The first click on a sortable column header sorts the table data in ascending order.
+2. The second click reverses the order (descending).
+3. The third click removes sorting from that column.
 :::
 
 ### initialFetcherParams
--- a/docs/components/table-view.md
+++ b/docs/components/table-view.md
@@ -333,6 +333,12 @@ const sortData = (sortData: TableSortPayload): void => {
 
 :::tip NOTE
 Notice that in the example above the _Username_ column is `sortable` and the _Email_ column is `hidable`.
+
+Sorting cycles through three states: ascending, descending, and unset.
+
+1. The first click on a sortable column header sorts the table data in ascending order.
+2. The second click reverses the order (descending).
+3. The third click removes sorting from that column.
 :::
 
 ### loading


src/types/table.ts
```diff
@@ -106,7 +106,7 @@ export type TableState = 'loading' | 'error' | 'success' | 'empty'
 export interface TableSortPayload<Key extends string = string> {
   prevKey: Key | ''
   sortColumnKey: Key | ''
-  sortColumnOrder: SortColumnOrder
+  sortColumnOrder: SortColumnOrder | ''
```
Justineo: I'd prefer to use `undefined` here but as we are already using`''` for keys I'm ok with this...

BTW. This is technically a breaking change on types so I think we need to at least create a preview PR on Konnect.

src/components/KTableView/KTableView.vue
```diff
@@ -797,16 +798,28 @@ const getHeaderClasses = (column: TableViewHeader<ColumnKey>, index: number): Re
...
-    let newSortColumnOrder: SortColumnOrder = 'asc'
+    let newSortColumnOrder: SortColumnOrder | '' = 'asc'
+    let newSortColumnKey: ColumnKey | '' = column.key
     if (column.key === sortColumnKey.value && sortColumnOrder.value === 'asc') {
       newSortColumnOrder = 'desc'
     }
+    previousSortKeys.value.unshift(column.key as TableReservedColumnKey)
+    if (previousSortKeys.value.length === 3) {
+      if (previousSortKeys.value[0] === previousSortKeys.value[1] && previousSortKeys.value[1] === previousSortKeys.value[2]) {
+        // user has clicked the same column 3 times in a row, reset sort
```
Justineo: We should determine the next sort state based on three factors: `sortColumnKey`, `sortColumnOrder`, and `column.key`.  
Maintaining a separate `previousSortKeys` array that relies solely on the sequence of user interactions is fragile, since `sortColumnKey` and `sortColumnOrder` may change when `tablePreferences` updates. This can cause the array to become out of sync with the actual state.

portikM: We don't really care about the actual state - it's the opposite, we only care about user interactions. The idea is that regardless of the initial state of sorting in the table, user can reset sorting state within 3 clicks. If the table isn't sorted initially, user can click on the column once to sort in ascending order, then second time to sort in descending order and then third time to reset sorting. Likewise if initial sorting for this table is set through table preferences, it will take user 3 clicks to reset sorting (e.g. a column sorted in ascending order initially: 1 click -> descending, 2 click ascending, 3 click - reset).

For context, we need this because we save sorting in table preference once it's applied once but after that users don't have a way to unset sorting since we persist table preferences.

src/types/table.ts
portikM: I realized there is no real reason why we would want to reset `sortColumnOrder` and instead made it go back to default (`'desc'`). This way we don't need to change any types. I picked up the change in Konnect anyways just to make sure it doesn't break anything: https://github.com/Kong/konnect-ui-apps/pull/9010.

src/components/KTableView/KTableView.vue
```diff
@@ -595,8 +595,10 @@ const isScrollableVertically = ref<boolean>(false)
 const isScrolledVertically = ref<boolean>(false)
 const isScrolledHorizontally = ref<boolean>(false)
 const isScrollableRight = ref<boolean>(false)
+const previousSortKeys = ref<ColumnKey[]>([])
 const sortColumnKey = ref(tablePreferences.sortColumnKey || '') as Ref<ColumnKey | ''>
-const sortColumnOrder = ref<SortColumnOrder>(tablePreferences.sortColumnOrder || 'desc')
+const DEFAULT_SORT_ORDER: SortColumnOrder = 'desc'
```
adamdehaven: A default of `desc` seems counter intuitive.

If we were to sort alphabetically, descending order would be Z-A instead of A-Z

src/components/KTableView/KTableView.vue
```diff
@@ -798,15 +800,27 @@ const getHeaderClasses = (column: TableViewHeader<ColumnKey>, index: number): Re
 const onHeaderClick = (column: TableViewHeader<ColumnKey>) => {
   if (column.sortable && column.key !== TableViewHeaderKeys.BULK_ACTIONS && column.key !== TableViewHeaderKeys.ACTIONS) {
     let newSortColumnOrder: SortColumnOrder = 'asc'
+    let newSortColumnKey: ColumnKey | '' = column.key
     if (column.key === sortColumnKey.value && sortColumnOrder.value === 'asc') {
       newSortColumnOrder = 'desc'
     }
+    previousSortKeys.value.unshift(column.key as TableReservedColumnKey)
+    if (previousSortKeys.value.length === 3) {
+      if (previousSortKeys.value[0] === previousSortKeys.value[1] && previousSortKeys.value[1] === previousSortKeys.value[2]) {
```
adamdehaven: I'm not sure I understand tracking the state by "number of clicks" -- could there just be a persistent sort state that cycles through the three states?

src/components/KTableView/KTableView.vue
```diff
@@ -1015,23 +1029,24 @@ watch(() => headers, (newVal: readonly Header[]) => {
 
 const isColumnSortable = (column: TableViewHeader<ColumnKey>): boolean => !column.hideLabel && !!column.sortable && column.key !== TableViewHeaderKeys.BULK_ACTIONS && column.key !== TableViewHeaderKeys.ACTIONS
 
-const sortClickHandler = (header: TableViewHeader<ColumnKey>): void => {
-  const { key } = header
-
-  if (sortColumnKey.value) {
-    if (key === sortColumnKey.value) {
+const sortClickHandler = (columnKey: ColumnKey | ''): void => {
```
adamdehaven, line 1028: it's hard to tell what's going on in this method now with all of the changes. Can you please add some comments to help with reviewing?

src/components/KTableView/KTableView.vue
Justineo: The problem isn’t the interaction itself but how it’s implemented.  
Using `previousSortKeys` relies on click history instead of the actual sort state and the currently clicked column.  
When `tablePreferences` change the sort, this history goes out of sync and can trigger an incorrect “third click reset.”  
It should be based on the current `sortColumnKey`, `sortColumnOrder`, and the clicked column instead.

Justineo: An example:

| Action | Result (current) | `previousSortKeys` | Result (expected) |
| -- | -- | -- | -- |
| Initial | — | [ ] | — |
| Click A | A↑ | [A] | A↑ |
| Click A | A↓ | [A, A] | A↓ |
| Set B ↑ | B↑ | [A, A] | B↑ |
| Click A | — | [A, A, A] → [ ] | A↑ |

The last click on A directly reset the order state.

src/components/KTableView/KTableView.vue
portikM: You're thinking about this wrong. The sorting order is toggled each time user clicks sort, between 2 states. However initially (while sort order is in default, "idle" state - `'desc'`) sort column is not set, so very first time user applies sorting, the sorting key is set and sorting order is toggled to `'asc'`. So the sequence of sorting order states is always ascending -> descending, which is intuitive.

src/components/KTableView/KTableView.vue
portikM: Added some comments here - lmk if you think I should clarify it even more

src/components/KTableView/KTableView.vue
adamdehaven: So does that mean this `DEFAULT_SORT_ORDER` is like, "when we change it the first time after a click, change it to `DEFAULT_SORT_ORDER`"?

src/components/KTableView/KTableView.vue
portikM: After some more thought I agree. Updated my implementation so that sorting state comes in sequence ascending -> descending -> none.

src/components/KTableView/KTableView.vue
portikM: After some more thought, I reworked my implementation so that sorting state comes in such sequence: ascending -> descending -> none. This goes along with your suggestion to stir away from relaying on additional ref tracking user interactions (got rid of `previousSortKeys`; new implementation determines the next sort state based on three factors: `sortColumnKey`, `sortColumnOrder`, and `column.key`). This also makes the result of next interaction with sorting interface more predictable: after descending order always comes reset when sorting by the same column. Please let me know if is better.

src/components/KTableView/KTableView.vue
portikM: No, `DEFAULT_SORT_ORDER` means "`DEFAULT_SORT_ORDER` is initial state but when we change it first time after a click, we go to next state in sequence after `DEFAULT_SORT_ORDER`". It's weird, I know, but that's how it's been since v8 so I've learned to live with this logic.

adamdehaven: 🙃 seems confusing tbh

portikM: That's cause it is. But we couldn't simplify this without introducing a breaking change in types. Ideally we'd want an unset state for sorting order - probably `undefined` or `null`. But currently it's `'asc' | 'desc'` and I don't feel like going that deep to untangle this lol

src/components/KTableView/KTableView.cy.ts
```diff
@@ -490,13 +490,49 @@ describe('KTableView', () => {
...
+          cy.wrap(Cypress.vueWrapper.emitted()).should('have.property', 'sort').and('have.length', 2)
+          cy.wrap(Cypress.vueWrapper.emitted('sort')?.[1]?.[0]).should('deep.equal', { prevKey: sortableColumnKey, sortColumnKey: sortableColumnKey, sortColumnOrder: 'desc' })
+
+          cy.getTestId(`table-header-${sortableColumnKey}`).click().then(() => {
+            cy.getTestId(`table-header-${sortableColumnKey}`).should('not.have.class', 'active-sort')
+            cy.wrap(Cypress.vueWrapper.emitted()).should('have.property', 'sort').and('have.length', 3)
+            cy.wrap(Cypress.vueWrapper.emitted('sort')?.[2]?.[0]).should('deep.equal', { prevKey: sortableColumnKey, sortColumnKey: '', sortColumnOrder: 'desc' })
+          })
+        })
+      })
```
Justineo, line 535: Can we use async/await here?

src/components/KTableView/KTableView.vue
```diff
@@ -798,15 +799,23 @@ const getHeaderClasses = (column: TableViewHeader<ColumnKey>, index: number): Re
...
+    }
+
     emit('sort', {
       prevKey: sortColumnKey.value,
-      sortColumnKey: column.key,
+      sortColumnKey: newSortColumnKey,
       sortColumnOrder: newSortColumnOrder,
     })
-    sortClickHandler(column)
+    sortClickHandler(newSortColumnKey)
```
Justineo: It’s not very clear to me why we need a separate `sortClickHandler` inside the header click handler, since it’s essentially serving as the sort handler itself. But since this wasn’t introduced in this PR, I’m fine with leaving it as is for now. I do think it would be better to extract a pure function that determines the next sort state based on the previous one. That way, the click handler could be much simpler. The current code looks good overall, but I’d suggest refactoring it later.

src/components/KTableView/KTableView.cy.ts
Justineo: We should also include [the example from the review](https://github.com/Kong/kongponents/pull/2962#discussion_r2481485866) to make sure it doesn't reset the sort state by accident.

src/components/KTableView/KTableView.vue
portikM: Yes, I agree that this function is redundant (all logic this function handles could be handled inside `onHeaderClick` function itself) but this was here before and I don't feel like refactoring it in this PR. Added a comment to refactor this later.

src/components/KTableView/KTableView.cy.ts
portikM: I'm not sure whether it's possible to use async/await syntax here.

Added a test scenario ensuring the correct sorting state order when switching sort columns like in your example.

Justineo: Oh I forgot we are using Cypress 😅

In theory, Cypress queues all commands automatically, so we usually don’t need to use `then` in most cases.

adamdehaven (changes requested):
If you look at the first example [here](https://deploy-preview-2962--kongponents-sandbox.netlify.app/#/table-view) on the sandbox.

- first table is default sorted by username
- if I click to change the sort order, it unsorts instead of reversing the sort order

The proper behavior would only be to unsort after going through the other two valid sort states

PR #2966, opened 2025-10-31
feat(packaging): update build config to preserve modules for chunked output [KHCP-18457]

arashsheyda:
# Summary

this PR updates the Vite build config to preserve modules for chunked output + register components globally in Nuxt module

the reason being when the Nuxt module has global components, the builded sourcemap file gets too big if components are not chunked

| Before | After |
|--------|--------|
| Screenshot 2025-10-31 at 11 50 20 AM | Screenshot 2025-10-31 at 10 40 00 AM |

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "build:cli": "rimraf ./bin && tsc --project ./cli/tsconfig.json && chmod u+x ./bin/index.js",
     "build:docs": "vitepress build docs",
     "build:kongponents:umd": "BUILD_UMD=true vite build",
-    "build:kongponents": "pnpm typecheck && vite build && vue-tsc -p './tsconfig.build.json' --emitDeclarationOnly && tsc-alias -p './tsconfig.build.json' && build:kongponents:umd && pnpm build:nuxt",
+    "build:kongponents": "pnpm typecheck && vite build && vue-tsc -p './tsconfig.build.json' --emitDeclarationOnly && tsc-alias -p './tsconfig.build.json' && pnpm build:kongponents:umd && pnpm build:nuxt",
     "build:nuxt": "tsc -p ./tsconfig.build.nuxt.json --outDir dist/nuxt",
     "build:visualize": "cross-env-shell BUILD_VISUALIZER=true vite build -m production",
     "build": "pnpm build:cli && pnpm stylelint && pnpm lint && pnpm build:kongponents && pnpm build:docs",


nuxt/index.ts
```diff
@@ -78,8 +78,8 @@ export default defineNuxtModule<ModuleOptions>({
         name,
         export: name,
         filePath: '@kong/kongponents',
-        // we don't need global registration here
-        global: false,
+        //!IMPORTANT: we must register components as global
```
adamdehaven: nit: explain why

package.json
```diff
@@ -8,7 +8,7 @@
     "url": "https://github.com/Kong/kongponents.git"
   },
   "homepage": "https://kongponents.konghq.com/",
-  "main": "./dist/kongponents.umd.js",
+  "main": "./dist/kongponents.cjs",
```
adamdehaven, line 11: issue: this would be a breaking change

package.json
```diff
@@ -20,7 +20,7 @@
     ".": {
       "types": "./dist/types/index.d.ts",
       "import": "./dist/kongponents.es.js",
-      "require": "./dist/kongponents.umd.js"
+      "require": "./dist/kongponents.cjs"
```
adamdehaven, line 23: issue: this would be a breaking change

vite.config.ts
```diff
@@ -50,7 +50,14 @@ export default defineConfig({
...
         entry: path.resolve(__dirname, 'src/index.ts'),
         name: 'Kongponents',
-        fileName: (format) => `kongponents.${format}.js`,
+        fileName: (format) => {
+          if (format === 'cjs') {
+            return 'kongponents.cjs'
+          } else {
+            return `kongponents.${format}.js`
+          }
+        },
```
adamdehaven, line 64: issue: this would be a breaking change

arashsheyda: do you mean we don't want this or use `!` in the commit 😄 ?

adamdehaven: We cannot change the exported file names/extensions.

arashsheyda: UMD and IIFE output formats are not supported for code-splitting builds

adamdehaven: Then you can add a _new_ format, but cannot change the existing

adamdehaven: Or... use the new format, and replace the existing filename but only if it is non-breaking for any existing supported usage pattern

package.json
adamdehaven, line 11: https://jsfiddle.net/r672om8t/

issue: The CDN example on the docs site no longer works

nuxt/index.ts
```diff
@@ -78,8 +78,8 @@ export default defineNuxtModule<ModuleOptions>({
         name,
         export: name,
         filePath: '@kong/kongponents',
-        // we don't need global registration here
-        global: false,
+        //!IMPORTANT: we must register components as global because the nuxt app might be using packages that rely on kongponents being global (e.g., @nuxt/content, @kong-ui-public/analytics-chart, ...)
```
adamdehaven:
```suggestion
        //!IMPORTANT: we must register components as global because the Nuxt app might be using packages that rely on Kongponents being globally available
```

vite.config.ts
```diff
@@ -64,7 +76,11 @@ export default defineConfig({
             'vue-router': 'VueRouter',
           },
         exports: 'named',
+        preserveModules: isUMDBuild ? false : true,
+        preserveModulesRoot: isUMDBuild ? undefined : 'src',
+        inlineDynamicImports: isUMDBuild ? true : false,
       },
+      preserveEntrySignatures: 'strict',
```
adamdehaven, line 84: Can you explain in the PR comments here why this is needed?

package.json
```diff
@@ -33,7 +33,8 @@
   "scripts": {
     "build:cli": "rimraf ./bin && tsc --project ./cli/tsconfig.json && chmod u+x ./bin/index.js",
     "build:docs": "vitepress build docs",
-    "build:kongponents": "pnpm typecheck && vite build && vue-tsc -p './tsconfig.build.json' --emitDeclarationOnly && tsc-alias -p './tsconfig.build.json' && pnpm build:nuxt",
+    "build:kongponents:umd": "pnpm typecheck && BUILD_UMD=true vite build && vue-tsc -p './tsconfig.build.json' --emitDeclarationOnly && tsc-alias -p './tsconfig.build.json'",
```
adamdehaven: We shouldn't need to build all the types again, etc. Will something like this work instead?
```suggestion
    "build:kongponents:umd": "BUILD_UMD=true vite build",
```

package.json
```diff
@@ -33,7 +33,8 @@
...
+    "build:kongponents": "pnpm typecheck && vite build && vue-tsc -p './tsconfig.build.json' --emitDeclarationOnly && tsc-alias -p './tsconfig.build.json' && pnpm build:nuxt && pnpm build:kongponents:umd",
```
adamdehaven:
```suggestion
    "build:kongponents": "pnpm typecheck && vite build && vue-tsc -p './tsconfig.build.json' --emitDeclarationOnly && tsc-alias -p './tsconfig.build.json' && pnpm build:kongponents:umd && pnpm build:nuxt",
```

package.json
arashsheyda: it's fixed now.

https://jsfiddle.net/r672om8t/

adamdehaven (approved):
lgtm

PR #2980, opened 2025-11-11
fix(ktoaster): prevent creating duplicate containers [KHCP-18535]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-18535

Prevent ToastManager from creating duplicate containers

### DIFF
--- a/.github/workflows/get-changed-files.yaml
+++ b/.github/workflows/get-changed-files.yaml
@@ -43,6 +43,7 @@ jobs:
         with:
           files: |
             src/**
+            nuxt/**
 
       - name: Check if docs (`docs` directory) files have changed
         uses: Kong/changed-files@1f31e0a3b1acd04d48dff8801048bb314c8a9e0d
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -136,7 +136,7 @@ jobs:
         - name: Semantic Release
           if: github.event_name == 'push'
           id: semantic_release
-          uses: cycjimmy/semantic-release-action@16ca923e6ccbb50770c415a0ccd43709a8c5f7a4 # v4.2.2
+          uses: cycjimmy/semantic-release-action@9cc899c47e6841430bbaedb43de1560a568dfd16 # v5.0.0
           env:
             # Since branch protections are on (pushing commits) you need to use a bot PAT
             GITHUB_TOKEN: ${{ secrets.KONGPONENTS_BOT_PAT }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,31 @@
+## [9.46.6](https://github.com/Kong/kongponents/compare/v9.46.5...v9.46.6) (2025-11-14)
+
+
+### Bug Fixes
+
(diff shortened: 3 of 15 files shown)


src/components/KToaster/ToastManager.ts
```diff
@@ -27,13 +29,30 @@ export default class ToastManager {
...
-    this.toastersContainer.id = toasterContainerId
-    document.body.appendChild(this.toastersContainer)
+    if (options?.zIndex) {
+      this.zIndex = options.zIndex
+    }
+
+    this.setupToastersContainer()
+  }
+
+  private setupToastersContainer(): void {
```
adamdehaven, line 39: you'll likely want to check that `document` exists just to be sure in case this is init during SSR

portikM: Added question mark - should be enough since we have another check in the `constructor` function

.github/workflows/test.yaml
```diff
@@ -68,3 +70,7 @@ jobs:
         with:
           name: cypress-videos
           path: cypress/videos
+
+      - name: Run Vitest unit tests
+        with:
+          command: pnpm test:unit
```
adamdehaven: question: why are we changing these?

adamdehaven:
@portikM did you notice the sandbox is broken on this PR?

portikM:
> https://deploy-preview-2980--kongponents-sandbox.netlify.app

No I didn't - I think it wasn't broken on some of the previous deployments.

.github/workflows/test.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
             ./README.md
             ./LICENSE
 
-      - name: Run Cypress component tests
+      - name: Run Cypress tests
```
adamdehaven: Can we revert this change so the required jobs don't change?

src/components/KToaster/ToastManager.cy.ts
```diff
@@ -0,0 +1,43 @@
...
+    expect(containers.length).equal(1)
+
+    // Clean up
+    toastManager.destroy()
+  })
+
+  it('should not create a duplicate toasters container if one already exists', () => {
+    const existingContainer = document.createElement('div')
+    existingContainer.id = toastersContainerId
+    document.body.appendChild(existingContainer)
```
adamdehaven, line 18: You should assert after this that this original container exists in the DOM before creating the new one

src/components/KToaster/ToastManager.cy.ts
```diff
@@ -0,0 +1,43 @@
...
+    // Clean up
+    toastManager.destroy()
+  })
+
+  it('should create a toasters container if one does not exist when open is called', () => {
+    const toastManager = new ToastManager()
+
+    const existingContainer = document.querySelectorAll(`#${toastersContainerId}`)
+    expect(existingContainer.length).equal(1)
+    document.querySelector(`#${toastersContainerId}`)?.remove()
```
adamdehaven: nit: you could use `getElementById` and not need the string interpolation

.github/workflows/test.yaml
portikM: Done

src/components/KToaster/ToastManager.cy.ts
portikM: Done

src/components/KToaster/ToastManager.cy.ts
portikM: Done

PR #2982, opened 2025-11-12
fix(vite): disable preserveModules for sandbox

arashsheyda:
# Summary

this PR disables the `preserveModules` config so the sandbox builds correctly
- also includes nuxt directory for changed files check so we get package-pr

### DIFF
--- a/.github/workflows/get-changed-files.yaml
+++ b/.github/workflows/get-changed-files.yaml
@@ -43,6 +43,7 @@ jobs:
         with:
           files: |
             src/**
+            nuxt/**
 
       - name: Check if docs (`docs` directory) files have changed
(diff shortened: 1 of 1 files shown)


vite.config.ts
```diff
@@ -76,7 +76,7 @@ export default defineConfig({
             'vue-router': 'VueRouter',
           },
         exports: 'named',
-        preserveModules: isUMDBuild ? false : true,
+        preserveModules: process.env.USE_SANDBOX ? false : !isUMDBuild,
```
adamdehaven, line 80: This logic looks correct, but please add a comment above to explain when this should be true vs. false

portikM (dismissed):

PR #2983, opened 2025-11-12
fix: display timezone for relative time ranges

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-4393

Display timezone in brackets for relative time ranges.

image


adorack (approved):
Looks good to me from a data team perspective, but I stand by the "may be a breaking change" assertion and can discuss further in Slack.

PR #2984, opened 2025-11-13
fix(ktableview): update table preferences parameters

2eha0:
# Summary

The `update:table-preferences` is emitted immediately after initialization, but the `columnVisibility` in parameter is always `{}`
```vue
<KTableView
  :table-preferences="{
     columnVisibility: { name: false }
  }"
 @update:table-preferences={console.log}
 // { "sortColumnKey": "",  "sortColumnOrder": "desc",  "columnVisibility": {} }
                                                                             ^ wrong value
>
```

### DIFF
--- a/src/components/KTableView/KTableView.vue
+++ b/src/components/KTableView/KTableView.vue
@@ -556,11 +556,31 @@ const getRowKey = (row: Row): string => {
   return ''
 }
 
+/**
+ * Reorder the headers to ensure bulk actions are first and actions are last
+ */
+const reorderHeaders = (headers: readonly Header[]) => {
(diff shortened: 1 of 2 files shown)


src/components/KTableView/KTableView.vue
```diff
@@ -556,11 +556,31 @@ const getRowKey = (row: Row): string => {
   return ''
 }
 
+/**
+ * Reorder the headers to ensure bulk actions are first and actions are last
```
adamdehaven, line 560: The changes in this file don't seem to match the PR description. Help me understand what I'm missing?

2eha0: At the line 583: `tableHeaders` initialized with a empty array, causing the `hasHidableColumns` to always be `false`, so the `computedTablePreferences` is missing the `columnVisibility`

adamdehaven (approved):

PR #2990, opened 2025-11-17
fix(ktoaster): avoid overflow when message or title is too long

kingyue737:
# Summary

[KM-2013]
When the message returned from the backend is long, the icon becomes elliptical, and the close button runs off the screen.

### Before

Screenshot 2025-11-17 134519

### After

Screenshot 2025-11-17 134748

[KM-2013]: https://konghq.atlassian.net/browse/KM-2013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KToaster/KToaster.cy.ts
+++ b/src/components/KToaster/KToaster.cy.ts
@@ -72,5 +72,17 @@ describe('KToaster', () => {
 
       cy.get('.k-toaster').should('have.css', 'z-index', '9999')
     })
+
+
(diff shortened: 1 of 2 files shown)


Justineo:
Should we set a limit on how many lines a toast message can display?
@adamdehaven @portikM

PR #3017, opened 2025-12-05
fix(ktoaster): prevent creating duplicate containers [KHCP-18535]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-18535

Second attempt at preventing `ToastManager` from creating duplicate containers. Fixing issues introduced in https://github.com/Kong/kongponents/pull/2980 and reverted in https://github.com/Kong/kongponents/pull/3015.

### DIFF
--- a/src/components/KToaster/ToastManager.ts
+++ b/src/components/KToaster/ToastManager.ts
@@ -22,13 +22,6 @@ export default class ToastManager {
   private zIndex: number = defaultZIndex
 
   constructor(options?: ToasterOptions) {
-    // For SSR, prevents failing on the build)
-    if (typeof document === 'undefined') {
-      console.warn('ToastManager should only be initialized in the browser environment. Docs: https://kongponents.konghq.com/components/toaster.html')
-
-      return
-    }
-
     if (options?.zIndex) {
       this.zIndex = options.zIndex
     }
@@ -37,7 +30,14 @@ export default class ToastManager {
   }
 
   private setupToastersContainer(): void {
-    const toastersContainerEl = document?.getElementById(toasterContainerId)
+    // For SSR, prevents failing on the build)
+    if (typeof document === 'undefined') {
+      console.warn('ToastManager should only be initialized in the browser environment. Docs: https://kongponents.konghq.com/components/toaster.html')
+
+      return
+    }
+
+    const toastersContainerEl = document.getElementById(toasterContainerId)
     if (toastersContainerEl) {
       this.toastersContainer = toastersContainerEl as HTMLElement
     } else {


portikM:
> Thanks Max!
> 
> Not sure if this helps, but we had a quick look at this yesterday and noticed that there's an opportunity for unintended behaviour here if you have 2 individual ToastManagers. One example is:
> 
> Given 2 ToastManagers:
> 
> 1. `open` is called on ToastManager 1
> 2. time passes
> 3. `open` is called on ToastManager 2
> 4. a tick passes
> 5. `destroy` is called from ToastManager 1
> 6. The toast from ToastManager 2 is removed from the DOM and was only "live" for a tick - i.e. it was never shown
> 
> I think the issue is worse on the current code on `main` (previous to this PR) because the DOM target it created on construction rather than `open` - but still I saw that some tests are expecting toasts to be in the DOM and they are failing due the tests not being able to find the toast. While this could explain the problem there, it doesn't explain why your upstream application tests were fine beforehand, unless I'm missing something 🤔
> 
> We found only two instances (in upstream applications aside from kuma) where the toast managers being created actually have destroy called on them. FYI: in kuma, we are not currently calling destroy at all but "ideally" things would get cleaned up properly.
> 
> I guess if we don't want to go down the lifecycle tracking route, another option would be to just provide the toast manager from a single "shell" application instead of having individual applications instantiate a toast manager themselves. Application would then access the manager provided by the "shell" and the shell would be responsible for cleaning up. Of course if we aren't super precious about cleaning up (after all we want the toast manager to be available throughout the lifetime of the app) we could skip that (ideally we wouldn't skip cleanup, but maybe pragmatism wins here if we don't want to manage lifecycle)
> 
> This option would solve it for our specific needs, but obvs not for others (not sure how many other folks are encountering this issue with duplicate toast managers) - I guess a "fix" there would be to document that ToastManager should be implemented as a singleton in your application and then destroy should only be called once you are sure all toast managers are "finished" (if at all) to prevent these sorts of shared state issues.
> 
> Anyway, thought I'd share our findings/thoughts from yesterday, none of the above are blocking asks or anything

Thanks for sharing your findings jc! Yeah I guess once we get this working there is a slight chance of mishap when 1 instance of ToastManager is being destroyed at the exact moment when another dispatches a toast. On the other hand, once this finalized and ToastManager only mounts one `kongponents-toaster-container`, do we really need to remove the element from the DOM when destroying an instance? 🤔 There's nothing wrong with leaving it there since we know the next created instance of ToastManager won't mount a new element.

johncowen:
Ah and we'd also have to communicate and change this in the two places we found it, if you need help finding ping me offline 👍 

> We found only two instances (in upstream applications aside from kuma) where the toast managers being created actually have destroy called on them

src/components/KToaster/ToastManager.ts
```diff
@@ -78,10 +95,17 @@ export default class ToastManager {
     this.toasts.value = []
   }
 
-  public destroy() {
-    if (this.toastersContainer) {
+  /**
+   * Destroys the ToastManager instance and removes the toasters container element from the DOM
+   * @param removeToastersContainer - Whether to remove the toasters container element from the DOM (defaults to false)
+   */
+  public destroy(removeToastersContainer: boolean = false) {
```
portikM, line 102: In Konnect where we can have multiple `ToastManager` on the same view, removing the toasters container element from the DOM can [very unlikely but can] lead to a race condition when one instance removes the container at the same moment as another instance is attempting to dispatch a notification.

Because a toaster instance will not create a new container if one already exists, I think it's no big deal if we just have the toasters container element in the DOM?

[Original conversation](https://github.com/Kong/kongponents/pull/3017#issuecomment-3728982397)

docs/components/radio.md
```diff
@@ -89,14 +89,27 @@ Use this prop to apply error styling to the component.
...
+  error
   v-model="radio"
   label="Input error"
-  error
   description="Some description text"
   :selected-value="true"
 />
 ```
 
+<KRadio card v-model="errorPropRadio" label="Input error" error description="Some description text" :selected-value="false" />
```
adamdehaven, line 100: FYI: This is unrelated but a valid docs example we requested 👍🏼

adamdehaven (commented):
question: is there any impact to the Nuxt module and the composable it exposes? (@arashsheyda @portikM )

portikM:
> question: is there any impact to the Nuxt module and the composable it exposes? (@arashsheyda @portikM )

I would imagine there's none but I can double check.

arashsheyda:
I don't think it it owuld effect at all, but also if it makes it easier we could setup a mini nuxt sandbox here for easier testing

src/components/KToaster/ToastManager.ts
schogges: Would it make sense to cover this in the tests by adding an assertion for making sure that the DOM element is not being removed after calling `destroy` with the default param (or setting it to `false`)? 🙂

johncowen: I think there's also a theoretical bug here also.

If I call `destroy` and then `destroy(true)` I don't think the DOM container will get cleaned up:

- destroy() > `this.toastersContainer = null`
- destroy(true) `this.toastersContainer && removeToastersContainer` > `null && true === false`

portikM: Both good points, pushed a fix for both.

src/components/KToaster/ToastManager.ts
```diff
@@ -27,13 +29,26 @@ export default class ToastManager {
...
-    document.body.appendChild(this.toastersContainer)
+    if (options?.zIndex) {
+      this.zIndex = options.zIndex
+    }
+
+    this.setupToastersContainer()
+  }
+
+  private setupToastersContainer(): void {
+    const toastersContainerEl = document?.getElementById(toasterContainerId)
```
adamdehaven: Can you pull the preview PR into portal? Just want to make sure this doesn't init on the server

src/components/KToaster/ToastManager.ts
```diff
@@ -27,13 +29,26 @@ export default class ToastManager {
...
+    if (toastersContainerEl) {
+      this.toastersContainer = toastersContainerEl as HTMLElement
+    } else {
+      this.toastersContainer = document.createElement('div')
+      this.toastersContainer.id = toasterContainerId
+      document.body.appendChild(this.toastersContainer)
```
adamdehaven, line 46: Above you're using optional chaining for document but here you're not. Should we just check once at the top if it doesn't exist and exit?

src/components/KToaster/ToastManager.ts
schogges: Probably I wasn't clear enough, but there is still a missing test/assertion for the case of calling `destroy()` and that the container still exists. There is only an assertion that the container is removed after calling `destroy(true)`. 🙂

portikM: Oh yup my miss. Added a test scenario for that now.

johncowen (approved):
Seeing as I've been involved in the convo a little here this LGTM! (considering we are fine not cleaning up of course)

Thanks for the extra detail in the work here Max 👍 

---

Separately and just out of interest (incase you are interested also) in regards to Adam's comment you have left to address, I implemented an improved guard for SSR over in my old suggestion PR over here https://github.com/Kong/kongponents/pull/3045/changes#diff-2adca4150f4c28f9c8c391a0ed2520d130301583843a5bb1890e526f835a8190R106-R113

Obvs we aren't doing that here, but I wanted to see what it would like. Its quite nice that you only have to guard in one place which is out of the main body of code at creation time, and then you can totally forget (and not worry about) the fact that you might be running in SSR everywhere else.

I quite like fixing things in a single place at the root and not having to use `?` and `if` statements everywhere in multiple places, but maybe thats a personal opinion thing 🤷

src/components/KToaster/ToastManager.ts
portikM: Good call, fixed.

src/components/KToaster/ToastManager.ts
portikM: Done: https://github.com/kong-konnect/portal/pull/1880

adamdehaven (approved):

PR #3023, opened 2025-12-11
fix(kdatetimepicker): time inputs on same line

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-4310

Each time input having its own row makes the time picker popover far taller than it needs to be. This change simply updates the time inputs to appear on the same row when picking a date range as per [figma](https://www.figma.com/design/Yze0SWXl5nKjR0rFdilljK/Components?node-id=2473-729&t=FC7Se6ce6Dy4hbCq-0)

Additionaly:
- Tighten up margins/padding
- Provide `trim-weeks` prop to v-calendar so an additional empty row of days is not rendered in some months.

A secondary issue related to the ticket is that the underlying KPopover, using "floating-ui" is not correctly detecting collision/overflow in some cases leading to it not opening in the "correct direction". This causes parts of the popover being cutoff.
- Further investigation and potential updates to KPopover to fix this can be done in a follow-up
- Reducing the height of the entire time picker should help mitigate the issue.

image

## Konnect ui
image

### DIFF
--- a/src/components/KDateTimePicker/KDateTimePicker.vue
+++ b/src/components/KDateTimePicker/KDateTimePicker.vue
@@ -605,7 +605,6 @@ $kDateTimePickerInputPaddingY: var(--kui-space-40, $kui-space-40); // correspond
 
     .popover-footer {
       margin: var(--kui-space-0, $kui-space-0);
-      margin-top: var(--kui-space-40, $kui-space-40);
 
       .datetime-picker-footer-container {
         display: flex;


src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -292,52 +298,70 @@ defineExpose({
...
+    border-radius: var(--kui-border-radius-20, $kui-border-radius-20);
 
-    .time-input {
-      border-radius: var(--kui-border-radius-10, $kui-border-radius-10);
+    .time-inputs-container {
       display: flex;
-      flex-direction: column;
+      flex-direction: row;
       gap: var(--kui-space-30, $kui-space-30);
       padding: var(--kui-space-30, $kui-space-30) var(--kui-space-60, $kui-space-60);
```
Chalks: This padding feels weird to me. I get that it's aligning with the calendar, but the gray background really throws me.

**Current**

Screenshot 2025-12-11 at 5 07 36 PM

**Suggestion**
Screenshot 2025-12-11 at 5 10 22 PM

Chalks (dismissed):
I won't block on my comment, but I feel pretty strongly about it.

Chalks (approved):

PR #3026, opened 2025-12-16
fix(ktextarea): fix mixin inclusion to prevent invalid css

Justineo:
# Summary

[KM-2105]

Current implementation would produce the following CSS:

```css
.input-textarea-wrapper.legacy[data-v-2080bcde]:after::-webkit-input-placeholder
```

which is an invalid CSS selector and will fail tools like LightningCSS.

[KM-2105]: https://konghq.atlassian.net/browse/KM-2105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KTextArea/KTextArea.vue
+++ b/src/components/KTextArea/KTextArea.vue
@@ -306,10 +306,8 @@ $kTextAreaPaddingY: var(--kui-space-40, $kui-space-40); // corresponds to mixin,
     }
   }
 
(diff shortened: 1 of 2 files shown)


adamdehaven (approved):
lgtm

PR #3027, opened 2025-12-16
fix(kcodeblock): ssr types check

portikM:
# Summary

Minor SSR fix for KCodeBlock

### DIFF
--- a/src/components/KCodeBlock/KCodeBlock.vue
+++ b/src/components/KCodeBlock/KCodeBlock.vue
@@ -482,6 +482,7 @@ onBeforeUnmount(function() {
 })
 
 function emitCodeBlockRenderEvent(): void {
+  // Guard for SSR environment (HTMLElement is undefined in SSR)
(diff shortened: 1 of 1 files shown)


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -605,7 +609,7 @@ function jumpToPreviousMatch(): void {
 }
 
 function jumpToMatch(direction: number): void {
-  if (matchingLineNumbers.value.length === 0 || !(codeBlockRef.value instanceof HTMLElement)) {
+  if (matchingLineNumbers.value.length === 0 || typeof HTMLElement === 'undefined' || !(codeBlockRef.value instanceof HTMLElement)) {
```
adamdehaven: can you move the HTMLElement check first, and then add a comment above each check you added to explain

portikM: Done

PR #3038, opened 2026-01-05
feat(kslider): make step labels clickable [KHCP-19044]

mptap:
# Summary

Addresses https://konghq.atlassian.net/browse/KHCP-19044
Make the KSlider's step labels clickable, ie. clicking a step label sets the slider value to that step's value.

# Demo
https://www.loom.com/share/8c56cf4f22144969a3f5be3f4d5ac180

### DIFF
--- a/src/components/KSlider/KSlider.vue
+++ b/src/components/KSlider/KSlider.vue
@@ -82,7 +82,8 @@
           v-for="mark in rangeMarks"
           :key="`mark-${mark.value}`"
           :data-testid="`mark-${mark.value}`"
-          :style="{ left: getValuePercent(mark.value), cursor: 'pointer' }"
+          role="button"
+          :style="{ left: getValuePercent(mark.value) }"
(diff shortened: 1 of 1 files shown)


kaiarrowood (commented):
Functional looks good 🚀

src/components/KSlider/KSlider.vue
```diff
@@ -82,7 +82,8 @@
           v-for="mark in rangeMarks"
           :key="`mark-${mark.value}`"
           :data-testid="`mark-${mark.value}`"
-          :style="{ left: getValuePercent(mark.value) }"
+          :style="{ left: getValuePercent(mark.value), cursor: 'pointer' }"
```
portikM: nit: can you move `cursor: pointer;` rule to the style block? Also I think we want `role="button"` attribute here

mptap: Done

PR #3039, opened 2026-01-05
chore(deps): bump cypress

portikM:
# Summary

Bump Cypress to `15.8.1` ([renovate PR](https://github.com/Kong/kongponents/pull/3014))

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -25,7 +25,7 @@ const buildFormats: Array<'es' | 'cjs' | 'umd'> = isUMDBuild ? ['umd'] : ['es', 
 export default defineConfig({
   plugins: [
     vue(),
-    ...(process.env.DISABLE_VUE_DEVTOOLS ? [] : [VueDevTools()]), // Cypress 14+ introduces an issue with VueDevTools when running tests so we need to disable it in the test environment only
+    ...(process.env.DISABLE_VUE_DEVTOOLS === 'true' ? [] : [VueDevTools()]), // Cypress 14+ introduces an issue with VueDevTools when running tests so we need to disable it in the test environment only
   ],
   resolve: {
     alias: {


cypress/support/index.ts
```diff
@@ -43,6 +49,12 @@ Cypress.Commands.add('mount', (component: ComputedOptions, options = {}): Chaina
 })
 
 Cypress.Commands.add('mountWithProdRouter', (component: ComputedOptions, options = {}) => {
+  Cypress.on('uncaught:exception', (err) => {
+    if (err.message.includes('Cannot read properties of undefined (reading \'app\')')) {
+      return false
+    }
+  })
```
portikM: Not sure this is ideal but it's the only way I could find to fix [this](https://github.com/Kong/kongponents/actions/runs/20725588032/job/59500237986?pr=3014#step:10:365) issue.

cypress/support/index.ts
```diff
@@ -27,6 +27,12 @@ Cypress.Commands.add('findTestId', { prevSubject: 'element' }, (subject, dataTes
 })
 
 Cypress.Commands.add('mount', (component: ComputedOptions, options = {}): Chainable => {
+  Cypress.on('uncaught:exception', (err) => {
+    if (err.message.includes('Cannot read properties of undefined (reading \'app\')')) {
+      return false
+    }
+  })
```
portikM: Not sure this is ideal but it's the only way I could find to fix [this](https://github.com/Kong/kongponents/actions/runs/20725588032/job/59500237986?pr=3014#step:10:365) issue.

adamdehaven (changes requested):
We need to figure out the issue if we're upgrading. This "fix" is a hack. 

I would imagine there are breaking changes between versions 13 > 15

portikM:
> We need to figure out the issue if we're upgrading. This "fix" is a hack.
> 
> I would imagine there are breaking changes between versions 13 > 15

I found a better way to fix the issue: the error s coming from Cypress itself, some function related to Vue devtools. Since we don't really need Vue devtools in tests, I disabled it in `vite.config.ts`.

vite.config.ts
```diff
@@ -25,7 +25,7 @@ const buildFormats: Array<'es' | 'cjs' | 'umd'> = isUMDBuild ? ['umd'] : ['es',
 export default defineConfig({
   plugins: [
     vue(),
-    VueDevTools(),
+    ...(process.env.DISABLE_VUE_DEVTOOLS ? [] : [VueDevTools()]), // Cypress 14+ introduces an issue with VueDevTools when running tests so we need to disable it in the test environment only
```
adamdehaven:
```suggestion
    ...(process.env.DISABLE_VUE_DEVTOOLS === 'true' ? [] : [VueDevTools()]), // Cypress 14+ introduces an issue with VueDevTools when running tests so we need to disable it in the test environment only
```

adamdehaven (approved):
lgtm

PR #3045, opened 2026-01-07
fix(ktoaster): prevent duplicate containers via a shared pool v2

johncowen:
Alternative to https://github.com/Kong/kongponents/pull/3041

This one makes the entire ToastManager a singleton instead of just its DOM target.

It does this by creating a new proxy ToastManager class which can be instantiated multiple times, this new ToastManager instantiates the old ToastManager as a singleton, so we only ever have one ToastManager.

The one caveat with this is the `zIndex` option can't really be guaranteed to be set it we have a ToastManager singleton, but looking at the code it never worked as intended anyway (it set the zIndex of the Toasters inside of the DOM target, which form what I see is pointless)

I found the PR where this zIndex option was added (https://github.com/Kong/kongponents/pull/2130) and it looks like it wasn't a specific ask anyway as its in amongst a more general refactor of "add zIndex to all overlaying components" so I'm guessing no-one is using this anyway (and if they are its not working as intended).

I deprecated the options argument here also, and I'd see this approach as backwards compatible because, if I'm not mistaken, the zIndexing wasn't working as intended anyway (and can't be supported properly if we want a shared singleton DOM target), although there maybe other opinions on this. Personally I prefer this approach to https://github.com/Kong/kongponents/pull/3041, but I'd be happy with either.

Small note: I spotted quite a few things/potential issues here with other things in the older code, but didn't want to mix that up with this singleton task.

### DIFF
--- a/sandbox/composables/useSandboxToaster.ts
+++ b/sandbox/composables/useSandboxToaster.ts
@@ -2,7 +2,17 @@ import { onBeforeUnmount } from 'vue'
 import { ToastManager } from '@/index'
 
 export default function useSandboxToaster() {
-  const toaster = new ToastManager()
+  const toasters = Array.from({ length: 20 }).map(() => new ToastManager())
+  const toaster = toasters.pop()!
+
+  toasters.forEach(item => setTimeout(() => item.destroy(), Math.ceil(Math.random() * 10000)))
+
+  Array.from({ length: 20 }).forEach(() => {
+    setTimeout(() => {
+      const item = new ToastManager()
+      setTimeout(() => item.destroy(), Math.ceil(Math.random() * 10000))
+    }, Math.ceil(Math.random() * 10000))
+  })
 
   onBeforeUnmount(() => {
     toaster.destroy()
--- a/sandbox/pages/SandboxToaster.vue
+++ b/sandbox/pages/SandboxToaster.vue
@@ -62,6 +62,11 @@
           KToaster
         </KButton>
       </SandboxSectionComponent>
+      <SandboxSectionComponent title="destroy manager">
+        <KButton @click="() => toaster.destroy()">
+          Destroy
+        </KButton>
+      </SandboxSectionComponent>
     </div>
   </SandboxLayout>
 </template>
--- a/src/components/KToaster/SharedPool.ts
+++ b/src/components/KToaster/SharedPool.ts
@@ -1,0 +1,55 @@
+type Transition<K, T> = (state: 'creating' | 'acquiring' | 'releasing' | 'destroying', key: K, item: T) => T
+type Entry<T> = {
+  value: T
+  references: Set<symbol>
+}
+export default class SharedPool<K, T> {
+  constructor(
+    protected transition: Transition<K, T>,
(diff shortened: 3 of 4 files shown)


Justineo (commented):
Why do we need a state machine here? Overall, this feels like it might be adding more complexity than necessary for what seems to be a fairly simple issue.

Could we just reuse the toast container if it’s already been created?

johncowen:
> Also, if a page ends up loading multiple Kongponents bundles (even if it’s not ideal), in-memory ref counting would be isolated per bundle while still operating on the same actual DOM container, which makes the lifecycle harder to coordinate. Using the DOM container as the source of truth keeps this straightforward.

I noted that here:

https://github.com/Kong/kongponents/pull/3041#issuecomment-3716024852

But to be fair, if multiple Kongponents bundles are installed adding two DOM nodes due to this are the least of your worries as its pretty likely all the CSS breaks. This has caused several incidents as far as I'm aware. We specifically have tests in place to prevent multiple kong dependencies from being installed and prevent the CSS from breaking, but we are getting into a different discussion here. My assumption was that we assume a single kongponents installation, but if thats not the case I would guess it would be pretty trivial to move pool somewhere where its guaranteed to be a per-app singleton itself.

> From my perspective, the actual requirements here are quite small: we just need to ensure the toast container exists when a toast is opened, and to avoid creating a new container if one already exists. A direct DOM-based check already handles both.

How are you going to handle destruction? Or are you just going to skip that?

FYI: There is a slightly simpler version that could be done seeing as we only need one singleton/destructor here, but seeing as I just copy/pasta'ed the code for SharedPool over from kuma (i.e. a poor mans dependency), that's actually much simpler than reinventing the wheel for a single usecase or implementing something new that doesn't require multiple singletons.

Anyway I'm going to go ahead and close these, thanks for the feedback! I'll keep an eye out for a different fix.

johncowen:
> I think we can start by revisiting Max’s draft to see if a simple solution already addresses the problem.

I think this was my point of the PRs I put up. I don't want to sounds dismissive, but as far as I'm aware the draft is the third go at getting this right with what I think you'd consider a "simple solution" and that's been up for a while and I'd assumed there was some reason for that - who knows 🤷 . Additionally from what I understood that Max has been pushed for time to get that through, and again not to sound dismissive but I don't follow why its not just an approve and merge given the simplicity 🤷.

We reported the issue a good while back (~2 months ago I believe) and I've seen others now reporting the same issue. So I thought I'd give a couple of options using reusable/purposeful code that was already written moons ago that I know works. Quick copy paste and issue solved.

But as I said if either of my approaches isn't desired then, I'll just continue to keep an eye out for an alternative fix 👍

johncowen:
I noticed this was accidentally left open, closing.

Further note:

> The one caveat with this is the zIndex option can't really be guaranteed to be set it we have a ToastManager singleton, but looking at the code it never worked as intended anyway (it set the zIndex of the Toasters inside of the DOM target, which form what I see is pointless)

This ^ was dealt with in https://github.com/Kong/kongponents/pull/3093

PR #3055, opened 2026-01-13
fix(*): add name attribute to inputs [KHCP-19041]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-19041

Set `name` attributes on KInputs used in other components to disable autosuggest in password managers for inputs that don't need it


src/components/KCodeBlock/KCodeBlock.vue
```diff
@@ -280,6 +281,8 @@ const {
   maxHeight = 'none',
 } = defineProps<CodeBlockProps>()
 
+const inputId = useId()
```
adamdehaven: issue: we shouldn't be utilizing random IDs as `name` attributes. You can give the field in the component an actual name if you'd like, but a dynamic id as the name attribute isn't ideal here (whole PR)

portikM: I was debating that. Fixed, using actual name in all components (generic)

src/components/KFileUpload/KFileUpload.vue
```diff
@@ -174,11 +175,13 @@ const { stripRequiredLabel } = useUtilities()
 
 const defaultId = useId()
 const fileInputId = computed((): string => attrs.id ? String(attrs.id) : defaultId)
+const fileInputName = computed((): string => attrs.name ? String(attrs.name) : 'file')
```
adamdehaven: nit
```suggestion
const fileInputName = computed((): string => attrs.name ? String(attrs.name) : 'uploaded-file')
```

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -176,6 +177,7 @@
                   class="multiselect-dropdown-input"
                   data-testid="multiselect-dropdown-input"
                   :model-value="filterString"
+                  name="multiselect"
```
adamdehaven: issue: you've got two inputs with the same name. let's be more specific

src/components/KPrompt/KPrompt.vue
```diff
@@ -48,6 +48,7 @@
           data-testid="confirmation-input"
           :error="displayErrorState"
           :error-message="errorMessage"
+          name="confirmation"
```
adamdehaven:
```suggestion
          name="prompt-confirmation-text"
```

src/components/KSelect/KSelect.vue
```diff
@@ -52,6 +52,7 @@
             :disabled="isDisabled"
             :error="error"
             :model-value="filterQuery"
+            name="select"
```
adamdehaven: be more specific

src/components/KMultiselect/KMultiselect.vue
portikM: Made names more specific. Both inputs serve the same purpose (items filter) but only one is present in the DOM in scope of one KMultiselect component (one is rendered when `collapsedContent` prop is `true`, the other is rendered when that prop is `false`) so it's okay to have the same value in `name` attribute for both inputs.

src/components/KSelect/KSelect.vue
portikM: Done

PR #3057, opened 2026-01-13
feat(kselect, kmultiselect): item grouping interface [KHCP-19008]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-19008

Introduces a new interface for defining grouped items in KSelect and KMultiselect components, allowing to control group ordering and structure items in a more explicit, nested format.

Behavior:

- Groups appear in the order defined in the array (not alphabetically sorted)
- Ungrouped items always render first, followed by groups
- Supports mixing `SelectGroup`/`MultiselectGroup` with ungrouped `SelectItem`/`MultiselectItem` entries
- Fully backwards compatible - existing group property on individual items still works (alphabetical sorting)

Konnect consuming PR: https://github.com/kong-konnect/konnect-ui-apps/pull/9755

### DIFF
--- a/docs/components/multiselect.md
+++ b/docs/components/multiselect.md
@@ -58,24 +58,28 @@ const clearIt = () => {
 
 ### items
 
-An array of items containing a `label` and `value`.
-
-You may also specify:
-- a certain item is `selected` by default
-- a certain item is `disabled`
-- certain items are grouped under a `group`
+Prop for providing multiselect item options.
 
 ```ts
 interface MultiselectItem {
+  /** Label for the item to be displayed in the multiselect dropdown. */
   label: string
   value: string
+  /** Optional parameter that will be used as the key for the item in the multiselect dropdown. */
+  key?: string
   selected?: boolean
   disabled?: boolean
-  group?: string // Deprecated, use `MultiselectGroup` instead
+  /**
+   * @deprecated The `group` property on individual items is deprecated.
+   * Instead, use the `SelectGroup` interface to structure grouped items.
+   */
+  group?: string
 }
 
 interface MultiselectGroup {
+  /** Label for the group to be displayed in the multiselect dropdown. */
   label: string
+  /** Optional parameter that will be appended with `-group` when grouped. */
   key?: string
   items: MultiselectItem[]
 }
@@ -87,37 +91,65 @@ type MultiselectEntry = MultiselectItem | MultiselectGroup
   <KMultiselect :items="deepClone(defaultItemsWithDisabledAndGroups)" />
 </ClientOnly>
 
-```html
-<KMultiselect :items="[{
+```vue
+<template>
+  <KMultiselect :items="multiselectItems" />
+</template>
+
+<script setup lang="ts">
+import type { MultiselectEntry } from '@kong/kongponents'
+
+const items: MultiselectEntry[] = [
+  /** Group 1 */
+  {
+    label: 'Birds', // Group 1 label
+    items: [ // Group 1 items
+      { label: 'Duck', value: 'duck' },
+      { label: 'Oriole', value: 'oriole' },
+    ] },
+  /** Group 2 */
+  {
+    label: 'Fish', // Group 2 label
+    items: [ // Group 2 items
+      { label: 'Salmon', value: 'salmon' },
+      { label: 'Trout', value: 'trout' },
+    ] },
+  /** Ungrouped items */
+  {
     label: 'Cats',
     value: 'cats',
-    selected: true
-  }, {
+    selected: true,
+  },
+  {
     label: 'Dogs',
     value: 'dogs',
     selected: true,
     disabled: true,
-  }, {
+  },
+  {
     label: 'Bunnies',
     value: 'bunnies',
-    selected: true
+    selected: true,
   },
   {
     label: 'Lions',
     value: 'lions',
     disabled: true,
-  }, {
+  },
+  {
     label: 'Tigers',
     value: 'tigers',
-    disabled: true
-  }, {
+    disabled: true,
+  },
+  {
     label: 'Bears',
-    value: 'bears'
-  }, {
+    value: 'bears',
+  },
+  {
     label: 'An extremely lengthy truncated item',
-    value: 'long'
-  }, ...]"
-/>
+    value: 'long',
+  }]
+</script>
 ```
 
 ### help
--- a/docs/components/select.md
+++ b/docs/components/select.md
@@ -38,7 +38,7 @@ Selected service: {{ selectValue || 'none' }}
 
 ### items
 
-Prop for providing select item options. Supports grouping items under one group name through providing optional `group` property.
+Prop for providing select item options.
 
 ```ts
 interface SelectItem {
@@ -106,7 +106,7 @@ type SelectEntry = SelectItem | SelectGroup
   }]" />
 </ClientOnly>
 
-```html
+```vue
 <template>
   <KSelect :items="selectItems" />
 </template>
@@ -129,7 +129,7 @@ const selectItems: SelectEntry[] = [
       { label: 'Service A2', value: 'a2' },
       { label: 'Service B2', value: 'b2' },
     ] },
-  /**Ungrouped items */
+  /** Ungrouped items */
   { label: 'Service A', value: 'a', selected: true },
   { label: 'Service B', value: 'b' },
   { label: 'Service F', value: 'f', disabled: true },
--- a/src/types/multi-select.ts
+++ b/src/types/multi-select.ts
@@ -1,18 +1,25 @@
 import type { LabelAttributes } from './label'
 import type { PopoverAttributes } from './popover'
 export interface MultiselectItem<T extends string = string> {
+  /** Label for the item to be displayed in the multiselect dropdown. */
   label: string
   value: T
+  /** Optional parameter that will be used as the key for the item in the multiselect dropdown. */
   key?: string
   selected?: boolean
   disabled?: boolean
   custom?: boolean
-  /** @deprecated Use `SelectGroup` instead. */
+  /**
+   * @deprecated The `group` property on individual items is deprecated.
+   * Instead, use the `SelectGroup` interface to structure grouped items.
+   */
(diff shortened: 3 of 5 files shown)


filipgutica (approved):

docs/components/multiselect.md
```diff
@@ -102,6 +102,51 @@ You may also specify:
 />
 ```
 
+### groupComparator
```
adamdehaven: Can we call this `sortHandlerFunction` to be more similar to what we offer in other components (e.g. KTable)?

adamdehaven: Help me understand why the component needs to handle sorting instead of the host application passing in the items in the desired order?

docs/components/select.md
```diff
@@ -113,6 +113,51 @@ interface SelectItem {
 }]" />
 ```
 
+### groupComparator
```
adamdehaven: Same comment on prop naming

src/components/KMultiselect/KMultiselectItems.vue
```diff
@@ -81,7 +82,15 @@ const SELECTABLE_ITEM_SELECTOR = '.multiselect-item button:not(:disabled)'
 const handleItemSelect = (item: MultiselectItem<T>) => emit('selected', item)
 
 const nonGroupedItems = computed((): Array<MultiselectItem<T>> => items?.filter(item => !item.group))
-const groups = computed((): string[] => [...new Set((items?.filter(item => item.group))?.map(item => item.group!))].sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase())))
+const groups = computed((): string[] => {
```
adamdehaven: Please add comments in this section

src/components/KSelect/KSelectItems.vue
```diff
@@ -83,7 +84,15 @@ const SELECTED_ITEM_SELECTOR = '.select-item[aria-selected="true"] button:not(:d
 const handleItemSelect = (item: SelectItem<T>) => emit('selected', item)
 
 const nonGroupedItems = computed((): Array<SelectItem<T>> => items?.filter(item => !item.group))
-const groups = computed((): string[] => [...new Set((items?.filter(item => item.group))?.map(item => item.group!))].sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase())))
+const groups = computed((): string[] => {
```
adamdehaven: Please add comments in this section

adamdehaven (changes requested):

docs/components/multiselect.md
portikM: I'll rework this to follow the approach Yiling suggested (resolving all comments).

docs/components/multiselect.md
```diff
@@ -102,6 +120,86 @@ You may also specify:
 />
 ```
 
+#### Group ordering
```
Justineo: I think we should recommend nested items as the default way to define groups, rather than only recommending it when we need to control group ordering

src/components/KMultiselect/KMultiselectItems.vue
```diff
@@ -4,44 +4,49 @@
...
-      </template>
-    </KMultiselectItem>
-
-    <div
-      v-for="group in groups"
-      :key="`${group}-group`"
-      class="multiselect-group"
+    <template
+      v-for="entry in items"
+      :key="isGroup(entry) ? `${entry.key || entry.label}-group` : entry.key"
```
Justineo: What if a non-group entry doesn't have a `key`? Maybe we should fallback to `value` and then `index`.

src/components/KMultiselect/KMultiselectItems.vue
```diff
@@ -76,14 +81,30 @@ const {
...
+
+// Helper to check if entry is a group (has 'items' property)
+const isGroup = (entry: unknown): entry is GroupEntry<T> => {
+  if (!entry || typeof entry !== 'object') {
+    return false
+  }
 
-const nonGroupedItems = computed((): Array<MultiselectItem<T>> => items?.filter(item => !item.group))
-const groups = computed((): string[] => [...new Set((items?.filter(item => item.group))?.map(item => item.group!))].sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase())))
+  return 'items' in entry && Array.isArray((entry as { items?: unknown }).items)
```
Justineo:
```suggestion
  return 'items' in entry && Array.isArray(entry.items)
```

We already checked `'items' in entry` so I believe we don't need to cast entry to `{ items?: unknown }`.

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -546,6 +598,51 @@ const filteredItems = computed(() => {
...
+    return normalizedItems.value
+  }
+
+  // STEP 1: Get filtered flat items
+  const filteredFlatItems = filteredItems.value
+
+  // STEP 2: Create a Set of filtered item values for quick lookup
+  const filteredValues = new Set(filteredFlatItems.map(item => item.value))
+
+  // STEP 3: Rebuild from normalizedItems with only filtered items
```
Justineo: The comments in this function look a bit verbose to me.

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -760,16 +875,65 @@ const handleAddItem = (): void => {
   filterString.value = ''
 }
 
-// Sort items. Non-grouped items are displayed first, then grouped items.
-// Within non-grouped and grouped items, selected items are displayed first.
+// Helper to check if entry is MultiselectGroup (has 'items' property)
+const isMultiselectGroup = (entry: unknown): entry is MultiselectGroup<Value> => {
```
Justineo: Why `entry` is `unknown` here? Shouldn't it be `MultiselectEntry<Value>`? Same below.

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -760,16 +875,65 @@ const handleAddItem = (): void => {
...
+
+// Helper to check if normalized entry is a group
+const isNormalizedGroup = (entry: unknown): entry is NormalizedGroup => {
+  if (!entry || typeof entry !== 'object') {
+    return false
+  }
+
+  const candidate = entry as { items?: unknown }
+  return Array.isArray(candidate.items)
+}
```
Justineo, line 879: Do we need both `isMultiselectGroup` and `isNormalizedGroup` here? What about using a single `hasItems` instead?

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -760,16 +875,65 @@ const handleAddItem = (): void => {
...
+    const unselectedGroupItems = group.items.filter(item => !item.selected)
+
+    result.push({
+      label: group.label,
+      key: group.key,
+      items: [...selectedGroupItems, ...unselectedGroupItems],
+    })
+  }
+
+  sortedItems.value = result
```
Justineo, line 919: Is this necessary? I would expect we just preserve the order of the input.

docs/components/multiselect.md
portikM: I was debating whether I should switch the documentation to use the new approach and thought it might confuse someone who's used to the old approach but I agree with you, we should start using the new approach going forward. Updated the documentation for both components promoting the new approach.

src/components/KMultiselect/KMultiselect.vue
portikM: Yeah prob don't need as many. Removed some

src/components/KMultiselect/KMultiselect.vue
portikM: Yes - that's how KMultiselect behaves currently - selected items are displayed first. I won't find the thread now as it was ages ago but basically we decided that's the behaviour we want.

src/components/KMultiselect/KMultiselect.vue
portikM: Good call, I need both originally but I think I simplified my approach and didn't notice I don't need both. Fixed

src/components/KMultiselect/KMultiselect.vue
portikM: `MultiselectItem<T> | GroupEntry<T>`, updated

src/components/KMultiselect/KMultiselectItems.vue
portikM: Good call, fixed.

src/components/KMultiselect/KMultiselectItems.vue
portikM: Fixed.

src/types/multi-select.ts
```diff
@@ -7,9 +7,20 @@ export interface MultiselectItem<T extends string = string> {
   selected?: boolean
   disabled?: boolean
   custom?: boolean
+  /** @deprecated Use `SelectGroup` instead. */
```
adamdehaven: This deprecated message is referencing an interface, but it should actually mention the property that should be used instead.

src/types/select.ts
```diff
@@ -8,13 +8,24 @@ export interface SelectItem<T extends string | number = string | number> extends
   key?: string
   selected?: boolean
   disabled?: boolean
+  /** @deprecated Use `SelectGroup` instead. */
```
adamdehaven: same comment

src/types/multi-select.ts
portikM: It's not a property that we're introducing instead, it's a different interface, `SelectGroup`.

docs/components/select.md
```diff
@@ -63,54 +103,38 @@ interface SelectItem {
...
+      { label: 'Service B1', value: 'b1' },
+    ] },
+  /** Group 2 */
+  {
+    label: 'Series 2', // Group 2 label
+    items: [ // Group 2 items
+      { label: 'Service A2', value: 'a2' },
+      { label: 'Service B2', value: 'b2' },
+    ] },
+  /**Ungrouped items */
```
adamdehaven:
```suggestion
  /** Ungrouped items */
```

PR #3058, opened 2026-01-15
fix(radio): disabled checked radio card style not match figma

jw-foss:
* Remove the background for disabled and checked radio card to match figma design.

### Before 

CleanShot 2026-01-15 at 11 03 06@2x

### After

CleanShot 2026-01-15 at 11 03 14@2x

### Docs

CleanShot 2026-01-15 at 11 11 35@2x

# Summary

### DIFF
--- a/docs/components/radio.md
+++ b/docs/components/radio.md
@@ -114,7 +114,7 @@ Use this prop to apply error styling to the component.
 
 Any valid attribute will be added to the input. You can read more about `$attrs` [here](https://vuejs.org/api/composition-api-setup.html#setup-context).
 
-<div style="display: flex; flex-direction: column; gap: 16px;">
+<div class="vertical-spacing">
   <KRadio
     checked
     description="Disabled radio card"
@@ -136,7 +136,7 @@ Any valid attribute will be added to the input. You can read more about `$attrs`
 </div>
 
 ```html
-<div style="display: flex; flex-direction: column; gap: 16px;">
+<div>
   <KRadio
     checked
     description="Disabled radio card"


docs/components/radio.md
```diff
@@ -114,16 +114,48 @@ Use this prop to apply error styling to the component.
 
 Any valid attribute will be added to the input. You can read more about `$attrs` [here](https://vuejs.org/api/composition-api-setup.html#setup-context).
 
-<KRadio v-model="disabledAttributeRadio" :selected-value="true" disabled>Disabled radio</KRadio>
+<div style="display: flex; flex-direction: column; gap: 16px;">
```
Justineo: Use a class name here for better consistency.

docs/components/radio.md
```diff
@@ -114,16 +114,48 @@ Use this prop to apply error styling to the component.
...
 
 ```html
-<KRadio
-  v-model="checked"
-  :selected-value="true"
-  disabled
->
-  Disabled radio
-</KRadio>
+<div style="display: flex; flex-direction: column; gap: 16px;">
```
Justineo: We don't need to include this in our code block.

src/components/KRadio/KRadio.vue
```diff
@@ -539,7 +539,6 @@ $kRadioDotSize: 6px;
 
       &.checked {
         .radio-card-wrapper {
-          background-color: var(--kui-color-background-disabled, $kui-color-background-disabled);
```
adamdehaven, line 542: @portikM fornsome reason I thought we added this on purpose. 

Am I misremembering, or is this on purpose and the Figma wasn't updated?

adamdehaven (changes requested):
Just blocking for discussion

src/components/KRadio/KRadio.vue
portikM: I think the Figma wasn't updated and we just eyeballed it. I'm ok with removing.

PR #3070, opened 2026-01-21
fix: add sameDayFullRange prop to allow valid time range for same day queries [MA-4646]

jordojordo:
Fix [MA-4646](https://konghq.atlassian.net/browse/MA-4646)

# Summary

This will add a new prop (`sameDayFullRange`) to the `KDateTimePicker` and `CalendarWrapper` components to allow for valid time ranges when a user selects the same day for a start and end time.

When enabled and the same date is chosen for the start and end, the time range will automatically be adjusted to `12:00` am to `11:59` pm of the same day, as opposed to the `12:00` am and `12:00` am for both start and end times.

Another callback prop has been added `customValidationRange` which will allow for custom validation of `start` and `end` times to disable the "Apply" button in KDateTimePicker.

This also refactors a little bit of repeated and messy type guarding within the CalendarWrapper.

[MA-4646]: https://konghq.atlassian.net/browse/MA-4646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sandbox/pages/SandboxDateTimePicker.vue
+++ b/sandbox/pages/SandboxDateTimePicker.vue
@@ -55,6 +55,26 @@
           range
         />
       </SandboxSectionComponent>
+      <SandboxSectionComponent title="sameDayFullRange">
+        <KComponent
+          v-slot="{ data }"
+          :data="{ sameDayFullRange: true, value: { start: null, end: null } }"
+        >
+          <KDateTimePicker
+            v-model="data.value"
+            mode="dateTime"
+            range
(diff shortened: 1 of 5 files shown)


filipgutica (approved):

adamdehaven (approved):

adamdehaven:
Approved based on @filipgutica 's review

PR #3076, opened 2026-01-27
fix(ktoaster): create unique toaster container element [KHCP-19222]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-19222

Fixed toast overlapping when multiple `ToastManager` instances exist on one page. Implemented a shared container pattern where all instances render into a single `#kongponents-toaster-container`, with each instance creating its own wrapper `div` inside. Reference counting via DOM data attributes ensures the container persists until all instances are destroyed. Fully backward compatible with existing API.

Consuming in Konnect: https://github.com/kong-konnect/konnect-ui-apps/pull/10005

### DIFF
--- a/package.json
+++ b/package.json
@@ -188,12 +188,12 @@
   },
   "engines": {
     "node": ">=v16.20.2 || >=18.12.1 || >=20.14.0 || >=22.14.0",
-    "pnpm": ">=9.11.0 || >=10.10.0"
+    "pnpm": ">=10.28.2"
   },
-  "packageManager": "pnpm@10.28.0",
+  "packageManager": "pnpm@10.28.2",
   "volta": {
     "node": "24.12.0",
-    "pnpm": "10.28.0"
+    "pnpm": "10.28.2"
   },
   "config": {
     "commitizen": {
@@ -231,7 +231,8 @@
       "tar@<=7.5.2": ">=7.5.3",
       "lodash@>=4.0.0 <=4.17.22": ">=4.17.23",
       "seroval@<=1.4.0": ">=1.4.2",
-      "seroval@<1.4.1": ">=1.4.2"
+      "seroval@<1.4.1": ">=1.4.2",
+      "tar@<7.5.7": ">=7.5.7"
     },
     "onlyBuiltDependencies": [
       "@evilmartians/lefthook",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -30,6 +30,7 @@ overrides:
   lodash@>=4.0.0 <=4.17.22: '>=4.17.23'
   seroval@<=1.4.0: '>=1.4.2'
   seroval@<1.4.1: '>=1.4.2'
+  tar@<7.5.7: '>=7.5.7'
 
 importers:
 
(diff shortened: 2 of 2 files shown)


src/components/KToaster/ToastManager.ts
```diff
@@ -19,36 +19,21 @@ export default class ToastManager {
...
-    const toastersContainerEl = document.getElementById(toasterContainerId)
-    if (toastersContainerEl) {
-      this.toastersContainer = toastersContainerEl as HTMLElement
-    } else {
-      this.toastersContainer = document.createElement('div')
-      this.toastersContainer.id = toasterContainerId
-      document.body.appendChild(this.toastersContainer)
-    }
+    this.toastersContainer = document.createElement('div')
+    this.toastersContainer.id = `${toasterContainerId}-${getUniqueStringId()}`
```
portikM: Thought of using `userId` here instead but the catch with that approach is that `useId` needs to run in the `setup()` function which we have no way of ensuring is done right in all host apps. Random string should be fine here.

src/components/KToaster/KToaster.vue
```diff
@@ -81,23 +81,46 @@ const getToastIcon = (appearance?: ToasterAppearance): ToastIcon => {
 }
 </script>
 
-<style lang="scss" scoped>
-.k-toaster {
+<style lang="scss">
```
johncowen, line 84: Did you mean to remove the `scoped` here? Might be related to the new comment you added, but thought I'd check

portikM: Yup this is intentional. The `scoped` style block is still there, down below - this is global styles for shared container and instance wrapper.

src/components/KToaster/ToastManager.cy.ts
```diff
@@ -1,63 +1,97 @@
...
+    expect(container!.getAttribute('data-instance-count')).to.equal('3')
+
+    // Verify three separate instance wrappers exist
+    const wrappers = document.querySelectorAll(`[id^="${toasterWrapperPrefix}-"]`)
+    expect(wrappers.length).to.equal(3)
+
+    // Cleanup
+    toastManager1.destroy()
+    toastManager2.destroy()
+    toastManager3.destroy()
```
johncowen, line 96: Do we want to assert the container doesn't exist anymore here?

src/components/KToaster/ToastManager.cy.ts
```diff
@@ -1,63 +1,97 @@
...
-    toastManager.open('Test Toast')
+    // Destroy second instance
+    toastManager2.destroy()
 
-    const containers = document.querySelectorAll(`#${toastersContainerId}`)
-    expect(containers.length).equal(1)
+    // Verify container is now removed
+    const containerAfterDestroy = document.getElementById(toastersContainerId)
+    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+    expect(containerAfterDestroy).to.be.null
```
johncowen, line 71: Ah sorry I guess this is a test to make sure it totally cleans up right? Feel free to ignore my previous comment if so.

src/components/KToaster/KToaster.vue
johncowen: gotcha 👍

src/components/KToaster/ToastManager.cy.ts
portikM: Yeah exactly- this one specifically checks the cleanup is done right.

johncowen (approved):
In my mind I'm happy with this as is, I left a little comment offline that could be addressed in the future, but its a bit of a nit to be fair.

(FYI: I didn't install this and try it in our application)

You might also want to get approval from others, but personally I'd be fine with this, thanks again for the work here 👍

src/components/KToaster/ToastManager.ts
```diff
@@ -14,46 +15,80 @@ const toasterDefaults = {
...
+ * Get or create the shared toaster container element.
+ * Increments the reference count for tracking active instances.
+ */
+function getOrCreateSharedContainer(): HTMLElement {
+  let container = document.getElementById(toasterContainerId)
+
+  if (!container) {
+    container = document.createElement('div')
+    container.id = toasterContainerId
+    container.setAttribute('data-instance-count', '0')
```
kaiarrowood: suggestion: store attribute name in a `const` at top of file so you only need to change in one place.

portikM: Done

kaiarrowood (approved):
LGTM! 🚀

PR #3084, opened 2026-02-04
feat(table-data): extract data fetching related code out of k-table-data

jw-foss:
# Summary

- Extract logics related to fetching data out of `KTableData.vue` to a dedicated file.
- Organize sort and filter logics into the same file to make the code cleaner.

## The goal

This PR is made for exporting the data fetching logics to be reused and not reinventing the wheels.

### Why

Right now for components that renders a list of data with `KTableData`, when the component wants to:

- Get a hold of the list of data fetched from the remote server, `KTableData` will need be rendered into the DOM (maybe without being seen) no matter the data is empty or not. 
- Or when we need to switch the table view to another type of list view, `KTableData` will need to be there in the DOM to provide fetching abilities.

### What this PR does

Extracting the logics for sorting, pagination, fetching, into a sharable composable that can be called outside the component itself, so that we could use the composable with `KTableView` to render the table view and still be able to get a hold of the fetched data, pagination properties, sort logics with other type of view for instance a catalog view.

This can also benefit `KCatalog` because `KCatalog` and `KTableData` share the same API for fetching data.

### Furthermore

Since we want to keep the scope of this PR to be small, when this gets merged to the main branch, we will start to extracting data fetching common logics from `KCatalog` to build a view only component like `KCatalogView` for the business components to encapsulate a switchable list/catalog view component for sharing.

### Potential benefited counterparts

When you need the flexibility to do certain operation and you need to manage the state on your own.

### DIFF
--- a/src/components/KTableData/useTableFetcher.ts
+++ b/src/components/KTableData/useTableFetcher.ts
@@ -1,0 +1,558 @@
+import { computed, ref, useId, toValue } from 'vue'
+import { getInitialPageSize, DEFAULT_PAGE_SIZE } from '@/utilities'
+import useUtilities from '@/composables/useUtilities'
+
+import type { ComputedRef, Ref, UnwrapRef, MaybeRefOrGetter } from 'vue'
+import type { TableDataProps, TableColumnKey, TableDataHeader, SortColumnOrder, TableState, TableSortPayload, TablePaginationAttributes, TableColumnVisibility, TableColumnWidths, TablePreferences } from '@/types/table'
+import type { SwrvStateData } from '@/types/swrv'
+
+export type TableDataRecord = ReadonlyArray<Record<string, any>>
+export type OffsetType = string | number
+
+/**
+ * Utility type that wraps all properties of an object type with MaybeRefOrGetter
+ */
+export type MaybeRefOrGetterProps<T> = {
+  [K in keyof T]: MaybeRefOrGetter<T[K]>
+}
+
+export type TableSortProps<
+  Header extends TableDataHeader,
+  Data extends TableDataRecord,
+  Offset extends OffsetType,
+> = MaybeRefOrGetterProps<Pick<
+  TableDataProps<Header, Data, Offset>,
+  | 'initialFetcherParams' | 'clientSort' | 'tablePreferences'
+  | 'paginationAttributes' | 'sortHandlerFunction' | 'sortable' | 'headers'
+>>
+
+export type SortParams<T extends TableDataHeader> = {
+  sortColumnKey: TableColumnKey<T>
+  sortColumnOrder: SortColumnOrder
+}
+
+const DEFAULT_SORT_ORDER: SortColumnOrder = 'desc'
+
+export const useTableSort = <Header extends TableDataHeader, Data extends TableDataRecord, Offset extends OffsetType>(
+  {
+    tablePreferences,
+    initialFetcherParams,
+    clientSort,
+    paginationAttributes,
+    sortHandlerFunction,
+  }: TableSortProps<Header, Data, Offset>,
+  {
+    tableHeaders,
+    page,
+    offsets,
+    tableData,
+    debouncedRevalidate,
+  }: {
+    tableHeaders: ComputedRef<readonly Header[]>
+    page: Ref<number>
+    offsets: Ref<Array<Offset | null>>
+    tableData: Ref<Array<Data[number]>>
+    debouncedRevalidate: () => void
+  },
+) => {
+
+  type ColumnKey = TableColumnKey<Header>
+
+  const sortColumnKey = ref(toValue(tablePreferences)?.sortColumnKey || toValue(initialFetcherParams)?.sortColumnKey || '') as Ref<ColumnKey>
+  const sortColumnOrder = ref(toValue(tablePreferences)?.sortColumnOrder || toValue(initialFetcherParams)?.sortColumnOrder || DEFAULT_SORT_ORDER) as Ref<SortColumnOrder>
+  const initialSortHandled = ref<boolean>(!(sortColumnKey.value && toValue(clientSort))) // For clientSort tables, if sortColumnKey is set, that means we need to handle initial sort
+  const sortParams = computed(() => ({
+    sortColumnKey: sortColumnKey.value,
+    sortColumnOrder: sortColumnOrder.value,
+  })) as Ref<SortParams<Header>>
+
+  const { clientSideSorter: defaultClientSideSorter } = useUtilities()
+
+  const sortHandler = ({ sortColumnKey: columnKey, prevKey, sortColumnOrder: sortOrder }: TableSortPayload<ColumnKey>, sortEvtEmitter?: (params: TableSortPayload<ColumnKey>) => void): void => {
+    initialSortHandled.value = true
+
+    const header = tableHeaders.value.find((header) => header.key === columnKey) || {} as UnwrapRef<typeof tableHeaders>[number]
+    const { useSortHandlerFunction } = header
+
+    sortEvtEmitter?.({
+      prevKey,
+      sortColumnKey: columnKey,
+      sortColumnOrder: sortOrder,
+    })
+
+    page.value = 1
+
+    if (!sortColumnKey.value || columnKey !== sortColumnKey.value) {
+      offsets.value = [null]
+    }
+
+    sortColumnKey.value = columnKey
+    sortColumnOrder.value = sortOrder as SortColumnOrder
+    const sortHandler = toValue(sortHandlerFunction)
+    if (toValue(clientSort)) {
+      if (useSortHandlerFunction && sortHandler) {
+        const sorted = sortHandler({
+          key: columnKey,
+          prevKey,
+          sortColumnOrder: sortColumnOrder.value,
+          data: tableData.value,
+        })
+
+        // As `sortHandlerFunction` was marked as returning an array but we didn't use the return value
+        // before, we can keep the old behavior when nothing is returned but use the returned value if it exists.
+        if (sorted) {
+          tableData.value = [...sorted]
+        }
+      } else {
+        defaultClientSideSorter(columnKey, prevKey, sortColumnOrder.value, tableData.value)
+      }
+    } else if (!toValue(paginationAttributes)?.offset) {
+      debouncedRevalidate()
+    }
+  }
+
+  return {
+    sortColumnKey,
+    sortColumnOrder,
+    sortParams,
+    initialSortHandled,
+
+    sortHandler,
+  }
+}
+
+export type TablePaginationProps<
+  Header extends TableDataHeader,
+  Data extends TableDataRecord,
+  Offset extends OffsetType,
+> = MaybeRefOrGetterProps<Pick<
+  TableDataProps<Header, Data, Offset>,
+  | 'fetcher' | 'paginationAttributes' | 'hidePaginationWhenOptional' | 'hidePagination' | 'tablePreferences'
+>>
+
+export const useTablePagination = <Header extends TableDataHeader, Data extends TableDataRecord, Offset extends OffsetType>(
+  props: TablePaginationProps<Header, Data, Offset>,
+  tableData: Ref<Array<Data[number]>>,
+  response: Ref<Awaited<ReturnType<TableDataProps<Header, Data, Offset>['fetcher']>>>,
+) => {
+  const page = ref<number>(1)
+  const pageSize = ref<number>(getInitialPageSize(toValue(props.tablePreferences) ?? {}, toValue(props.paginationAttributes) ?? {}))
+  const offset = ref(null) as Ref<Offset | null>
+  const offsets = ref([]) as Ref<Array<Offset | null>>
+  const hasNextPage = computed(() => {
+    const res = response.value
+    return (res?.pagination && 'hasNextPage' in res.pagination) ? res.pagination.hasNextPage || false : true
+  })
+
+  const previousOffset = computed((): Offset | null => offsets.value[page.value - 1] || null)
+  const nextOffset = computed(() => {
+    const res = response.value
+    const paginationAttrs = toValue(props.paginationAttributes)
+
+    if (paginationAttrs?.offset) {
+      if (!res?.pagination?.offset) {
+        return null
+      } else {
+        return res?.pagination.offset
+      }
+    }
+
+    return null
+  })
+
+  const total = computed(() => {
+    const res = response.value
+    const paginationAttrs = toValue(props.paginationAttributes)
+    return paginationAttrs?.totalCount || res?.total || res?.data?.length || 0
+  })
+
+  const showPagination = computed((): boolean => {
+    const fetcherValue = toValue(props.fetcher)
+    const hidePaginationValue = toValue(props.hidePagination)
+    const paginationAttrs = toValue(props.paginationAttributes)
+    const hidePaginationWhenOptionalValue = toValue(props.hidePaginationWhenOptional)
+
+    // if fetcher is not defined or hidePagination is true, don't show pagination
+    if (!fetcherValue || hidePaginationValue) {
+      return false
+    }
+
+    const minPageSize = paginationAttrs?.pageSizes?.[0] ?? DEFAULT_PAGE_SIZE
+
+    // this logic is built around min page size so that pagination doesn't disappear when a higher value is selected and hidePaginationWhenOptional is true
+    if (hidePaginationWhenOptionalValue && page.value === 1) {
+      if (!paginationAttrs?.offset) {
+        // if using cursor-based pagination, hide pagination when number of items is less than min page size
+        return total.value > minPageSize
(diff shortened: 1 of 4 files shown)


src/components/KTableData/useTableFetcher.ts
```diff
@@ -0,0 +1,558 @@
...
+  const page = ref<number>(1)
+  const pageSize = ref<number>(getInitialPageSize(toValue(props.tablePreferences) ?? {}, toValue(props.paginationAttributes) ?? {}))
+  const offset = ref(null) as Ref<Offset | null>
+  const offsets = ref([]) as Ref<Array<Offset | null>>
+  const hasNextPage = computed(() => {
+    const res = response.value
+    return (res?.pagination && 'hasNextPage' in res.pagination) ? res.pagination.hasNextPage || false : true
+  })
+
+  const previousOffset = computed((): Offset | null => offsets.value[page.value - 1] || null)
```
Justineo, line 147: Reviewed by Codex: This regresses offset-based pagination. `previousOffset` depends on `offsets[page - 1]`, but after this refactor the returned `res.pagination.offset` is no longer pushed into `offsets`; `offsets` is only seeded during `initData()`. After clicking Next from page 1, `page` is 2 while `offsets` still only contains the initial `null`, so the Previous button remains disabled and multi-page back navigation uses the wrong offset. Please preserve the old response watcher behavior that appended each returned offset, and add a test that clicks Next and then Previous for `paginationAttributes.offset`.

jw-foss: Updated per suggestion

adamdehaven (changes requested):
There are a few glaring issues with this pull request:

1. What’s the benefit of extracting the fetching logic that is only utilized in the component? I don’t see reasoning in the pull request description
2. The new composable mixes type imports, exports, and doesn’t follow a clear pattern for either 
3. 30 exported properties from the composable arguably make this less readable/maintainable than the original code
4. Explicit type casting rather than proper interfaces
5. Virtually no comments in the code 
6. Zero test coverage over the composable

jw-foss:
> There are a few glaring issues with this pull request:
> 
> 1. What’s the benefit of extracting the fetching logic that is only utilized in the component? I don’t see reasoning in the pull request description
> 2. The new composable mixes type imports, exports, and doesn’t follow a clear pattern for either
> 3. 30 exported properties from the composable arguably make this less readable/maintainable than the original code
> 4. Explicit type casting rather than proper interfaces
> 5. Virtually no comments in the code
> 6. Zero test coverage over the composable

Hi @adamdehaven sorry for not marking this WIP, this is still under development, I only used this for creating PR releases that we could test out the "Result". I will add **WHY** and **comments** to the code, also I was going to send a message in the Slack channel for what I am doing here.

> 4. Explicit type casting rather than proper interfaces

I only moved code and reorganized them into different composables, the typing were kept as-is.

| Before | After |
|--------|--------|
| CleanShot 2026-04-22 at 18 04 40@2x | CleanShot 2026-04-22 at 18 04 48@2x |

You can see in the screenshot that type casting is not invented by me in this PR here :bow:

Justineo:
@adamdehaven Let me fill in some context.
This improvement has been on my wishlist for Kongponents for quite a while. I had also discussed possible implementation approaches with Jeremy before, and I think moving the list data fetching logic out of the component and into a composable makes a lot of sense. It could solve quite a few issues we’ve run into on list pages where external parts of the page depend on the table’s state, especially after the earlier empty state revamp.

From what I understand, this PR mainly extracts the non-KTableView logic in KTableData into a public composable. That said, I feel we could probably reduce the API surface a bit, since some parts may not need to be split out so granularly. Also, similar to the component side, we could move the type definitions into separate files to make the composable itself a bit clearer. @jw-foss

adamdehaven:
> sorry for not marking this WIP

@jw-foss No worries. Would you mind putting the PR back to Draft state?

It was marked as Open and ready for review.

src/components/KTableData/useTableFetcher.ts
johncowen, line 1: In the Slack post to go with this the following was said:

> Fetching, pagination, sorting, loading, empty, and error state move into a reusable logic layer.

I was wondering exactly what that "reusable logic layer" was. Is this `useTableFetcher` what is meant be a "reusable logic layer"?

johncowen:
FYI: I was asked to look at this from a Slack thread, I had a quick look at the description of the PR and an even quicker look to get an idea of what the "reusable logic layer" was and leave a comment.

But I just wanted to say at a general level, I agree with the idea that this logic shouldn't be wrapped up in a Table rendering component. In my opinion, the Table component should only ever be about rendering a table and nothing more. KTableData tries to do far too much at the moment and thats why its nigh on impossible to use for anything past first steps.

> This PR is made for exporting the data fetching logics to be reused and not reinventing the wheels.

👍 Whilst I would go further myself (😅 ) this sounds like a first step towards what I think we should be doing.

Anything about data loading, pagination, sorting etc should have nothing to do with a table. At the very least this PR extracts this sort of stuff out of the KTableData component and it sounds like then its possible to use it to fetch some JSON and render that into a table.

Although, I have a question:

Q: How come you can't use KTableView?

FYI: we use KTableView in mesh and wrap it in really simple:

- pagination (i.e. `?page=1` etc plus KPagination for rendering)
- loading state (`typeof data === 'undefined'` plus KSkeleton) 
- error state (`error !== 'undefined'` plus KEmptyState)
- loading data can easily use fetch or WebSocket or ServerSideEvents etc etc (we happen to use a "state-while-revalidate" plus "request sharing" lib we use, but its not tied to this at all which is the point).

As I said above no data loading should be wrapped up in a Table component. That just doesn't make sense to me (which are positive first steps I see in this PR)

Anyway @jw-foss thanks for initiating this and putting the PR up 👍 I'll take a better look at the code tomorrow

adamdehaven:
> Anything about data loading, pagination, sorting etc should have nothing to do with a table. At the very least this PR extracts this sort of stuff out of the KTableData component and it sounds like then its possible to use it to fetch some JSON and render that into a table.
> ...
> As I said above no data loading should be wrapped up in a Table component. That just doesn't make sense to me (which are positive first steps I see in this PR)

This built-in functionality is the exact purpose of `KTableData` so that data fetching across an application can be easier and essentially drop-in.

If an implementation does not want this functionality, they should use the display-only `KTableView` instead.

johncowen:
> If an implementation does not want this functionality, they should use the display-only KTableView instead.

Totally, this is why I asked @jw-foss the question, lets see what he says 👍 

Apart from that and I'm guessing is part of @jw-foss intention here

> This built-in functionality is the exact purpose of KTableData so that data fetching across an application can be easier and essentially drop-in.

If KTableData is so difficult to use for anything other than the most simple of cases (reportedly, listening to other folks as far as I understand it) , why do we have it? Why not just have a separate logic layer (i.e. data loading, loading/error states, pagination sorting etc) and just use a component that only does rendering.

adamdehaven:
> If KTableData is so difficult to use for anything other than the most simple of cases (reportedly, listening to other folks as far as I understand it) , why do we have it?

Again, this is an assumption you're making.

`KTableData` is utilized hundreds of places in our apps and shared component libraries without issue. 

Yes, I'm sure there are some use cases where it shouldn't be part of the current implementation, but this isn't the majority.

johncowen:
> Again, this is an assumption you're making.

I said `as far as I understand it`:

> reportedly, listening to other folks as far as I understand it

Its how I understand it from the reports over the years of people having issues with it. Yes, maybe it is being used in "hundreds" of places. But it feels to me like its not ideal, otherwise why do we need this PR? (rhetorical: Adam I'm not asking you to answer that)

Now - please can we stick to the important things. The PR is here. There is a reason why the PR was made and @jw-foss and @Justineo (and others in the past including myself) are reporting this stuff and trying to improve it for all of us.

---

I've asked several questions that I'm waiting for @jw-foss to get a chance to answer and explain his thinking:

> I was wondering exactly what that "reusable logic layer" was. Is this useTableFetcher what is meant be a "reusable logic layer"?

> Q: How come you can't use KTableView?

> Why do we have it? (edit: "it" being KTableData, if you think its difficult to use and therefore are splitting out the data loading functionality)

Please give @jw-foss a chance, lets discuss the important stuff.

---

@jw-foss and @Justineo, sorry about this but its getting close to me feeling like I shouldn't participate further, I don't want the PR to go off on a tangent. If I get to that point I'll have to bow out of this thread and take it up elsewhere. If you could take a look at my questions that would be very much appreciated!

src/components/KTableData/useTableFetcher.ts
jw-foss: It should be, for now, in this PR I extracted the fetching related logic into a dedicated composable so that we could share them within the codebase, for controlling the entire lifecycle of the data.

Justineo:
> If an implementation does not want this functionality, they should use the display-only `KTableView` instead.

We do need this functionality. The point is not to avoid what `KTableData` provides, but to make that same fetching/pagination/sorting logic available outside of the table component.

We would still use `KTableView` for rendering. Extracting the logic into a composable gives us a more flexible composition model: the composable owns the data lifecycle, `KTableView` renders the table, and other parts of the page can share the same data/loading/error/pagination/sorting state.

So this is not “we do not want `KTableData` functionality”. It is “we want that functionality without requiring the table component itself to own the state”.

jw-foss:
@johncowen 

> Q: How come you can't use KTableView?

The answer for this is: I only created this PR/idea **because I want to utilize KTableView.**.
Here’s my thinking:

- KTableView did a great job extracting the rendering logic from KTableData, but we still need a way to manage state.
- KTableData keeps its data management logic internally and syncs part of that data back to the parent component through events, as I mentioned in the Slack thread. What I want is both: KTableView for rendering, and the management logic from KTableData. Since that logic already exists, I thought we could simply extract and export it as composables.
- KCatalog follows the same pattern as KTableData, using the same APIs such as fetch key and fetcher, so it could also benefit from this shared composition approach.

By extracting the logic into reusable composables, we would get:

- KTableData can remain intact.
- KCatalog can reuse the extracted fetching logic and avoid duplication where it exists.
- Any other place that needs to control the lifecycle of data fetching can reuse the same logic instead of reimplementing it.
- Better code organization by separating concerns like data fetching, sorting, and pagination.

Justineo:
> If KTableData is so difficult to use for anything other than the most simple of cases (reportedly, listening to other folks as far as I understand it)

It works mostly fine, but since we have so many tables, even a small percentage of issues can end up feeling like a lot.

johncowen:
@jw-foss / @Justineo - I've had a brief read of this and unless I'm mistaken this is exactly what I was hoping/expecting to hear. Thankyou!

> FYI: we use KTableView in mesh and wrap it in really simple:

FYI: The approach I explained above that we use in Mesh also uses KTableView, _not_ KTableData because we found it too difficult to use for anything but the most simple of usecases. So it sounds to me like you are encountering the same issues that we and other folks have encountered and this solution here solves those issues for you. I believe all three of us are pretty much on the same page.

I have a busy day today, but I'll try and come back later on today once I've digested your responses in more detail, but just wanted to say I very much like that way this conversation is going, lets keep it on track 👍 

I can see this is a WIP PR so I don't think there is any particular urgency here, but if there is please let me know

johncowen:
Ok, having read over your responses and looked at the PR in a bit more detail now, I think I understand what you are doing here, and I agree with the approach.

I understand the approach is (please correct me if I misunderstood): 

- Pretty much copy/paste (i.e. extract) the data related stuff out of `KTableData` and put it in `useTableFetcher`. I'm guessing the important thing here is to do it like for like - i.e. extract it and keep it as close as possible to as it was when within `KTableData`. I can see @Justineo has pointed an LLM at this and because of that I'm assuming that they are like for like.
- In your application itself (i.e. not in this PR) you will use `useTableFetcher` and feed its return data/properties into a `KTableView`
- (I know you get this, but just to say) The most important thing here is that `KTableData` functions exactly as before but you have access to the "data" parts.

If I understood right, this PR on its own seems perfectly reasonable to me, and more or less I agree with the approach  considering how things were before this PR.

I do have some slightly unrelated concerns, but they are mostly historical and not really related to the work you've done in this PR. But I do have some questions about this PR:

I noticed that `KTableData` internally isn't quite "a call to `useTableFetcher` and pass the result to `KTableView`" i.e. there is a bit of extra code there doing "stuff". I don't have enough context to understand exactly what this "stuff" is doing, but I'm sure you folks do. 

Q: If I were to use `useTableFetcher` plus `KTableView` without using `KTableData` in my application, am I missing functionality that I would get if I just use `KTableData`?

- If yes, Q: should we avoid that somehow?
- If no, Q: should we deprecate KTableData if the functionality is covered by `useTableFetcher` plus `KTableView`

Q: I noticed there a pieces of code that existed in `KTableData` and are now in `useTableFetcher` that seem to be dependent on the shape of a HTTP response. Should these details be present in this code? (caveat: I get that we are doing like for like at the moment, so this question is more theoretical)

 I noticed that the interface for KTableView is very large (roughly 44 attributes to set):

https://github.com/Kong/kongponents/blob/6a7cde33ce780dd9e24cfb8882a1826febc7959d/src/components/KTableData/KTableData.vue#L3-L47

I would guess that, not only would this make using KTableData (seems to also have a large amount of props) and KTableView difficult to understand and use, but it probably makes it difficult to author the component itself without introducing further issues.

Q: Will you have to set all of these attributes (or a large amount of them) whenever you use `useTableFetcher` plus `KTableView`?

Q: Do you think it might be an improvement to make these interfaces simpler by reducing the amount of properties. (caveat: I get that we are doing like for like at the moment, so this question is more theoretical)

At a super high-level level:

Q: I felt like there is a lot of code to do pagination, sorting, and data fetching (considering it also "uses/is tightly coupled to" a 3rd library to actually fetch data). To my eyes it looks large and complex for implementing this sort of stuff. What's your take on this? (caveat: I get that we are doing like for like at the moment, so this question is more theoretical)

Lmk!

jw-foss:
Hi @johncowen,

Thanks for your comments here. Before I started to address/answer your questions in the thread, I wanted to emphasis the following:

The main goal for this PR is to **Separate the logics from KTableData for reuse**, rather than refactor the entire component.

>Q : If I were to use useTableFetcher plus KTableView without using KTableData in my application, am I missing functionality that I would get if I just use KTableData?

I left some code inside `KTableData` on purpose, but yeah, I think some of them should be moved to `useTableData` to prevent losing some functionalities, thanks for bring that up, and this PR was a on going one, so I will address that for sure.

 > Q: I noticed there a pieces of code that existed in KTableData and are now in useTableFetcher that seem to be dependent on the shape of a HTTP response. Should these details be present in this code? (caveat: I get that we are doing like for like at the moment, so this question is more theoretical)

I think to keep the response in a specific shape made sense, we want this component to be used by anyone, so the fetcher might need to have some transformation overhead to align with the data model useKTableData uses otherwise we will need to add params for users to pass in adapters for handling pagination, sort, and stuff.

> I noticed that the interface for KTableView is very large (roughly 44 attributes to set):  Will you have to set all of these attributes (or a large amount of them) whenever you use useTableFetcher plus KTableView?

Yeah I guess KTableView/KTableData has that "problem", I guess we could simplify the APIs by giving them default values, if you do not need that many functionalities you can just ignore it by default, only pass those you need.
I will try to keep it as simple as possible w/o breaking the functionalities.

> Do you think it might be an improvement to make these interfaces simpler by reducing the amount of properties. (caveat: I get that we are doing like for like at the moment, so this question is more theoretical)

Yes, but again, that was not the purpose of this PR so I think we should be able to create another dedicated PR/Epic for this work.

> felt like there is a lot of code to do pagination, sorting, and data fetching (considering it also "uses/is tightly coupled to" a 3rd library to actually fetch data). To my eyes it looks large and complex for implementing this sort of stuff. What's your take on this? (caveat: I get that we are doing like for like at the moment, so this question is more theoretical)

Yes, I did what I could to organize the code into the same block like `useSort` `usePagination` to keep the readability as best as it could be. But the amount of features this component supports requires code to be this complex. 

To simplify the component and the fetcher and other hooks would be another topic, for what we (the AI Manager and Gateway Manager) need right now is we could expose this composable in the MFEs to allow us to control the full lifecycle of the internal state for some special needs. By exposing the composable without causing regressions would be the top priority for this PR, thanks again for your attention and inputs with this 🙏 .

johncowen:
Hey @jw-foss 

Thanks for your in-depth and thoughtful answers 🙇 

I had a brief read and think I totally agree with all you are saying, wanted to ping of a response real quick because I know we are on different timezones, but I will re-read to make sure I've understood.

Totally get that this PR is ongoing, and I think you've picked out the question that was relevant to this PR (I think I totally get your scope here), and you also picked out the things that was not for this PR.

Anyway, as I said just wanted to ping off a quick message before the end of your day, lets catch up later 👍

johncowen:
(sorry for the noise, I just remembered about this and remembered that I said I'd come back with more comment, and I totally forgot! 🤦 sorry! I don't think I had much to add anyway, but please don't let me hold you up if I don't get back here in time, from your last comments I think we are on the same page)

PR #3092, opened 2026-02-12
feat: kfiltergroup base functionality [khcp-19485, khcp-19482]

Chalks:
**Please note that this is **NOT** being merged into main in this PR, this is merging into `feat/filter-group-khcp-19257` in an attempt to reduce review burden**

Addresses [khcp-19485](https://konghq.atlassian.net/browse/KHCP-19485) and [khcp-19482](https://konghq.atlassian.net/browse/KHCP-19482).

# Summary

`KFilterGroup` is a new component that is described at length in [the engineering portion of this doc](https://docs.google.com/document/d/1bjwdNJya58oj-1qFLkbzeW5Srkox5s1WlKKxfd6QN5M/edit?tab=t.2yvqby8z6dj) and has designs available [in this figma](https://www.figma.com/design/Yze0SWXl5nKjR0rFdilljK/Components?node-id=8825-19&p=f&t=cZXWCKh4YgQDcpFW-0)

## How to review

Sorry this PR is a bit long. However, each commit _should_ be fairly atomic, so if you would like to review each of those individually that shouldn't be too bad. If an individual component doesn't make sense to you though, it's probably best to review it within the context of the whole thing.

## How to test

* pull down the code
* run `pnpm sandbox:dev`
* go to http://localhost:5173/kongponents/#/filtergroup
* play around with the component

Please check for:

* reasonable keyboard controls (mostly working, the popover content being placeholder makes it a little hard to get perfect)
* pixel perfect styles (compare "What's in this PR" screenshot/figma to the sandbox)
* reasonable behavior (tooltip behavior for truncated elements, adding a filter but then canceling immediately, etc)

### Remaining work (todo in future PRs)

* filter popover content and associated slots
* documentation
* fleshed out sandbox page
* finalized tests on all components (there are some `.skip`'d in this PR)
* slight css tweaks on the popover positioning
* update KTableData to support a "search" slot in the toolbar (I'm not sure if we captured this well in the brief, but it was a conversation we went through while preparing all this)

#### (from figma): The eventual goal
Screenshot 2026-02-12 at 3 33 32 PM

#### (from figma): What's in this PR

Screenshot 2026-02-12 at 3 34 40 PM

#### (from figma): What's NOT in this PR

Screenshot 2026-02-12 at 3 35 11 PM

#### (from this PR): Working in the sandbox

Note, the last bit of this video includes using the keyboard to tab through elements and enter to select them. It's partially not-quite-working-right because the content in the popover isn't included in this PR.

https://github.com/user-attachments/assets/ec073e0a-9d98-49b8-8ff5-0eadfb6f975a

### DIFF
--- a/src/components/KFilterGroup/KFilterGroup.vue
+++ b/src/components/KFilterGroup/KFilterGroup.vue
@@ -46,6 +46,11 @@ const {
 } = defineProps<FilterGroupProps>()
 const selection = defineModel<FilterGroupSelection>({ required: true })
 
+/**
+ * tracks any filter that's currently open. filters are only rendered as pills
+ * if they have a selection, are pinned, or are active (e.g. just added but not
+ * yet applied)
+ */
 const activeFilterKey = ref<string | undefined>()
 
 /**
@@ -151,10 +156,11 @@ watch(selection, (newSelection) => {
   // mutation of the selection.
   Object.entries(newSelection)
     .forEach(([key, value]) => {
-      const addedIndex = unpinnedSelectionKeys.value.indexOf(key)
+      // if the key is pinned we can short circuit as pinned filters are always displayed
       const pinned = filters[key]?.pinned
       if (pinned) return
 
+      const addedIndex = unpinnedSelectionKeys.value.indexOf(key)
       if (addedIndex !== -1 && value === undefined) {
         // remove undefined selections if it's already in the list
         unpinnedSelectionKeys.value = [
--- a/src/types/filter-group.ts
+++ b/src/types/filter-group.ts
@@ -17,7 +17,7 @@ export type FilterOperator = 'eq' | 'neq' | 'contains' | 'exists' | 'lt' | 'lte'
 
 export interface Filter {
   /**
-   * The filter's label, displays in the filter selection dropdown and in the pill
+   * Displays in the filter selection dropdown and in the pill
    */
   label: string
 
@@ -29,7 +29,7 @@ export interface Filter {
 
   /**
    * The list of options the user can choose from in the filter. If unset, the
-   * filter renders an input box instead
+   * filter renders a text input instead
    * @default undefined
    */
   options?: SelectEntry[]
@@ -48,7 +48,7 @@ export interface Filter {
   pinned?: boolean
 
   /**
-   * The placement this filter's KPop should use
+   * The placement this filter's popover should use
    * @default 'bottom-start'
    */
   placement?: PopPlacement
@@ -63,7 +63,7 @@ export interface FilterSelection {
   /**
    * The value input by the user
    */
-  value: string | string[] | number | number[] | boolean | boolean[]
+  value: string | string[] | number | number[]
 
   /**
    * The user facing display string for this selection. Displays in the pill and
@@ -92,7 +92,7 @@ export interface FilterPillProps {
   filter: Filter
 
   /**
-   * Whether to initialize the pill with its KPop open
+   * Whether to initialize the pill with its popover open
    * @default false
    */
   initOpen?: boolean
@@ -121,7 +121,7 @@ export interface FilterGroupProps {
 
 export interface FilterGroupEmits {
   /**
-   * When a user triggers "apply" in a filter's KPop, this is fired. Provides
+   * When a user triggers "apply" in a filter's popover, this is fired. Provides
    * the key of the filter that was triggered and the KFilterGroup's new
    * selection object.
    */
@@ -135,13 +135,13 @@ export interface FilterGroupEmits {
   clear: [clearedFilterKey: string, selection: FilterGroupSelection]
 
   /**
-   * When a user triggers a filter's KPop to open, this is fired. Provides the
+   * When a user triggers a filter's popover to open, this is fired. Provides the
    * key of the filter that was triggered.
    */
   open: [openedFilterKey: string]
 
   /**
-   * When a user triggers a filter's KPop to close, this is fired. Provides the
+   * When a user triggers a filter's popover to close, this is fired. Provides the
    * key of the filter that was triggered.
    */
   close: [closedFilterKey: string]


src/components/KFilterGroup/KFilterGroup.vue
```diff
@@ -0,0 +1,160 @@
...
+    })
+}, { immediate: true, deep: true })
+</script>
+
+<style lang="scss" scoped>
+.k-filter-group {
+  align-items: center;
+  display: flex;
+  flex-wrap: wrap;
+  gap: $kui-space-40;
```
adamdehaven: For all design tokens, we must utilize the CSS variable along with the fallback. Applies to whole PR
```suggestion
  gap: var(--kui-space-40, $kui-space-40);
```

src/types/filter-group.ts
```diff
@@ -0,0 +1,33 @@
+import type { SelectEntry } from './select'
+import type { PopPlacement } from './popover'
+
+export type Operator = 'eq' | 'neq' | 'contains' | 'exists' | 'lt' | 'lte' | 'gt' | 'gte'
```
adamdehaven: nit: can we be a touch more specific here `FilterOperator`

src/types/filter-group.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface Filter {
```
adamdehaven, line 18: nit: please add TSDoc comments above each property (all types)

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -0,0 +1,138 @@
...
+      />
+    </template>
+
+    <template #content>
+      <div data-testid="filter-pill-content">
+        <div @click="onApply">
+          @apply with fake selection
+        </div>
+        <div @click="onCancel">
+          cancel
```
adamdehaven, line 27: Not sure I understand the text here?

And shouldn't these be actual buttons?

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -0,0 +1,138 @@
...
+
+<script setup lang="ts">
+import { computed, onMounted, ref, useAttrs, useTemplateRef } from 'vue'
+import type { Filter, FilterSelection } from '@/types'
+import KPop from '@/components/KPop/KPop.vue'
+import InteractivePill from './InteractivePill.vue'
+
+// we want to manually specify where the attrs get inherited to
+defineOptions({ inheritAttrs: false })
+const $attrs = useAttrs()
```
adamdehaven: nit: you don't need the dollar sign here

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -0,0 +1,138 @@
...
+
+const {
+  filter,
+  initOpen = false,
+  selection = undefined,
+} = defineProps<{
+  filter: Filter
+  initOpen?: boolean
+  selection?: FilterSelection
```
adamdehaven: Can the type interface here be defined in the types file with the others?

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -0,0 +1,138 @@
...
+}>()
+
+const isOpen = ref(false)
```
adamdehaven: nit: always add explicit types
```suggestion
const isOpen = ref<boolean>(false)
```

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -0,0 +1,138 @@
...
+const filterPopper = useTemplateRef('filterPopper')
```
adamdehaven: issue(?): I could be misremembering but we've ran into issues in the past where the literal constant name and the template ref string were identical. 

You should rename, something like this and then update usage
```suggestion
const filterPopperRef = useTemplateRef('filterPopper')
```

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -0,0 +1,138 @@
...
+    case 'lt':
+      return ' < '
+    case 'lte':
+      return ' ≤ '
+    case 'gt':
+      return ' > '
+    case 'gte':
+      return ' ≥ '
+    default:
+      return undefined
```
adamdehaven: why undefined instead of `=`?

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -0,0 +1,138 @@
...
+const emit = defineEmits<{
+  (e: 'apply', selection?: FilterSelection): void
+  (e: 'close'): void
+  (e: 'open'): void
+  (e: 'clear'): void
+}>()
+
+const onApply = () => {
+  closeFilter()
+  emit('apply', { operator: 'eq', value: 'apple', text: 'Apple' })
```
adamdehaven, line 114: This shouldn't be hard-coded

src/components/KFilterGroup/FilterSelector.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+  <KDropdown
+    ref="filterSelectorRef"
+    :items="items"
+    @toggle-dropdown="onToggle"
+  >
+    <template #default>
+      <InteractivePill
+        data-testid="filter-selector"
+        label="Add filter"
```
adamdehaven: question: should this string be a prop on the main component that can be customized?

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,281 @@
+<template>
+  <div>
```
adamdehaven, line 2: question: is the pill itself clickable?

src/components/KFilterGroup/FilterSelector.vue
```diff
@@ -0,0 +1,70 @@
...
+        :pill-focus="pillFocus"
+        @click.prevent.stop
+        @trigger="onTrigger"
+      >
+        <template #openIcon>
+          <AddIcon />
```
adamdehaven:
```suggestion
          <AddIcon decorative />
```

src/components/KFilterGroup/FilterSelector.vue
```diff
@@ -0,0 +1,70 @@
...
+</template>
+
+<script setup lang="ts">
+import { computed, useTemplateRef, ref } from 'vue'
+import type { DropdownItem, FilterGroupFilters } from '@/types'
+import KDropdown from '@/components/KDropdown/KDropdown.vue'
+import InteractivePill from './InteractivePill.vue'
+import { AddIcon } from '@kong/icons'
+
+const filterSelectorRef = useTemplateRef('filterSelectorRef')
```
adamdehaven: Same comment here on template fee const naming

src/components/KFilterGroup/FilterSelector.vue
```diff
@@ -0,0 +1,70 @@
...
+    value: key,
+    onClick: () => onItemClick(key),
+  })))
+
+const emit = defineEmits<{
+  (e: 'selectFilter', filterKey: string): void
+}>()
+
+const onItemClick = (key: string) => {
+  emit('selectFilter', key)
```
adamdehaven: Could these just be "select"? We already know it's a filter component

src/components/KFilterGroup/FilterSelector.vue
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const onToggle = (open: boolean) => {
```
adamdehaven, line 58: nit: add TSDoc comments for methods in PR

src/components/KFilterGroup/KFilterGroup.vue
```diff
@@ -0,0 +1,160 @@
...
+      v-for="key in visibleFilterKeys"
+      :key="`filter-pill-${key}`"
+      :data-testid="`filter-group-pill-${key}`"
+      :filter="filters[key]"
+      :init-open="key === activeFilterKey"
+      :selection="selection[key]"
+      @apply="(selected) => onFilterApply(key, selected)"
+      @clear="() => onFilterClear(key)"
+      @close="() => onFilterClose(key)"
+      @open="() => onFilterOpen(key)"
```
adamdehaven: Why do you need `() =>` for these?

src/components/KFilterGroup/KFilterGroup.vue
```diff
@@ -0,0 +1,160 @@
...
+    ...acc,
+    [key]: filters[key],
+  }), {}))
+
+const onSelectFilter = (filterKey: string) => {
+  setTimeout(() => {
+    // we want to give the select dropdown time to finish animating its close
+    // before we insert a new filter with a competing popover already open.
+    activeFilterKey.value = filterKey
+  }, 200)
```
adamdehaven: Is the animation timing tied to a token? If yes, we could use it here

src/components/KFilterGroup/KFilterGroup.vue
```diff
@@ -0,0 +1,160 @@
...
+
+  emit('apply', filterKey, selection.value)
+}
+
+watch(selection, () => {
+  // in order to preserve filter add order, we have to watch the selection. As
+  // it changes we add/remove keys from the `unpinnedSelectionKeys` list. This
+  // allows us to respond to both our mutation of the selection and any host app
+  // mutation of the selection.
+  Object.entries(selection.value)
```
adamdehaven: issue: don't reference the ref directly, utilize the param from the watcher
```suggestion
watch(selection, (newSelection) => {
  // in order to preserve filter add order, we have to watch the selection. As
  // it changes we add/remove keys from the `unpinnedSelectionKeys` list. This
  // allows us to respond to both our mutation of the selection and any host app
  // mutation of the selection.
  Object.entries(newSelection)
```

...etc.

src/components/KFilterGroup/KFilterGroup.vue
Chalks: ah dangit, I knew that and I forgot to do it. will adjust.

src/components/KFilterGroup/FilterPill.vue
Chalks: this is the one part that's just actual placeholder. I needed a way to trigger these events to get the rest of it working.

src/components/KFilterGroup/FilterSelector.vue
Chalks: yes, if only because otherwise it's impossible to localize it.

src/components/KFilterGroup/KFilterGroup.vue
adamdehaven: We have a stylelint plugin in the works that would fix it for you 🙃 but it hasn't merged yet in another repo

src/components/KFilterGroup/KFilterGroup.vue
Chalks: old habit. I always misread `@foo="onFoo(key)"` as "use the result of `onFoo(key)` as the `@foo` handler" even though it doesn't actually work that way. Will adjust.

src/components/KFilterGroup/KFilterGroup.vue
Chalks: hmm, there is `KUI_ANIMATION_DURATION_20` which is `0.2s`. I'm not sure if it's more clear to translate that to `200` though.

src/components/KFilterGroup/InteractivePill.vue
Chalks: yes. I added this div wrapper as the root element because if the root element is `<KTooltip>` instead and then some parent component wraps this in a `<KPop>`, the popover goes to the top left corner of the screen because that's where the `<KTooltip>` is when it's not visible.

I don't like this solution though. Is there a better way to handle those two components and how they interact?

src/components/KFilterGroup/FilterPill.vue
Chalks: this is related to the other part you called out [here](https://github.com/Kong/kongponents/pull/3092#discussion_r2801458787)... it's placeholder. Anything inside the filter popover I left for the next PR. This snippet is hardcoded just so you can interact with the component in review.

src/components/KFilterGroup/FilterPill.vue
Chalks: but I like how fancy it looks (and I spent a lot of time in vue2. lol). Will adjust.

src/components/KFilterGroup/FilterPill.vue
Chalks: yes. I was under the impression that we were only defining/exporting types for props that the end user interacts with (i.e. just `KFilterGroup`), but I don't mind putting all of it there.

src/components/KFilterGroup/InteractivePill.vue
adamdehaven: I was asking bcc if clickable I was going to ask should it be a button?

src/components/KFilterGroup/FilterPill.vue
Chalks: Could we add [vue/require-typed-ref](https://eslint.vuejs.org/rules/require-typed-ref) to @kong/eslint-config-kong-ui?

src/components/KFilterGroup/FilterPill.vue
Chalks: the delimiter is used by `<InteractivePill>` which has a default when this is `undefined`. But... yeah, I could have it be explicit. (the default is `: ` fwiw)

src/components/KFilterGroup/InteractivePill.vue
Chalks: oh hmm... maybe. I always default to div just because how easy it is to style. I'll look into it in the morning though and see if that makes more sense.

src/components/KFilterGroup/FilterPill.vue
Chalks: I think we did. I can't remember what the actual issue was though. :|

src/components/KFilterGroup/FilterPill.vue
adamdehaven: I actually don't think it would help here (see the `number` example at the bottom of that page) 


Since you're initializing as a boolean, it's inferred, but in our component repos we try to explicitly type to guard against someone changing the init value later and unknowingly causing a breaking change

src/components/KFilterGroup/FilterSelector.vue
```diff
@@ -0,0 +1,70 @@
...
+  >
+    <template #default>
+      <InteractivePill
+        data-testid="filter-selector"
+        label="Add filter"
+        :pill-focus="pillFocus"
+        @click.prevent.stop
+        @trigger="onTrigger"
+      >
+        <template #openIcon>
```
portikM:
```suggestion
        <template #open-icon>
```
nit: we use `kebab-case` for slots throughout the repo

src/components/KFilterGroup/FilterSelector.vue
```diff
@@ -0,0 +1,70 @@
...
+import { computed, useTemplateRef, ref } from 'vue'
+import type { DropdownItem, FilterGroupFilters } from '@/types'
+import KDropdown from '@/components/KDropdown/KDropdown.vue'
+import InteractivePill from './InteractivePill.vue'
+import { AddIcon } from '@kong/icons'
+
+const filterSelectorRef = useTemplateRef('filterSelectorRef')
+const isOpen = ref(false)
+
+const pillFocus = ref(false)
```
portikM:
```suggestion
const isOpen = ref<boolean>(false)

const pillFocus = ref<boolean>(false)
```

suggestion: explicit types

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,281 @@
...
+          @keydown.enter.prevent.stop="onClear"
+        >
+          <div
+            class="clear-focus-highlight"
+            data-testid="interactive-pill-clear-focus"
+          >
+            <slot name="clearIcon">
+              <CloseIcon
+                decorative
+                size="16"
```
portikM: please use `KUI_ICON_SIZE_*` token here

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,281 @@
...
+        </div>
+        <div
+          v-else
+          class="pill-icon open-icon"
+          data-testid="interactive-pill-open-icon"
+        >
+          <slot name="openIcon">
+            <ChevronDownIcon
+              decorative
+              size="16"
```
portikM: same here, please use icon size token

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,281 @@
...
+
+const onClear = () => {
+  emit('clear')
+  if (browserClearFocus.value) {
+    clearRef.value?.blur()
+  }
+}
+</script>
+
+<style lang="scss" scoped>
```
portikM, line 191: issue: throughout this PR, please use CSS tokens with SCSS token fallback, e.g.: `border: var(--kui-border-width-10, $kui-border-width-10) solid var(--kui-color-border-transparent, $kui-color-border-transparent);`

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,281 @@
...
+$kui-shadow-focus-narrow: 0 0 0 2px rgba($kui-color-background-primary, 0.2);
+
+.interactive-pill {
+  align-items: stretch;
+  border: 1px solid transparent;
```
portikM: issue tokens for border width and color (we have a token for transparant)

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,281 @@
...
+  border-radius: $kui-border-radius-20;
+  box-shadow: 0 0 0 0 transparent;
+  cursor: pointer;
+  display: inline-flex;
+  font-size: $kui-font-size-20;
+  max-width: 240px;
+  overflow: hidden;
+  transition: box-shadow 0.1s ease-in;
```
portikM:
```suggestion
  transition: box-shadow $kongponentsTransitionDurTimingFunc;
```
we have a variable transition timing and function

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,281 @@
...
+  white-space: nowrap;
+
+  .label {
+    line-height: 16px;
```
portikM: issue: line height token

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,281 @@
...
+    overflow: hidden;
+    padding: $kui-space-20 $kui-space-20 $kui-space-20 $kui-space-40;
+    text-overflow: ellipsis;
+
+    .base-label {
+      font-weight: 600;
```
portikM: issue: font weight token

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  .label,
+  .pill-icon {
+    transition: all 0.2s ease-in;
+  }
+
+  &.no-content {
+    background-color: $kui-color-background-neutral-weakest;
+    border: 1px dashed $kui-color-border;
```
portikM: issue: border width token

src/components/KFilterGroup/KFilterGroup.vue
portikM: I'd say it's better to use token here

src/components/KFilterGroup/KFilterGroup.vue
portikM: Here's the PR: https://github.com/Kong/design-tokens/pull/613

Might be quicker to fix it using that rule:
1. Install the preview version: `@kong/design-tokens@pr-613`
2. Add this line to `.stylelintrc.mjs`: `'@kong/design-tokens/token-var-usage': true,`
3. Run `pnpm stylelint:fix`
4. Don't forget to undo the changes in `package.json`, lockfile and `.stylelintrc.mjs`

src/components/KFilterGroup/InteractivePill.vue
Chalks: > should it be a button?

I'm going to go with no. Technically, yes, maybe it should be a button... but then the obvious next question is should the X be a separate button? Also there are a lot of styles to override if I do make it a button and the popper doesn't play nicely with this on first attempt... I'd prefer to leave it as divs.

src/components/KFilterGroup/FilterSelector.vue
Chalks: hmmm, I'm not sure what you're looking for here exactly. None of the functions in these components are exposed, so an end user isn't going to see this.

It feels a little unnecessary to add "onToggle happens when the thing is toggled" comments, but I'm guessing that's not really what you're looking for?

adamdehaven: Sorry, just meant whatever for the next developer that comes in here to look at the code to understand what "toggling" does, etc.

Maybe for some it's not necessary

Chalks: I'll go through and add comments to anything more complicated.

Chalks: added a few more comments in `KFilterGroup.vue` as that's where most of the complicated logic happens.

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,290 @@
...
+
+  &.clear-focused .pill-icon .clear-focus-highlight {
+    // more specific styles for clear focus when clearFocus true
+    border-radius: var(--kui-border-radius-20, $kui-border-radius-20);
+    box-shadow: $shadowFocusNarrow;
+  }
+
+  &.unfocused.no-content:hover {
+    background-color: var(--kui-color-background-primary-weakest, $kui-color-background-primary-weakest);
+    border: 1px solid var(--kui-color-border-primary, $kui-color-border-primary);
```
portikM:
```suggestion
    border: var(--kui-border-width-10, $kui-border-width-10) solid var(--kui-color-border-primary, $kui-color-border-primary);
```
suggestion: border width token

src/components/KFilterGroup/InteractivePill.vue
portikM: Is the wrapping div is meant to be clickable? I Think the answer is yes and if so I think it should be a `button` element. You can reset the default button styles using `defaultButtonReset` mixin (just do `@include defaultButtonReset;` in the element style block). Same applies to the wrapper element of the clear icon - as far as I can tell it's a div but it should be a button (see comment below)

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,290 @@
...
+        <div
+          v-if="hasContent"
+          ref="clearRef"
+          class="pill-icon clear-icon"
+          data-testid="interactive-pill-clear-icon"
+          tabindex="0"
+          @blur="onClearBlur"
+          @click.prevent.stop="onClear"
+          @focus="onClearFocus"
+          @keydown.enter.prevent.stop="onClear"
```
portikM: I think you should make this element a button. You can reset the default button styles using `defaultButtonReset` mixin. This way you can get rid of all the keyboard event bindings - you're getting them out of the box with just `@click` event binding

src/components/KFilterGroup/InteractivePill.vue
```diff
@@ -0,0 +1,301 @@
...
+const emit = defineEmits<{
+  (e: 'trigger'): void
+  (e: 'clear'): void
+}>()
+
+const fullLabel = computed<string>(() => {
+  return hasContent.value ? `${label}${delimiter}` : label
+})
+
+const truncated = ref(false)
```
adamdehaven, line 123: (unsure where to place this comment)

Image

When a long filter has a value, should the truncation occur separately on the `key` and `value` so that we can always see the operator in the middle, as well as the start of both the key and the value?

src/types/filter-group.ts
```diff
@@ -0,0 +1,148 @@
...
+ * lt - less than
+ * lte - less than or equal to
+ * gt - greater than
+ * gte - greater than or equal to
+ */
+export type FilterOperator = 'eq' | 'neq' | 'contains' | 'exists' | 'lt' | 'lte' | 'gt' | 'gte'
+
+export interface Filter {
+  /**
+   * The filter's label, displays in the filter selection dropdown and in the pill
```
adamdehaven: Can we mention the default here?

src/types/filter-group.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * The list of FilterOperators supported by this filter
+   * @default ['eq']
+   */
+  operators?: FilterOperator[]
+
+  /**
+   * The list of options the user can choose from in the filter. If unset, the
+   * filter renders an input box instead
```
adamdehaven: nit
```suggestion
   * filter renders a text input instead
```

src/types/filter-group.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * If true the filter's pill is always rendered, regardless of whether it has
+   * an active selection.
+   * @default false
+   */
+  pinned?: boolean
+
+  /**
+   * The placement this filter's KPop should use
```
adamdehaven: nit: We don't need to expose the component implementation here, so say "popover" instead of the internal component name being used
```suggestion
   * The placement this filter's popover should use
```

src/types/filter-group.ts
```diff
@@ -0,0 +1,148 @@
...
+export interface FilterSelection {
+  /**
+   * The operator this selection should be evaluated with
+   */
+  operator: FilterOperator
+
+  /**
+   * The value input by the user
+   */
+  value: string | string[] | number | number[] | boolean | boolean[]
```
adamdehaven: how would `boolean[]` ever be valid here?

src/types/filter-group.ts
```diff
@@ -0,0 +1,148 @@
...
+ * @internal
+ */
+export interface FilterPillProps {
+  /**
+   * The filter this pill renders
+   */
+  filter: Filter
+
+  /**
+   * Whether to initialize the pill with its KPop open
```
adamdehaven: nit
```suggestion
   * Whether to initialize the pill with its popover open
```

adamdehaven: question: do we actually have a use-case for this? Why would we ever render the popover open on page/component load?

src/types/filter-group.ts
```diff
@@ -0,0 +1,148 @@
...
+   * @default false
+   */
+  initOpen?: boolean
+
+  /**
+   * This filter's current selection
+   * @default undefined
+   */
+  selection?: FilterSelection
```
adamdehaven, line 104: This is an internal prop, right?

src/types/filter-group.ts
```diff
@@ -0,0 +1,148 @@
...
+   * in the order they are defined. All other filters are displayed in the
+   * FilterSelector in the order in which they are defined.
+   */
+  filters: FilterGroupFilters
+
+  /**
+   * The label to display on the FilterSelector
+   * @default 'Add filter'
+   */
+  selectorLabel?: string
```
adamdehaven, line 119: question: why is this "selectorLabel" instead of just "label" ?

src/types/filter-group.ts
```diff
@@ -0,0 +1,148 @@
...
+}
+
+export interface FilterGroupEmits {
+  /**
+   * When a user triggers "apply" in a filter's KPop, this is fired. Provides
+   * the key of the filter that was triggered and the KFilterGroup's new
+   * selection object.
+   */
+  apply: [appliedFilterKey: string, selection: FilterGroupSelection]
```
adamdehaven, line 128: Would `update` be closer to reg. semantics since we're using `v-model`?

src/components/KFilterGroup/InteractivePill.vue
adamdehaven: I'm still a bit confused why this wrapper element isn't a button

src/components/KFilterGroup/KFilterGroup.vue
```diff
@@ -0,0 +1,185 @@
+<template>
+  <div class="k-filter-group">
+    <div
+      class="label"
+      data-testid="filter-group-label"
+    >
+      Filters
```
adamdehaven, line 7: issue: this needs to be mapped to the prop text I think?

src/components/KFilterGroup/KFilterGroup.vue
```diff
@@ -0,0 +1,185 @@
...
+} from '@/types'
+import { KUI_ANIMATION_DURATION_20 } from '@kong/design-tokens'
+
+const {
+  filters,
+  selectorLabel = undefined,
+} = defineProps<FilterGroupProps>()
+const selection = defineModel<FilterGroupSelection>({ required: true })
+
+const activeFilterKey = ref<string | undefined>()
```
adamdehaven, line 54: nit: add a comment on what this one is for

src/components/KFilterGroup/KFilterGroup.vue
```diff
@@ -0,0 +1,185 @@
...
+
+  emit('apply', filterKey, selection.value)
+}
+
+watch(selection, (newSelection) => {
+  // in order to preserve filter add order, we have to watch the selection. As
+  // it changes we add/remove keys from the `unpinnedSelectionKeys` list. This
+  // allows us to respond to both our mutation of the selection and any host app
+  // mutation of the selection.
+  Object.entries(newSelection)
```
adamdehaven, line 157: can `newSelection` here ever be undefined?

src/components/KFilterGroup/KFilterGroup.vue
```diff
@@ -0,0 +1,185 @@
...
+        // remove undefined selections if it's already in the list
+        unpinnedSelectionKeys.value = [
+          ...unpinnedSelectionKeys.value.slice(0, addedIndex),
+          ...unpinnedSelectionKeys.value.slice(addedIndex + 1),
+        ]
+      } else if (addedIndex === -1 && value !== undefined) {
+        // add defined selections if it's not already in the list
+        unpinnedSelectionKeys.value.push(key)
+      }
+    })
```
adamdehaven: nit: a few more descriptive comments would be helpful here

src/components/KFilterGroup/InteractivePill.vue
Chalks: addressed with design already. We tried it both ways and doing it separately looked... weird. We settled on this pattern intentionally.

src/types/filter-group.ts
Chalks: there is no default for `label`, it's required.

src/components/KFilterGroup/InteractivePill.vue
adamdehaven: This seems unusable to me. There's no way to see what the key/value is without opening the popover. What about `title` text on hover (or tooltip if they prefer) to expose the full string?

src/types/filter-group.ts
Chalks: it's not. this is what we had in the engineering brief and I hadn't looked at it since. will update.

src/components/KFilterGroup/InteractivePill.vue
adamdehaven: image
 
already done - resolving this 👍🏼

src/types/filter-group.ts
adamdehaven: Maybe a better comment here from me would be "what is this used for"?" It's not obvious 😅

adamdehaven: "The filter's label" -- does that mean the text before the operator? Maybe an example here in the comment

src/types/filter-group.ts
Chalks: this is an internal component to the filter group (this type is annotated with `@internal`). It's rendered open when you first select it from the list of available filters. This:

https://github.com/user-attachments/assets/c45b07b2-7f29-4dd5-9468-e6969baf1b5f

src/types/filter-group.ts
Chalks: yes. marked as such on line 86.

src/types/filter-group.ts
Chalks: you previously asked if the "Add filter" should be something we can override. this is how.

src/types/filter-group.ts
Chalks: maybe. Apply doesn't always mean the model has been mutated though because of the custom content possibilities. If you have custom content then in that case `@apply` just means "the user clicked the apply button, go update the model now"

adamdehaven: Ok that's fair; good with this

src/components/KFilterGroup/KFilterGroup.vue
Chalks: this isn't something you can change. but maybe it should be. The prop label is for the `Add filter` button.

adamdehaven: Since we don't want to deal with i18n if we can map this to a prop with `Filters` as the default, that would be ideal

src/components/KFilterGroup/KFilterGroup.vue
Chalks: no. selection is a required model.

adamdehaven: Right I just meant if it was clearable or something, this would error. Sounds like no-op

src/types/filter-group.ts
Chalks: "the filter's label" meaning "literally `label`". I think I can just remove that phrase and it makes sense.

```
  /**
   * displays in the filter selection dropdown and in the pill
   */
  label: string
```

adamdehaven: Is this like, "The label of the thing you're filtering" ?

PR #3093, opened 2026-02-17
fix(ktoaster): deprecate zindex prop, set z-index on toaster container

portikM:
# Summary

* Deprecate zIndex prop in KToaster component
* Instead of setting provided `zIndex` on nested `.k-toaster` element, set it on shared toaster container (top-level element)


sandbox/pages/SandboxToaster.vue
```diff
@@ -152,7 +154,7 @@ const openToaster = (argument: string) => {
         title: 'Timeout',
         message: 'This is toaster with timeout (10s)',
         appearance: 'info',
-        timeoutMilliseconds: 10000,
+        timeoutMilliseconds: 1000000,
```
adamdehaven: issue: why is this changing to 1,000 seconds? (I know this is sandbox only)

src/types/toaster.ts
```diff
@@ -33,8 +33,7 @@ export interface ToasterProps {
   toasterState: Array<Toast & { key: string }>
 
   /**
-   * The z-index of the toaster.
-   * @default 10000
+   * @deprecated zIndex provided through ToasterOptions is set on the shared container on initialization. This prop is no longer used.
```
adamdehaven, line 36: Leave the `@default 10000` comment in place along with this new `@deprecated`

src/components/KToaster/ToastManager.cy.ts
```diff
@@ -95,4 +95,23 @@ describe('ToastManager', () => {
...
     toastManager3.destroy()
   })
+
+  it('creates toasters container with correct default z-index', () => {
+    const toastManager = new ToastManager()
+    const container = document.getElementById(toastersContainerId)
+    expect(container).to.have.css('z-index', '10000')
+
+    // Cleanup
+    toastManager.destroy()
```
adamdehaven, line 105: nit: would it make sense to call `destroy` in an `afterEach` hook for the whole file?

src/components/KToaster/ToastManager.ts
```diff
@@ -20,13 +20,14 @@ const defaultZIndex = 10000
  * Get or create the shared toaster container element.
  * Increments the reference count for tracking active instances.
  */
-function getOrCreateSharedContainer(): HTMLElement {
+function getOrCreateSharedContainer(zIndex: number): HTMLElement {
```
adamdehaven: suggestion: make this an object interface in case we need to add more properties later

```suggestion
function getOrCreateSharedContainer({ zIndex }: { zIndex: number }): HTMLElement {
```

Also, should the default value be enforced here?

sandbox/pages/SandboxToaster.vue
portikM: Oh prob was testing something. Reverted

src/components/KToaster/ToastManager.cy.ts
portikM: Each test scenario initializes it's own instance (or a few) and then cleans them up.

src/components/KToaster/ToastManager.ts
portikM: Done

src/types/toaster.ts
portikM: Done

adamdehaven (approved):

PR #3095, opened 2026-02-18
feat: KFilterGroup content [KHCP-19483]

Chalks:
**Please note that this is merging into `feat/filter-group-khcp-19257`, NOT `main`**

# Summary

* Adds the content to the filter pills within KFilterGroup.
* Updates some types to more cleanly handle the different types of filters

### Demo

https://github.com/user-attachments/assets/f306388b-4f4b-4cd8-8a37-da9e536962af

### DIFF
--- a/src/components/KFilterGroup/FilterPill.vue
+++ b/src/components/KFilterGroup/FilterPill.vue
@@ -288,10 +288,10 @@ const operatorSelectItems = computed((): SelectItem[] => {
     // 'contains' means 'exists in this selection' for (multi)select
     contains: filterType.value === 'input' ? 'Contains' : 'In',
     exists: 'Exists',
-    lt: '<',
-    lte: '≤',
-    gt: '>',
-    gte: '≥',
+    lt: 'Less than',
+    lte: 'Less than or equal to',
+    gt: 'Greater than',
+    gte: 'Greater than or equal to',
   }
 
   return operators.value.map((operator) => ({


src/components/KMultiselect/KMultiselect.vue
```diff
@@ -1283,6 +1283,7 @@ $kMultiselectInputHelpTextHeight: var(--kui-line-height-20, $kui-line-height-20)
 
   .expanded-selection-empty {
     @include inputText;
+    @include truncate;
```
Chalks, line 1286: oh. this probably shouldn't be included in this PR.

I'm leaving it here for discussion though. Basically, multiselect placeholder breaks if your placeholder is too long _and_ the multiselect doesn't have anything selected.

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -1,8 +1,12 @@
 <template>
   <KPop
     ref="filterPopper"
+    :data-testid="`${attrs['data-testid'] ?? 'filter-pill'}-popover`"
```
adamdehaven: Since this component isn't directly exposing a data-testid to the consumer, I have an issue with making this interior attribute dynamic based on a data-testid applied to the parent component. Can this be static?

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -18,23 +22,90 @@
...
-          @apply with fake selection
-        </div>
-        <div @click="onCancel">
-          cancel
-        </div>
+      <div
+        class="filter-content"
+        data-testid="filter-pill-content"
+      >
+        <slot name="content">
```
adamdehaven: question: if there's only 1 slot in the component, why can't this be `default` instead of `content`?

issue: you're not utilizing `defineSlots` down in the `script setup`

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -18,23 +22,90 @@
...
+                v-model="userMultiselect"
+                data-testid="filter-pill-multiselect"
+                :items="filter.multiselectOptions"
+                label="Value"
+                placeholder="Select values"
+              />
+              <KInput
+                v-else
+                ref="inputField"
+                v-model="userInput"
```
adamdehaven: Can this input only accept a string? If yes, please use `v-model.trim`

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -18,23 +22,90 @@
...
+        >
+          Cancel
+        </KButton>
+        <KButton
+          appearance="primary"
+          data-testid="filter-pill-apply"
+          :disabled="applyDisabled"
+          @click="onApply"
+        >
+          Apply
```
adamdehaven, line 92: should these strings be applied via props somehow or are we so deep we just hard-code them until it's an issue?

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -45,11 +116,92 @@ const {
   filter,
   initOpen = false,
   selection = undefined,
+  isCustom = false,
 } = defineProps<FilterPillProps>()
 
 const isOpen = ref<boolean>(false)
 const filterPopperRef = useTemplateRef('filterPopper')
 
+const userOperator = ref<FilterOperator>()
```
adamdehaven, line 144: Don't hate me, but please add comments for all of these refs and methods in the script block so future-us doesn't have questions

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -45,11 +116,92 @@ const {
...
+    case 'custom':
+    default:
+      return false
+  }
+})
+
+const operators = computed<FilterOperator[]>(() => {
+  const defaultOperators: FilterOperator[] = filterType.value === 'input'
+    ? ['eq', 'contains']
+    : ['eq', 'contains', 'exists']
```
adamdehaven: Help me understand: why not just `eq`?

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -45,11 +116,92 @@ const {
...
+
+  return filter.operators && filter.operators.length
+    ? filter.operators
+    : defaultOperators
+})
+
+const userSelection = computed<FilterSelection | undefined>(() => {
+  let value
+  let text
```
adamdehaven: Not obvious what these are for, or their type, etc.

adamdehaven: Also, let's not use `value` since we have enough of that with Vue refs already 😅

src/components/KFilterGroup/KFilterGroup.vue
```diff
@@ -40,6 +45,8 @@ import type {
 } from '@/types'
 import { KUI_ANIMATION_DURATION_20 } from '@kong/design-tokens'
 
+const slots = useSlots()
```
adamdehaven, line 53: Are we using defineSlots anywhere?

src/types/filter-group.ts
```diff
@@ -22,23 +23,27 @@ export interface Filter {
...
   /**
-   * Whether a user can select more than one `option` or not.
-   * @default false
+   * The list of options the user can choose multiple item from in the filter.
+   * If both `selectOptions` and `multiselectOptions` are unset, the filter
+   * renders an input box instead
+   * @default undefined
    */
-  multiple?: boolean
+  multiselectOptions?: MultiselectEntry[]
```
adamdehaven: Why are these changing? 🤔 I'm not sure I like this

src/types/filter-group.ts
```diff
@@ -102,6 +107,12 @@ export interface FilterPillProps {
    * @default undefined
    */
   selection?: FilterSelection
+
+  /**
+   * Whether the filter's content slot was overridden by the host app
+   * @default false
+   */
+  isCustom?: boolean
```
adamdehaven: why can't this be keyed off of the existence of slot content?

src/components/KFilterGroup/FilterPill.vue
portikM: The later one. We hard-code them across all components

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -45,11 +116,92 @@ const {
...
 const isOpen = ref<boolean>(false)
 const filterPopperRef = useTemplateRef('filterPopper')
 
+const userOperator = ref<FilterOperator>()
+const userInput = ref<string>()
+const userSelect = ref<string | number>()
+const userMultiselect = ref<string[]>()
+const inputFieldRef = useTemplateRef('inputField')
+
+const filterType = computed<'custom' | 'input' | 'select' | 'multiselect'>(() => {
```
portikM:
```suggestion
const filterType = computed((): 'custom' | 'input' | 'select' | 'multiselect' => {
```
nit: syntax we use across the repo

src/components/KMultiselect/KMultiselect.vue
portikM: I think this change is ok

src/types/filter-group.ts
portikM: I agree. I think it should just be:

```
options?: SelectEntry[] | MultiselectEntry[]
multiple?: boolean
```

SelectEntry and MultiselectEntry types are relatively compatible

src/types/filter-group.ts
portikM: Agree but even if for some reason it really can't be, can we remove the prefix in the name? So `isCustom -> custom`

src/components/KFilterGroup/FilterPill.vue
Chalks: yeah, this was a convenience I put in just to make testing easier. Because the filterpill is wrapped in a KPop, the actual content testid isn't the _whole_ pill, so it was awkward to target the correct element in the test.

I can just bite the bullet and make the test a little more complicated instead.

src/components/KFilterGroup/FilterPill.vue
Chalks: > issue: you're not utilizing `defineSlots`

ah, wasn't aware that was a thing. Will add.

Chalks: > question: if there's only 1 slot in the component, why can't this be default instead of content?

as I was building it, it felt reaaaaally weird that that was the only slot. It really feels like the cancel/apply should be slots too. But I didn't want to deviate too far from the plan. So yeah, I guess this is fair. I can change to default.

src/components/KFilterGroup/FilterPill.vue
Chalks: do we generally prefer this stuff to be props or do we generally prefer to use slots? Related to [this comment](https://github.com/Kong/kongponents/pull/3095#discussion_r2833524381).

I can easily do either. I think props is a reasonable way to do it but I'm tempted to slot it.

src/components/KFilterGroup/FilterPill.vue
Chalks: no hate. I could stand to comment more frequently.

src/components/KFilterGroup/FilterPill.vue
portikM: With static strings like Apply button text - neither. We Won't really need to change the text in this button ever. With things that we do want to be configurable we prefer a prop unless it's an element that often will have an icon or a tooltip or anything else other than text - then slot is better approach.

src/components/KFilterGroup/FilterPill.vue
Chalks: I actually am not sure what to do here.

text input - `['eq', 'contains']` seems like a reasonable default to me. I could see an argument for including `'neq'` but I'm not convinced.

select input - `['eq']` makes sense to me, but that would make it so the operator selector doesn't appear. I think that's okay? It seems a little different from what design had though.

multiselect input - `['eq']` seems reasonable too, same reasons as select.

---

I'm going to ask in #ask-kong-design-system. I'll update with a link to the thread when I do.

src/components/KFilterGroup/FilterPill.vue
Chalks: I can update them with a comment, but I was using `value` and `text` because those are the property names for `FilterSelection` and I return them on lines 198-202.

src/components/KFilterGroup/KFilterGroup.vue
Chalks: nope. I'll update.

src/components/KFilterGroup/FilterPill.vue
adamdehaven: What is the array here defining? Defaults?

src/types/filter-group.ts
Chalks: Because `KFilterGroup` is the component that _actually_ exposes the slot to the end user with:

```
<FilterPill
  v-for="key in visibleFilterKeys"
  …
  :is-custom="!!slots[`${key}-content`]"
  …
>
  <template #content>
    <slot :name="`${key}-content`" />
  </template>
</FilterPill>
```

When you expose the slot this way, the child component `FilterPill` thinks that `useSlots().content` is always truthy, regardless of whether the user actually put content in `<template #keyname-content>` or not.

So unfortunately it's not possible for the child component to know whether the slot content is provided by the end user or not without inspecting the actual slot content, which felt significantly messier/fragile than just using a prop.

adamdehaven: Couldn't the parent component detect if it's custom or not and then use `provide/inject` to send that info down the chain?

src/types/filter-group.ts
Chalks: Mostly because of the difficulty of communicating the right types to the end user.

With the original setup, options would have been typed like `options?: SelectEntry[] | MultiselectEntry[]`. Then, based on whether `multiple` is true or not, we have to make sure the correct type got used in `options`. This was _really_ painful to get typed correctly and required a lot of type narrowing and got VERY confusing very fast.

The real issue is that `SelectEntry` and `MultiselectEntry` aren't the same. If they were (they should be imo) then none of this would have mattered.

Choices I considered:

1. change `SelectEntry` and `MultiselectEntry` to be the same - not possible in this PR, outside the scope of this work.
2. create an `OptionsEntry` for this component that is a subset of both `SelectEntry` and `MultiselectEntry` then use the data to build either of those types based on `multiple`'s truthiness. - not unreasonable, but I didn't want to create yet-another-select-options-abstraction. I could be argued into this though.
3. switch to just `multiselectOptions` and `selectOptions` and drop `multiple` - this felt like the cleanest approach to me, and keeps strict typing while still being clear.

I'm not sure I like it either. I'm curious what you think is a better solution.

src/components/KFilterGroup/FilterPill.vue
Chalks: thread in chat about this https://kongstrong.slack.com/archives/C053JL4CE9K/p1771600543198849

src/types/filter-group.ts
portikM: > change SelectEntry and MultiselectEntry to be the same

This is not feasible

portikM: I see, thanks for explaining. I'm ok with `selectOptions` and `multiselectOptions` approach

src/components/KFilterGroup/FilterPill.vue
Chalks: okay, defaulting to `['eq']` in all cases. This is going to look more clean anyways. Thanks for poking on this.

src/types/filter-group.ts
adamdehaven: No, I don't think this approach makes sense. There should be a single "options" interface

adamdehaven: We shouldn't be exposing implementation details of the underlying "single select" or "multi select" component interfaces to users of this component. Otherwise, they should just be using the slot.

src/types/filter-group.ts
Chalks: yes, but I'm not sure I see how that's better than a prop.

adamdehaven: As long as this prop is internal-only (meaning not exposed to the user) I don't care -- but we don't typically prefix with `is-`

src/types/filter-group.ts
Chalks: what I was hoping to accomplish by exposing the underlying component interfaces was to get the additional functionality of the select field (disabled options, grouping, etc) for free. Otherwise, you _have_ to use the slot. If you think that's not worth it, and the slot is sufficient for all that, then using just `options` and `multiple` is fine, and I can type it as `options?: Array<{ value: string, label: string }>`

src/components/KFilterGroup/FilterPill.vue
Chalks: I'm fine with just hard code.

src/types/filter-group.ts
adamdehaven: IMO we don't care about grouping, and all of the complexities that come from exposing all of those actual component props. The component should have a single interface -- if users need grouping and more complexity, they are free to slot it

src/components/KFilterGroup/FilterPill.vue
Chalks: oh, hmm. should I change all the computed things to that syntax? I think I did `computed<T>` everywhere.

edit: eh, yeah I'll just change 'em all.

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -1,8 +1,11 @@
 <template>
   <KPop
     ref="filterPopper"
+    :max-width="filterType === 'custom' ? 'none' : '400px'"
```
adamdehaven: question: why wouldn't we still enforce a max width here on slotted content?

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -18,25 +21,94 @@
...
-        </div>
-        <div @click="onCancel">
-          cancel
-        </div>
+      <div
+        class="filter-content"
+        data-testid="filter-pill-content"
+      >
+        <slot name="default">
+          <div class="internal-layout">
```
adamdehaven: nit: `internal-layout` is vague -- could we name this class better?

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -45,12 +117,144 @@ const {
   filter,
   initOpen = false,
   selection = undefined,
+  custom = false,
```
adamdehaven, line 121: it seems like we're only using `custom` for the maxWidth, emit, filterType, etc.

I'm still questioning why this prop is needed and can't be inferred from the other props/slots

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -72,14 +276,74 @@ const delimiter = computed<string | undefined>(() => {
...
+const operatorSelectItems = computed((): SelectItem[] => {
+  const filterOperatorLabels = {
+    eq: 'Equals',
+    neq: 'Not equal to',
+    contains: 'Contains',
+    exists: 'Exists',
+    lt: 'Less than',
+    lte: 'Less than or equal to',
+    gt: 'Greater than',
+    gte: 'Greater than or equal to',
```
adamdehaven: Image

nit: most of these are so long that they will never be legible in the dropdown. Anything we can do here?

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -72,14 +276,74 @@ const delimiter = computed<string | undefined>(() => {
...
+ */
+const resetUserSelection = () => {
+  userOperator.value = selection?.operator ?? operators.value?.[0]
+  userInput.value = typeof selection?.value === 'string' ? selection.value : undefined
+  userSelect.value = filter.options?.find(({ value }) => value === selection?.value)?.value
+  userMultiselect.value = filter.options?.filter(({ value }) => {
+    if (Array.isArray(selection?.value)) {
+      return (selection.value).includes(`${value}`)
+    }
+  }).map(({ value }) => value)
```
adamdehaven, line 332: Add comments to help us understand the logic here

src/components/KFilterGroup/FilterPill.vue
Chalks: Some of our earlier designs included content that was significantly longer than 400px. We didn't include those things in this component, but I expect they'll be used for a custom filter. I didn't want to hamstring custom filters to be forced to be 400 pixels wide.

adamdehaven: should it be exposed as a prop then, with a default 400px?

src/components/KFilterGroup/FilterPill.vue
Chalks: I don't know of any way around this. Here's the issue that's being resolved by it:

First case

```
<KfilterGroup
  v-model="selection"
  :filters="{ myFilter: { label: 'My Filter' } }"
/>
```

Second case

```
<KfilterGroup
  v-model="selection"
  :filters="{ myFilter: { label: 'My Filter' } }"
>
  <template #filter-myFilter>hi</template>
</KFilterGroup>
```

in both cases, within `FilterPill.vue`
```
const slots = useSlots();
console.log(!!slots.default) // prints true
```

adamdehaven: Maybe I need to see a better example somewhere of what using `custom` looks like

src/components/KFilterGroup/FilterPill.vue
Chalks: yeah, I need to get with design and figure out better language and/or use symbols instead. Helen and I will chat.

src/components/KFilterGroup/FilterPill.vue
Chalks: yeah, that would work. Will change.

src/components/KFilterGroup/FilterPill.vue
Chalks: `default-layout`? It's just putting a flex box around the default slot content.

adamdehaven: I mean in the scope of the component, like what's this the layout of?

portikM: Something like `.filter-popover-content-wrapper` perhaps?

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -1,8 +1,11 @@
 <template>
   <KPop
     ref="filterPopper"
+    :max-width="filterType === 'custom' ? 'none' : '400px'"
+    :offset="KUI_SPACE_30"
```
portikM, line 5: Is there a reason you needed to reduce the default offset? IMO the default (`16px`, `KUI_SPACE_60` token) looks better. When I look at the Figma I think the offset we have there is `12px` (`KUI_SPACE_50` token). I don't think we should go lower than either of those tokens because it looks like it's overlaying the pill too much (screenshot).

Image

src/components/KFilterGroup/FilterPill.vue
portikM: Sorry missed this replies. Thanks for changing them!

src/types/filter-group.ts
portikM: > then using just options and multiple is fine, and I can type it as options?: Array<{ value: string, label: string }>

This is great, thank you!

src/components/KFilterGroup/FilterPill.vue
Chalks: yes. This was explicitly called out by @two24studios in a design overview meeting with Helen and I. Jason said that overlap was intentional. [Design in figma](https://www.figma.com/design/Yze0SWXl5nKjR0rFdilljK/Components?node-id=10964-5036&t=zE6vuWZR1zfdDfm0-4)

Screenshot from figma

Screenshot 2026-02-23 at 12 54 59 PM

---

Screenshot with `KUI_SPACE_30` (current implementation)

Screenshot 2026-02-23 at 12 56 54 PM

---

Screenshot with `KUI_SPACE_40`

Screenshot 2026-02-23 at 12 57 37 PM

---

Screenshot with `KUI_SPACE_50`

Screenshot 2026-02-23 at 12 58 21 PM

---

I went with `KUI_SPACE_30` because to my eye it most closely matches the design. Probably worth noting that the popover looks a little different from the design anyways, as the arrow indicator is fatter and more to the left than what's in figma.

portikM: Ah I see, thanks for explaining.

src/components/KFilterGroup/FilterPill.vue
Chalks: it's the layout of the operator and the value. It forces them to render side by side and take up the entire width of the popover.

src/components/KFilterGroup/FilterPill.vue
Chalks: To create a filter that looks like this:

edit: my screenshot was from an old version by accident. This is the correct screenshot:

Screenshot 2026-02-24 at 9 27 06 AM

You have this:

```
<KFilterGroup
  v-model="selection"
  :filters="{
    myCustom: {
      label: 'Custom slotted filter',
    },
  }"
>
  <template #filter-myCustom>
    <KInputSwitch
      v-model="foo"
      label="This is in the #filter-custom slot"
    />
  </template>
</KFilterGroup>
```

src/components/KFilterGroup/FilterPill.vue
Chalks: Talked with Helen on this:

Screenshot 2026-02-23 at 1 16 02 PM

&nbsp;

Screenshot 2026-02-23 at 1 16 08 PM

src/components/KFilterGroup/FilterPill.vue
adamdehaven: Why wouldn't this use the label from the prop instead via slotProps? Why would you redefine the label?

src/components/KFilterGroup/FilterPill.vue
adamdehaven: I don't think this is acceptable. We either use text, or we use symbols, but both here isn't the way

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -72,14 +278,83 @@ const delimiter = computed<string | undefined>(() => {
...
+  const filterOperatorLabels = {
+    eq: 'Equals',
+    neq: 'Not equals',
+    // 'contains' means 'exists in this selection' for (multi)select
+    contains: filterType.value === 'input' ? 'Contains' : 'In',
+    exists: 'Exists',
+    lt: '<',
+    lte: '≤',
+    gt: '>',
+    gte: '≥',
```
adamdehaven: this isn't viable. We need to standardize on symbols or text, not both. I think text is better for readability

src/components/KFilterGroup/FilterPill.vue
Chalks: sorry, my screenshot was bad and I didn't realize. I'm not redefining anything, the actual output of that code is:

Screenshot 2026-02-24 at 9 27 06 AM

src/components/KFilterGroup/FilterPill.vue
Chalks: I agree with your point, but I'm not sure how to accomplish it in a way that doesn't completely break the design. Can we make a ticket to resolve this (before merging into main), and bring it up with design instead of trying to handle it in this PR (which is going into the staging branch)?

src/components/KFilterGroup/FilterPill.vue
Chalks: We definitely do both in other places (see screenshot below), but I don't disagree with your point. [I suggested having a separate convo/ticket about this here](https://github.com/Kong/kongponents/pull/3095#discussion_r2847555117)

Screenshot 2026-02-24 at 9 32 19 AM

src/components/KFilterGroup/FilterPill.vue
adamdehaven: Sure, but for now, switch them all back to text

Chalks: done. [Ticket is here](https://konghq.atlassian.net/browse/KHCP-19681)

PR #3098, opened 2026-02-19
docs: KFilterGroup documentation [KHCP-19484]

Chalks:
**Please note that this is merging into `feat/filter-group-khcp-19257`, NOT `main`**

# Description

Documentation for the KFilterGroup.

### how to test

Strongly recommend pulling this branch down and `pnpm docs:dev` then looking at the page that way.

### demo of documentation

https://github.com/user-attachments/assets/501bd7c3-1d3d-4b31-ba6b-4f63c78fe756

### DIFF
--- a/docs/components/filter-group.md
+++ b/docs/components/filter-group.md
@@ -1,6 +1,6 @@
 # Filter Group
 
-KFilterGroup is a component that displays provided filters and allows a user to set the values of those filters.
+KFilterGroup is a component that provides an interface for displaying and applying filter values.
 
 <ClientOnly>
   <KFilterGroup
@@ -10,10 +10,10 @@ KFilterGroup is a component that displays provided filters and allows a user to 
 </ClientOnly>
 
 ```html
-  <KFilterGroup
-    v-model="selection"
-    :filters="filters"
-  />
+<KFilterGroup
+  v-model="selection"
+  :filters="filters"
+/>
 ```
 
 ## Props
@@ -60,7 +60,6 @@ type FilterGroupSelection = Record<string, FilterSelection | undefined>
 </template>
 
 <script setup lang="ts">
-import { ref } from 'vue'
 import type {
   FilterGroupFilters,
   FilterGroupSelection,
@@ -82,11 +81,12 @@ const selection = ref<FilterGroupSelection>({
 const resetModelSelection = () => {
   selection.value.name = defaultNameSelection,
 }
+</script>
 ```
 
 ### filters
 
-The `filters` prop is of type `FilterGroupFilters`.
+Prop for passing list of available filters. Accepts object of type `FilterGroupFilters`.
 
 ```ts
 /*** All filters the KFilterGroup can render */
@@ -126,11 +126,11 @@ export interface FilterOption {
 }
 ```
 
-There are four kinds of filters that can be rendered
+KFilterGroup supports four filter types, determined by the filter configuration:
 
 #### Select filter
 
-A select filter renders if `options` is set and `multiple` is falsey
+A select filter is rendered when `options` is provided and `multiple` is `false` or `undefined`.
 
 <ClientOnly>
   <KFilterGroup
@@ -140,25 +140,25 @@ A select filter renders if `options` is set and `multiple` is falsey
 </ClientOnly>
 
 ```html
-  <KFilterGroup
-    v-model="selection"
-    :filters="{
-      status: {
-        label: 'Status',
-        options: [
-          { value: 'todo', label: 'Todo' },
-          { value: 'inprogress', label: 'In Progress' },
-          { value: 'done', label: 'Done' },
-        ],
-        pinned: true,
-      },
-    }"
-  />
+<KFilterGroup
+  v-model="selection"
+  :filters="{
+    status: {
+      label: 'Status',
+      options: [
+        { value: 'todo', label: 'Todo' },
+        { value: 'inprogress', label: 'In Progress' },
+        { value: 'done', label: 'Done' },
+      ],
+      pinned: true,
+    },
+  }"
+/>
 ```
 
 #### Multiselect filter
 
-A multiselect filter renders if `multiple` is true and `options` is set.
+A multiselect filter is rendered when `multiple` is true and `options` is provided.
 
 <ClientOnly>
   <KFilterGroup
@@ -168,27 +168,27 @@ A multiselect filter renders if `multiple` is true and `options` is set.
 </ClientOnly>
 
 ```html
-  <KFilterGroup
-    v-model="selection"
-    :filters="{
-      tag: {
-        label: 'Tag',
-        multiple: true,
-        options: [
-          { value: 'foo', label: 'Foo' },
-          { value: 'bar', label: 'Bar' },
-          { value: 'baz', label: 'Baz' },
-          { value: 'bat', label: 'Bat' },
-        ],
-        pinned: true,
-      },
-    }"
-  />
+<KFilterGroup
+  v-model="selection"
+  :filters="{
+    tag: {
+      label: 'Tag',
+      multiple: true,
+      options: [
+        { value: 'foo', label: 'Foo' },
+        { value: 'bar', label: 'Bar' },
+        { value: 'baz', label: 'Baz' },
+        { value: 'bat', label: 'Bat' },
+      ],
+      pinned: true,
+    },
+  }"
+/>
 ```
 
 #### Text filter
 
-A text filter renders if `options` is unset:
+A text filter is rendered when `options` is not provided.
 
 <ClientOnly>
   <KFilterGroup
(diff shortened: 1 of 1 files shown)


docs/components/filter-group.md
```diff
@@ -1,69 +1,538 @@
...
+    @apply="onCustomApply"
+    @clear="onCustomClear"
+    @open="onCustomOpen"
+  >
+    <template #customName-content>
+      <div style="width: 300px; transform: rotate(5deg); padding: 10px 5px;">
+        <KInput
+          v-model="customName"
+          label="Value"
+          placeholder="Your name, but rotated 5 degrees"
```
Chalks: don't think this is actually what should be used as an example of custom content (though I think it's funny), but was struggling to come up with something reasonable but also easy to implement. help!

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
 # Filter Group
 
-**KFilterGroup** should be described here in the first paragraph.
+KFilterGroup is a component that displays provided filters and allows a user to set the values of those filters.
```
portikM:
```suggestion
KFilterGroup is a component that provides an interface for displaying and applying filter values.
```

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
-.KFilterGroup-wrapper {
-  --KFilterGroup-wrapperBorderColor: lime;
+const defaultNameSelection: FilterSelection = {
+  operator: 'eq',
+  text: 'My name',
+  value: 'My name',
 }
-</style>
+
+export default defineComponent({
```
portikM: Please use `<script setup lang="ts">` and Vue3 syntax

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
+    v-model="basicSelection"
+    :filters="basicFilters"
+  />
+</ClientOnly>
+
+```html
+  <KFilterGroup
+    v-model="selection"
+    :filters="filters"
+  />
```
portikM:
```suggestion
<KFilterGroup
  v-model="selection"
  :filters="filters"
/>
```
issue: indentation (all code blocks in this PR)

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
+  <KButton @click="resetModelSelection">Reset</KButton>
 </template>
 
 <script setup lang="ts">
-  // Example code goes here
-</script>
+import { ref } from 'vue'
```
portikM:
```suggestion
```
nit: you can omit this in the example to keep it shorter

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
+
+const filters: FilterGroupFilters = { … } // see `filters` below
+
+const selection = ref<FilterGroupSelection>({
+  name: defaultNameSelection,
+})
+
+const resetModelSelection = () => {
+  selection.value.name = defaultNameSelection,
+}
```
portikM, line 83:
```suggestion
}
</script>
```
nit: closing tag

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
 ```
 
-## Props
+### filters
+
+The `filters` prop is of type `FilterGroupFilters`.
```
portikM:
```suggestion
Prop for passing list of available filters. Accepts object of type `FilterGroupFilters`.
```

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
+
+export interface FilterOption {
+  /** Label for the item to be displayed in the (multi)select dropdown. */
+  label: string
+  /** value for the item to be displayed in the (multi)select dropdown. */
+  value: string
+}
+```
+
+There are four kinds of filters that can be rendered
```
portikM:
```suggestion
KFilterGroup supports four filter types, determined by the filter configuration:
```

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
+
+#### Select filter
+
+A select filter renders if `options` is set and `multiple` is falsey
```
portikM:
```suggestion
A select filter is rendered when `options` is provided and `multiple` is `false` or `undefined`.
```

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
+        ],
+        pinned: true,
+      },
+    }"
+  />
+```
+
+#### Multiselect filter
+
+A multiselect filter renders if `multiple` is true and `options` is set.
```
portikM:
```suggestion
A multiselect filter is rendered when `multiple` is `true` and `options` is provided.
```

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
+        ],
+        pinned: true,
+      },
+    }"
+  />
+```
+
+#### Text filter
+
+A text filter renders if `options` is unset:
```
portikM:
```suggestion
A text filter is rendered when `options` is not provided:
```

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
+        label: 'Name',
+        pinned: true,
+      },
+    }"
+  />
+```
+
+#### Custom filter
+
+Custom filters render if there is content in the [filter-\*](#filter-content) slot.
```
portikM:
```suggestion
A custom filter is rendered when content is provided in the [filter-\*](#filter-content) slot.
```

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
 ```
 
+
 ## Slots
 
-- `default` - default slot description
-- `slotName` - slot description
+### Filter Content
+
+You can slot your custom content into each filter. Use the corresponding filter key from the [`filters` prop](#filters) to target the appropriate `filter-*` slot. When you do this, the KFilterGroup is no longer able to determine what the user's selection is for that filter therefore you must manage the `selection.*` for that filter.
```
portikM:
```suggestion
Custom content can be slotted into each filter. The corresponding filter key from the [`filters` prop](#filters) should be used to target the appropriate `filter-*` slot. When custom content is provided, KFilterGroup is no longer able to determine the user's selection for that filter, therefore the `selection.*` for that filter must be managed by the host app.
```

docs/components/filter-group.md
portikM: It is funny, first time I opened this popover it threw me off completely lol. I think we should have a useful example here, something resembling our own use case. I think repurposing the KDateTimePicker example you have above is a good option for one filter, for another one you could use a [KSlider](https://kongponents.konghq.com/components/slider.html) (something like number of nodes in the control plane filter? - hypothetical use case for control planes table filter).

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
 ## Events
 
-- `@change` - Emitted when...
+### apply
+
+Emitted when the apply button is clicked. `@apply` returns the applied filter's key and the value of the entire KFilterGroup's selection ([described in v-model](#v-model)).
+
+### clear
+
+Emitted when the `X` icon is clicked. `@clear` returns the cleared filter's key and the value of the entire KFilterGroup's selection ([described in v-model](#v-model)).
```
portikM:
```suggestion
Emitted when the `X` icon is clicked in filter pill. Event payload is the cleared filter's key and the value of the entire KFilterGroup's selection ([described in v-model](#v-model)).
```

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
+}
+</script>
 ```
 
 ## Events
 
-- `@change` - Emitted when...
+### apply
+
+Emitted when the apply button is clicked. `@apply` returns the applied filter's key and the value of the entire KFilterGroup's selection ([described in v-model](#v-model)).
```
portikM:
```suggestion
Emitted when the apply button is clicked. Event payload is the applied filter's key and the value of the entire KFilterGroup's selection ([described in v-model](#v-model)).
```

docs/components/filter-group.md
```diff
@@ -1,65 +1,548 @@
...
+
+### clear
+
+Emitted when the `X` icon is clicked. `@clear` returns the cleared filter's key and the value of the entire KFilterGroup's selection ([described in v-model](#v-model)).
+
+### close
+
+Emitted when the filter's popover content is closed for any reason (including `@apply`). Returns the closed filter's key.
```
portikM:
```suggestion
Emitted when the filter's popover is closed (including when selection is applied). Event payload is the closed filter's key.
```

portikM (dismissed):

docs/components/filter-group.md
Chalks: I wanted to do the KDateTimePicker, but the code for it is shockingly verbose. KSlider might be better, I'll take a look.

docs/components/filter-group.md
Chalks: oh right, I did this because I was cribbing from `multiselect.md` and never thought to check if `setup` worked. 🤦

docs/components/filter-group.md
Chalks: okay. All of your comments addressed, and updated custom demo to be like so:


https://github.com/user-attachments/assets/d968c955-a149-4b57-8b8f-93ae5343c493

PR #3162, opened 2026-04-13
chore: package bumps for renovate

Chalks:
# Summary

Does #3052 but with more vim, pun intended.


sandbox/netlify.toml
```diff
@@ -2,4 +2,5 @@
 
 # Ignore so that the build is always processed for the sandbox
 [build]
+  environment = { NODE_VERSION = "24.14.0" }
```
Chalks, line 5: Netlify uses this toml file when building the sandbox, and needs a more recent node version, otherwise it fails on `crypto.hash is not a function` [as these logs show](https://app.netlify.com/projects/kongponents-sandbox/deploys/69dd6073302ffd0008cf5cae)

adamdehaven (approved):

PR #2707, opened 2025-04-18
fix(ktabs): use kbutton for ktabs

Justineo:
# Summary

[KM-1192](https://konghq.atlassian.net/browse/KM-1192)

Reuse the logic in `<KButton>` that handles `to`, `disabled`, etc.

- Preview: https://github.com/Kong/konnect-ui-apps/pull/6533

[KM-1192]: https://konghq.atlassian.net/browse/KM-1192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3171, opened 2026-04-24
Revert "fix(ktabs): use kbutton for ktabs (#2707)"

johncowen:
This reverts commit fdd17e25ff11678fff15ee7552c5f15556afdc8a.

The current behaviour results in an anchor element inside of a button element. I'm not sure why a KButton was added for something that looks like a Tab as there isn't much info in the PR (nor the JIRA) that caused this. cc @Justineo 

https://github.com/Kong/kongponents/pull/2707

Screenshot 2026-04-24 at 10 22 59

FYI: I ws hoping to do a fully clean revert, but there were some tiny conflicts in the revert due to https://github.com/Kong/kongponents/pull/2781 coming in just after the PR that caused the issue. I thin kI've resolved those correctly and I checked the sandbox. But please let me know if you spot anything amiss.

# Summary

### DIFF
--- a/src/components/KTabs/KTabs.vue
+++ b/src/components/KTabs/KTabs.vue
@@ -15,16 +15,15 @@
         :class="{ active: activeTab === tab.hash }"
         :data-testid="`${tab.hash.replace('#', '')}-tab`"
       >
-        <KButton
-          appearance="none"
+        <component
+          :is="tabComponent(tab).tag"
           :aria-controls="hidePanels ? undefined : `panel-${tab.hash}`"
           :aria-selected="hidePanels ? undefined : (activeTab === tab.hash ? 'true' : 'false')"
           class="tab-link"
           :class="{ disabled: tab.disabled }"
-          :disabled="tab.disabled"
           role="tab"
           :tabindex="getAnchorTabindex(tab)"
-          :to="tab.to"
+          v-bind="tabComponent(tab).attributes"
           @click="!tab.disabled ? handleTabChange(tab.hash) : undefined"
           @click.prevent="!tab.disabled ? handleTabChange(tab.hash) : undefined"
           @keydown.enter.prevent="!tab.disabled ? handleTabChange(tab.hash) : undefined"
@@ -33,7 +32,7 @@
           <slot :name="`${getTabSlotName(tab.hash)}-anchor`">
             <span>{{ tab.title }}</span>
           </slot>
-        </KButton>
+        </component>
       </li>
     </ul>
 
@@ -57,7 +56,6 @@
 
 <script lang="ts" setup generic="const Hash extends string = string">
 import { ref, watch } from 'vue'
-import KButton from '@/components/KButton/KButton.vue'
 import type { StripHash, Tab, TabsEmits, TabsProps, TabsSlots } from '@/types'
 
 const {
@@ -92,6 +90,18 @@ const getAnchorTabindex = (tab: Tab): string => {
   }
 
   return typeof anchorTabindex === 'number' && anchorTabindex >= -1 && anchorTabindex <= 32767 ? String(anchorTabindex) : '0'
+}
+
+const tabComponent = (tab: Tab) => {
+  if (tab.to) {
+    if (typeof tab.to === 'string') {
+      return { tag: 'a', attributes: { href: tab.disabled ? undefined : tab.to } }
+    } else if (typeof tab.to === 'object') {
+      return { tag: 'router-link', attributes: { to: tab.disabled ? undefined : tab.to } }
+    }
+  }
+
+  return { tag: 'div', attributes: {} }
 }
 
 watch(() => modelValue, (newTabHash) => {
--- a/src/components/KTabs/KTabs.vue
+++ b/src/components/KTabs/KTabs.vue
@@ -93,11 +93,11 @@ const getAnchorTabindex = (tab: Tab): string => {
 }
 
 const tabComponent = (tab: Tab) => {
-  if (tab.to) {
+  if (tab.to && !tab.disabled) {
     if (typeof tab.to === 'string') {
-      return { tag: 'a', attributes: { href: tab.disabled ? undefined : tab.to } }
+      return { tag: 'a', attributes: { href: tab.to } }
     } else if (typeof tab.to === 'object') {
-      return { tag: 'router-link', attributes: { to: tab.disabled ? undefined : tab.to } }
+      return { tag: 'router-link', attributes: { to: tab.to } }
     }
   }
 
--- a/src/components/KTabs/KTabs.cy.ts
+++ b/src/components/KTabs/KTabs.cy.ts
@@ -1,6 +1,16 @@
 import { h } from 'vue'
 import type { TabsAppearance } from '@/types'
 import KTabs from '@/components/KTabs/KTabs.vue'
+import { createRouter, createWebHistory } from 'vue-router'
+
+const router = createRouter({
+  history: createWebHistory(),
+  routes: [{
+    name: 'home',
+    path: '/',
+    component: { template: '<div />' },
+  }],
+})
 
 const TABS = [
   { hash: '#pictures', title: 'Pictures' },
@@ -19,6 +29,9 @@ describe('KTabs', () => {
             tabs: TABS,
             appearance,
           },
+          global: {
+            plugins: [router],
+          },
         })
         cy.get('.k-tabs').should('have.class', appearance)
       })
@@ -28,6 +41,9 @@ describe('KTabs', () => {
           props: {
             tabs: TABS,
             appearance,
+          },
+          global: {
+            plugins: [router],
           },
         })
 
@@ -41,6 +57,9 @@ describe('KTabs', () => {
             modelValue: '#books',
             appearance,
           },
+          global: {
+            plugins: [router],
+          },
         })
 
         cy.get('.tab-item').eq(2).should('have.class', 'active')
@@ -51,6 +70,9 @@ describe('KTabs', () => {
           props: {
             tabs: TABS,
             appearance,
+          },
+          global: {
+            plugins: [router],
           },
         })
 
@@ -70,6 +92,9 @@ describe('KTabs', () => {
             tabs: TABS,
             hidePanels: true,
             appearance,
+          },
+          global: {
+            plugins: [router],
           },
           slots: {
             pictures: h('div', {}, picturesSlot),
@@ -104,6 +129,9 @@ describe('KTabs', () => {
             tabs,
             appearance,
           },
+          global: {
+            plugins: [router],
+          },
         })
 
         cy.get('.tab-item .tab-link').eq(1).should('have.class', 'disabled')
@@ -124,6 +152,9 @@ describe('KTabs', () => {
             tabs,
             appearance,
           },
+          global: {
+            plugins: [router],
+          },
         })
 
         cy.get('.tab-item .tab-link').eq(1).should('have.attr', 'href', '/movies')
@@ -131,15 +162,18 @@ describe('KTabs', () => {
 
       it('renders the tab as a link with no href attribute if tab.to is present and tab.disabled is true', () => {
         const tabs = [
-          { hash: '#pictures', title: 'Pictures' },
-          { hash: '#movies', title: 'Movies', to: '/movies', disabled: true },
-          { hash: '#books', title: 'Books' },
+          { hash: '#pictures', title: 'Enabled' },
+          { hash: '#movies', title: 'String to, disabled', to: '/movies', disabled: true },
+          { hash: '#books', title: 'Object to, disabled', to: { name: 'home' }, disabled: true },
         ]
 
         cy.mount(KTabs, {
           props: {
             tabs,
             appearance,
+          },
+          global: {
+            plugins: [router],
           },
         })
 
@@ -152,6 +186,9 @@ describe('KTabs', () => {
             tabs: TABS,
             beforeChange: () => false,
             appearance,
+          },
+          global: {
+            plugins: [router],
           },
         })
 
@@ -170,6 +207,9 @@ describe('KTabs', () => {
             props: {
               tabs: TABS,
               appearance,
+            },
+            global: {
+              plugins: [router],
             },
             slots: {
               pictures: h('div', {}, picturesSlot),
@@ -198,6 +238,9 @@ describe('KTabs', () => {
             props: {
               tabs: TABS,
               appearance,
+            },
+            global: {
+              plugins: [router],
             },
             slots: {
               'pictures-anchor': h('div', {}, picturesSlot),


Justineo:
I agree the invalid DOM needs to be addressed, and I understand that rendering links inside the `*-anchor` slot may be legacy behavior that some consumers depend on.

That said, I’m not convinced we should continue optimizing for anchors rendered inside that slot long term. Ideally, `KTabs` should own the interactive tab trigger and determine the rendered element based on `tab.to`:

- no `tab.to` → button-like tab trigger  
- string `tab.to` → native `<a>`  
- object `tab.to` → `<router-link>`

The `*-anchor` slot should focus on customizing the label or content, not introducing its own interactive element. Allowing consumers to render `<a>` or `<router-link>` inside the slot makes it unsafe for `KTabs` to wrap it with a `<button>` or `<a>`, and shifts responsibility for correct navigation semantics back to the consumer.

Before optimizing for the slotted-anchor case, can we confirm whether there are real use cases that cannot be expressed with `tab.to`? If such patterns exist and can’t be replicated with the current API, we should support them deliberately, with tests, and consider a deprecation strategy. Otherwise, I’d prefer to document and deprecate interactive content inside the anchor slot, and continue with the `KButton` or shared trigger approach. That keeps concerns like `to`, disabled state, keyboard interaction, and accessibility centralized within the component rather than relying on each consumer to implement them correctly.

johncowen:
> I’m not convinced we should continue optimizing for anchors rendered inside that slot long term

Let's put that aside for the moment? Removing it like this is a breaking change, and its breaking our application currently.

Would you agree that we should undo the bug with this PR? There are no changes here, just reverting the PR that caused the issue.

Once we've done that, happy to discuss more 👍 , let me know where is best.

Justineo:
Could you help me understand how this is breaking your consumer app? I ran [a quick test](https://codepen.io/Justineo/pen/dPOypbj), and even when an `<a>` is nested inside a `<button>`, it remains clickable and still triggers navigation. It seems more likely that the issue stems from this:

https://github.com/Kong/kongponents/blob/0bef6ec46c0de96aec5219c372c80571d659d658/src/components/KTabs/KTabs.vue#L26

...which was introduced almost two years ago. Has the tab been broken for that entire time?

johncowen:
There must be no interactive content inside buttons:

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#technical_summary

Maybe a good analogy, maybe not:

If I was an engineer building a bridge and I'd used the wrong bolts for a bridge and it then fell down, and then I was trying to argue that it "Looked Good To Me! I drove my own car over it and it was fine!" - people would be pointing at the instructions saying "you didn't follow the instructions"

---

Analogies aside, please could we revert the change and discuss later. I still haven't understood why this was changed in the first place.

johncowen:
Oh sorry missed a question:

> Has the tab been broken for that entire time?

Not that I know of.

But again can we revert to how it was please and we can discuss.

Justineo:
Have you read my last comment? I mean how is the application broken? Do you mean clicking the tab doesn’t trigger a navigation?

johncowen:
Sorry, yes I did, I tried to explain - our HTML is now broken.

This has accessibility concerns at least, and whilst might be fine in your browser right now I'm not totally any others now or in the future will continue to work because we are not "following the instructions".

Does that answer your question?

Also could you let me know why this was changed in the first place? At the moment I don't understand reasons why we can't just approve this revert and continue the discussion.

I would very much appreciate it if we could concentrate on/prioritise the important things.

Let me know if I've missed something

johncowen:
FYI here's another example of an issue where the HTML was wrong and we went ahead and fixed it

https://github.com/Kong/kongponents/pull/3076

My team and others were very appreciative from what I remember

johncowen:
Also, you said:

> I agree the invalid DOM needs to be addressed,

Just a few comments up?

Justineo:
> At the moment I don't understand reasons why we can't just approve this revert and continue the discussion.

Because I want to make sure I fully understand the issue and confirm whether this revert would actually fix it. As I understand it, if you place an `<a>` inside `KTabs`’s `anchor` slot, it won’t trigger navigation because there’s a `@click.prevent` on it. So the problem isn’t the `button > a` structure, but the use of the `preventDefault` modifier.

I tested this revert PR locally and anchor elements inside the tab still don't work. I wonder how does your call site look like, is it using a plain `<a>` or `<router-link>`?

johncowen:
> Because I want to make sure I fully understand the issue and confirm whether this revert would actually fix it.

Ok thats a good ask/query, let me come back to you with info 👍

johncowen:
~Hmm, I just tried building this from source and got a bunch of errors about swrv I must be doing something wrong there also.~

~But I tried the resulting build and you're right, this revert it didn't seem to fix the issue either 🤔~

~I need to jump for 30mins but I need to look further at this also, just wanted to let you know asap, I'll be back soon~

edit: Sorry! Totally got "branch-confused" I had the wrong branch checked out. See https://github.com/Kong/kongponents/pull/3171#issuecomment-4344408291

Justineo:
I can see two separate issues with `KTabs` right now:

1. If a user uses an anchor element in the `*-anchor` slots, the rendered HTML can end up with an invalid structure. This doesn’t directly break navigation, but it’s still incorrect.

2. If the anchor is a plain `<a>`, navigation becomes unusable because there is a `@click.prevent` applied outside the slot. In contrast, `<router-link>` may still work since it doesn’t rely on the default anchor behavior for navigation.

Even if we are using `<router-link>`, I think fixing the call sites would be cleaner. Instead of passing `<router-link>` through slots, we can move the same `to` prop onto the tab items themselves.

Justineo:
If the issue you mentioned is simply the broken HTML structure, reverting the change should fix it.

If the issue is that navigation no longer works, I believe the correct fix would be to remove `@click.prevent`.

> Also could you let me know why this was changed in the first place?

The tab buttons share similar behavior with `KButton`, eg. both support the same `to` and `disabled` behavior and conditionally render into a link. Also, `button` is actually the recommended semantic element for a tab button. See: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/

`KTabs` tab items also support a `to` prop, so I would expect this to be the recommended way to make tabs support navigation, similar to anchors or router links.

johncowen:
> Hmm, I just tried building this from source and got a bunch of errors about swrv I must be doing something wrong there also.

> But I tried the resulting build and you're right, this revert it didn't seem to fix the issue either 🤔

> I need to jump for 30mins but I need to look further at this also, just wanted to let you know asap, I'll be back soon

Sorry totally mistaken, I had a branch checked out already that had tabs in the name and I just assumed it was this one.

Let me strike out the above comment, sorry!

---

I'm pretty sure the revert works for us. I just rebuilt this revert branch (the correct one!) and tried it in our app, and the button/a issue is no longer there (we have div > a) and our tabs continue to work.

Screenshot 2026-04-29 at 14 55 04

I'm pretty sure look I looked into this when I made the revert, but I've totally forgotten the detail, I guess I should have gone into more detail in the description here!

Let me examine this again and give you some better reasoning

johncowen:
Ah yeah so it was this

```ts
return { tag: 'div', attributes: {} }
```

https://github.com/Kong/kongponents/pull/3171/changes#diff-01f83cc2092a263bfd47c1d0a4ba5e96dea98264d85282662a63f4fa18c44dedR100

which was removed. Instead of dynamically using different nodeNames depending on "things" it only ever uses a KButton now.

I can look into exactly what the "things" are if you like, but the fact that its only ever a KButton now means it doesn't really help to know that in order to revert the PR. But let me know either way.

johncowen:
P.S. the further information in your additional comments is interesting, but I'm still keen to revert (once we are sure that the revert fixes things 😅 ) and then continue discussing.

Justineo:
> Instead of dynamically using different nodeNames depending on "things" it only ever uses a KButton now.

Because `KButton` already provides this behavior. It renders a `<button>` since no `to` prop is passed to the tab items. If we leverage the built-in `to` support, there’s no need to slot `*-anchor` with router links.

Justineo:
Given #2707 landed exactly a year ago, I don’t think we need to treat this as requiring an immediate revert if we can ship the more robust fix directly.

I agree the invalid HTML needs to be fixed. My preference would be to address it in a way that moves `KTabs` toward owning the interactive trigger semantics, with the anchor slot reserved for label/content customization only.

Is there a specific consuming use case that cannot be represented with `tab.to` or another supported API path? If not, I’d prefer that we fix the invalid HTML while also moving toward the intended API.

johncowen:
Ok, I'm totally happy to discuss further once its reverted, as I've said a few times now. We could have been approved/merged/deployed and then discussing this by now 🤷 

> I agree the invalid DOM needs to be addressed,

You agree that this need to be addressed.

> Because I want to make sure I fully understand the issue and confirm whether this revert would actually fix it.

~We~ I have confirmed the revert will actually fix it

Let's revert?

---

Again, I really want to concentrate on the important things, and that to me is to revert the change and thus fix the issue.

I'm then totally happy to discuss further without feeling as though I need to agree with what you are saying in order to get a fix in - that just feels wrong.

Sounds like a plan? Revert, let's discuss.

adamdehaven:
> Ok, I'm totally happy to discuss further once its reverted, as I've said a few times now. We could have been approved/merged/deployed and then discussing this by now 🤷

This code has been in place for a year -- I don't think we need to rush making a revert here that could have broader impact across Konnect.

@johncowen can you provide some context around @Justineo's question here:

> Is there a specific consuming use case that cannot be represented with `tab.to` or another supported API path? If not, I’d prefer that we fix the invalid HTML while also moving toward the intended API.

---

> I agree the invalid HTML needs to be fixed. My preference would be to address it in a way that moves KTabs toward owning the interactive trigger semantics, with the anchor slot reserved for label/content customization only.

I'm aligned with this strategy -- let's fix forward.

Justineo:
I don’t think a full revert should be treated as the default solution here.

- This behavior has been in place for a year, so it doesn’t look like something that requires an immediate rollback of upstream behavior.
- A revert is not automatically a safe fix. It would undo a year-old change and restore older behavior for all consumers, which carries its own risk.
- If there is a cleaner way to fix the issue directly, I don’t see why we need to revert first before considering that.
- If a consuming app is currently blocked, it can patch locally and then migrate to `tab.to`.

johncowen:
Adam, please can we all try to stick to the important things.

There is a bug (we've agreed on that).

This is a clean revert that would fix the bug (we've agreed on that).

My team have noticed the bug in 3 separate pieces of software. We do not know as yet what else it affects, we do not know for sure that this bug also affected multiple pieces of software, but we should assume that is the case.

As this is a revert it is highly unlikely to cause further issues. Reverts of understood bugs are generally considered a safer approach to adding more and more code on top in an attempt to fix things.

That said, and to underline, please can we avoid taking this thread off on a tangent, but I'd like to do you the courtesy of addressing your comment at least. I understand that part of this might come across as critical but I need to be open and give you some insight into my concerns. I hope you understand.

---

This PR had been up for a while with no review.

Offline, I posted in the dedicated channel for kongponents with a PR link asking if someone could review. You responded saying something like: it looks like you need @Justineo to review. You didn't want to help me review, you wanted someone else to do it. Maybe you were too busy at the time, but could I kindly suggest that if you feel strongly about something, that you get involved from the outset? It felt to me like we had exhausted all avenues here and we were about to revert, and that is now why you have felt the need to get involved, maybe I have misunderstood the situation?

The code has been broken for a year. A full year. My team have noticed and reported it. It is broken in at least 3 separate applications, one CNCF/OSS serving that community, one purchased by enterprise customers and another subscribed to as a SaaS). Being responsible, we should assume that it could be broken in countless others. For a full year.

We have a release, for our paying customers who purchase this software, happening very soon indeed (a matter of days). I'd like to make sure this fix is in there. I'll say more on this below.

You are suggesting adding more code ontop of already broken code instead of reverting the change that was previously working just fine in all our applications, and we can assume countless others suggested this is "fixing it forward"

What is it to say that if we add more code ontop of this in an attempt to "fix" this that we will not break things even more. It is far more likley (and pretty much proven) that the previous code worked for far more time than a full year.

If you do break things further and we release our enterprise product, then that release will be broken. We are hoping this is the final point release of our v2 software and we can move onto v3. I wouldn't like our final point release of v2 to be broken.

I can't help but be reminded of one of my team mates interactions with you yesterday, when instead of approving a single line fix he made for an issue that was blocking us. You made further commits to a bot PR to fix a security issue with no explanation of why, a further PR to "forward fix" (?) the problems caused by those unexplained commits. This "forward fix" caused us further issues that blocked us differently. So then we were blocked in two separate places. 

My team had to put up a further PR to finally fix the issue correctly. It seemed to me that instead of a simple LGTM!, after a long thread, several strange commits and several PRs later we finally got unblocked by undoing the what had been done to "fix" the issue.

This happened just yesterday and it feels like you are trying to push through a "forward fix" again. So I'm sorry, but you specifically telling me this is "fixing forward" makes me less likely to want to do that, I feel that I wouldn't be doing anyone any favours (including yourself) if I was to accept your "let's fix forward" statement. I'm saying this for all of us.

I still haven't understood why this original amend was made, as I've said in the description of the PR and in several comments, the reason for the original PR is "unexplained"

---

On Konnect. I've have been in conversations with you where you push back on several engineers' opinions saying that Kongponents is an open source library used in I guess hundreds of codebases. If you are now saying we can't fix things that is broken in other software because of Konnect, it feels like you have different opinions depending on which opinion suits your argument of the time. I dunno maybe I'm getting the wrong end of the stick?

I have more to say on Konnect, but again I'm trying to not to take let people take this off on a tangent. Let's concentrate on the important things.

---

Lastly, because I am trying to be as courteous and respectful as I can whilst concentrating on the important things (see above)

> Is there a specific consuming use case that cannot be represented with tab.to or another supported API path? If not, I’d prefer that we fix the invalid HTML while also moving toward the intended API.

I'm not totally sure, we are using `<router-link :to="">` which is standard and documented `vue-router` functionality. We are trying not to reinvent the wheel or add too much unnecessary (potentially buggy) javascript on top of that.

This is my point. I'm not totally sure, and I don't think other people here are either - which is why I'd rather revert and continue discussing.

My biggest fear is, recollecting another PR/discussion we'd had about rendering JSON to tables (and I bring it up to explain my concern here) is that we end up with a component that is overly difficult to use to render some anchor links to a page. I know the "rendering something to something" sounds like I'm simplifying things, or may dismissing things (? some folks may think?) but I don't think I am. In my opinion this stuff should be easier than it currently is and to me thats a problem.

This should be `nav > li > a` with some CSS - I honestly don't understand what all the javascript is for.

If you have more to say on the above, please can we take the conversation elsewhere, at least temporarily.

---

Back to the important things. Lets revert this and continue the discussion? I'm more than happy to do that, just let me know where is best.

I do have more to say on the important things here. So I'll do that in a separate comment to try to avoid this particular conversation taking us all off on a tangent. I've put great care and effort into explaining my viewpoint in an hope that we can avoid the tangent but answer your questions, I sincerely hope its taken as what it is: a further attempt to find a way for us to work together better.

Thankyou for reading.

Justineo:
I wanted to summarize the technical context here, because I think the thread has been missing some important details about the actual affected usage, the relevant history, and what the revert actually fixes.

Since the affected call sites and migration constraints were not clearly described in the discussion, I looked into `kuma-gui` (hopefully I looked in the right place) and the relevant `kongponents` history myself so we can keep the discussion grounded in concrete behavior.

What I found:

1. The affected `kuma-gui` usage is not simply an arbitrary unsupported consumer pattern.

   `kuma-gui` has an internal `XTabs` wrapper around `KTabs`. `XTabs` forwards its `*-tab` slots into `KTabs`’ `*-anchor` slots.

   Inside those slots, many call sites render `XAction :to`. `XAction` renders a Vue `RouterLink` when `to` is provided and the appearance is the default anchor style.

   So the current effective rendering chain is:

   ```text
   XTabs *-tab slot
     -> KTabs *-anchor slot
       -> XAction :to
         -> RouterLink
   ```

2. This pattern predates the current `KTabs tab.to` API.

   `XAction` was added in `kuma-gui` in March 2024. Before that, `kuma-gui` was already using `RouterLink` inside `KTabs` through `NavTabs`.

   Earlier, the chain looked like this:

   ```text
   NavTabs
     -> KTabs *-anchor slot
       -> RouterLink
   ```

   `XTabs` was introduced in April 2024 and then rolled out across the app shortly after. At that time, Kongponents did not yet have `tab.to` support in `KTabs`.

   Kongponents added `tab.to` later, in December 2024, in the PR that also removed the old docs/sandbox example showing `router-link` inside the `*-anchor` slot and deprecated `anchorTabindex`.

   So this `kuma-gui` pattern was consistent with the older Kongponents example at the time it was introduced. I think it is better described as legacy supported usage that was later superseded by `tab.to`.

3. In April 2025, in the PR that this PR aims to revert, I moved `KTabs` to use `KButton`.

   The reason was to reuse the shared `to` and `disabled` handling from `KButton`. At the time, `tab.to` appeared to be the documented path for route-based tabs, and the older `router-link`-inside-`*-anchor` example had already been removed from the docs.

   That change makes sense if `KTabs` owns the interactive trigger and consumers express navigation through `tab.to`.

   However, it changes the behavior for legacy `*-anchor` slot usage. If a consumer renders an interactive element such as `RouterLink` or `<a>` inside the anchor slot, `KTabs` now wraps that slotted content with a button-like trigger, which can produce invalid interactive nesting such as:

   ```text
   button
     -> RouterLink / a
   ```

4. There are multiple real `kuma-gui` call sites, but the impact appears to be internal and fairly localized.

   I found 17 `XTabs` blocks in `kuma-gui`, and all of them use the `XTabs -> XAction :to` pattern.

   `XAction` is exported from `@kumahq/x`, but `@kumahq/x` is a private workspace package, not a public npm package like Kongponents. So I would describe the downstream risk as affecting Kuma GUI and products/users consuming that built app, not external customer code directly importing `XAction`.

   If this is the main affected usage, then the problem appears to be reasonably contained around the `XTabs` abstraction rather than spread across arbitrary `KTabs` consumers.

5. One important nuance: the revert works for `kuma-gui` largely because `XTabs` already performs DOM-level post-processing.

   `XTabs` renders `KTabs`, then manually removes attributes from the generated DOM:

   ```ts
   const $tablist = $el.querySelector('[role="tablist"]')
   const $tabs = $el.querySelectorAll('[role="tab"]')
   const $tabindexed = $el.querySelectorAll('[tabindex="0"]')
   const $id = $el.querySelectorAll('[id]')

   $tablist && ['role', 'aria-label'].forEach(attr => $tablist.removeAttribute(attr))
   ;['role'].forEach(attr => Array.from($tabs).forEach(item => item.removeAttribute(attr)))
   ;['tabindex'].forEach(attr => Array.from($tabindexed).forEach(item => item.removeAttribute(attr)))
   ```

   This explains why the revert fixes the known `kuma-gui` usage: when the outer wrapper is a `div`, this post-processing can turn the structure into something close to:

   ```text
   div
     -> RouterLink
   ```

   But for direct `KTabs` consumers, reverting does not necessarily produce a fully clean accessibility structure if they put interactive content inside the `*-anchor` slot. It changes the structure from:

   ```text
   button
     -> a / RouterLink
   ```

   back to something closer to:

   ```text
   div[role=tab][tabindex=0]
     -> a / RouterLink
   ```

   That is still not the ideal model, because there can still be nested focus/interactive semantics unless the consumer also handles `anchorTabindex` or otherwise removes the outer tab semantics.

   So I would treat the revert as a compatibility fix for this legacy usage, not as the final accessibility fix.

6. Given the concrete `kuma-gui` usage, I think there is a feasible fix-forward path.

   If the main affected usage is the `kuma-gui` `XTabs -> XAction :to` pattern, then the longer-term fix could be to move that route information out of the slotted `XAction` and into the `KTabs` tab data via `tab.to`.

   `XTabs` currently derives its `KTabs` items from slot names. The route target currently lives inside the slot content:

   ```vue
   <template #[`${name}-tab`]>
     <XAction :to="{ name }">
       ...
     </XAction>
   </template>
   ```

   That means `XTabs` cannot currently pass `tab.to` to `KTabs`, because `XTabs` does not know what `to` is.

   However, the current `XTabs` usages already generate those slots from route data, usually `route.children`.

   Current usage:

   ```vue
   <XTabs :selected="route.child()?.name">
     <template
       v-for="{ name } in route.children"
       :key="name"
       #[`${name}-tab`]
     >
       <XAction :to="{ name }">
         {{ t(`services.routes.item.navigation.${name}`) }}
       </XAction>
     </template>
   </XTabs>
   ```

   So the data already exists at the call site. It is just encoded into a slot name and hidden inside the slotted `XAction`.

7. Proposed `XTabs` API direction: move from slot-derived tabs to data-driven tabs.

   Instead of having `XTabs` enumerate slots to discover tabs, it could accept tab data directly.

   Proposed usage for simple route tabs:

   ```vue
   <XTabs
     :selected="route.child()?.name"
     :tabs="route.children.map(({ name }) => ({
       name,
       label: t(`services.routes.item.navigation.${name}`),
       to: { name },
     }))"
   />
   ```

   Proposed usage for tabs that need query state:

   ```vue
   <XTabs
     :selected="route.child()?.name"
     :tabs="route.children.map(({ name }) => ({
       name,
       label: t(`connections.routes.item.navigation.${name.split('-')[5]}`),
       to: {
         name,
         query: {
           inactive: route.params.inactive,
         },
       },
     }))"
   />
   ```

   `XTabs` could then convert these into `KTabs` tab items:

   ```ts
   const kTabs = computed<Tab[]>(() => props.tabs.map(tab => ({
     title: tab.label,
     hash: `#${tab.name}`,
     to: tab.to,
   })))
   ```

   And slots could become optional label/content customization only:

   ```vue
   <XTabs
     :selected="route.child()?.name"
     :tabs="tabs"
   >
     <template #overview-tab="{ tab }">
       <span>
         {{ tab.label }}
         <XBadge>New</XBadge>
       </span>
     </template>
   </XTabs>
   ```

   This changes the API shape from:

   ```text
   route.children
     -> v-for creates named slots
     -> XTabs enumerates slot names
     -> XTabs derives KTabs tabs
     -> slot renders XAction :to
     -> XAction renders RouterLink
   ```

   to:

   ```text
   route.children
     -> XTabs receives tabs prop
     -> XTabs passes tab.to to KTabs
     -> KTabs owns the single interactive trigger
     -> optional slot customizes label/content only
   ```

8. If my understanding is correct, the `kuma-gui` fix may be relatively contained.

   It would not require hunting through arbitrary downstream code or rewriting unrelated navigation. The main work would be to update the `XTabs` API/implementation and then adjust its known call sites to stop putting `XAction :to` inside the tab slot.

   Based on the current usages, this looks feasible:
   - most call sites already iterate over `route.children`,
   - most route targets are simply `{ name }`,
   - a few need query state preserved in `to`,
   - and the change would be centered on `XTabs`, which is already the abstraction responsible for adapting Kuma GUI navigation tabs to `KTabs`.

   This is still more work than a direct revert, and it may still carry release risk if the timing is tight, but it does not look like an unbounded migration if the affected usage is indeed the `XTabs` pattern.

9. Moving `XTabs` to `tab.to` would also let `kuma-gui` remove some of its DOM-level hacks.

   Today `XTabs` has to render `KTabs` and then strip `role`, `tabindex`, and generated IDs from the DOM after render. That is fragile because `XTabs` is fighting the semantics produced by `KTabs`.

   If `XTabs` passed route targets into `KTabs` via `tab.to`, then `KTabs` could own the single interactive element directly. `XTabs` would no longer need to render `XAction`/`RouterLink` inside `*-anchor`, and ideally would no longer need to strip `role`, `tabindex`, or generated IDs from the DOM after render.

   So I still think `tab.to` is the cleaner long-term model.

Given this updated understanding, I understand why a revert feels simpler and safer from the `kuma-gui` release perspective.

@johncowen, does this match the actual problem you are seeing?

Now that I have looked through the code, I understand the Kuma GUI compatibility concern better. I can also understand why reverting may feel simpler for Kuma GUI right now, because changing `XTabs` and its call sites would take some work.

That said, I still think reusing `KButton` in `KTabs` is the correct direction for Kongponents. It keeps `to` and `disabled` behavior centralized and gives `KTabs` a single owned interactive trigger. Reverting upstream would make the known Kuma GUI legacy usage work again, but it would not make interactive content inside `*-anchor` slots a generally correct accessibility model for direct `KTabs` consumers.

Given that, I will not keep blocking this revert if that is what is needed to unblock the immediate release.

However, my recommendation is still that Kuma GUI should patch this behavior on its side temporarily, then remove that patch once `XTabs` is migrated to pass route targets through `KTabs tab.to`. That keeps the urgent release unblock scoped to the known affected app, while preserving the upstream Kongponents direction that I still believe is correct.

Could you confirm whether the affected usage is indeed the `XTabs -> XAction :to` pattern described above, and whether there are other affected call-site shapes that I missed?

johncowen:
Hey @Justineo I'm just going over this PR again now with an eye to getting a response to your previous comment (the one above your last one).

Let me read over this comment also, I had a glance but it looks like its full of useful information and I want to read over it properly. Might I suggest that we get on a zoom call to discuss?

I'm keen on bringing this to a conclusion where we either approve and merge this PR or close it and I'd like that to be the objective of the conversation.

Please let me know and I can schedule a call 👍

Justineo:
In fact, I’ve already organized all the points I could think of in my replies above. Hopefully you’ll be able to find answers there to any questions that may come up. Our China team has just entered a five-day holiday, so I may not be available for a Zoom meeting during this time.

johncowen:
I'm going to quickly answer your questions here, but lets try and continue on a call:

> @johncowen, does this match the actual problem you are seeing?

edit: ~Partially yes~, Partially yes, but partially no, I do have a problem with the term "legacy", unless I'm misremembering this is supported behaviour, and from what I remember when we had issues with this component previously the current usage is what we were advised to do, but I can try check on that, and there has been no breaking change bump since.

Also I'm not fully sure I understand your statement on the accessibility tree, I think you say it moves a a nodeName from button to a div, but I think we only remove attributes again as a result of the last change we were asked/advised to use.

---

edit: I just checked this statement for myself:

> One important nuance: the revert works for kuma-gui largely because XTabs already performs DOM-level post-processing.

Whilst its not clear what is meant by "largely because", I tried the revert and removed the "DOM-level post-processing" that you describe (that remove the incorrect aria attributes). From what I could see, the revert continues to function as before and the incorrect HTML is fixed. Unless you are seeing differently, we should discard that statement as I believe it is misleading/not correct. Please could you check and let me know.

edit:
> slots a generally correct accessibility model for direct KTabs consumers.

This is also an incorrect statement.

You are confusing "behavioural" ARIA tabs, not horizontal navigation which is what KTabs is being used for. It's a horizontal navigation bar with some underlines. ARIA is about behaviour for visually impaired users (and others), its not about how things look.

---

> Could you confirm whether the affected usage is indeed the XTabs -> XAction :to pattern described above, and whether there are other affected call-site shapes that I missed?

Yes, XTabs is our usage, but the API to XTabs/NavTabs is in a good few places. We could change it, but as you point out I was hoping to avoid changing all of that for reasons we've touched on and I won't go into again here.

You do not mention KongMesh nor Mesh Manager above, for these things we need specific functionality and people often miss the point that most people using kongponents have a single application to support (or they have two essentially "forked" ones). Very commonly people don't realise these additional things that we have to take into account.

Additionally we've just done a lot of work around this component again to get our applications to work as required. So not to get into the weeds, but all of this conversation affects this.

This was just to give you some immediate answers, but I would still prefer a zoom call.

---

I've just seen your further response, but let me post the above anyway.

If you have your holiday marked in your calendar (if not please let me know when you are back), and I can schedule a call, does that sound ok? Could you let me know either way, thanks!

Hopefully we can find a way forwards 👍

Justineo:
> I do have a problem with the term "legacy"

Fair. To clarify, I don’t mean “unsupported hack”. I mean a previously supported slot-based pattern that has since been superseded by `tab.to`.

Kongponents now documents `tab.to` for link tabs, removed the old `router-link`-inside-`*-anchor` example, and deprecated `anchorTabindex` with the note that the old links-inside-tab-buttons use case is now covered by `to`. So I think “previously supported, now superseded” is the accurate framing.

> I’m not fully sure I understand your statement on the accessibility tree...

My point is not that Kuma’s DOM patch is irrelevant. My point is that Kongponents’ public API cannot rely on downstream apps patching generated DOM after render.

If `KTabs` is meant to be ARIA tabs, Kongponents should render a correct `tablist` / `tab` / `tabpanel` model itself. If it is meant to be horizontal navigation, Kongponents should expose a navigation mode/API and render navigation/link semantics directly. Either way, the correctness should come from Kongponents, not from Kuma removing attributes after the fact.

> the revert works for kuma-gui largely because XTabs already performs DOM-level post-processing

The revert itself changes the structure back from `button > RouterLink` to `div > RouterLink`, so it does remove that invalid nesting for your current usage even without the Kuma DOM patch.

The broader point still stands, though: the revert fixes the immediate invalid nesting, but it does not make `KTabs`’ own accessibility model correct. It restores an older behavior that Kuma can work around, but it does not resolve whether `KTabs` is ARIA tabs, horizontal navigation, or both.

> You are confusing "behavioural" ARIA tabs, not horizontal navigation

I just don’t think the answer is “slot an interactive element and patch the DOM”. The answer should be an explicit Kongponents API for the intended semantic mode.

For context, I checked several mainstream implementations:

| Library | Default trigger | Navigation / route support |
|---|---|---|
| MUI | `button[role="tab"]` | `component={Link}` / `href`; nav example uses `Tabs[role="navigation"]` |
| Radix | `button[role="tab"]` | `asChild` lets the consumer provide the trigger root |
| Chakra | `button[role="tab"]` | `asChild` + `navigate` |
| Ark UI | `button[role="tab"]` | `asChild` + `navigate` |
| Base UI | `button[role="tab"]` | `render` prop for custom root |
| Bootstrap | `button[role="tab"]` for tab panels | plain nav links for navigation |
| Headless UI | `button[role="tab"]` | `as` to change root |

So defaulting an ARIA tab trigger to a button is not unusual. The part we need to improve is the navigation/composition story.

My updated view is that `tab.to` and slotted `RouterLink` support do not have to be mutually exclusive.

`tab.to` should remain the built-in, recommended routing API because it lets `KTabs` own the trigger element and centralize `to`, disabled state, keyboard behavior, and accessibility semantics.

But for advanced/custom use cases, we can still support a slotted `RouterLink` path deliberately, as long as it is modeled as composition rather than nesting. In other words, the slot should be able to provide the trigger root, and `KTabs` should pass the necessary attrs/handlers/state into that slot instead of wrapping the slotted link in another interactive element.

Conceptually:

```vue
<template #overview-anchor="{ triggerProps, selected, disabled }">
  <RouterLink
    v-bind="triggerProps"
    :to="{ name: 'overview' }"
  >
    Overview
  </RouterLink>
</template>
```

That gives us two supported paths:

1. Recommended/simple path: `tab.to`
2. Advanced/custom path: slotted trigger root

Both preserve the key rule: there should only be one interactive trigger.

> Yes, XTabs is our usage, but the API to XTabs/NavTabs is in a good few places

Understood. I still think the issue is relatively contained in the sense that it is centered around the `XTabs`/`NavTabs` abstraction rather than arbitrary direct `KTabs` call sites, but I agree that “contained” does not mean “trivial” or “risk-free”.

> You do not mention KongMesh nor Mesh Manager

Fair point. I had a look at `kong-mesh-gui` and found a `XTabs` copy and two additional usages.

Given that, I won’t block this revert. After this revert, I plan to open a follow-up Kongponents PR that moves the default implementation back to `KButton`, while adding a new explicit API for the cases where consumers need the slotted element itself to be responsible for the interactive trigger. That should preserve the default `tab.to` direction while also giving Kuma/KongMesh-style usage a supported path without relying on invalid nesting or downstream DOM patching.

For the follow-up, I’d prefer to continue async rather than move this to a Zoom call. This discussion depends on concrete DOM output, ARIA semantics, component API shape, mainstream library patterns, and downstream usage. Written examples and references make it easier to reason precisely and avoid relying on memory or assumptions.

src/components/KTabs/KTabs.vue
```diff
@@ -90,6 +88,18 @@ const getAnchorTabindex = (tab: Tab): string => {
...
+  if (tab.to) {
+    if (typeof tab.to === 'string') {
+      return { tag: 'a', attributes: { href: tab.disabled ? undefined : tab.to } }
+    } else if (typeof tab.to === 'object') {
+      return { tag: 'router-link', attributes: { to: tab.disabled ? undefined : tab.to } }
+    }
+  }
+
+  return { tag: 'div', attributes: {} }
+}
```
Justineo, line 105: For `disabled` tabs with an object route, this passes `to: undefined` into `<router-link>`. `RouterLink` requires `to`, and `vue-router` warns that resolving `undefined` will fail in production. In dev it resolves to the current route, so a `disabled` route tab can still render as a real link without the click-capture prevention that `KButton` provided. We should keep the original to and explicitly suppress `disabled` link interaction, or reuse the shared `KButton` handling.

Justineo (changes requested):

src/components/KTabs/KTabs.vue
johncowen: Ok great. This sounds like it might answer why the PR I was reverting was done in the first place.

Let me take a look.

johncowen: Ok, lets keep concentrating on the important stuff here:

TL;DR: Uses `!tab.disabled` to avoid rendering anchors.

- I've added `!tab.disabled` which makes any disabled tabs use a `div` instead of an anchor to render. A disabled anchor is just a piece of text, which in most cases means a `span`. But, I've pragmatically chosen to reuse the existing `div` element so as to not over-complicate things further.
- Following the above, the previous forks/ternaries for `tabs.disabled` are unnecessary so I've removed those.

Please can you let me know if you are happy with the above or you have any further suggestions. Please take into account I'm still keen on keeping any changes as minimal as possible.

If you are happy with this, it would be great if you could check this change in the applications in which you have some sort of ownership (which is where I understand original the problem was). Then let me know either way whether it looks good or not 👍 

Additionally:

- I noticed there was an existing test `renders the tab as a link with no href attribute if tab.to is present and tab.disabled is true` that only tests the `to: string` variant. I added the object based variant and found that the tests do not render `router-link` correctly - `router-link` is pretty much mocked to a noop.
- I've tried to make the test more realistic and reduce any mocking of javascript functionality by adding a full vue router which then removes implicit noop mocking of router-link.
- Once I had done this I checked that the test previous to my amend failed, and then passed once I'd added my amend back in.
- I then proceeded to add the router-link plugin to all of the other tests - hopefully they are all a bit closer to an actual behavioural test now rather than a unit test i.e. hopefully they are a little more valuable. Whilst they continue to pass, I didn't look at them in great detail. This shouldn't effect actual functionality of course and this isn't strictly necessary for the revert but hopefully will surface similar issues in the future.

johncowen: Hey @Justineo 👋 Sorry to pester!:

- FYI: The two applications that we ship to customers are entering stabilisation phase this coming Monday (18th May). This means we try to avoid making any large changes from then on until after things are released. As I explained this release should be the last minor release before we move to another major.
- You "Requested Changes" over a week ago and I made some changes around a week ago as requested (see above comment and commit) I've not heard back as yet, apologies if you have been afk!
- Before the changes were requested the conversation with pretty fast flowing, so I'm a little concerned in case this has been forgotten about.

Do you know whether you'll be able to review the changes I made soon? I'm keen to do "something" before our release.

Lmk!

Justineo (approved):
Approving this as a tactical compatibility fix.

The pattern of passing interactive elements into the `*-anchor` slot has never been semantically handled correctly by `KTabs` itself. Downstream usages like `kuma-gui` made it work by applying DOM-level patches as a workaround.

This should not be considered a correctness fix for `KTabs` itself. `KTabs` still needs a follow-up that addresses the a11y model directly: clearly defining which element owns the interactive role, and handling `tabs` vs. `navigation` semantics explicitly.

johncowen:
Awesome thanks @Justineo 👍 

Sorry, just to keep concentrating on the important stuff here, could you confirm you did (or didn't?) do this for me: 

> it would be great if you could check this change in the applications in which you have some sort of ownership (which is where I understand original the problem was). Then let me know either way whether it looks good or not 👍

It's not clear from your response whether you'd done this or not? I'm looking to get this merged tomorrow hopefully, and I'm aware we are on different timezones so wanted to check on this before tomorrow comes around for me.

lmk!

Justineo:
Would you mind creating an adoption PR for `konnect-ui-apps` with the preview version (`@kong/kongponents@pr-3171`)?

johncowen:
Thanks for the response! Ok so from that I'm assuming that you didn't do this:

> it would be great if you could check this change in the applications in which you have some sort of ownership (which is where I understand original the problem was). Then let me know either way whether it looks good or not 👍

Glad I asked/checked!

Let me see if I can figure out exactly where the original problem occurred (that prompted the original PR I'm reverting), figure out the reproduction and check this revert/fix that I've done.

In the meantime, do I need to check anywhere in kong-manager do you know or just konnect-ui-apps? lmk!

Justineo:
I think `konnect-ui-apps` alone should be fine. The original PR wasn’t tied to any specific issue and was mainly intended to align the behavior with `KButton`, including handling of `to`, `disabled`, etc.

johncowen:
awesome thankyou 🙏

👀

PR #3173, opened 2026-04-27
fix: bumping actions versions to prevent Node20 warning [KHCP-20416]

ValeryG:
# Summary

Those are to prevent Node20 depiction warnings


johncowen (approved):
Not sure if I have approval power here, but lets give it a go

adamdehaven (approved):

PR #3175, opened 2026-04-28
fix(deps): correct peer dependencies

adamdehaven:
# Summary

Fix the peer dependencies from #3174

PR #3177, opened 2026-04-29
fix(deps): vue-router semver range

schogges:
Unfortunately I didn't check back the correction of peer range in https://github.com/Kong/kongponents/pull/3175 well enough. The range `>4 || >5` is the same as `>=5` which means that versions of major 4 and below are excluded and all version from 5 upwards are included (`v5`, `v6`, `v7`, ...). I've changed it now to `4 || 5` which should include all versions of 4 and 5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
   },
   "peerDependencies": {
     "vue": ">= 3.5.0 < 4",
-    "vue-router": ">4 || >5"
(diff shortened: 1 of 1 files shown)


johncowen (approved):
Nice spot! 👏 

I hit approve but I don't think my approval counts 😭 , but lets see!

adamdehaven (approved):

PR #3180, opened 2026-05-01
fix(kcodeblock): add copy code prop [KHCP-20471]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-20471

Add `copyCode` prop in KCodeBlock


kaiarrowood (approved):
LGTM! 🚀

Issue #2678, opened 2025-04-01
KTable headers are on an incorrect z-index


johncowen:
A `z-index: 2` added in https://github.com/Kong/kongponents/pull/2629 which causes the following to happen:

![Screenshot 2025-04-01 at 19 16 52](https://github.com/user-attachments/assets/259b49e8-8f07-4bce-97ee-8b4692b69d25)

Related: https://github.com/kumahq/kuma-gui/pull/3729

johncowen:
Hey @portikM 👋 thanks for taking a look!

You might want to use CSS `isolation:` to ensure the z-indexing here doesn't interfere with consuming applications.

https://developer.mozilla.org/en-US/docs/Web/CSS/isolation

 I've applied this in our application as its a better fix and doesn't overwrite the z-index added here. I also gave it a quick try in the kongponent itself and it also seems to solve the issue but still support the z-indexing you need due to the custom KInput icons, but I'll leave it up to you.

Let me know if you apply it here and see if we want to remove it from our application 👍 , alternatively let me know and we decide whether to close this issue or keep it open so others encountering issues are aware of this workaround.

adamdehaven:
Is this issue now resolved?

PR #3191, opened 2026-05-07
fix(ktableview): actions dropdown popover attributes prop [KHCP-20535]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-20535

Adding `actionsDropdownPopoverAttributes` prop in KTableView which aims to solve this problem:

When using KTableView / KTableData inside of KModal / KPrompt, the modal element `z-index` is higher than default KPop `z-index`. That means actions dropdown (which is teleported to `body`) will appear underneath the modal. There is no easy way to override it - targeting the `.popover` with global CSS overrides may result in other dropdown elements being affected and potentially causing other `z-index` related issues.

### DIFF
--- a/docs/components/table-view.md
+++ b/docs/components/table-view.md
@@ -808,6 +808,18 @@ A boolean to disable some of the table features like column visibility, column r
 
 Prop for hiding toolbar. Useful when elements provided in the toolbar are not actionable (for example, when toolbar contains filter controls, however the fetcher returns no results as there are no records to filter by).
 
+### actionsDropdownPopoverAttributes
+
+Attributes (of type [`PopoverAttributes`](/components/popover#popoverattributes)) to be applied to the row actions dropdown popover. Useful for overriding things like `zIndex` or `maxHeight` on the actions popover. Defaults to `{}`.
+
+```html
+<KTableView
+  :data="tableData"
+  :headers="headers"
+  :actions-dropdown-popover-attributes="{ zIndex: 9999 }"
+/>
+```
+
 ### tooltipTarget
 
 Teleport target element selector for tooltips rendered by KTableView (e.g. column header tooltip, disabled row bulk action checkbox tooltip). Defaults to `body`.
--- a/src/components/KTableView/KTableView.cy.ts
+++ b/src/components/KTableView/KTableView.cy.ts
@@ -316,6 +316,22 @@ describe('KTableView', () => {
       cy.get('th').eq(options.headers.indexOf(options.headers.find((header => header.key === 'actions'))!)).find('.table-header-label').should('have.class', 'sr-only')
     })
 
+    it('applies actionsDropdownPopoverAttributes to the actions dropdown popover', () => {
+      cy.mount(KTableView, {
+        props: {
+          headers: options.headers,
+          data: options.data,
+          actionsDropdownPopoverAttributes: { zIndex: 9999 },
+        },
+        slots: {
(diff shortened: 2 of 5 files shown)


johncowen:
Hey @portikM sorry for the random suggestion but this PR reminded me of

https://github.com/Kong/kongponents/issues/2678#issuecomment-2796614178

Which suggests using CSS `isolation` to avoid "zindex battles".

I'm not sure if its the same/similar issue but as its zindex related and the same component I thought I'd mention it

Also briefly reading the description here it does sound to me like using CSS `isolation` might avoid the upping the zindex/adding further passthrough props etc.

Anyway lemme know I'd be interested to know if `isolation` would help here?

aanchalm01:
> sorry for the random suggestion but this PR reminded me of
> 
> [#2678 (comment)](https://github.com/Kong/kongponents/issues/2678#issuecomment-2796614178)
> 
> Which suggests using CSS `isolation` to avoid "zindex battles".
> 
> I'm not sure if its the same/similar issue but as its zindex related and the same component I thought I'd mention it
> 
> Also briefly reading the description here it does sound to me like using CSS `isolation` might avoid the upping the zindex/adding further passthrough props etc.
> 
> Anyway lemme know I'd be interested to know if `isolation` would help here?

Hey @johncowen -  I am picking up this PR while Max is out until tomorrow.

This PR adds an optional actionsDropdownPopoverAttributes prop on KTableView so host can pass extra popover settings through to the row actions KDropdown. That includes things like zIndex on the popover type, so apps can fix stacking when the menu is rendered with target: 'body' and needs to sit above their own layout (modals, nav, etc.). 

For [https://github.com/Kong/kongponents/issues/2678#issuecomment-2796614178](https://github.com/Kong/kongponents/issues/2678#issuecomment-2796614178) my read is that issue is about the sticky header using z-index **inside the table** so it stays above the scrolling body, which can leak into the page’s stacking order. Your isolation idea sounds like a good fit for that `table vs rest of page` case.

For the row actions dropdown specifically, we **still** need **explicit control** on the body-target popover: because the menu is teleported to body, it’s **outside the table subtree**, so putting isolation on the table wouldn’t replace tuning popover zIndex and fix the issue in hand.

johncowen:
Aaah gotcha, the poppy.js type thing. This bit right?:

> it’s outside the table subtree,

So yeah agree, I think you are right in that case, that it's not something that would be solved by `isolation`

Just thinking out loud: I guess its the case that a level for popovers is needed for normal content that is higher than all the normal content (but not modal content). After that, another level is needed for popovers inside modal content (which is higher than normal content) that is higher than all the modal content.

TL;DR Please ignore me! Thanks for the extra info! 🙇

aanchalm01:
> Aaah gotcha, the poppy.js type thing. This bit right?:
> 
> > it’s outside the table subtree,
> 
> So yeah agree, I think you are right in that case, that it's not something that would be solved by `isolation`
> 
> Just thinking out loud: I guess its the case that a level for popovers is needed for normal content that is higher than all the normal content (but not modal content). After that, another level is needed for popovers inside modal content (which is higher than normal content) that is higher than all the modal content.
> 
> TL;DR Please ignore me! Thanks for the extra info! 🙇

No worries. Yes, this is for the cases when the the dropdown is part of table inside a modal.

aanchalm01 (approved):
LGTM! Thank You for fixing this.

PR #3211, opened 2026-05-21
fix: cleanup overrides in package.json

ValeryG:
# Summary

override section of package.json is a mess. Cleaned up from scratch


jillztom (approved):
Approved for force merge

PR #3214, opened 2026-05-22
feat: add slots for filter items [ma-5062]

Chalks:
satisfies [ma-5062](https://konghq.atlassian.net/browse/MA-5062)

# Summary

If you have too many filters, the filter selection dropdown bleeds off the edge of the screen. Additionally, in some cases you want to be able to modify the content of the filter selection dropdown items.

## Screenshots

**bleed issue, before**

Screenshot 2026-05-22 at 4 52 57 PM

**bleed issue, after**

Screenshot 2026-05-22 at 4 55 12 PM

**Using the new slots to decorate the "API" item**

Screenshot 2026-05-22 at 4 56 50 PM

### DIFF
--- a/src/components/KFilterGroup/FilterSelector.vue
+++ b/src/components/KFilterGroup/FilterSelector.vue
@@ -1,7 +1,7 @@
 <template>
   <KDropdown
     ref="filterSelector"
-    :items="items"
+    width="auto"
     @toggle-dropdown="onToggle"
   >
     <template #default>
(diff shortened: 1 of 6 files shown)


Chalks:
Can see it in use at https://github.com/kong-konnect/konnect-ui-apps/pull/11609

src/components/KFilterGroup/FilterSelector.vue
```diff
@@ -64,11 +88,25 @@ const onToggle = (open: boolean) => {
...
+    // we need to scroll the item wrapper to the top.
+    await nextTick()
+    itemWrapperRef.value?.scrollTo({ top: 0 })
   }
 }
 </script>
+
+<style lang="scss" scoped>
+.items-wrapper {
+  max-height: 50vh;
```
portikM: question: so essentially we're setting maximum dropdown height to 50% of viewport height? Is that not too much? Maybe something like `500px` would be enough?

Chalks: Maybe. I would want to do it more like `max-height: min(500px, 90vh);` though. Then on screens that are less than 500px tall we get 90% of the space anyways?

portikM: Perfect 👍

portikM (approved):

PR #3219, opened 2026-05-26
feat(kbutton): button component tokens

adamdehaven:
# Summary

Implement `button` component tokens from https://github.com/Kong/design-tokens/pull/662 (replicating changes from #3190)

Since these are additive:

- Non-breaking for v9 of Kongponents utilized in Konnect and Dev Portal
- Dev Portal gains the ability to enable customers to customize their `primary`, `secondary`, `tertiary`, and `danger` buttons globally, rather than per-button.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ docs/.vitepress/.temp
 docs/.vitepress/.cache
 docs/.vitepress/cache
 docs/.vitepress/dist
+
+.claude/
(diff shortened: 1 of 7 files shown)


adamdehaven:
Two things I noticed:

- `line-height-small` leak — `.small` reuses `--kui-button-line-height-medium`; there's no `-small` token. Should add?
- Single `--kui-button-border-radius/-border-width` with divergent fallbacks — base falls back to `radius-30`, small to `radius-20`. The moment a theme sets `--kui-button-border-radius`, small buttons snap to the base radius. Either split per-size or normalize. (Same for border-width.)

PR #3232, opened 2026-06-10
feat: support operator delimiter overrides [MA-5120]

Chalks:
Satisfies [MA-5120](https://konghq.atlassian.net/browse/MA-5120)

# Summary

If you want to use an operator like `exists` (which generates the delimiter `': '`), you get this output:

Screenshot 2026-06-10 at 4 19 16 PM

There are cases where you want a different delimiter than the default. That's where the new `operatorDelimiter` comes in. It ONLY affects the delimiter. With `operatorDelimiter: ' in: '` you get:

Screenshot 2026-06-10 at 4 22 39 PM

[MA-5120]: https://konghq.atlassian.net/browse/MA-5120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/components/filter-group.md
+++ b/docs/components/filter-group.md
@@ -29,6 +29,7 @@ interface FilterSelection {
   operator: FilterOperator // 'eq' | 'neq' | 'contains' | 'exists' | 'lt' | 'lte' | 'gt' | 'gte'
   value: string | string[] // the user's selection
   text: string // the display string for that selection
+  operatorDelimiter?: string // override the display string for the delimiter displayed in the pill
 }
 ```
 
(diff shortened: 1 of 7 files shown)


adamdehaven (changes requested):
The implementation looks fine; however, did this change go through Product Design?

My concern is that this could easily lead to inconsistencies in the platform. 

Can you please get signoff from Katrina's team? 

(I'm only blocking for this discussion, code lgtm)

adamdehaven (approved):
lgtm

PR #3248, opened 2026-06-23
chore(deps): bump pnpm, node, axios override

portikM:
Renovate PR takeover: https://github.com/Kong/kongponents/pull/3165

- pnpm: 10.33.0 → 10.34.2 (packageManager + volta)
- node: 24.14.0 → 24.16.0 (.nvmrc + volta)
- axios override: >=1.16.0 → >=1.17.0 (pnpm-workspace.yaml)

# Summary


package.json
```diff
@@ -191,10 +191,10 @@
     "node": ">=v16.20.2 || >=18.12.1 || >=20.14.0 || >=22.14.0",
     "pnpm": ">=10.28.2"
   },
-  "packageManager": "pnpm@10.33.0",
+  "packageManager": "pnpm@10.34.2",
   "volta": {
-    "node": "24.14.0",
-    "pnpm": "10.33.0"
```
kaiarrowood, line 194: Do you need to update `engines` too?

portikM: No, see the original renovate PR

arashsheyda: in my experience renovate sometimes misses the engines or CI node versions so we do it manually, what's the reason that we want to only bump the volta node and not the engine 🤔

PR #3276, opened 2026-07-02
fix(kpop, kmodal, kslideout): teleported kpop z-index [KHCP-20474]

portikM:
# Summary

Addressing: https://konghq.atlassian.net/browse/KHCP-20474

When a KPop was used with a teleported popover (via the `target` prop) inside a KModal, KPrompt, or KSlideout, the popover rendered _underneath_ the container because its default `z-index` (`1000`) was lower than the modal's (`1100`) or slideout's (`9999`), and a CSS fix was impossible since teleporting removes the popover from the parent's DOM subtree. To solve this cleanly, KModal and KSlideout now provide their `z-index` through a shared injection key, and KPop injects it to automatically resolve to one above its parent when no explicit `zIndex` is set — leveraging the fact that Vue's `provide` / `inject` follows the component tree rather than the DOM, so it works through `<Teleport>`. An explicit `zIndex` prop still takes precedence, and KPop re-provides its resolved value so nested popovers continue stacking correctly.

### DIFF
--- a/docs/components/popover.md
+++ b/docs/components/popover.md
@@ -415,6 +415,8 @@ Attributes to be applied to `.popover` element.
 
 Pass a number to use for the `z-index` property. Default value is `1000`.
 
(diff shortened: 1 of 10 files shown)


kaiarrowood (approved):
LGTM! 🚀

PR #3281, opened 2026-07-08
fix(kcatalog, ktableview): render dynamic toolbar slots [KM-2835]

Justineo:
# Summary

[KM-2835]

Fixes toolbar rendering when consumers conditionally add the `#toolbar` slot after `KCatalog` or `KTableView` has already mounted.

The toolbar wrappers now read slot presence during render instead of caching slot keys in computed state. `KTableView` keeps the existing default toolbar controls behavior, and `KTableData` is covered through forwarding regression tests.

[KM-2835]: https://konghq.atlassian.net/browse/KM-2835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KCatalog/KCatalog.cy.ts
+++ b/src/components/KCatalog/KCatalog.cy.ts
@@ -183,6 +183,39 @@ describe('KCatalog', () => {
       cy.get('.k-catalog .catalog-toolbar button').should('contain.text', 'Toolbar button')
     })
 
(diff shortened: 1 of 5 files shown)


portikM (approved):
LGTM 🚀

PR #3282, opened 2026-07-09
fix(ktableview): resize handle NaN height on mount [KM-2845]

kingyue737:
# Summary

[KM-2845]

Preview: https://github.com/kong-konnect/konnect-ui-apps/pull/12557

`KTableView`'s resize handle height (`headerHeight`) was computed once from `getComputedStyle(th).height` inside a Vue `computed`, with no reactive dependency on actual DOM layout. Two compounding issues:

1. `getComputedStyle().height` can still return `"auto"` right when the table mounts, before layout has settled. `"auto"` is truthy, so it passed the guard and got `parseInt`'d into `NaN`, rendering `.resize-handle { height: NaNpx }` (handle invisible/undraggable).
2. Even with an `isNaN` guard, the `computed` has no reactive dependency on layout — it only reruns when the header row's template ref changes, not when the browser finishes laying out the row. So once it settled on a bad value, it would never recompute.

Replaced the one-shot `getComputedStyle` read with `useElementSize(headerRowRef)` from `@vueuse/core`, which is backed by a `ResizeObserver` and returns a reactive numeric height — no string parsing, no `NaN` path, and it stays in sync with the header row's real height for the lifetime of the component (window resizes, column resizing, content changes, etc).

[KM-2845]: https://konghq.atlassian.net/browse/KM-2845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KTableView/KTableView.cy.ts
+++ b/src/components/KTableView/KTableView.cy.ts
@@ -252,8 +252,11 @@ describe('KTableView', () => {
         cy.get('.resize-handle').first().should(([handle]) => {
           const row = handle.closest('tr')
 
+          // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+          expect(row).to.exist
+
           expect(parseFloat(window.getComputedStyle(handle).height))
-            .to.be.closeTo(row?.getBoundingClientRect().height ?? NaN, 1)
+            .to.be.closeTo(row!.getBoundingClientRect().height, 1)
         })
       }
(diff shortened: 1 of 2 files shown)


portikM (approved):
LGTM 🚀

PR #3283, opened 2026-07-09
fix: calendar only tries to reset times when appropriate [MA-5071]

Chalks:
Satisfies [MA-5071](https://konghq.atlassian.net/browse/MA-5071)

# Summary

`KDateTimePicker` tries to munge the start/end times to `00:00:00` and `23:59:59` respectively when a user picks the same date twice. It was doing that too greedily in too many cases which in some odd edge cases was causing a valid time choice of 4am - 10am on the same day to become 4am - 11:59:59pm instead.

[MA-5071]: https://konghq.atlassian.net/browse/MA-5071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KDateTimePicker/CalendarWrapper.vue
+++ b/src/components/KDateTimePicker/CalendarWrapper.vue
@@ -82,9 +82,9 @@
 <script lang="ts" setup>
 import type { DatePickerModel, DatePickerRangeObject, DateTimePickerMode, TimeGranularity } from '@/types'
 
-import { computed, nextTick, onMounted, ref, useId, watch } from 'vue'
+import { computed, onMounted, ref, useId, watch } from 'vue'
 import { DatePicker } from 'v-calendar'
-import { format, isBefore, isSameDay, startOfToday } from 'date-fns'
+import { format, isBefore, isSameSecond, isSameDay, parse, startOfToday } from 'date-fns'
 
 const {
   isRange,
@@ -118,7 +118,6 @@ const calendarVModel = defineModel<DatePickerModel>({ required: true })
 const hasError = defineModel<boolean>('error', { default: false })
 const startTimeValue = ref<string>(formatTimeForInput(new Date(), timeGranularity))
 const endTimeValue = ref<string>(formatTimeForInput(new Date(), timeGranularity))
-const isApplyingFullDayRange = ref(false)
 const originalTimeValues = ref<{ start: string, end: string }>({
   start: formatTimeForInput(new Date(), timeGranularity),
   end: formatTimeForInput(new Date(), timeGranularity),
@@ -197,7 +196,7 @@ const isValidDateRange = (
  * Automatically adjust times to create a full-day range (00:00:00 - 23:59:59)
  * @returns true if full-day range was applied, false otherwise
  */
-const applyFullDayRangeIfSameDay = async (): Promise<boolean> => {
+const applyFullDayRangeIfSameDay = (): boolean => {
   if (!sameDayFullRange || !isRange) {
     return false
   }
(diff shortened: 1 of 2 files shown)


src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -303,16 +307,18 @@ watch(() => endTimeValue.value, (newTime) => {
...
 
     hasError.value = isInvalidRange(calendarVModel.value.start, calendarVModel.value.end)
   } else if (calendarVModel.value instanceof Date) {
     startTimeValue.value = formatTimeForInput(calendarVModel.value, timeGranularity)
   }
-}, { immediate: true, deep: true })
+  // we don't want to calculate this immediately, because we need to wait for
+  // initTimeInputs to execute. Otherwise it can mistakenly reset the times to
+  // midnight/midnight when a valid time had already been selected (e.g. 5-10am)
+}, { immediate: false, deep: true })
```
filipgutica, line 327: Nit: not a blocker
vue `watch` is lazy by default, I think we can just leave `immediate: false` out altogether since that's the default.

Chalks: yeah, that's correct. The only reason I left it in was because I felt like the comment made more sense with it explicitly set to false.

I'm going to push to this branch again in a few minutes, so if you feel like that reason isn't good enough, I'll remove it.

src/components/KDateTimePicker/CalendarWrapper.vue
```diff
@@ -303,16 +307,18 @@ watch(() => endTimeValue.value, (newTime) => {
 
 watch(() => calendarVModel.value, () => {
   if (isRange && isValidDateRange(calendarVModel.value)) {
-    if (!applyFullDayRangeIfSameDay()) {
-      startTimeValue.value = formatTimeForInput(calendarVModel.value.start, timeGranularity)
-      endTimeValue.value = formatTimeForInput(calendarVModel.value.end, timeGranularity)
-    }
+    applyFullDayRangeIfSameDay()
+    startTimeValue.value = formatTimeForInput(calendarVModel.value.start, timeGranularity)
+    endTimeValue.value = formatTimeForInput(calendarVModel.value.end, timeGranularity)
```
filipgutica, line 318: Not a blocker, just clarification.

Could this lose the intended end-of-day value?

Before this change, when the component auto-expanded a same-day selection into a full-day range, it set the model end to `23:59:59.999` and blocked the time-input watcher from writing the displayed `23:59` back into the model.

After this change, we still set the model end to `23:59:59.999`, but then format that value into the input as `23:59`, and the time-input watcher can write it back to the model as `23:59:00.000`.

That means we may lose the last `59.999` seconds of the day. Not sure that is necessarily an issue for downstream consumers. Just wanted to call it out because this can result in the emitted range ending one minute earlier than the intended end of day boundary.

Chalks: hmm, this is a good point, but I'm not sure what to do about it.

The only way to get the last 59.999 currently is with this "exactly pick the end of the day" behavior. You can't manually request it any other way.

Maybe a better solution is `if (timeValue === 23:59) then set it to 23:59:59.999` explicitly? Then even if you just type in `23:59` anywhere it works that way. (edit: this is a bad idea probably)

Chalks: worth noting that this isn't tested anywhere either. :|

Chalks: also, I don't think the last minute matters in a 24 hour query, right? We don't provide second granularity at that scale.

filipgutica: Yeah I'm inclined to say it might not really matter for us

Chalks: I'll check to see if it makes a difference in the query result at least, I don't _think_ it does, but doesn't hurt to be sure.

jordojordo (dismissed):

src/components/KDateTimePicker/CalendarWrapper.vue
Chalks: you can see deploy preview at https://pr-12551.cloud-preview.konghq.tech/us/analytics

adamdehaven (approved):
lgtm

PR #3275, opened 2026-07-01
test(kdatetimepicker): first of month test failure

adamdehaven:
# Summary

Fix `KDateTimePicker` `sameDayFullRange` test to keep both selected dates within the current calendar month, preventing failures when run on the first day of the month.

PR #3285, opened 2026-07-10
fix(tests): uses deterministic dates for KDatePicker tests

johncowen:
I did this last week and only just remembered today.

I don't want it to block or complicate other in progress PRs I won't be merging this until the time is right (unintended pun!)

I'll come back and write a better desc also, but TLDR;

- Mostly uses a date helper function to make all dates deterministic during testing.
- In the end I added `cy.clock` also for reasons I'll explain later. Ideally I would have liked to not use runner specific things.

Related: https://github.com/Kong/kongponents/pull/3275

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,9 +54,9 @@
     "lint:fix": "eslint --fix",
     "stylelint": "stylelint './src/**/*.{css,scss,sass,vue}'",
     "stylelint:fix": "stylelint './src/**/*.{css,scss,sass,vue}' --fix",
-    "test": "cross-env DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome",
-    "test:open": "cross-env DISABLE_VUE_DEVTOOLS=true cypress open --component -b chrome",
-    "test:spec": "cross-env DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome --spec",
+    "test": "cross-env TZ=UTC DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome",
+    "test:open": "cross-env TZ=UTC DISABLE_VUE_DEVTOOLS=true cypress open --component -b chrome",
+    "test:spec": "cross-env TZ=UTC DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome --spec",
     "semantic-release": "semantic-release"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -54,9 +54,9 @@
     "lint:fix": "eslint --fix",
     "stylelint": "stylelint './src/**/*.{css,scss,sass,vue}'",
     "stylelint:fix": "stylelint './src/**/*.{css,scss,sass,vue}' --fix",
-    "test": "cross-env DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome",
-    "test:open": "cross-env DISABLE_VUE_DEVTOOLS=true cypress open --component -b chrome",
-    "test:spec": "cross-env DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome --spec",
+    "test": "cross-env TZ=UTC DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome",
+    "test:open": "cross-env TZ=UTC DISABLE_VUE_DEVTOOLS=true cypress open --component -b chrome",
+    "test:spec": "cross-env TZ=UTC DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome --spec",
```
adamdehaven, line 59: Rather than adding the timezone into the command here, can you just add this to the consumer (in the test file) or in the Cypress config if applicable

johncowen: Surely its best to set it once and the very top of all the tests?

That way its the same for all the tests and nobody forgets to add it causing tests to be different for different folks in different timezones

lmk!

adamdehaven: Why would we want to inject an ENV variable that is only needed for 1 test file into all tests? 

I'm just suggesting set the variable in the test that needs it, not globally for all tests.

johncowen: What happens if another test file somewhere somehow starts to use a Date? Does the author have to remember to change the timezone also, and if that author forgets we have to follow up with another PR?

Q: Whats the issue we are solving by not putting the TZ here?

adamdehaven: Then let's just set it in the cypress config: https://github.com/Kong/kongponents/blob/52dfcd5682ee9531532bbe38f1c11a5f915e89ec/cypress.config.ts#L6-L9

Just trying to keep the package.json clean

johncowen: Shall I move this one also?

`DISABLE_VUE_DEVTOOLS=true`

?

adamdehaven: Sure that would be great

johncowen: Actually I'm not totally sure setting TZ in Cypress config will work 🤔

adamdehaven: it makes it available in the test environment; where is it evaluated? It should be the same, no?

PR #3288, opened 2026-07-15
feat(kfiltergroup): expose scoped slot props for custom filter content

kingyue737:
[KM-2898]

## Motivation

Custom `KFilterGroup` filter content rendered via `filter-*` slots is currently not integrated with the pill's pending selection state. Consumers must keep separate local state, listen to `@open` / `@apply` / `@clear`, branch by filter key, and manually patch a `FilterSelection` into `v-model`.

This makes custom filters difficult to encapsulate and reuse because each one needs parent-level state and key-specific event handling.

## Approach

This PR lets custom filter slot content participate in the same pending-value + Apply/Cancel flow used by built-in input/select/multiselect filters.

The `filter-*` slot now receives scoped props:

```ts
interface FilterPillSlotProps {
  value?: string | string[]
  text?: string
  options?: FilterOption[]
  multiple?: boolean
  operators: FilterOperator[]
  operator?: FilterOperator
  setOperator: (op: FilterOperator) => void
  setValue: (value: string | string[], text: string) => void
}
```

A custom filter can bind its control directly to `value` / `setValue` and `operator` / `setOperator`. The existing footer Apply button then builds the `FilterSelection` and updates `v-model`, so consumers no longer need to reconstruct the selection object manually.

The slot intentionally exposes only pending-state APIs; apply, cancel, and clear continue to use the existing pill footer and clear behavior.

`value` remains `string | string[]` to match the existing `FilterSelection.value` and `FilterOption.value` contracts. Custom boolean/number-like values should use stable string representations such as `'true'` / `'false'`.

## Compatibility

This is additive. Existing bare `filter-*` slots can continue using host-managed `@open` / `@apply` / `@clear` handling. The new scoped props are opt-in for consumers that want `KFilterGroup` to manage the custom filter's pending value.

## What's included

- `src/types/filter-group.ts` — adds `FilterPillSlotProps` and threads it through `FilterPillSlots` / `FilterGroupSlots`.
- `src/components/KFilterGroup/FilterPill.vue` — lets custom filters use the same pending selection flow as built-in filters.
- `src/components/KFilterGroup/KFilterGroup.vue` — forwards scoped props to host-facing `filter-*` slots.
- `sandbox/pages/SandboxFilterGroup.vue` / `docs/components/filter-group.md` — add examples showing scoped-prop custom filters.
- Cypress coverage for prop forwarding, apply via `setValue`, and legacy bare-slot behavior.

[KM-2898]: https://konghq.atlassian.net/browse/KM-2898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/KFilterGroup/FilterPill.vue
```diff
@@ -189,8 +202,9 @@ const filterType = computed((): 'custom' | 'input' | 'select' | 'multiselect' =>
 
 /**
  * Whether the apply button is disabled or not. If the filter is a custom filter
- * we can't detect whether it should be disabled, so in that case it defaults to
- * false.
+ * we can't reliably detect whether it should be disabled (slot content may not
+ * use `setValue` at all and manage its own state instead), so in that case it
+ * defaults to false.
```
Chalks: **nice to have, not a blocker**

Since we're augmenting this with things like `setOperator` and `setCustomValue` I think it might be worthwhile to add `setApplyState` so that we can manage that for custom filters too. It was something I ran into several times when implementing custom filters and really missed having.

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -216,6 +230,39 @@ const operators = computed((): FilterOperator[] => {
...
+}
+
+/**
+ * The props passed into the `#default` slot so custom filter content can read
+ * the filter's options/operators and update the pending value/operator, which
+ * then flow through the existing Apply/Cancel footer like any other filter type.
+ */
+const filterPillSlotProps = computed((): FilterPillSlotProps => ({
+  value: userCustomValue.value,
+  text: userCustomText.value,
```
Chalks, line 270: **blocker?**

I think these should be

```suggestion
  value: userSelection.value.value,
  text: userSelection.value.text,
```

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -216,6 +230,39 @@ const operators = computed((): FilterOperator[] => {
...
+  options: filter.options,
+  multiple: filter.multiple,
+  operators: operators.value,
+  operator: userOperator.value,
```
Chalks, line 274: **blocker?**

Similarly, I believe this should be

```suggestion
  operator: userSelection.value.operator,
```

src/components/KFilterGroup/FilterPill.vue
```diff
@@ -411,14 +466,7 @@ const onPopClose = () => {
...
 const onApply = () => {
-  if (filterType.value === 'custom') {
-    // always emit `undefined` when the filter is custom because the host app
-    // must handle the selection when they've created a custom filter.
-    emit('apply', undefined)
-  } else {
-    emit('apply', userSelection.value)
-  }
-
+  emit('apply', userSelection.value)
```
Chalks, line 486: **question, not a blocker**

This can still emit `undefined`, right? I think that's okay, I'm just making sure because I think some custom filter implementations rely on it.

src/components/KFilterGroup/FilterPill.vue
kingyue737: Added `setApplyState(enabled: boolean)`

src/components/KFilterGroup/FilterPill.vue
kingyue737: I kept `operator` sourced from `userOperator.value` directly rather than `userSelection.value.operator` — `userSelection.value` is only defined once value+text are *also* set, so switching this would make `operator` report `undefined` until the user has also picked a value, even though `userOperator` is meaningfully defaulted (e.g. to `'eq'`).

src/components/KFilterGroup/FilterPill.vue
kingyue737: Yep, still emits `undefined` — unchanged from before. `userSelection` returns `undefined` whenever operator/value/text aren't all set, which is exactly what happens for a custom filter that never calls the new `setValue` (unmigrated slot content). Covered by the regression test in `FilterPill.cy.ts` ("still emits `@apply` with undefined if setValue is never called").

src/components/KFilterGroup/FilterPill.vue
kingyue737: Kept these as the raw `userCustomValue`/`userCustomText` refs — `userSelection` can be `undefined` (e.g. before `operator` has a default, or before all of operator/value/text are set), so reading through it here could show `undefined` even when a value has actually been set via `setValue`.

Chalks: ah, I was missing that custom value and text are [both set here](https://github.com/Kong/kongponents/pull/3288/changes#diff-0c62cf8846138cb78547c3283e3456dddd1e5c1063a6e66b51187b79cb77e26eR427-R428).

I was worried that you wouldn't be able to see the state set by something external if you were also using `setValue` but I see now that you can.

Chalks (approved):
Works well, thanks for tackling this. I think this will make custom filters way easier to use.

adamdehaven (approved):

PR #3290, opened 2026-07-16
fix: re-evaluate cached slot-presence checks on every render

kingyue737:
[KM-2920]

## Summary
- Several components cached slot-presence checks (`slots.title`, `slots.tooltip`, etc.) in a `computed()`. Since `computed` doesn't track raw property access on the `slots` object, the cached value never updates if a parent later starts conditionally passing that slot — a slot absent at initial mount could never appear afterward.
- Converts each affected check from a `computed` ref to a plain function invoked in the template, matching the fix already applied to `KCatalog`/`KTableView` in #3281.
- Affected: `KCard`, `KTooltip`, `KRadio`, `KCheckbox`, `KTable` (legacy), `KInput`, `KTextArea`, `KLabel`, `KCollapse`, `KFileUpload`, `KMultiselect`, `KSelect`, `KTreeItem`, `KEmptyState`, `KModal` (`showHeader` only — `maxWidthValue`/`maxHeightValue` use `v-bind()` CSS vars driven by a separate `watchEffect` and need a different fix, left as-is).
- Added a Cypress test per component reproducing "slot added after mount" and confirming it now renders.

## Test plan
- [x] Added regression tests for each fixed component (mount without the slot, dynamically enable it via `v-if`, assert it renders)
- [x] Verified each fix reproduces the bug when reverted (test fails) and passes once applied
- [x] Full spec run across all 15 touched components: 263/263 passing

[KM-2920]: https://konghq.atlassian.net/browse/KM-2920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/KSelect/KSelect.vue
+++ b/src/components/KSelect/KSelect.vue
@@ -23,7 +23,7 @@
     <KToggle v-slot="{ toggle, isToggled }">
       <KPop
         ref="popperElement"
-        v-bind="boundKPopAttributes"
+        v-bind="createKPopAttributes()"
         close-on-popover-click
         hide-close-icon
         :offset="`var(--kui-space-40, ${KUI_SPACE_40})`"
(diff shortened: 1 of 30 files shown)


Justineo (changes requested):
Thanks for the fix! Could you please address these two remaining cases?

src/components/KSelect/KSelect.vue
```diff
@@ -264,7 +264,7 @@ const isDropdownOpen = ref<boolean>(false)
 
 const resizeObserver = ref<ResizeObserverHelper>()
 
-const hasLabelTooltip = computed((): boolean => !!(labelAttributes?.info || slots['label-tooltip']))
+const hasLabelTooltip = (): boolean => !!(labelAttributes?.info || slots['label-tooltip'])
```
Justineo, line 267: Could we also update the `dropdown-footer-text` checks in `KSelect` and `KMultiselect`? They still cache slot presence when building popover classes.

src/components/KTable/KTable.vue
```diff
@@ -561,7 +561,7 @@ const offsets: Ref<any[]> = ref([])
 const hasNextPage = ref(true)
 const isClickable = ref(false)
 const hasInitialized = ref(false)
-const hasToolbarSlot = computed((): boolean => !!slots.toolbar || hasColumnVisibilityMenu.value)
+const hasToolbarSlot = (): boolean => !!slots.toolbar || hasColumnVisibilityMenu.value
```
Justineo: Since KTable is deprecated, could we remove the KTable-specific changes from this PR?

Justineo (approved):
LGTM

PR #3302, opened 2026-07-23
fix(ktableview): slotted link default styles

portikM:
# Summary

Apply default link styles to slotted links in KTableVIew table cells

### DIFF
--- a/sandbox/pages/SandboxTableView/SandboxTableView.vue
+++ b/sandbox/pages/SandboxTableView/SandboxTableView.vue
@@ -106,7 +106,7 @@
           <template #username="{ rowValue }">
             <KExternalLink
               class="username-link"
-              href="https://kongponents.konghq.com/components/alert.html"
+              href="https://kongponents.konghq.com/components/table-view.html"
             >
               @{{ rowValue }}
             </KExternalLink>
@@ -129,7 +129,7 @@
           <template #username="{ rowValue }">
             <KExternalLink
               class="username-link"
-              href="https://kongponents.konghq.com/components/alert.html"
+              href="https://kongponents.konghq.com/components/table-view.html"
             >
               @{{ rowValue }}
             </KExternalLink>
@@ -183,6 +183,14 @@
           </template>
           <template #tooltip-email>
             Id: <code>8576925e-d7e0-4ecd-8f14-15db1765e69a</code>
+          </template>
+          <template #username="{ rowValue }">
+            <a
+              href="https://kongponents.konghq.com/components/table-view.html"
+              target="_blank"
+            >
+              @{{ rowValue }}
+            </a>
           </template>
           <template #email="{ rowValue }">
             <KCopy :text="rowValue" />
@@ -683,7 +691,7 @@ onBeforeMount(() => {
   .horizontal-container {
     display: flex;
     flex-wrap: wrap;
-    gap: $kui-space-50;
+    gap: var(--kui-space-50, $kui-space-50);
   }
 
   .username-link {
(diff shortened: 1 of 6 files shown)


src/styles/mixins/_links.scss
```diff
@@ -0,0 +1,25 @@
+@use "@kong/design-tokens/tokens/scss/variables" as *;
+
+@mixin link {
+  color: var(--kui-link-color-text, var(--kui-color-text-primary, $kui-color-text-primary));
+  font-size: inherit;
+  font-weight: var(--kui-link-font-weight, var(--kui-font-weight-regular, $kui-font-weight-regular));
```
adamdehaven: I'm unsure if we should include font-weight in the mixin. If an anchor tag appears in a title, etc. this would override the font weight.

Can we move this back into KExternalLink? Other places would inherit the font-weight

src/styles/mixins/_links.scss
```diff
@@ -0,0 +1,25 @@
...
+  list-style: inherit;
```
adamdehaven: I'm not sure `list-style` belongs in the `link` mixin?

src/styles/mixins/_tables.scss
```diff
@@ -230,6 +231,10 @@
               line-height: var(--kui-table-cell-line-height, var(--kui-line-height-30, $kui-line-height-30));
               white-space: nowrap;
 
+              :deep(a) {
+                @include link;
+              }
```
adamdehaven, line 236: question: will this properly style links that are and are not slotted?

src/styles/mixins/_links.scss
portikM: I was debating this too. Done

src/styles/mixins/_links.scss
portikM: Moved back to KExternalLink

src/styles/mixins/_tables.scss
portikM: Cell can only have slotted links - by default the table just renders the raw cell value. So this only applies to slotted links and those are properly styled (example in sandbox)

adamdehaven: image
 
 When I put a link in a cell it's not styled

adamdehaven: are they supposed to have text-decoration: underline, or no?

portikM: That's styled link in classic mode. Switch to electric lime day and it will be grey and have underline.

> are they supposed to have text-decoration: underline, or no?

In classic mode - no. In day / night modes - yes.

adamdehaven: realized that after I commented

sandbox/pages/SandboxTableView/SandboxTableView.vue
```diff
@@ -184,6 +184,14 @@
...
             Id: <code>8576925e-d7e0-4ecd-8f14-15db1765e69a</code>
           </template>
+          <template #username="{ rowValue }">
+            <a
+              href="https://kongponents.konghq.com/components/table-view.html"
+              target="_blank"
+            >
+              @{{ rowValue }}
+            </a>
+          </template>
```
adamdehaven, line 199: I don't see links rendered in the table here?

portikM: https://deploy-preview-3302--kongponents-sandbox.netlify.app/#/table-view

Screenshot 2026-07-23 at 11 43 14 AM

adamdehaven: I kind of think it would be nice to have 1 table example with both internal and external links. You want to do that in this PR?

portikM: Done

adamdehaven (approved):
lgtm

jillztom (approved):

PR #3304, opened 2026-07-24
fix(kselect, kmultiselect): new dropdown footer slot [KHCP-21063]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-21063

Changes:

- Added a new `dropdown-footer` slot to KSelect and KMultiselect that supports interactive content (buttons, links), replacing the `pointer-events: none;` restriction on the footer
- In both components deprecated the `dropdown-footer-text` slot in favour of `dropdown-footer` (the new slot takes precedence over both the deprecated slot and the `dropdownFooterText` prop)
- In both components renamed the `dropdownFooterTextPosition` prop to `dropdownFooterPosition`, keeping the old prop working as a deprecated alias (new prop wins when both are set)
- Updated types, docs, sandbox examples, and component tests to cover the new slot / prop, precedence, and backwards compatibility

### DIFF
--- a/src/components/KMultiselect/KMultiselect.cy.ts
+++ b/src/components/KMultiselect/KMultiselect.cy.ts
@@ -605,6 +605,114 @@ describe('KMultiselect', () => {
     cy.get('.dropdown-footer').should('be.visible').should('contain.text', dropdownFooterText)
   })
 
+  it('renders interactive content in the dropdown-footer slot', () => {
+    const labels = ['Label 1', 'Label 2', 'Label 3']
+    const vals = ['label1', 'label2', 'label3']
+    const buttonText = 'Footer action'
+
+    cy.mount(KMultiselect, {
+      props: {
+        items: [{
+          label: labels[0],
+          value: vals[0],
+        }, {
+          label: labels[1],
+          value: vals[1],
+        }, {
+          label: labels[2],
+          value: vals[2],
+        }],
+      },
+      slots: {
+        'dropdown-footer': `<button data-testid="footer-button">${buttonText}</button>`,
+      },
+    })
+
+    cy.get('.multiselect-trigger').trigger('click')
+
+    cy.get('.dropdown-footer')
+      .should('be.visible')
+      .should('have.css', 'pointer-events', 'auto')
+    // the interactive content should be clickable
+    cy.getTestId('footer-button').should('be.visible').click()
+  })
+
+  it('dropdown-footer slot takes precedence over dropdownFooterText prop and dropdown-footer-text slot', () => {
+    const labels = ['Label 1', 'Label 2', 'Label 3']
+    const vals = ['label1', 'label2', 'label3']
+
+    cy.mount(KMultiselect, {
+      props: {
+        items: [{
+          label: labels[0],
+          value: vals[0],
+        }, {
+          label: labels[1],
+          value: vals[1],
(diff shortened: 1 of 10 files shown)


src/components/KSelect/KSelect.cy.ts
```diff
@@ -383,6 +383,114 @@ describe('KSelect', () => {
...
+      },
+    })
+
+    cy.getTestId('select-input').trigger('click')
+
+    cy.get('.dropdown-footer')
+      .should('be.visible')
+      .should('contain.text', 'New footer slot')
+      .should('not.contain.text', 'Deprecated footer slot')
+      .should('not.contain.text', 'Footer text prop')
```
kaiarrowood: nit: store strings you assert on in vars

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -222,12 +222,19 @@
               </div>
             </div>
             <div
-              v-if="dropdownFooterText || $slots['dropdown-footer-text']"
+              v-if="hasDropdownFooter"
```
kaiarrowood: q: do you need to also check for `resolvedDropdownFooterPosition` like you did in `KSelect`?

src/components/KMultiselect/KMultiselect.cy.ts
```diff
@@ -605,6 +605,114 @@ describe('KMultiselect', () => {
...
+          label: labels[0],
+          value: vals[0],
+        }, {
+          label: labels[1],
+          value: vals[1],
+        }, {
+          label: labels[2],
+          value: vals[2],
+        }],
+        dropdownFooterText: 'Footer text prop',
```
kaiarrowood: nit: store strings you assert on in vars

kaiarrowood (commented):
Functional and code looks good to me with a couple nits 🚀

src/components/KMultiselect/KMultiselect.cy.ts
portikM: Done

src/components/KMultiselect/KMultiselect.vue
portikM: No, it's correct as-is. The two components are structured differently

kaiarrowood (dismissed):
LGTM! 🚀

src/components/KMultiselect/KMultiselect.vue
```diff
@@ -543,13 +552,16 @@ const modifiedAttrs = computed(() => {
   return $attrs
 })
 
-const hasDropdownFooterTextSlot = (): boolean => !!(dropdownFooterText || slots['dropdown-footer-text'])
+const hasDropdownFooter = computed((): boolean => !!(dropdownFooterText || slots['dropdown-footer-text'] || slots['dropdown-footer']))
```
Justineo: Please take a look at https://github.com/Kong/kongponents/pull/3290
We shouldn’t cache the slot existence check in computed values.

Same applies to `KSelect`

Justineo (dismissed):

src/components/KMultiselect/KMultiselect.vue
portikM: Addressed

kaiarrowood (approved):
LGTM! 🚀

PR #3314, opened 2026-07-30
fix(kmultiselect): reflect modelValue order in selectedItem badges

filipgutica:
closes: https://konghq.atlassian.net/browse/MA-5270

# Summary

Fixes `KMultiselect` so selected badges follow the value order in programmatic `modelValue` updates.

# Changes

- Rebuild internal selected items and badge overflow state in `modelValue` order.
- Wait for updated items before reconciling simultaneous `items` and `modelValue` changes.
- Skip reconciliation when a user selection already matches `modelValue`.
- Add focused tests for badge ordering and stable dropdown option order.
- Add interactive sandbox and documentation examples.

### DIFF
--- a/src/components/KMultiselect/KMultiselect.vue
+++ b/src/components/KMultiselect/KMultiselect.vue
@@ -1101,8 +1101,11 @@ watch(value, (newVal, oldVal) => {
 
     const previouslySelectedItems = new Set<string>(oldVal)
     const currentlySelectedItems = new Set<string>(newVal)
+    const internallySelectedItems = new Set(currentSelectedValues)
 
-    const selectedAndPresentItems = unfilteredItems.value.filter((item: Item) => currentlySelectedItems.has(item.value))
+    const selectedAndPresentItems = unfilteredItems.value.filter((item: Item) =>
+      currentlySelectedItems.has(item.value) && !internallySelectedItems.has(item.value),
+    )
     const deselectedItems = selectedItems.value.filter((item: Item) => !currentlySelectedItems.has(item.value) && previouslySelectedItems.has(item.value))
 
     if (deselectedItems.length) {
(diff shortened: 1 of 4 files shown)


src/components/KMultiselect/KMultiselect.vue
```diff
@@ -1076,6 +1088,12 @@ watch(filteredNormalizedItems, () => {
 // watch for programmatic changes to model
 watch(value, (newVal, oldVal) => {
   if (JSON.stringify(newVal) !== JSON.stringify(oldVal)) {
+    const currentSelectedValues = selectedItems.value.map(item => item.value)
+
+    if (JSON.stringify(currentSelectedValues) === JSON.stringify(newVal)) {
+      return
+    }
```
portikM, line 1095: question: why do we need a return here? Wouldn't it mean we're skipping the `handleMultipleItemsDeselect` and `handleMultipleItemsSelect` logic if the order of selected items matches `modelValue`?

filipgutica: There are two different state comparisons:

```ts
JSON.stringify(newVal) !== JSON.stringify(oldVal)
```

This says the parent model changed.

```ts
JSON.stringify(currentSelectedValues) === JSON.stringify(newVal)
```

This says `KMultiselect` already applied that exact membership and order internally.

That second case occurs during the normal `v-model` round trip:

1. The user selects an item.
2. `KMultiselect` updates `selectedItems`.
3. It emits `update:modelValue`.
4. The parent passes the new value back.
5. The watcher runs because `oldVal` and `newVal` differ.
6. The internal selection already matches `newVal`, so no reconciliation is needed.

Therefore, the early return does skip `handleMultipleItemsDeselect` and `handleMultipleItemsSelect`, but only when there is nothing to deselect or select. If membership or order differs, the equality check fails and reconciliation continues.

portikM (approved):
LGTM 🚀

PR #3317, opened 2026-07-31
feat: expose label slot on KInput, KTextArea, KSelect, KMultiselect, KFileUpload, KSlider

2eha0:
## Summary

- Add a `label` slot to 6 input-style components that internally use `KLabel` but previously had no way to customize label content
- The slot follows the same pattern as `KInputSwitch` (which already had a `label` slot): takes precedence over the `label` prop when both are provided, and triggers label rendering even without the prop
- No breaking changes — purely additive

## Components updated

| Component | Change |
|---|---|
| `KInput` | Added `label` slot |
| `KTextArea` | Added `label` slot |
| `KSelect` | Added `label` slot |
| `KMultiselect` | Added `label` slot |
| `KFileUpload` | Added `label` slot |
| `KSlider` | Added `label` slot + new `SliderSlots` type + `defineSlots` |

## Implementation pattern

```vue
<KLabel v-if="label || $slots.label" ...>
  <slot name="label">
    {{ strippedLabel }}
  </slot>
  ...
</KLabel>
```

### DIFF
--- a/src/components/KFileUpload/KFileUpload.vue
+++ b/src/components/KFileUpload/KFileUpload.vue
@@ -8,13 +8,15 @@
     @drop.prevent="onDrop"
   >
     <KLabel
-      v-if="label"
+      v-if="label || $slots.label"
       v-bind="labelAttributes"
       :id="isAppearanceDropzone ? `file-upload-dropzone-label-${fileInputId}` : undefined"
       :for="fileInputId"
       :required="isRequired"
     >
-      {{ strippedLabel }}
+      <slot name="label">
+        {{ strippedLabel }}
+      </slot>
 
       <template
         v-if="hasLabelTooltip()"
--- a/src/components/KSlider/KSlider.vue
+++ b/src/components/KSlider/KSlider.vue
@@ -4,11 +4,13 @@
     data-testid="k-slider"
   >
     <KLabel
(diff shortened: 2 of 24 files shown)


adamdehaven (changes requested):
Holding for discussion: https://kongstrong.slack.com/archives/CTAMPSTNX/p1785489831311119

src/components/KFileUpload/KFileUpload.vue
```diff
@@ -8,13 +8,15 @@
     @drop.prevent="onDrop"
   >
     <KLabel
-      v-if="label"
+      v-if="label || $slots.label"
```
Justineo, line 11:
```suggestion
      v-if="label || slots.label"
```

Let’s prefer `slots` over this implicit binding. The same applies to other occurrences as well.

src/components/KSlider/KSlider.vue
```diff
@@ -117,6 +119,7 @@ const {
 } = defineProps<SliderProps>()
 
 const emit = defineEmits<SliderEmits>()
+defineSlots<SliderSlots>()
```
Justineo, line 122:
```suggestion
const slots = defineSlots<SliderSlots>()
```

PR #3323, opened 2026-08-04
fix(kradio): reset card label margin

Justineo:
# Summary

Reset the KRadio card label's `margin-bottom` to `0` so host normalize styles cannot add unintended spacing.

image

image

### DIFF
--- a/src/components/KRadio/KRadio.vue
+++ b/src/components/KRadio/KRadio.vue
@@ -347,6 +347,7 @@ $kRadioDotSize: 6px;
       box-shadow: var(--kui-shadow-border, $kui-shadow-border);
       cursor: pointer;
       height: 100%;
+      margin-bottom: 0;
       outline: none;
       padding: var(--kui-space-70, $kui-space-70);
       transition: box-shadow $kongponentsTransitionDurTimingFunc, background-color $kongponentsTransitionDurTimingFunc;


adamdehaven (dismissed):

src/components/KRadio/KRadio.vue
```diff
@@ -347,6 +347,7 @@ $kRadioDotSize: 6px;
       box-shadow: var(--kui-shadow-border, $kui-shadow-border);
       cursor: pointer;
       height: 100%;
+      margin-bottom: 0;
```
adamdehaven, line 350: Going to sound dumb but can you use the space token please

Justineo: I’m not sure we really want to use a token for zero values, since there don’t seem to be any meaningful overrides for them. I’m actually thinking in the opposite direction: should we discourage using a token for `0px`?

adamdehaven: Technically the contract we have is to tokenize everything. If there's no token, there isn't a way to customize if desired, e.g.

```scss
.radio-card {
  --kui-space-0: 12px;
}
```

I'm not saying this really would make sense to do, but there's nothing against it, so I think we should always use tokens.

Justineo: I mean, using `0` here and making it uncustomizable should be intentional by design. Even if we want to expose a way to redefine it eventually, overriding `--kui-space-0` wouldn’t be the right approach because we can’t afford the blast radius. It’s the same reason we introduced component tokens in the first place.

adamdehaven: > I mean, using `0` here and making it uncustomizable should be intentional by design. Even if we want to expose a way to redefine it eventually, overriding `--kui-space-0` wouldn’t be the right approach because we can’t afford the blast radius. It’s the same reason we introduced component tokens in the first place.

The opposite actually. Konnect isn't the only consumer of Kongponents. Everything that can be tokenized should be.

Justineo: It has nothing to do with Konnect. The component itself should be designed to ensure that it doesn’t have a bottom margin. Using a token here means it's expected to be customizable. Even though in my opinion, `--kui-space-0` should never be customized even if it's already a token.

Justineo: F.Y.I.

---

# Research: How Design Systems Handle Zero Spacing

This report was researched with Codex using current official documentation and component source code.

## Executive summary

Mainstream design systems do not consistently require every `0` spacing value to use a design token. The dominant pattern is intent-based:

- Use spacing tokens for variable design decisions.
- Use literal `0` for browser resets and layout invariants.
- Use a component-specific token when a property defaults to zero but should remain customizable.

A generic zero token may exist for API completeness, but this does not imply that redefining zero as a non-zero value is a valid theming use case.

## Findings by design system

| Design system | Treatment of zero |
| --- | --- |
| **Atlassian Design System** | Defines `space.0 = 0px` and actively uses it in layout APIs and component styles. This is the clearest example of systematic zero tokenization. [Spacing documentation](https://atlassian.design/foundations/grid-beta/applying-grid) |
| **Fluent UI v9** | Defines `spacingHorizontalNone` and `spacingVerticalNone` as `0`, but still uses literal `margin: 0` for Button resets. Zero tokens exist without being mandatory for every reset. [Token definitions](https://github.com/microsoft/fluentui/blob/c83bfa7666778640350f54991c8ebe712a8f389b/packages/tokens/src/global/spacings.ts#L4-L34), [Button source](https://github.com/microsoft/fluentui/blob/c83bfa7666778640350f54991c8ebe712a8f389b/packages/react-components/react-button/library/src/components/Button/useButtonStyles.styles.ts#L29-L38) |
| **GitHub Primer** | Maps `0px` to literal `0`. Its former `$spacer-0` token is deprecated rather than replaced with a zero CSS variable. [Migration guide](https://primer.style/product/primitives/migrating/#spacing) |
| **IBM Carbon** | Its spacing scale begins at 2px. Official examples combine spacing tokens with literal zero values, such as `margin: $spacing-07 0 $spacing-04 0`. [Spacing documentation](https://carbondesignsystem.com/elements/spacing/overview/) |
| **Ant Design** | Its smallest general spacing token is 4px. Component resets use literal values such as `margin: 0` and `padding: 0`. [Token scale](https://github.com/ant-design/ant-design/blob/5ea586b559f47a72bb7d812b1d4c8174610b64a6/components/theme/themes/shared/genSizeMapToken.ts#L6-L16), [Checkbox source](https://github.com/ant-design/ant-design/blob/5ea586b559f47a72bb7d812b1d4c8174610b64a6/components/checkbox/style/index.ts#L119-L130) |
| **Material Web / Material 3** | Does not provide a formal global spacing-token scale. Components use literal `padding: 0`, `margin: 0`, and `inset: 0` for resets and invariants. [Button source](https://github.com/material-components/material-web/blob/23b638c1d2338b9fcaccef81cf54298cce9c27ed/button/internal/_shared.scss#L94-L107) |
| **MUI** | Supports zero through `theme.spacing(0)`, but does not expose a named zero token. Core component resets still use literal `0`. [Spacing API](https://mui.com/material-ui/customization/spacing/), [ButtonBase source](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/ButtonBase/ButtonBase.js#L40-L52) |
| **Adobe Spectrum CSS** | Uses literal zero for resets, but introduces component-specific zero-valued properties when customization is intentional, for example, `--spectrum-tooltip-margin: 0px`. [Button reset](https://github.com/adobe/spectrum-css/blob/37620864c60c4c142a506017e1a15348a26abb0e/components/commons/basebutton.css#L27-L30), [Tooltip property](https://github.com/adobe/spectrum-css/blob/37620864c60c4c142a506017e1a15348a26abb0e/components/tooltip/index.css#L17-L22) |

## Conclusion

There is little support for the rule that every zero value must use a global spacing token.

A token such as `space.0` can be useful as a typed layout option, but its semantic meaning remains “no space.” Overriding it to a non-zero value would change every use within its scope and create unintended coupling.

For a CSS reset or invariant, the appropriate implementation is:

```css
margin-bottom: 0;
```

If the property should be customizable, use a semantic component token instead:

```css
margin-bottom: var(--radio-card-margin-bottom, 0);
```

The decision should follow the property’s intent, not whether its numeric value happens to exist in the spacing scale.

adamdehaven: You know the conventions that we use in the repository; please adhere to them until those conventions change

adamdehaven: I don't disagree that the overall conventions don't match what I'd normally expect.

Justineo: I agree that following established conventions is important. However, this repository already uses literal `0`/`0px` extensively, including `margin: 0` in this component, so I don’t see a consistent convention requiring every zero spacing value to use a token. Mechanically changing this line to `--kui-space-0` would not improve consistency. If that change is required to merge the PR, I can make it, but I don’t consider it a well-founded requirement.

Justineo: We do have a convention to use design tokens. But IMO it should be “use design tokens wherever possible to accurately express design intent”, rather than mechanically replacing literal values with tokens even when the literal value is intentionally a constant.

adamdehaven: I agree that the repo isn’t perfectly consistent today, but that doesn’t mean we should use existing literal values as the standard going forward.

The goal here is to encourage consistency with the established design system primitives rather than reinforce inconsistencies that already exist in the codebase.

If this change was driven by the Design team, I'm sure they'd have utilized the existing token here, not a raw value, as that is the convention we agreed to build upon.

Justineo: I think my core point is still being missed. I’m not arguing that existing inconsistencies should define the standard going forward. I’m questioning the premise that our convention is to mechanically replace every tokenizable literal value with a token.

The convention should be to use design tokens where they accurately express design intent. Here, `0` is an invariant reset, not a themeable spacing decision. Using `--kui-space-0` does not improve consistency in any meaningful sense. It merely makes the syntax uniform while misrepresenting the intent.

I've made the change to unblock the PR, but I don’t agree that this follows a sound convention.

---

It’s not driven by the design team. It’s simply fixing a style issue introduced by Konnect’s global styles.

adamdehaven: I tried reaching out to you on Slack to connect on this but it seems like you'd rather keep this in the public forum.

adamdehaven (approved):

sumimakito (approved):

hangrao (approved):

PR #3349, opened 2026-08-11
fix(kcopy): truncate when the root cannot shrink [parked]

merrywhether:
# Summary

`KCopy` with `truncate` + `truncation-limit="auto"` renders the full text and truncates only when it measures `offsetWidth < scrollWidth` on the inner `.copy-text`. Where the component can't shrink to the width it was given, that measurement never trips — so the text renders full width, overflows its container (typically clipped mid-character by an ancestor's `overflow: hidden`), and the copy button is pushed out of view.

Measured against a long URL in a 200px column. `overflow` is how far the copy button lands past the column's right edge:

| Shape | truncates? | overflow | |
|---|---|---|---|
| `.k-copy` is a grid item | no | **+315px** | ✗ |
| Consumer sets `display: block` on `.k-copy` | no | **+351px** | ✗ |
| Plain block parent | yes | none | ✓ |
| Flex parent | yes | none | ✓ |

So it isn't universal — it's specific to shapes where the root box is held at its content width:

- **`.k-copy` as a grid item.** `.k-copy` is `overflow: visible`, so its `min-width: auto` resolves to a content-based minimum and the box won't shrink into its track. The flex case already works, because `.copy-element` is `overflow: hidden` and its automatic minimum resolves small.
- **Consumer makes the root a block.** The badge box (`.copy-container.badge-styles`) is `inline-flex; width: fit-content`, which sizes to its content and won't shrink to fit a block parent.

## What changed

All in `KCopy.vue` styles — no template or script changes, no API change:

- `min-width: 0` on `.k-copy`, so it can shrink as a grid or flex item
- `max-width: 100%` and `box-sizing: border-box` on `.copy-element`, so the fit-content badge box is capped at the width it was given, padding included
- `min-width: 0` on `.copy-tooltip-wrapper` and `.copy-text`, since flex children default to `min-width: auto` and the measured element has to be able to shrink

Each is load-bearing — reverting any one of them fails the new tests.

That last one also closes a latent feedback loop: `KTooltip` renders no wrapper element until it has text, and it only gets text once `isTruncated` flips. Without `min-width: 0` on the wrapper, the wrapper appearing can stop `.copy-text` shrinking, which unsets `isTruncated`, which removes the wrapper again.

## Tests

Three added to `KCopy.cy.ts`, covering both broken shapes plus a negative control that the component still leaves text alone when it fits. The two positive tests were each confirmed to fail without the fix (the second in isolation, since `cypress-fail-fast` masks it behind the first).

An earlier draft of these tests mounted into a plain narrow block parent — everything passed with the fix fully reverted, because that shape was never broken. Worth knowing if you go to extend them: the parent's display type is the whole ballgame.

## Verification

- `pnpm test` — 747 passing. One `KFileUpload` spec fails in full-suite runs, but it fails identically on a clean `main` and passes in isolation; it's pre-existing and unrelated.
- `pnpm stylelint`, `pnpm typecheck`, `pnpm test:unit` (87) — clean. `pnpm lint` has 2 warnings, both pre-existing in `KComponent`/`KTable`.
- Docs page checked in a browser: all 15 `KCopy` instances render with zero overflow past their parent and a visible copy button; the two `auto` demos truncate.

### DIFF
--- a/src/components/KCopy/KCopy.cy.ts
+++ b/src/components/KCopy/KCopy.cy.ts
@@ -122,6 +122,98 @@ describe('KCopy', () => {
     cy.get(container).find('.text-icon').should('be.visible')
   })
 
+  describe('truncation', () => {
+    // `truncationLimit: 'auto'` renders the full text and truncates only when it measures an
+    // overflow on `.copy-text`, so the component has to be given less room than the text needs.
+    const longText = 'https://example.konghq.com/a-very-long-path-that-does-not-fit/1234567890ABCDEFG'
+    const column = '[data-testid="column"]'
+
+    const expectTruncated = () => {
+      cy.get(container).find('.copy-text').should(($text) => {
+        expect($text[0].scrollWidth).to.be.greaterThan($text[0].offsetWidth)
+      })
+    }
+
+    const expectWithinColumn = () => {
+      cy.get(column).then(($column) => {
+        const columnRight = $column[0].getBoundingClientRect().right
+
+        cy.get(container).should(($root) => {
+          expect($root[0].getBoundingClientRect().right).to.be.at.most(columnRight)
+        })
+
+        cy.getTestId('copy-to-clipboard').should(($button) => {
+          expect($button[0].getBoundingClientRect().right).to.be.at.most(columnRight)
+        })
+      })
+    }
+
+    it('truncates when it is a grid item narrower than its text', () => {
+      cy.mount({
+        components: { KCopy },
+        data: () => ({ longText }),
+        template: `
+          <div
+            data-testid="column"
+            style="display: grid; grid-template-columns: 1fr; width: 200px"
+          >
+            <KCopy
+              badge
+              truncate
+              truncation-limit="auto"
+              :text="longText"
+            />
+          </div>
+        `,
+      })
+
+      expectTruncated()
+      expectWithinColumn()
+    })
+
+    // Consumers laying the badge out in a column commonly make the root a block so it fills the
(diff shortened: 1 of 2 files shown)


merrywhether:
Parking this as a draft — no consumer is currently blocked on it.

The Konnect case that prompted it turned out to be self-inflicted: the consumer had a `display: block` rule on the root that flipped the badge out of the flex context KCopy sets up for itself. Deleting that fixes it there without any change here.

What remains genuinely a KCopy gap is the **grid item** shape — `.k-copy` is `overflow: visible`, so `min-width: auto` holds it at content width and it won't shrink into a track. The fix and tests for that are verified and ready here whenever it's worth landing.

The `max-width: 100%` / `box-sizing: border-box` half covers the "consumer makes the root a block" shape, which now has no real-world consumer behind it — worth re-litigating before this lands.

adamdehaven:
Would it be worth adding an example of what this resolves in the Sandbox? (We use the sandbox to house edge-cases, etc. that don't necessarily need to exist on the docs site)

merrywhether:
Closing — no consumer for this.

The Konnect case that prompted it turned out to be self-inflicted: a `display: block` rule on the root that flipped the badge out of the flex context KCopy sets up for itself. Removing that fixed it with no change here.

That leaves only the grid-item shape, which nothing currently hits and which we don't expect to. Not worth carrying an open PR against KCopy for a hypothetical.

The branch stays if it's ever wanted: the one-line part is `min-width: 0` on `.k-copy` (its `overflow: visible` makes `min-width: auto` hold it at content width, so it won't shrink into a track), plus component tests that were verified to fail without it.

PR #3350, opened 2026-08-12
ci: setup vitest browser component tests [KHCP-21503]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-21503

Setting up vitest browser component testing


adamdehaven:
/muthur --model sonnet review the vitest integration and browser testing setup

.github/workflows/test.yaml
```diff
@@ -35,6 +35,11 @@ jobs:
       - name: Lint
         run: pnpm lint
 
+      # Test files are excluded from `tsconfig.build.json` so their declarations don't
+      # ship in `dist/types`, so they need their own typecheck pass.
+      - name: Typecheck tests
+        run: pnpm typecheck:tests
```
adamdehaven: A single `pnpm typecheck` command should check everything; we shouldn't need something additional here so I'm guessing it's the tsconfig

test/setup.ts
```diff
@@ -0,0 +1,33 @@
...
+// them is the point rather than a shortcut. The Cypress suite relies on CSS selectors
+// roughly 1,100 times for that reason.
+//
+// `locators.extend` is the sanctioned escape hatch. Returning a selector string (rather
+// than wrapping an element via `page.elementLocator`) yields a real lazy locator, so it
+// retries like any built-in selector and chains off both `page` and another `Locator` —
+// which is what makes `cy.get(...).find(...)` translate cleanly.
+//
+// Prefer `getByRole`/`getByTestId` where a semantic selector genuinely fits. Reach for
+// `getByCSS` when the class itself is the contract under test.
```
adamdehaven: let's make these comments more concise - applies to PR

test/tsconfig.json
```diff
@@ -0,0 +1,40 @@
...
+      ]
+    }
+  },
+  // Only the tests themselves. Components are typechecked by `pnpm typecheck` against
+  // `tsconfig.build.json`; vue-tsc still follows imports from here, so a test that
+  // misuses a component's props is caught.
+  "include": [
+    "../test/**/*.ts",
+    "../test/**/*.d.ts",
+    "../src/**/*.spec.ts"
```
adamdehaven: the tests themselves should be included in the normal typecheck

test/utils.ts
```diff
@@ -0,0 +1,100 @@
...
+ * Locate an element by its `data-testid` attribute. Counterpart of `cy.getTestId`.
+ */
+export const getTestId = (dataTestId: string): Locator => page.getByTestId(dataTestId)
+
+/**
+ * Locate an element by `data-testid` within another locator. Counterpart of
+ * `cy.findTestId`, which was chained off an existing subject.
+ */
+export const findTestId = (subject: Locator, dataTestId: string): Locator =>
+  subject.getByTestId(dataTestId)
```
adamdehaven: question: why do we need these instead of just calling `page.getByTestId` etc.?

package.json
```diff
@@ -51,15 +51,18 @@
     "docs:serve": "vitepress serve docs",
     "docs:preview": "vitepress preview docs --port 8080",
     "typecheck": "vue-tsc -p './tsconfig.build.json' --noEmit",
+    "typecheck:tests": "vue-tsc -p './test/tsconfig.json' --noEmit",
```
adamdehaven: If you're absolutely sure we need separate tsconfig and commands, then include this in the base typecheck command so that CI and other callers only need to call 1 
```suggestion
"typecheck": "vue-tsc -p './tsconfig.build.json' --noEmit && pnpm typecheck:tests",
    "typecheck:tests": "vue-tsc -p './test/tsconfig.json' --noEmit",
```

package.json
```diff
@@ -51,15 +51,18 @@
...
     "stylelint:fix": "stylelint './src/**/*.{css,scss,sass,vue}' --fix",
     "test": "cross-env DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome",
     "test:open": "cross-env DISABLE_VUE_DEVTOOLS=true cypress open --component -b chrome",
     "test:spec": "cross-env DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome --spec",
-    "test:unit": "vitest run",
-    "test:unit:watch": "vitest",
+    "test:unit": "vitest run --project unit",
+    "test:unit:watch": "vitest --project unit",
+    "test:component": "vitest run --project component",
+    "test:component:watch": "vitest --project component --browser.headless=false",
```
adamdehaven: question: what's the difference here? `test:unit` doesn't run in browser mode?

Long-term we'll want to remap `test:unit` to just `test` -- IMO `test:unit` and `test:component` are a bit confusing here in the short-term.

Could we do `test:browser` for now?

tsconfig.build.json
```diff
@@ -1,7 +1,12 @@
 {
   "extends": "./tsconfig.json",
+  // Tests are excluded from the published build. Without the `*.spec.ts` and `test/**`
+  // entries below, `vue-tsc --emitDeclarationOnly` emits declarations for test files
+  // into `dist/types/` and ships them to consumers.
```
adamdehaven: These comments aren't valid in the tsconfig file

vitest.config.ts
```diff
@@ -1,22 +1,78 @@
...
+          // `vitest-browser-vue` unmounts rendered components between tests. Without
+          // it, mounted trees accumulate in the page and locators match stale nodes.
+          setupFiles: ['vitest-browser-vue', './test/setup.ts'],
+          // Matches the Cypress `retries.runMode: 1` we're replacing.
+          retry: process.env.CI ? 1 : 0,
+          browser: {
+            enabled: true,
+            headless: true,
+            provider: playwright(),
+            instances: [{ browser: 'chromium' }],
```
adamdehaven: should we test in chrome and firefox and safari? maybe depends on speed?

.github/workflows/test.yaml
portikM: Oops missed this. Removed

test/setup.ts
portikM: Done

test/tsconfig.json
portikM: Done

test/utils.ts
portikM: The comment explains it, it's a counterpart of `cy.findTestId`.

package.json
portikM: No I consolidated typecheck for source code files and tests into one typecheck command now

package.json
portikM: Yeah I thought of that too. I agree we should rename them at some point but I suggest we do it after we remove Cypress completely. For now renamed `test:component -> test:browser`

test/utils.ts
adamdehaven: No, I'm saying there are native ways to do this in Playwright - this shouldn't be needed

adamdehaven: https://vitest.dev/api/browser/locators.html#getbytestid

portikM:
/muthur --model sonnet review the vitest integration and browser testing setup

test/utils.ts
portikM: Locate an element by `data-testid` **within another locator**. The usage would look something like this:
```
// Cypress
cy.get('.empty-state-icon').findTestId(testId).should('be.visible')

// vitest
const slottedIcon = findTestId(page.getByCSS('.empty-state-icon'), 'slotted-icon')

await expect.element(slottedIcon).toBeVisible()
```

vitest.config.ts
portikM: Let's try and see. Even if it takes longer, I see value in that for library like Kongponents. Added `firefox` and `webkit`

adamdehaven:
/muthur review the vitest browser testing setup

package.json
```diff
@@ -50,16 +50,18 @@
...
     "stylelint": "stylelint './src/**/*.{css,scss,sass,vue}'",
     "stylelint:fix": "stylelint './src/**/*.{css,scss,sass,vue}' --fix",
     "test": "cross-env DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome",
     "test:open": "cross-env DISABLE_VUE_DEVTOOLS=true cypress open --component -b chrome",
     "test:spec": "cross-env DISABLE_VUE_DEVTOOLS=true cypress run --component -b chrome --spec",
-    "test:unit": "vitest run",
-    "test:unit:watch": "vitest",
+    "test:unit": "vitest run --project unit",
+    "test:unit:watch": "vitest --project unit",
+    "test:browser": "vitest run --project browser",
```
adamdehaven, line 64: issue: when I run `test:browser` I get an error that playwright browsers aren't installed (this is expected on first run).

We should expose a `test:browser:install` command that has this:
```json
"test:browser:install": "playwright install --with-deps chromium firefox webkit"
```

Then, also update the `action.yaml` for installing dependencies to call `pnpm test:browser:install` instead

tsconfig.build.json
```diff
@@ -1,7 +1,18 @@
...
   "extends": "./tsconfig.json",
+  /**
+   * Declaration emit only — `pnpm typecheck` uses the root config.
+   *
+   * `--emitDeclarationOnly` emits a `.d.ts` for every file in the program, so the test
+   * entries below matter even though no bundled code imports a test. `test` especially:
+   * it lives outside `src`, so including it moves the inferred root to the repo root and
+   * declarations land in `dist/types/src/` instead of `dist/types/`, breaking the `types`
+   * entry point.
+   */
```
adamdehaven: issue: Remove all of these block comments from the JSON files; they aren't valid. Check all the tsconfig files modified.

portikM: Done

package.json
portikM: Done

vitest.config.ts
```diff
@@ -1,22 +1,100 @@
...
+            'src/components/KThemeProvider/KThemeProvider.spec.ts',
+          ],
+        },
+      },
+      {
+        extends: true,
+        test: {
+          name: 'browser',
+          include: ['src/components/**/*.spec.ts'],
+          exclude: ['src/components/KThemeProvider/KThemeProvider.spec.ts'],
```
adamdehaven: Rather than needing these individual ignores, could we do this instead:

- Tests without browser mode: `*.spec.ts`
- Tests with browser mode: `*.browser.spec.ts`

Then just update the include/exclude rules

adamdehaven: Then you'd have e.g. `KBadge.browser.spec.ts` for browser mode, `KThemeProvider.spec.ts` that doesn't need browser mode

portikM: Updated

.github/workflows/test.yaml
```diff
@@ -52,6 +52,30 @@ jobs:
...
         run: pnpm test:unit
 
+      - name: Cache Playwright browsers
+        uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5.1.0
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ hashFiles('**/pnpm-lock.yaml') }}
+
+      # Downloads the browser binaries and installs the system libraries the runner
+      # image lacks. `--with-deps` runs apt-get on every job; only the binaries cache.
```
adamdehaven: nit: remove

src/components/KBadge/KBadge.browser.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    renderComponent(KBadge, {
+      props: {
+        appearance: variant,
+      },
+      slots: {
+        default: () => variant,
+      },
+    })
+
+    await expect.element(page.getByCSS('.k-badge')).toHaveClass(variant)
```
adamdehaven, line 19: question: why wouldn't we use the data-testid attributes in the tests? Applies to whole file

.gitignore
```diff
@@ -9,6 +9,10 @@ bundle-analyzer
 cypress/screenshots
 cypress/videos
 
+# Screenshots Vitest Browser Mode captures for failing component tests
+**/.vitest-attachments/
+**/__screenshots__
```
adamdehaven, line 13: question: if we were to enable screenshot testing in the repo (e.g. per component) would this prevent those screenshots from being checked in? (that would not be ideal)

.github/workflows/test.yaml
portikM: Removed

src/components/KBadge/KBadge.browser.spec.ts
portikM: Trying to match the original Cypress test as closely as possible. KBadge doesn't have `data-testid` attribute on top-level element.

.gitignore
portikM: I'm not sure what you mean by screenshot testing? We need this to avoid checking in screenshots generated by tests

adamdehaven (dismissed):

.github/workflows/test.yaml
```diff
@@ -52,6 +52,28 @@ jobs:
...
         run: pnpm test:unit
 
+      - name: Cache Playwright browsers
+        uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5.1.0
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ hashFiles('**/pnpm-lock.yaml') }}
+
+      - name: Install Playwright browsers
+        run: pnpm test:browser:install
```
adamdehaven: Let's change this so the verbose output is grouped and suppressed a bit
```suggestion
      - name: Install Playwright browsers
        run: |
          echo "::group::Install Playwright browsers"
          pnpm test:browser:install
          echo "::endgroup::"
```

adamdehaven (approved):

ValeryG (approved):

PR #3351, opened 2026-08-12
fix(ktabs): minimal appearance horizontal overflow

portikM:
# Summary

Make KTabs (both `minimal` and `default` appearance) scroll horizontally when overflowing

Before

Screenshot 2026-08-12 at 5 09 15 PM

After

Screenshot 2026-08-12 at 5 15 59 PM

### DIFF
--- a/src/components/KTabs/KTabs.vue
+++ b/src/components/KTabs/KTabs.vue
@@ -111,6 +111,8 @@ watch(() => modelValue, (newTabHash) => {
     gap: var(--kui-space-40, $kui-space-40);
     list-style: none;
     margin-top: var(--kui-space-0, $kui-space-0);
+    overflow-x: auto;
+    overflow-y: hidden;
     padding: var(--kui-space-0, $kui-space-0);
 
     .tab-item {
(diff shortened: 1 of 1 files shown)


src/components/KTabs/KTabs.vue
```diff
@@ -165,8 +168,6 @@ watch(() => modelValue, (newTabHash) => {
       border-bottom-style: solid;
       border-bottom-width: var(--kui-tabs-border-width, var(--kui-border-width-10, $kui-border-width-10));
       margin-bottom: var(--kui-space-70, $kui-space-70);
-      overflow-x: auto;
-      overflow-y: hidden;
```
adamdehaven, line 169: do we still need this in case the tab-item overflows?

portikM: No, I moved it from `default` appearance styles to common styles for both `default` and `minimal` appearances so it applies to both

adamdehaven (approved):

jillztom (approved):

PR #3365, opened 2026-08-24
test(kbutton): convert kbutton cypress test to vitest [KHCP-21620]

arora-r:
# Summary

Using the new `convert-cypress-to-vitest-component` skill to convert the cypress tests for KButton to vitest browser equivalents

The skill worked great and even had a workaround created for the `mountWithProdRouter` test


portikM (approved):
LGTM 🚀

PR #3367, opened 2026-08-24
chore(convert-cypress-to-vitest-component): update location & routing

arora-r:
# Summary
Update the convert cypress to vitest skill to specifically handle any `mountWithProdRouter` cases more consistently


arora-r:
not necessary

PR #3368, opened 2026-08-25
test(kpop, kdropdown): convert component tests to vitest [KHCP-21642]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-21642

Converting KPop and KDropdown Cypress component tests to vitest


test/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+ * a window in which the component mounts under the stale cursor and shows a popover transiently,
+ * which is long enough for a retrying `toBeVisible()` to catch.
+ *
+ * ```ts
+ * await resetPointer()
+ * await render(KPop, { props: { trigger: 'hover' } })
+ * await page.getByCSS('.slottedEl').hover()
+ * ```
+ */
+export const resetPointer = async (): Promise<void> => {
```
adamdehaven: I'm not sure it's obvious what this is for, and if it's really needed? Let's try to make the comments more concise. 

For the actual placement of the file, could we create a subdirectory `test/utils/` and then have this in its own file?

skills/convert-cypress-to-vitest-component/SKILL.md
```diff
@@ -31,16 +31,19 @@ Convert all of them, or say which you skipped and why.
...
 
-There is deliberately **no test-utils module**. The team's standing decision is to add
-shared helpers only once something actually needs them, so use the library APIs directly:
-`render` from `vitest-browser-vue`, and `page` from `vitest/browser`. If you find yourself
-wanting a helper, say so in your report rather than adding one — that's a decision for the
-user, not a side effect of a conversion.
+Shared helpers are added only once something actually needs them, so reach for the library
+APIs first: `render` from `vitest-browser-vue`, and `page` from `vitest/browser`. If you
+find yourself wanting a helper beyond the one below, say so in your report rather than
+adding one — that's a decision for the user, not a side effect of a conversion.
```
arora-r, line 37: Do we want to stop the AI before continuing. We could have the AI prompt the user right away by telling it to prompt the user by using the ask user tool


```suggestion
find yourself wanting a helper beyond the one below, ask the user by using the ask user tool instead of adding one with reasoning behind why its necessary — that's a decision for the user, not a side effect of a conversion.
```

portikM: I prefer that it tells me that after the skill has ran - maybe I don't want a helper while it's running. Once it finished and came up with a working solution without a helper, it tells me how widespread the usage of the new helper will be and if it's significant - only then I want to make a helper.

test/utils.ts
portikM: Made the comment more concise and moved the file. This helper will be needed in KTooltip, KSelect and KMultiselect tests too so yeah I'd say we can keep it.

adamdehaven (approved):

PR #3373, opened 2026-08-26
test(kselect, kmultiselect, ktooltip): convert tests [KHCP-21618]

portikM:
# Summary

Addresses:

- https://konghq.atlassian.net/browse/KHCP-21687
- https://konghq.atlassian.net/browse/KHCP-21682
- https://konghq.atlassian.net/browse/KHCP-21702

Convert KSelect, KMultiselect, KTooltip component tests to vitest


.github/workflows/test.yaml
```diff
@@ -65,6 +65,7 @@ jobs:
           echo "::endgroup::"
 
       - name: Run Vitest browser tests
+        timeout-minutes: 10
```
portikM, line 68: Step-level timeout so a stalled browser instance fails this step instead of running the whole job out of its 20 minute budget. A job cancelled on timeout skips `if: failure()` steps, so the screenshot upload below never runs and the stall leaves nothing to debug.

adamdehaven (approved):

PR #3413, opened 2026-09-09
fix(kinput): reveal safari generated passwords [TDX-8840]

arashsheyda:
# Summary

ticket: https://konghq.atlassian.net/browse/TDX-8840

when using safari and `new strong password` feature, the reveal password button would not work because of safari's mask


adamdehaven:
/muthur review the safari browser password fix

portikM (approved):
LGTM 🚀 
Verified in sandbox

PR #3416, opened 2026-09-10
feat(ktabs): optionally cache visited tab panels

filipgutica:
# Summary

Add opt-in `cache-tabs` to KTabs so dashboards retain their component instances, DOM nodes, and local state across tab switches. Each panel mounts on its first activation, then Vue's native `KeepAlive` caches it while inactive. The default behavior is unchanged.

## Changes

- Add an internal panel component that supports arbitrary slot content, including multiple roots.
- Preserve `onActivated()` and `onDeactivated()` hooks for panel components and their descendants.
- Discard removed tabs and inactive content when caching is disabled, while preserving the active instance. `hidePanels` clears all panels.
- Document the caching behavior and add seven Vitest browser regression tests covering state, DOM identity, lifecycle hooks, programmatic activation, navigation guards, and cleanup.
- Add a `cacheTabs` sandbox example with Draft and Ideas inputs and a caching toggle.

## How to test/verify

1. Run `pnpm sandbox:dev` and open `/tabs`.
2. Find the `cacheTabs` section. Caching is enabled initially.
3. Type in Draft, switch to Ideas, enter different text, then return to Draft. Each tab retains its text.
4. Turn off **Cache visited tabs**, switch away from the active tab, then return. Its text resets.

Local validation passed: all 29 KTabs tests in Chromium, Firefox, and WebKit (87 checks), typecheck, ESLint, and stylelint. The sandbox build also passed.

## Risks or limits

Cached panels retain memory. `KeepAlive` does not automatically pause timers, polling, or watchers; consumers must manage background work in lifecycle hooks.

### DIFF
--- a/docs/components/tabs.md
+++ b/docs/components/tabs.md
@@ -211,6 +211,23 @@ If you want to keep your `v-model` in sync so that you can programmatically chan
 <KButton @click="currentTab = '#tab2'">Activate Tab 2</KButton>
 ```
 
+### cacheTabs
+
+A `boolean` that keeps visited tab content mounted when switching tabs. Defaults to `false`.
+
+When enabled, each tab mounts on its first activation. Inactive visited panels are hidden with CSS and retain their component instances, DOM, and local state. Unvisited tabs do not mount until selected.
+
+```html
+<KTabs :tabs="tabs" cache-tabs>
+  <template #tab1>Tab 1 content</template>
+  <template #tab2>Tab 2 content</template>
+</KTabs>
+```
+
+This prevents repeated mount-triggered requests. Cached content still uses memory, and its watchers and polling can continue while hidden. Charts may need to resize when their panel becomes visible again.
+
+Removing a tab discards its cached content. Disabling `cacheTabs` unmounts inactive content. Setting `hidePanels` to `true` or unmounting KTabs discards all panel content.
+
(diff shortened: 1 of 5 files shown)


adamdehaven (changes requested):
Help me understand why this shouldn't be solved by Vue's native `KeepAlive` component with slotted content?

filipgutica:
> Help me understand why this shouldn't be solved by Vue's native `KeepAlive` component with slotted content?


By solving it with KeepAlive, what do you mean? Consumers wrapping their slotted content with KeepAlive, or using KeepAlive internally in KTabs as an alternative to this PR's implementation? 

The consumer wrapper would currently be unmounted by KTabs’ slot-level v-if, losing its cache. Internally it could work, though we’d need a component wrapper to support arbitrary slot content.

adamdehaven:
My concern with the current approach isn't the mounting mechanics, it's that `v-show` gives consumers no activated/deactivated hook to pause work while hidden, which matters for the polling/chart case this feature targets.

filipgutica:
> My concern with the current approach isn't the mounting mechanics, it's that `v-show` gives consumers no activated/deactivated hook to pause work while hidden, which matters for the polling/chart case this feature targets.

Ok I made a change to use KeepAlive internally so that work can be paused

PR #3427, opened 2026-09-16
fix(build): externalize runtime dependencies from the ES/CJS builds

merrywhether:
> **Jira:** [KHCP-21889](https://konghq.atlassian.net/browse/KHCP-21889) — externalize runtime deps from ES/CJS; UMD still inlines

Every runtime dependency was inlined into the ES/CJS builds, so consumers got a vendored copy alongside the one they resolve themselves. swrv is the case that forced it: 1.3.0's injectable cache only works if there is a single shared instance, and `KTableData` could never reach one. `external` is now derived from `dependencies` + `peerDependencies`, matching subpaths too, since a string array only matches exact ids and would keep vendoring `date-fns/locale`. UMD is untouched.

Two exclusions:

- `lodash-es` — ESM-only with no `exports` map, so a `require`-condition consumer cannot resolve it.
- `.css` ids — externalizing `v-calendar/dist/style.css` leaves a bare `require()` of a stylesheet that Vitest cannot load. `dist/kongponents.css` is byte-identical to main's either way, so the stylesheet was never the problem.

Also pins the `rollup` alias override to `^4.43.0`; it carried no version, so every install re-resolved it and changed the bundler a publish was built with.

| | before | after |
|---|---|---|
| modules | 621 | 318 |
| ES bytes | 1,368,203 | 632,784 |
| CJS bytes | 1,048,314 | 516,548 |

UMD is byte-identical at 885,516.

The preview package was installed into Kong repos. That is what caught the CSS case; all three are back to exact baseline test counts with it fixed.

Deferred to the next major: promote the externals to `peerDependencies`, forward `@popperjs/core` and `sortablejs` as peers rather than direct dependencies, and revisit `lodash-es`.

## Validation

Calendar is the most interesting from a dep perspective (direct `v-calender` dep, CSS imports, etc). Verified working in Konnect: 
image

[KHCP-21889]: https://konghq.atlassian.net/browse/KHCP-21889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adamdehaven (commented):
I know this is still a draft, but you’ll need to use `fix` rather than `chore` commit to trigger a release

vite.config.ts
```diff
@@ -17,6 +18,22 @@ const buildVisualizerPlugin = process.env.BUILD_VISUALIZER
...
+
+const { dependencies, peerDependencies } = pkg
+
+const externalDependencies = [
+  ...Object.keys(dependencies),
+  ...Object.keys(peerDependencies),
+].filter((dep) => !bundledDependencies.includes(dep))
+
+// Match the package itself and any of its subpaths, e.g. `date-fns/locale`. Style
+// imports are excluded.
```
adamdehaven, line 32: nit: readability
```suggestion
// Match the package itself and any of its subpaths, e.g. `date-fns/locale`. 
// Style imports are excluded.
```

adamdehaven (approved):
lgtm - 1 nit but can be done in followup

PR #3429, opened 2026-09-17
chore(build): externalize runtime dependencies in the ES/CJS builds

merrywhether:
# Summary

Every runtime dependency was inlined into the ES/CJS builds, so a consumer that also
depends on `date-fns`, `@floating-ui/vue` or `@kong/icons` shipped two copies: its own
deduped one from the lockfile, plus kongponents' vendored one. `external` is now derived
from `dependencies`, so the consumer resolves each of them.

- `lodash-es` stays bundled. It is ESM-only with no `exports` map, so `require('lodash-es')`
  throws `ERR_REQUIRE_ESM` on Node below 22.12 — the one dependency a `require`-condition
  consumer could not resolve. 4.18.1 is current; a version bump does not change this.
- `@popperjs/core` and `sortablejs` never appear as externals because nothing in `src/`
  imports them directly — they arrive through `v-calendar` and `vue-draggable-next`.
  `focus-trap` is type-only.
- The UMD build is untouched: `dist/kongponents.umd.js` is byte-identical at 877,271 bytes.

## Bundle effect

`vite build` (ES + CJS), totals across all `preserveModules` output:

| | before | after |
|---|---|---|
| modules | 614 | 318 |
| ES bytes | 1,340,315 | 636,284 |
| CJS bytes | 1,028,783 | 519,269 |

Externals now reached as bare specifiers: `vue`, `vue-router`, `@kong/icons`, `@floating-ui/vue`,
`date-fns`, `date-fns-tz`, `nanoid`, `swrv`, `v-calendar`, `virtua`, `vue-draggable-next`,
`focus-trap-vue`.

## Validation

Needs a consumer check against the preview package before this leaves draft — a bundled
consumer is the path essentially every consumer takes, and it is the one this change
alters. `require`-condition resolution under plain Node is unaffected for the 13 externals
(all have `require` conditions) and unchanged for `lodash-es`, which stays inlined.

## Not in scope

The externals stay in `dependencies`. Promoting them to `peerDependencies` would make
consumers install them explicitly, which is a breaking change to the install contract and
belongs in a major.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


merrywhether:
Folded into #3427 to keep this to a single kongponents publish.

PR #3432, opened 2026-09-18
feat(mcp): POC - mcp server in Kongponents [KHCP-21913]

LukeSwierlik:
# Summary
- Feat - Adds a version-matched MCP server for Kongponents documentation, component source, styles, and theme tokens.
- Feat - Supports Codex, Claude Code, GitHub Copilot, and other stdio MCP clients.
- Chore - Packages the MCP executable and documentation snapshot with every Kongponents release.
- Chore - Adds setup guides and a new UI for Agents documentation section.
- Fix - Pins the local MCP Inspector and scopes its pnpm trust exception to the exact affected dependency.
- Test - Covers component lookup, documentation search, path protection, theme variables, protocol discovery, and packaged stdio usage.
---
- Task in Jira: https://konghq.atlassian.net/browse/KHCP-21913

## How it works

- The MCP server is shipped inside the regular `@kong/kongponents` npm package.
- Applications such as `konnect-apps` install Kongponents in exactly the same way as before.
- A small project-level configuration tells Codex, Claude Code, or GitHub Copilot how to start the bundled `kongponents-mcp` command.
- The AI client starts and stops the local MCP process automatically whenever it needs it.
- The bundled snapshot always matches the installed Kongponents version.
- Updating the Kongponents dependency also updates the documentation and component information available to agents.
- Consumers do not need to clone the Kongponents repository, run a separate service, or start MCP manually.
- No API key or hosted documentation service is required.

## Why this helps

- Agents can find component properties and examples without repeatedly searching the repository or `node_modules`.
- They receive documentation that matches the version used by the application.
- Less repository scanning means faster discovery and less unnecessary context.
- Structured tools reduce guessed props, outdated APIs, and incorrect component recommendations.
- Agents can compare components, inspect theme tokens, and check deprecated APIs before changing application code.
- The same source of truth works across multiple supported AI tools.

### Doc page:
- https://deploy-preview-3432--kongponents.netlify.app/guide/mcp-server.html

#### Documentation payload

| Component | Tested API | Before MCP | With MCP | Token reduction |
|---|---|---:|---:|---:|
| `KButton` | `to` | 1,540 | 237 | 84.6% |
| `KInput` | `modelValue` | 3,133 | 147 | 95.3% |
| `KSelect` | `v-model` | 5,993 | 198 | 96.7% |
| `KModal` | `visible` | 5,897 | 258 | 95.6% |
| `KTableView` | `headers` | 15,807 | 1,498 | 90.5% |
| **Total** | — | **32,370** | **2,338** | **92.8%** |

#### End-to-end usage

| Component | Before MCP | With MCP | Difference |
|---|---:|---:|---:|
| `KButton` | 25,851 | 39,815 | +54.0% |
| `KInput` | 27,422 | 24,616 | −10.2% |
| `KSelect` | 30,289 | 39,997 | +32.1% |
| `KModal` | 30,382 | 24,734 | −18.6% |
| `KTableView` | 34,814 | 26,179 | −24.8% |
| **Total** | **148,758** | **155,341** | **+4.4%** |

#### Summary

| Metric | Before MCP | With MCP | Difference |
|---|---:|---:|---:|
| Documentation payload | 32,370 | 2,338 | **−92.8%** |
| Non-cached input tokens | 48,080 | 38,300 | **−20.3%** |
| Total end-to-end tokens | 148,758 | 155,341 | **+4.4%** |

> MCP substantially reduces the documentation context sent to the model. The
> end-to-end result also includes the fixed cost of MCP instructions, tool
> schemas, tool execution rounds, and prompt-cache variability.

> Token counts are estimates based on approximately four characters per token. Actual usage depends on the model and tokenizer. Response generation remained below 1 ms in every tested scenario.

### MCP Inspector

https://github.com/user-attachments/assets/8b759115-943b-41ab-a673-cb51ce5ea811


adamdehaven (changes requested):
Needs review

mcp/catalog.ts
```diff
@@ -0,0 +1,66 @@
...
+  title,
+  docPath: `/components/${slug}`,
+  exports,
+  sourceDirectories: options.sourceDirectories ?? [`src/components/${exports[0]}`],
+  typeFiles: typeFile ? [`src/types/${typeFile}.ts`] : [],
+  aliases: options.aliases ?? [],
+  deprecated: options.deprecated,
+})
+
+export const COMPONENT_CATALOG: ComponentCatalogDefinition[] = [
```
adamdehaven, line 21: issue: shouldn't this be dynamic so that new components are discovered automatically?

adamdehaven (commented):
Overall, the implementation doesn't provide enough context. As an example, I can't review `mcp/generate-snapshot.ts` as there's no concise way on the PR to understand what each method is responsible for, etc.

Can you implement concise comments in the code to help with maintenance and review?

Justineo:
It would be worth checking the parser against the different structures in our docs. Props, slots, and HTML attributes aren’t always organized at the same heading level, and some APIs use tables or reference other sections. For example, the current extraction can pick up `KButton`’s `default` slot as a prop, or return parent-component props for `KDropdownItem`.

Could we also compare this with a Skill backed by a component-to-document index and version-matched docs? That could provide similar progressive disclosure without a separate MCP runtime. MCP may still be useful, but this would help clarify its benefits over a simpler approach.

For the retrieval granularity, I’d be interested in how this compares with reading a component doc directly during a complete coding task. Smaller snippets save tokens per response, but additional calls also consume context and may lead to repeated lookups across the same document. Reading isolated sections could also make relationships and constraints easier to miss. Comparing task quality and end-to-end token usage would give us a clearer picture than response size alone.

LukeSwierlik:
> It would be worth checking the parser against the different structures in our docs. Props, slots, and HTML attributes aren’t always organized at the same heading level, and some APIs use tables or reference other sections. For example, the current extraction can pick up `KButton`’s `default` slot as a prop, or return parent-component props for `KDropdownItem`.
> 
> Could we also compare this with a Skill backed by a component-to-document index and version-matched docs? That could provide similar progressive disclosure without a separate MCP runtime. MCP may still be useful, but this would help clarify its benefits over a simpler approach.
> 
> For the retrieval granularity, I’d be interested in how this compares with reading a component doc directly during a complete coding task. Smaller snippets save tokens per response, but additional calls also consume context and may lead to repeated lookups across the same document. Reading isolated sections could also make relationships and constraints easier to miss. Comparing task quality and end-to-end token usage would give us a clearer picture than response size alone.

@Justineo Thanks for the feedback — I addressed the parser issues, including the `KButton` slot and `KDropdownItem` parent-prop cases, and added regression tests for different documentation structures.

I also benchmarked the preview package in Konnect-Apps across four Vue tasks:

| Strategy | Correctness | Avg. time | Non-cached input |
| --- | ---: | ---: | ---: |
| Direct docs | 20/20 | 77.2s | 198,014 |
| Skill | 20/20 | 62.0s | 106,805 |
| MCP | 20/20 | 86.5s | 137,574 |

All approaches produced correct results. The Skill was used in 4/4 tasks, while MCP was selected in 2/4. When used, MCP returned correct version-matched documentation.

My takeaway is that MCP is still valuable as a structured interface, while a lightweight Skill can improve its discoverability and consistent usage.

> This was one run per task and strategy, so the results indicate direction rather than statistical significance.

mcp/catalog.ts
LukeSwierlik: Good point — this is dynamic now. New components are discovered automatically, with manual entries only needed for special cases.

LukeSwierlik:
> Overall, the implementation doesn't provide enough context. As an example, I can't review `mcp/generate-snapshot.ts` as there's no concise way on the PR to understand what each method is responsible for, etc.
> 
> Can you implement concise comments in the code to help with maintenance and review?

Good call — I added concise comments to explain the purpose of the less obvious methods and make the snapshot generation flow easier to follow 👍

adamdehaven:
/ksai review the MCP implementation and also provide recommendation on where the MCP should live

mcp/tools.spec.ts
```diff
@@ -0,0 +1,168 @@
+import { readFileSync } from 'node:fs'
+import path from 'node:path'
+import { describe, expect, it } from 'vitest'
+import { createToolHandlers } from './tools'
+import { extractComponentProperty, extractDocumentationScope, listComponentProperties } from './lib/markdown'
+import type { McpSnapshot } from './types'
+
+const snapshot = JSON.parse(readFileSync(path.resolve('bin/mcp-data/snapshot.json'), 'utf8')) as McpSnapshot
```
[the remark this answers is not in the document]
LukeSwierlik: Fixed

mcp/stdio.ts
```diff
@@ -0,0 +1,9 @@
+import { serveStdio } from '@modelcontextprotocol/server/stdio'
+import { readFileSync } from 'node:fs'
+import { createServer } from './server'
+import type { McpSnapshot } from './types'
+
+const snapshot = JSON.parse(readFileSync(new URL('../mcp-data/snapshot.json', import.meta.url), 'utf8')) as McpSnapshot
+const packageJson = JSON.parse(readFileSync(new URL('../../package.json', import.meta.url), 'utf8')) as { version: string }
+
+void serveStdio(() => createServer(snapshot, packageJson.version))
```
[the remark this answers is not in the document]
LukeSwierlik: Fixed

adamdehaven:
I'm not sure we want to expose a `get_theme_variables` tool for now; we don't want our internal consumers customizing the existing Konnect themes.

docs/guide/mcp-server.md
```diff
@@ -0,0 +1,346 @@
...
+| Approach | Strengths | Tradeoffs |
+| --- | --- | --- |
+| Read Markdown directly | No additional runtime or tool schemas; the agent sees the complete page and its relationships. | The agent must locate files and may load substantially more context than a focused question needs. |
+| Version-matched Skill and document index | Progressive disclosure with lightweight instructions and references; well suited to a Codex-specific workflow. | The Skill must be installed or copied into a location the client discovers and kept aligned with the application's Kongponents dependency. Other MCP clients do not automatically consume a Codex Skill. |
+| Bundled MCP server | Uses the exact installed package version and exposes the same structured, read-only interface to multiple stdio MCP clients. It can query docs, source, styles, and theme tokens. | Tool schemas, server instructions, and additional tool calls have a fixed context and latency cost. Focused payload savings do not guarantee lower total token usage. |
+
+Skills and MCP can also be complementary: a Skill can describe a team workflow while delegating version-matched Kongponents lookup to this server.
+
+### Evaluation methodology
```
adamdehaven: I'm not sure why this section needs to be included?

docs/guide/mcp-server.md
```diff
@@ -0,0 +1,346 @@
...
+```json
+{
+  "components": ["KButton"],
+  "sections": ["Props", "Events"]
+}
+```
+
+Large Markdown and source payloads are returned once as tool text. Structured results contain compact metadata instead of a second copy of the same content. This keeps the context smaller without removing information from the agent.
+
+### MCP, Skills, and direct documentation access
```
adamdehaven: I'm not sure this type of documentation belongs in the Kongponents repo (and similar sections in the new docs)

We shouldn't be explaining "Why MCP" etc. in our component documentation. 

If we're offering a MCP server, ok, but we shouldn't be explaining the technology, eval methods, token efficiency, etc.

docs/plans/mcp-server.md
```diff
@@ -0,0 +1,61 @@
+# Kongponents MCP Server Implementation Plan
```
adamdehaven: We shouldn't be adding this document to the docs site.

package.json
```diff
@@ -105,6 +114,7 @@
     "@vitejs/plugin-vue": "^5.2.4",
     "@vitest/browser-playwright": "^4.1.10",
     "@vue/compiler-core": "^3.5.40",
+    "@vue/compiler-sfc": "3.5.40",
```
adamdehaven: why is this dependency needed?

package.json
```diff
@@ -170,7 +184,17 @@
...
+        "@semantic-release/commit-analyzer",
+        {
+          "releaseRules": [
+            {
+              "type": "docs",
+              "release": "patch"
+            }
+          ]
+        }
+      ],
```
adamdehaven: what's the purpose of this addition?

vitest.config.ts
```diff
@@ -64,7 +64,8 @@ export default defineConfig({
         test: {
           name: 'unit',
           environment: 'jsdom',
-          include: ['src/**/*.spec.ts'],
+          globalSetup: ['./mcp/test-global-setup.ts'],
```
adamdehaven: question: why inject the mcp test setup into all of the existing component tests that are unrelated?

.github/workflows/get-changed-files.yaml
```diff
@@ -36,14 +36,16 @@ jobs:
...
-      - name: Check if component (`src` directory) files have changed
+      - name: Check if publishable runtime files have changed
         uses: Kong/changed-files@cd69b79ca4e2a1198064c5e6564cce68920df311
         id: component-files-changed
         # Return 'true' for any directories listed here that changed
         with:
           files: |
             src/**
             nuxt/**
+            cli/**
```
adamdehaven: why add the cli/ directory here?

docs/.vitepress/config.ts
```diff
@@ -7,6 +7,7 @@ const KONG_SVG = '<svg viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/
 
 export default defineConfig({
   base: '/',
+  srcExclude: ['plans/**'],
```
adamdehaven: issue: this should be removed. "plans" shouldn't be checked in to the docs directory

adamdehaven (changes requested):

PR #3433, opened 2026-09-18
refactor(build): break the externals comment and drop a swrv deep import

merrywhether:
Two follow-ups to #3427.

The `isExternalDependency` comment now breaks at the sentence rather than mid-sentence, per [review](https://github.com/Kong/kongponents/pull/3427#discussion_r4050608819).

`useUtilities.ts` imported `IKey` and `fetcherFn` from `swrv/dist/types`. Type-only, no runtime change, so `refactor:` — this publishes nothing on its own.

[TEST_COVERAGE_EXEMPT_REASON] The only prod code change is moving the location of type imports which is verified by type-checking.


adamdehaven (approved):
lgtm

